PR #233, opened 2014-09-05
Modify SCC parser's show/hide detection logic.

tparnell:
- Find exact frame within a line that a show or hide command is sent.
- Allow an initial show prior to any text being buffered
- Fix existing test case.  Test case was trivially passing b/c no parsing
  was occuring (missing tab character is regex.)


niksedk:

PR #328, opened 2014-11-13
Also show progress in taskbar.

FroggieFrog:
First commit with support for showing the progress in taskbar. (VobSubOcr)
I can't test other formats because I don't have them. (e.g. DvdSubRip)

### DIFF
--- a/src/Forms/VobSubOcr.cs
+++ b/src/Forms/VobSubOcr.cs
@@ -19,6 +19,7 @@ using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using System.Xml;
+using Microsoft.WindowsAPICodePack.Taskbar;
 
 namespace Nikse.SubtitleEdit.Forms
 {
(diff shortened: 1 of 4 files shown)


niksedk:
I think it possible to do without included dll files...

FroggieFrog:
Why reinvent the wheel, when there is a official library from Microsoft?
It would end up in a bunch of PInvokes. Is this really better/preferable?

niksedk:
I think it's included in .net 4...

FroggieFrog:
Yes, but only in WPF. The app is Winforms...
http://msdn.microsoft.com/en-us/library/system.windows.shell.taskbariteminfo%28v=vs.100%29.aspx

niksedk:
Perhaps I'm old-fashioned but I don't like bloat and extra dlls - I would *much* rather have only the needed functionality.
Minimal code: https://psycodedeveloper.wordpress.com/2013/01/29/displaying-progress-in-the-windows-7-taskbar/

PR #388, opened 2015-01-04
Fix for RemoveUneeded spaces

ivandrofly:


ivandrofly:
@XhmikosR Here

XhmikosR:
LGTM. /CC @niksedk for review

niksedk:
I don't think it 100% sure... if a subtitle has an quote end (from a previous line) + start a new quote (completed in a later line) - but still probably mostly good. Perhaps detection of this can be improved...

PR #389, opened 2015-01-04
Patch 3.0

ivandrofly:


src/Forms/GoToLine.cs
```diff
@@ -85,10 +85,7 @@ private void TextBox1KeyDown(object sender, KeyEventArgs e)
                     e.KeyCode == Keys.End ||
                     (e.KeyValue >= 96 && e.KeyValue <= 105))
                 {
-                }
-                else if (e.KeyData == (Keys.Shift | Keys.Home) || e.KeyData == (Keys.Shift | Keys.End) ||
```
ivandrofly: This code isn't needed 'cause it would never run... the first condition will cover it

niksedk: Hm, looks good to me... could you explain a bit more?

ivandrofly: https://github.com/ivandrofly/subtitleedit/blob/patch-3.0/src/Forms/GoToLine.cs#L70
Will cover the second condition which would capture the Shift+Home...

niksedk:
Seems to work okay except the "e.Handled = true;" should not be there as it disables the the arrow/home/end keys (alternately replace it with "return" which might look better than empty brackets)

ivandrofly:
@niksedk I'll close this pull and pull a new one which contain **return** instead

PR #390, opened 2015-01-05
Patch 3.0

ivandrofly:


ivandrofly:
@niksedk Handled replaced with return

PR #391, opened 2015-01-05
Preserve Subtitle start number

ivandrofly:


ivandrofly:
Ah I thought you missed it... :/ can be closed then. @niksedk

PR #397, opened 2015-01-07
Better check for line length [Faster]

ivandrofly:
I have done some test and the split performed better the replace

### DIFF
--- a/src/Logic/Paragraph.cs
+++ b/src/Logic/Paragraph.cs
@@ -132,7 +132,7 @@ namespace Nikse.SubtitleEdit.Logic
             {
                 if (string.IsNullOrEmpty(Text))
                     return 0;
-                return Text.Length - Text.Replace(Environment.NewLine, string.Empty).Length;
+                return Text.Split(Utilities.NewLineChars, StringSplitOptions.RemoveEmptyEntries).Length;
             }
         }
 


niksedk:
It's not faster on my computer... but the old code actually gives the wrong number of lines!

This seems to be the fastest code in my tests:
                int lines = 1;
                int idx = Text.IndexOf('\n');
                while (idx != -1)
                {
                    lines++;
                    idx = Text.IndexOf('\n', idx + 1);
                }
                return lines;

This is shorter but not as fast: 
                return Text.Split('\n').Length;

ivandrofly:
Here is the code I used to test it http://pastebin.com/vh3cqj3A

niksedk:
thx for sharing :)
Your test shows that the second method is slightly faster on my machine.

A few pointers: always run the test many, many times + always compile to release + always run outside VS/debug (e.g. with Ctrl+F5)
Here's the modified code + test results: http://pastebin.com/KpQQnneW

ivandrofly:
I think it has something to do with CPU clock speed... when I set loop max to 1000000 the second method performed better when otherwise the first method did perform better. have you tried with http://www14.zippyshare.com/v/31333125/file.html what was the result?!

niksedk:
Your zippshare file showed the first one to be slightly faster... but I believe it was compiled to "debug"...
Did you try my example from pastebin? What was your results in release mode outside VS?

niksedk:
Do let me know if you find something even faster :)

ivandrofly:
yeah that was compiled to debug... the example from pastbin the 3rd method was faster :+1:

PR #403, opened 2015-01-08
MultipleReplace remove line if only tag remains

ivandrofly:


ivandrofly:
@niksedk  here

niksedk:
Nice :)

PR #404, opened 2015-01-09
Handle possible error that can happen when BatchConverting

ivandrofly:
If you set a file to read-only and try batch converting it you will get a unhandled exception same thing would happen if the file is read-only and **Overwrite original file** is uncheck


niksedk:
Could you make it to a try-catch instead?

PR #408, opened 2015-01-09
Fix/patch 1

ivandrofly:


niksedk:
Nice - that should cover all situations :)

ivandrofly:
:+1:

PR #414, opened 2015-01-12
Update de-DE translation

xylographe:
add Settings.**AdjustSetEndTimeAndGoToNext** tag (3.4.5)
add Settings.**WaveformTextFontSize** tag (3.4.6)

### DIFF
--- a/src/Languages/de-DE.xml
+++ b/src/Languages/de-DE.xml
@@ -1527,6 +1527,7 @@ gleiche Untertiteldatei bearbeiten können (Zusammenarbeit)</Information>
     <WaveformSelectedColor>Farbe markiert</WaveformSelectedColor>
     <WaveformBackgroundColor>Hintergrundfarbe</WaveformBackgroundColor>
     <WaveformTextColor>Textfarbe</WaveformTextColor>
+    <WaveformTextFontSize>Schriftgröße</WaveformTextFontSize>
     <WaveformAndSpectrogramsFolderEmpty>Ordner für Spektogramme und Wellenform leeren</WaveformAndSpectrogramsFolderEmpty>
     <WaveformAndSpectrogramsFolderInfo>Ordner für Spektogramme und Wellenform enthält {0} Dateien ({1:0.00} MB)</WaveformAndSpectrogramsFolderInfo>
     <Spectrogram>Spektrogramm</Spectrogram>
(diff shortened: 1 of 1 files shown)


xylographe:
Why did AppVeyor fail on this PR???

niksedk:
Perhaps some appVeyor bug? Just says "Not found".

xylographe:
OK. As long as it didn't break anything…

Can I run these unit tests locally (in shell)? Or would I need AppVeyor?

niksedk:
You can use VS Express if you don't have a full VS - I've not tried the "mstest" command tool but the UI runner is pretty easy to run...

xylographe:
Thank you!

Via [mstest.exe](http://msdn.microsoft.com/en-us/library/ms182489.aspx) (which I couldn't find in VSE12.0), I found [vstest.console.exe](http://msdn.microsoft.com/en-us/library/jj155796.aspx). Works like a charm. :smile:

Only the __FixCommonOcrErrorsSlashIsL__ unit test fails.
`FixCommonErrorsTest.cs:497`: _// will fail if English dictionary is not found._
Strange, `~\AppData\Roaming\Subtitle Edit\Dictionaries` contains `en_US.dic` & `en_US.aff`.
Will investigate that later.

niksedk:
Can you set it to run in x64 mode? The last unit test might pass in X64 mode...

xylographe:
I tried `vstest.console Test.dll /platform:x64` (uses framework40) and `vstest.console Test.dll /platform:x64 /framework:framework45`. Same test still fails.

The error message is a bit strange:
```
Failed   FixCommonOcrErrorsSlashIsL
Error Message:
   Assert.AreEqual failed. Expected:<The font is ita/ic!>. Actual:<The font is italic!>.
Stack Trace:
   bij Test.FixCommonErrorsTest.FixCommonOcrErrorsSlashIsL() in c:\Users\Eric\Documents\Source\SubtitleEdit\src\Test\FixCommonErrorsTest.cs:regel 497
```
Shouldn't that be `Expected:<The font is italic!>. Actual:<The font is ita/ic!>.`

xylographe:
*Almost* forgotten. :grin: 

For the record: test failure was caused by missing `vstest.console` argument `/settings:Local.testsettings`

PR #416, opened 2015-01-12
Update: en_US_names

ivandrofly:


niksedk:
Atm SE cannot have normal words as 'names'... it will conflict with 'Change casing'.
Perhaps this could be changed so that 'Change casing' has a 'false' list - but this has to tested all places where word lists are used *very* carefully.

ivandrofly:
:+1:

PR #420, opened 2015-01-13
Bugfix from previous patch

ivandrofly:
+1 Test added


niksedk:
I'm not to sure about this... does this fix any bugs?
It will need a lot of testing, like sentences with multiple dash'es.

ivandrofly:
yeah
"Foobar.\r\n<font color='red'>- Mm-hmm.</font>"
I'll add more test then

PR #431, opened 2015-01-17
Unneded codes removed

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -1888,11 +1888,6 @@ public static void GetLineLengths(Label label, string text)
 
         public static bool IsValidRegex(string testPattern)
         {
-            if (string.IsNullOrEmpty(testPattern))
```
ivandrofly: Let the Try Catch handle it...

alfaproject: If the `testPattern` is empty, there won't be any error.

ivandrofly: :+1: @alfaproject

PR #446, opened 2015-01-22
RemoveStartEndNoise speeded up

ivandrofly:
Test => http://pastebin.com/LPfeRYei
Speed improved +10 seconds
If see that this method isn't been used for now...

### DIFF
--- a/src/Forms/RemoveTextFromHearImpaired.cs
+++ b/src/Forms/RemoveTextFromHearImpaired.cs
@@ -76,31 +76,21 @@ namespace Nikse.SubtitleEdit.Forms
 
         public static string RemoveStartEndNoise(string text)
         {
-            string s = text.Trim();
-            if (s.StartsWith("<b>") && s.Length > 3)
-                s = s.Substring(3);
-            if (s.StartsWith("<i>") && s.Length > 3)
-                s = s.Substring(3);
-            if (s.StartsWith("<u>") && s.Length > 3)
-                s = s.Substring(3);
-            if (s.StartsWith("<B>") && s.Length > 3)
-                s = s.Substring(3);
-            if (s.StartsWith("<I>") && s.Length > 3)
-                s = s.Substring(3);
-            if (s.StartsWith("<U>") && s.Length > 3)
-                s = s.Substring(3);
+            const int i = ('i' | 'I');
+            const int b = ('b' | 'B');
+            const int u = ('u' | 'U');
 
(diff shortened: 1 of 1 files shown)


src/Forms/RemoveTextFromHearImpaired.cs
```diff
@@ -76,31 +76,21 @@ private void FixLargeFonts()
...
-                s = s.Substring(0, s.Length - 4);
-            if (s.EndsWith("</I>") && s.Length > 4)
-                s = s.Substring(0, s.Length - 4);
-            if (s.EndsWith("</U>") && s.Length > 4)
+            // <i> <I> | <b> <B> | <u> <U>
+            if (text.Length > 3 && text[0] == '<' && text[2] == '>' &&
+                ((text[1] & i) == text[1] || (text[1] & b) == text[1] || (text[1] & u) == text[1]))
+            {
+                s = s.Substring(0, 3);
+            }
```
xylographe:
```
if (s.Length > 3 && s[0] == '<' && s[2] == '>' && "ibuIBU".Contains(s[1])
{
    s = s.Substring(3)
}
```
Perhaps?

But it won't handle `text = "<i><b>Lorem ipsum dolor sit amet</b></i>"`

src/Forms/RemoveTextFromHearImpaired.cs
```diff
@@ -76,31 +76,21 @@ private void FixLargeFonts()
...
+            // </i> </I> | </b> </B> | </u> </U>
+            int idx = (s.Length - 4);
+            if(s.Length > 4 && s[idx + 2] == '/' && s[idx] == '<' &&
+                 ((text[idx + 3] & i) == text[idx] || (text[idx + 3] & b) == text[idx] || (text[idx + 3] & u) == text[idx]))
                 s = s.Substring(0, s.Length - 4);
```
xylographe:
```
if (idx > 0 && s[idx] == '<' && s[idx+1] == '/' && s[idx+3] == '>' && "ibuIBU".Contains(s[idx+2])
{
    s = s.Remove(idx)
}
```

ivandrofly:
@xylographe as you can see I'm trying to avoid using string which will decrease speed

ivandrofly:
@xylographe http://pastebin.com/LPfeRYei which your  tweaks and report the results you got

xylographe:
Sure, optimization is good. But, the functionality of the code should not be altered.

PR #447, opened 2015-01-22
RemoveStartEndNoise speeded up

ivandrofly:


src/Forms/RemoveTextFromHearImpaired.cs
```diff
@@ -76,32 +76,24 @@ private void FixLargeFonts()
...
-            if (s.EndsWith("</U>") && s.Length > 4)
+            const int i = ('i' | 'I');
+            const int b = ('b' | 'B');
+            const int u = ('u' | 'U');
+
+            var s = text.Trim();
+            // <i> <I> | <b> <B> | <u> <U>
+            if (s.Length > 3 && s[0] == '<' && s[2] == '>' &&
+                ((s[1] & i) == text[1] || (s[1] & b) == s[1] || (s[1] & u) == s[1]))
+            {
```
xylographe: The expression `(s[1] & ('i' | 'I')) == s[1]` doesn't seem to be the same as `(s[1] == 'i' || s[1] == 'I')`.

A powershell example:
```
PS> base2 ([int]'i'[0])
1101001
PS> base2 ([int]'I'[0])
1001001
PS> base2 ([int]'i'[0] -bor [int]'I'[0])
1101001
PS> base2 (0x40)
1000000
PS> (0x40 -band ([int]'i'[0] -bor [int]'I'[0])) -eq 0x40
True
PS> [char]0x40
@
```

ivandrofly: yeah I know but since the subrip only use <i>, <p>, <u>... I think it wouldn't make difference that is why I check for '<' && '>' first

PR #452, opened 2015-01-24
Fixed languageOriginal would always be empty

ivandrofly:
Fails building because of: 
[01:49:29] EXEC : error : Object reference not set to an instance of an object. [C:\projects\subtitleedit\src\SubtitleEdit.csproj]
[01:49:41] EXEC : error : Something went wrong when generating the revision number... [C:\projects\subtitleedit\src\SubtitleEdit.csproj]

I thinks the problem were caused by language file updates from last commites
try fixing it in your file


niksedk:
Nice catch.

Perhaps better error handling in the get revision-number code + a re-try... my guess is a timeout.

PR #454, opened 2015-01-24
More refact in Utilities

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -848,7 +848,7 @@ public static string RemoveHtmlTags(string s, bool alsoSsaTags)
         public static string RemoveSsaTags(string s)
         {
             int k = s.IndexOf('{');
-            while (k >= 0)
+            while (k > -1)
```
niksedk: I think k >= 0 is better...

PR #455, opened 2015-01-25
Feat/patch 2

ivandrofly:
This patch also contains fix for inserting unicode to selected text...

PR #458, opened 2015-01-26
New helper method to remove Unicode  chars + Refact

ivandrofly:
Commit from #455... I'll fix the feat and re-pull it later

### DIFF
--- a/src/Forms/Main.cs
+++ b/src/Forms/Main.cs
@@ -2315,7 +2315,7 @@ namespace Nikse.SubtitleEdit.Forms
                         {
                             ImportSubtitleFromMatroskaFile(fileName);
                             return;
-                        }                        
+                        }
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


src/Forms/Main.cs
```diff
@@ -6591,10 +6591,10 @@ private void NormalToolStripMenuItemClick(object sender, EventArgs e)
                     if (p != null)
                     {
                         int indexOfEndBracket = p.Text.IndexOf('}');
-                        if (p.Text.StartsWith("{\\") && indexOfEndBracket > 1 && indexOfEndBracket < 6)
-                            p.Text = p.Text.Remove(0, indexOfEndBracket + 1);
+                        if (p.Text.StartsWith("{\\") && indexOfEndBracket > 1 && indexOfEndBracket == 5)
```
niksedk: < 6 is on purpose... could also be {\a5}...

PR #461, opened 2015-01-26
Fixed typos + minor refact

xylographe:
The two if-expressions in `src/Forms/DvdSubRipChooseLanguage.cs` look like 'always true'. But, they're probably there for a reason.


src/Forms/DvdSubRipChooseLanguage.cs
```diff
@@ -143,7 +143,7 @@ private void ComboBoxLanguagesSelectedIndexChanged(object sender, EventArgs e)
 
         private void ButtonOkClick(object sender, EventArgs e)
         {
-            if (listBox1.Items.Count > -1)
+            if (listBox1.Items.Count >= 0)
```
xylographe: always true?

src/Forms/DvdSubRipChooseLanguage.cs
```diff
@@ -175,7 +175,7 @@ private void DvdSubRipShowSubtitles_KeyDown(object sender, KeyEventArgs e)
 
         private void buttonSaveAs_Click(object sender, EventArgs e)
         {
-            if (listBox1.Items.Count > -1)
+            if (listBox1.Items.Count >= 0)
```
xylographe: always true?

src/Forms/Beamer.cs
```diff
@@ -288,7 +288,7 @@ private Bitmap GenerateImageFromTextWithStyle(string text)
                     sb = new StringBuilder();
 
                     int endIndex = text.Substring(i).IndexOf('>');
-                    if (endIndex == -1)
```
ivandrofly: I think this isn't need since the method is built only to return -1 or >

xylographe: I think, the assumption is, that comparing against 0 is more efficient than comparing against -1 ([here for example](https://github.com/SubtitleEdit/subtitleedit/commit/61d8bda6fc4a137152c81d1371b9e767b9aaab90#diff-65a9c06f48cd76263659510678e33a72R852)). Although [my own test](https://www.dropbox.com/s/recumj0rm8nxl6f/MinusOne-Zero.cs?dl=0) gave inconclusive results.

src/Forms/DvdSubRipChooseLanguage.cs
ivandrofly: Nice one :1

niksedk: he, yes - I guess the if statement can be removed

src/Forms/DvdSubRipChooseLanguage.cs
niksedk: Yeah...

PR #462, opened 2015-01-27
Remove unneeded spaces for <b>, <u> + Refact

ivandrofly:
/cc @niksedk


src/Logic/Utilities.cs
```diff
@@ -2972,16 +2971,15 @@ public static string RemoveUnneededSpaces(string text, string language)
             const string zeroWhiteSpace = "\u200B";
             const string zeroWidthNoBreakSpace = "\uFEFF";
             const string noBreakSpace = "\u00A0";
-            const string char160 = " "; // Convert.ToChar(160).ToString()
```
niksedk: Why did you remove this one?

ivandrofly: 160d = A0hex which is noBreakSpace

PR #463, opened 2015-01-27
Set allowMultLine's default value to false

ivandrofly:


src/Forms/Main.cs
```diff
@@ -11956,14 +11956,14 @@ private void InsertUnicodeGlyph(object sender, EventArgs e)
...
-                PasteIntoActiveTextBox(item.Text, false);
+                PasteIntoActiveTextBox(item.Text);
         }
 
         private void InsertUnicodeGlyphAllowMultiLine(object sender, EventArgs e)
         {
             var item = sender as ToolStripItem;
             if (item != null)
-                PasteIntoActiveTextBox(item.Text, true);
+                PasteIntoActiveTextBox(item.Text);
```
xylographe: Perhaps, this should remain `true`?

ivandrofly: :+1: @xylographe

PR #465, opened 2015-01-28
SE build depency + UpdateLanguageFiles refact

xylographe:
LanguageDeserializer.cs must be generated before building SE.

No need to always generate LM — only if Language has changed.

### DIFF
--- a/src/UpdateLanguageFiles/Program.cs
+++ b/src/UpdateLanguageFiles/Program.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+using System.Text;
 
 namespace UpdateLanguageFiles
 {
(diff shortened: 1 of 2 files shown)


src/UpdateLanguageFiles/Program.cs
```diff
@@ -25,11 +26,10 @@ private static int Main(string[] args)
                 string oldLanguageAsXml = File.ReadAllText(args[0]);
                 if (oldLanguageAsXml != currentLanguageAsXml)
                 {
-                    File.WriteAllText(args[0], currentLanguageAsXml);
+                    language.Save(args[0]);
```
xylographe: Alternatively, `File.WriteAllText(args[0], currentLanguageAsXml, Encoding.UTF8);`

PR #466, opened 2015-01-28
Update cs-CZ.xml

Trottel:
Updated for version 3.4.5.132


xylographe:
It would appear, that [ExtractDateTimeInfo.Description](https://github.com/Trottel/subtitleedit/blob/patch-11/src/Languages/cs-CZ.xml#L455) has been ‘replaced’ by [ExtractDateTimeInfo.OpenVideoFile](https://github.com/SubtitleEdit/subtitleedit/blob/master/LanguageMaster.xml#L455). Next update, perhaps? :smile:

PR #469, opened 2015-01-28
Fix + Refact (AllowFix)

ivandrofly:
/cc @niksedk

### DIFF
--- a/src/Forms/FixCommonErrors.cs
+++ b/src/Forms/FixCommonErrors.cs
@@ -799,11 +799,11 @@ namespace Nikse.SubtitleEdit.Forms
                             p.Text = p.Text.Replace(Environment.NewLine, " ");
 
                             string stripped = Utilities.RemoveHtmlTags(prev.Text).TrimStart();
-                            if (!stripped.StartsWith("- "))
+                            if (!stripped.StartsWith("- ", StringComparison.Ordinal))
                                 prev.Text = "- " + prev.Text.TrimStart();
 
                             stripped = Utilities.RemoveHtmlTags(p.Text).TrimStart();
-                            if (!stripped.StartsWith("- "))
+                            if (!stripped.StartsWith("- ", StringComparison.Ordinal))
                                 p.Text = "- " + p.Text.TrimStart();
 
                             prev.Text = prev.Text.Trim() + Environment.NewLine + p.Text;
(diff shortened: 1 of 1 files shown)


src/Forms/FixCommonErrors.cs
```diff
@@ -67,7 +67,7 @@ public sealed partial class FixCommonErrors : Form
         private static readonly Regex UrlOrg = new Regex(@"\w\.org\b", RegexOptions.Compiled);
 
         private static readonly Regex ReAfterLowercaseLetter = new Regex(@"[a-zæøåäöéùáàìéóúñüéíóúñü]I", RegexOptions.Compiled);
-        private static readonly Regex ReBeforeLowercaseLetter = new Regex(@"I[a-zæøåäöéùàìéóúñüéíóúñü]", RegexOptions.Compiled);
+        private static readonly Regex ReBeforeLowercaseLetter = new Regex(@"I[a-zæøåäöéùáàìéóúñüéíóúñü]", RegexOptions.Compiled);
```
ivandrofly: <code>à</code> added

src/Forms/FixCommonErrors.cs
```diff
@@ -2745,8 +2733,10 @@ public static string FixAloneLowercaseIToUppercaseLine(Regex re, string oldText,
                 s = s.Replace(">" + target + "</", ">I</");
             if (s.Contains(">" + target + " "))
                 s = s.Replace(">" + target + " ", ">I ");
-            if (s.Contains(">" + target + "" + Environment.NewLine))
```
ivandrofly: "" is equal to string.Empty

src/Forms/FixCommonErrors.cs
```diff
@@ -2854,19 +2849,12 @@ private void Fix3PlusLines()
             for (int i = 0; i < Subtitle.Paragraphs.Count; i++)
             {
                 Paragraph p = Subtitle.Paragraphs[i];
-                string text = p.Text;
-
-                if (Utilities.CountTagInText(text, Environment.NewLine) > 1)
+                if (AllowFix(p, fixAction) && Utilities.CountTagInText(p.Text, Environment.NewLine) > 1)
```
ivandrofly: Check if AllowFix is <code>true</code> before doing any changes which may slow the performance

src/Forms/FixCommonErrors.cs
niksedk: I like string.Empty better - not important but String.Empty is used in SE

src/Forms/FixCommonErrors.cs
```diff
@@ -2207,27 +2207,27 @@ public static string FixStartWithUppercaseLetterAfterParagraph(Paragraph p, Para
 
             if (p.Text != null && p.Text.Contains(Environment.NewLine))
             {
-                string[] arr = p.Text.Replace("\r\n", "\n").Replace("\r", "\n").Split('\n');
+                var arr = p.Text.Split(Utilities.NewLineChars, StringSplitOptions.RemoveEmptyEntries);
```
niksedk: Not correct... we should probably make a extension method for string splitting...

ivandrofly: So should anything be changed in this patch?!

niksedk: Yes, this is buggy - the split does not work as it removes empty entries... a new extension method "SplitToLines" could be made so avoid duplicate code for line splitting

ivandrofly: Like:
<code>
        public static string[] SplitToLines(this string source)
        {
            return source.Replace("\r\n", "\n").Replace("\r", "\n").Split('\n');
        }
</code>

niksedk: :)

PR #473, opened 2015-01-29
Use SE exMethod to split lines part 2

ivandrofly:
I think there is no more: StringSplitOptions.RemoveEmptyEntries


niksedk:
he, that was a lot :)

ivandrofly:
:) yeah

PR #478, opened 2015-01-30
Remove unneeded spaces for '<font' tags + refact

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -3069,6 +3069,17 @@ public static string RemoveUnneededSpaces(string text, string language)
...
+            {
+                var idx = text.IndexOf("<font ", StringComparison.OrdinalIgnoreCase);
+                var endIdx = text.IndexOf('>', idx);
+                if (endIdx > idx && endIdx < text.Length)
+                {
+                    var color = text.Substring(idx, (endIdx - idx) + 1);
+                    text = RemoveSpaceBeforeAfterTag(text, color);
+                }
+            }
             text = text.Trim();
```
xylographe: Perhaps, it might be more efficient to combine `.Contains()` and `.IndexOf()`?
The `endIdx < text.Length` expression looks like an 'always true' condition?
```
{
    var idx = text.IndexOf("<font ", StringComparison.OrdinalIgnoreCase);
    if (idx >= 0)
    {
        var endIdx = text.IndexOf('>', idx);
        if (endIdx > idx) // or (endIdx > 0)
        {
            ...
        }
    }
}
```

src/Logic/Utilities.cs
```diff
@@ -3175,6 +3199,9 @@ private static string RemoveSpaceBeforeAfterTag(string text, string openTag)
                     break;
             }
 
+            if (closeTag == string.Empty && openTag.Contains("<font "))
+                closeTag = "</font>";
+
```
xylographe: IIRC, `closeTag.Length == 0` would be slightly more efficient than `closeTag == string.Empty`?

ivandrofly: :+1: => #480 @xylographe

PR #482, opened 2015-01-31
Minor corrections + Translation updates

xylographe:
[2] BatchConvertOptions moved because the Dutch FixCommonErrors text needed one more pixel. Else, it ‘disturbs’ the [Settings] button to the right of it.


niksedk:
Nice stuff :)

PR #487, opened 2015-02-02
New Unicode chars added [Normalize no Unicode Subtitle]

ivandrofly:


src/Forms/Main.cs
```diff
@@ -3045,11 +3045,31 @@ private DialogResult SaveSubtitle(SubtitleFormat format)
...
+                    allText = allText.Replace("\uFF1A", defColon); // ： Fullwidth Colon (\uFF1A)
+                    allText = allText.Replace("\uF313", defColon); // ︓ Presentation Form for Vertical Colon (\uF313)
+
+                    // Others
                     allText = allText.Replace("…", "...");
                     allText = allText.Replace("♪", "#");
                     allText = allText.Replace("♫", "#");
+
+                    // Spaces
+                    allText = allText.Replace("\u00A0", " "); // No-Break Space
```
ivandrofly: I'm confused for this one, should we keep it like this our replace it with empty string? /cc @niksedk

niksedk: I guess that the "zero width" spaces should be replaced with string.Empty, right?

ivandrofly:
Done. @niksedk

PR #490, opened 2015-02-04
Added new method to update fix status + minor Refact

ivandrofly:


src/Forms/FixCommonErrors.cs
```diff
@@ -2654,20 +2633,19 @@ private void DialogsOnOneLine()
             for (int i = 0; i < Subtitle.Paragraphs.Count; i++)
             {
                 Paragraph p = Subtitle.Paragraphs[i];
-                string oldText = p.Text;
-                var text = FixCommonErrorsHelper.FixDialogsOnOneLine(oldText, language);
```
niksedk: FixCommonErrorsHelper.FixDialogsOnOneLine(oldText, language) is actually much faster than the AllowFix...

ivandrofly: I think since its going to be run even after FixCommonErrorsHelper.FixDialogsOnOneLine(..., ...) why not run it first to check if it should proceed or not

niksedk: Ah, yes - perhaps the "if (AllowFix(p, fixAction) && oldText != text)" asked about oldText first?

ivandrofly: :+1:

PR #500, opened 2015-02-07
Various refact commits

xylographe:
[1] These Visible props are not set anywhere else.

[2] Same, but these are not the result of commit 1eddc4e. Lines commented out, not removed. Perhaps, you were planning to use them elsewhere later?


ivandrofly:
:+1:  /cc @niksedk

xylographe:
Rebased.

PR #510, opened 2015-02-09
UpdateLanguageFiles / LanguageDeserializer

xylographe:


ivandrofly:
Are you using Recharper/Codemaid or just doing cleaning manually?!

xylographe:
I 'bumped into' these by accident. For white space formatting I use a powershell script.

I think, the experiment with Codemaid wasn't satisfying; at least it seems Nikolaj is not using it at the moment. The problem with all code formatters (in my experience, at least) is, that they cannot recognize, that sometimes small format inconsistencies lead to improved legibility.

ivandrofly:
Oh, cool :+1: I'll start learning PS scripting soon :), I have played with Batch it's fun :)

PR #514, opened 2015-02-10
UpdateLanguageFiles / LanguageDeserializer

xylographe:


xylographe:
Adjusted and rebased.

PR #520, opened 2015-02-12
Increase efficiency with XmlReaderSettings (LanguageDeserializer)

xylographe:
According to [this test](https://gist.github.com/xylographe/1f8d736ff311365a8ef6), ignoring whitespace nodes in the reader would be _slightly_ more efficient than ignoring them in the while-loop. But, perhaps, not enough to justify the extra ‘clutter’?

### DIFF
--- a/src/Logic/LanguageDeserializer.cs
+++ b/src/Logic/LanguageDeserializer.cs
@@ -19,7 +19,9 @@ namespace Nikse.SubtitleEdit.Logic
             var stream = new FileStream(fileName, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
             var language = new Language();
 
-            using (XmlReader reader = XmlReader.Create(stream))
+            using (XmlReader reader = XmlReader.Create(stream, new XmlReaderSettings {
+                   IgnoreWhitespace = true, IgnoreProcessingInstructions = true, IgnoreComments = true,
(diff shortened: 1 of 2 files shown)


src/Logic/LanguageDeserializer.cs
```diff
@@ -19,7 +19,8 @@ public static Language CustomDeserializeLanguage(string fileName)
             var stream = new FileStream(fileName, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
             var language = new Language();
 
-            using (XmlReader reader = XmlReader.Create(stream))
+            using (XmlReader reader = XmlReader.Create(stream, new XmlReaderSettings() {
+                   IgnoreWhitespace = true, IgnoreProcessingInstructions = true, IgnoreComments = true, CloseInput = true }))
```
niksedk: Cool, this seems slightly faster: 
'using (XmlReader reader = XmlReader.Create(stream, new XmlReaderSettings { IgnoreWhitespace = true, IgnoreProcessingInstructions = true, IgnoreComments = true, DtdProcessing = DtdProcessing.Ignore, CheckCharacters = false }))'

xylographe: Thx, Nikolaj.
Updated.

PR #530, opened 2015-02-14
Bug fixed in Spellcheck + Minor refact

ivandrofly:
All ***http*** will automatically be changed to uppercase (***HTTP***).
This is Safe /cc @niksedk

### DIFF
--- a/src/Forms/SpellCheck.cs
+++ b/src/Forms/SpellCheck.cs
@@ -1009,7 +1009,6 @@ namespace Nikse.SubtitleEdit.Forms
 
             LoadDictionaries(dictionaryFolder, dictionary);
 
-
             _currentIndex = 0;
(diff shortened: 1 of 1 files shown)


src/Forms/SpellCheck.cs
```diff
@@ -696,22 +695,19 @@ private void PrepareNextWord()
                             }
                             else
                             {
-                                if (_currentWord.ToUpper() != "LT'S" && _currentWord.ToUpper() != "SOX'S") //TODO: get fixed nhunspell
+                                if (_currentWord.ToUpper() != "LT'S" && _currentWord.ToUpper() != "SOX'S" && !_currentWord.ToUpper().StartsWith("HTTP", StringComparison.Ordinal)) //TODO: get fixed nhunspell
```
ivandrofly: The problem was here

niksedk:
cool, nice catch!

PR #536, opened 2015-02-16
Fix/tags bold under line

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -505,7 +505,9 @@ public static string AutoBreakLineMoreThanTwoLines(string text, int maximumLineL
                         s = ReInsertHtmlTags(s, htmlTags);
                         s = s.Replace(" " + Environment.NewLine, Environment.NewLine);
                         s = s.Replace(Environment.NewLine + " ", Environment.NewLine);
-                        s = s.Replace(Environment.NewLine + "</i>", "<i>" + Environment.NewLine);
```
ivandrofly: Possible bug /cc @niksedk

niksedk:
nice catch :)

PR #542, opened 2015-02-17
Equals overrode inside Paragraph class

ivandrofly:
/cc @niksedk


src/Logic/Paragraph.cs
```diff
@@ -148,5 +148,12 @@ public double WordsPerMinute
                 return (60.0 / Duration.TotalSeconds) * wordCount;
             }
         }
+
+        public override bool Equals(object obj)
```
ivandrofly: Useful in future

niksedk:
Equals should use all properties...not really sure this is needed

PR #543, opened 2015-02-18
Removed always-false tests + other minor refact

xylographe:


src/Forms/SetMinimumDisplayTimeBetweenParagraphs.cs
```diff
@@ -72,8 +70,10 @@ public Subtitle FixedSubtitle
         public void Initialize(Subtitle subtitle)
         {
             _subtitle = subtitle;
-            numericUpDownMinMillisecondsBetweenLines.Value = Configuration.Settings.General.MininumMillisecondsBetweenLines;
-            //            GeneratePreview();
+            numericUpDownMinMillisecondsBetweenLines.Value = Configuration.Settings.General.MinimumMillisecondsBetweenLines != 0
+                                                           ? Configuration.Settings.General.MinimumMillisecondsBetweenLines
+                                                           : 1;
+            //GeneratePreview();
```
xylographe: Did throw ArgumentOutOfRangeException when `MinimumMillisecondsBetweenLines == 0`

src/Logic/Settings.cs
```diff
@@ -1316,9 +1316,11 @@ private static Settings CustomDeserialize(string fileName)
             subNode = node.SelectSingleNode("SubtitleMaximumDisplayMilliseconds");
             if (subNode != null)
                 settings.General.SubtitleMaximumDisplayMilliseconds = Convert.ToInt32(subNode.InnerText);
-            subNode = node.SelectSingleNode("MininumMillisecondsBetweenLines");
+            subNode = node.SelectSingleNode("MinimumMillisecondsBetweenLines");
+            if (subNode == null) // TODO: Remove in 3.5
+                subNode = node.SelectSingleNode("MininumMillisecondsBetweenLines");
```
xylographe: Backwards compatible

src/Logic/Settings.cs
```diff
@@ -2051,9 +2053,11 @@ private static Settings CustomDeserialize(string fileName)
             subNode = node.SelectSingleNode("SpectrogramAppearance");
             if (subNode != null)
                 settings.VideoControls.SpectrogramAppearance = subNode.InnerText;
-            subNode = node.SelectSingleNode("WaveformMininumSampleRate");
+            subNode = node.SelectSingleNode("WaveformMinimumSampleRate");
+            if (subNode == null) // TODO: Remove in 3.5
+                subNode = node.SelectSingleNode("WaveformMininumSampleRate");
```
xylographe: Backwards compatible

src/Logic/Settings.cs
niksedk: Probably better just to fix the language files...

xylographe: ???
I'm afraid, I don't understand what you mean.
These two 'compatibility lines' are there to be able to use the value from the existing Settings.xml file (with the tag named Mini**n**umMillisecondsBetweenLines).
When the Settings.xml file is re-generated, the new tag (Mini**m**umMillisecondsBetweenLines) will be used.

niksedk: ups - I thought it was language tags...

PR #545, opened 2015-02-18
Improved: FixEllipsesStartHelper

ivandrofly:
/cc @niksedk  the problem from https://github.com/SubtitleEdit/subtitleedit/commit/147c366025b74e386b60804fa547137a6cb9e113 should be fixed now :wink:


src/Logic/Forms/FixCommonErrorsHelper.cs
```diff
@@ -12,106 +12,47 @@ public static string FixEllipsesStartHelper(string text)
...
-                if (pre.Length < 2)
-                    return text;
-
-                text = text.Remove(0, index + 1).TrimStart();
+                text = text.Substring(index + 1).TrimStart();
                 text = FixEllipsesStartHelper(text);
-                if (pre.Length > 0)
-                    pre += " ";
             }
+            if (pre.Length > 0 && !" >\"".Contains(pre[pre.Length - 1]))
```
ivandrofly: The problem was here shouldn't be ```pre.Length > 1```

ivandrofly:
@niksedk I think this is safe now

PR #544, opened 2015-02-18
Refact only

ivandrofly:
/cc @niksedk

PR #546, opened 2015-02-18
Refact only

ivandrofly:
Re-pull from https://github.com/SubtitleEdit/subtitleedit/pull/544 + Rebased /cc @niksedk

### DIFF
--- a/src/Forms/Main.cs
+++ b/src/Forms/Main.cs
@@ -6190,12 +6190,13 @@ namespace Nikse.SubtitleEdit.Forms
         {
             var format = GetCurrentSubtitleFormat();
             bool useExtraForStyle = format.HasStyleSupport;
+            var formatType = format.GetType();
             var styles = new List<string>();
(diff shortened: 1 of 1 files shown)


src/Forms/Main.cs
```diff
@@ -6625,11 +6618,7 @@ private void ButtonAutoBreakClick(object sender, EventArgs e)
 
         private static void FixVerticalScrollBars(TextBox tb)
         {
-            var lineCount = Utilities.CountTagInText(tb.Text, Environment.NewLine) + 1;
-            if (lineCount > 3)
-                tb.ScrollBars = ScrollBars.Vertical;
-            else
-                tb.ScrollBars = ScrollBars.None;
+            tb.ScrollBars = Utilities.GetNumberOfLines(tb.Text) + 1 > 3 ? ScrollBars.Vertical : ScrollBars.None;
```
ivandrofly: Changed :wink:

niksedk: Are you sure the "+ 1" should be there?

ivandrofly: Btw I found it strange too but didn't what do change it so...

PR #553, opened 2015-02-20
Fix/patch 0

ivandrofly:


niksedk:
"l" is okay and "fLen" is worse than fi in my opinion...

PR #556, opened 2015-02-20
Fix/patch 0

ivandrofly:
Repull from https://github.com/SubtitleEdit/subtitleedit/pull/553 /cc @niksedk


niksedk:
Added via 77de6562c2de4b5ebfd12c6728e64d9c538c8ac1

PR #557, opened 2015-02-20
Fix/start with upper after prgph

ivandrofly:


src/Logic/StripableText.cs
```diff
@@ -61,7 +65,7 @@ public StripableText(string text, string stripStartCharacters, string stripEndCh
             }
 
             Post = string.Empty;
-            if (text.Length > 0 && !Utilities.AllLettersAndNumbers.Contains(text[text.Length - 1]))
+            if (stripEndCharacters != null && text.Length > 0 && !Utilities.AllLettersAndNumbers.Contains(text[text.Length - 1]))
```
ivandrofly: Most of time ``Post`` isn't needed so better skip it

PR #558, opened 2015-02-21
Refact in UpdateAssemblyInfo

ivandrofly:


src/UpdateAssemblyInfo/Program.cs
```diff
@@ -103,7 +103,7 @@ private static int Main(string[] args)
         private static string GetGitPath()
         {
             var envPath = Environment.GetEnvironmentVariable("PATH");
-            if (!string.IsNullOrWhiteSpace(envPath))
+            if (!string.IsNullOrWhiteSpace(envPath) && envPath.IndexOf("git.exe", StringComparison.Ordinal) > 0)
```
xylographe: PATH contains directory paths separated by `';'` (Windows)  or `':'` (Posix).

ivandrofly: yeah but instead of looping looking for git.ext why not checking in path first

xylographe: Like I said, PATH contains *directory* paths. PATH does not contain *file* paths.

ivandrofly: Ah... okay :+1:

src/UpdateAssemblyInfo/Program.cs
```diff
@@ -113,7 +112,7 @@ private static string GetGitPath()
                 }
             }
 
-            var gitPath = Path.Combine("Git", "bin", "git.exe");
+            var gitPath = "Git\\bin\\git.exe";
```
niksedk: I think it's just nicer to always use the platform independent Path.Combine method

PR #560, opened 2015-02-22
Fixed: Always true + Empty lines removed

ivandrofly:
/cc @niksedk

### DIFF
--- a/src/Logic/StripableText.cs
+++ b/src/Logic/StripableText.cs
@@ -42,7 +42,7 @@ namespace Nikse.SubtitleEdit.Logic
                     if (text.StartsWith("{\\", StringComparison.Ordinal))
                     {
                         int endIndex = text.IndexOf('}');
-                        if (endIndex > 0 && (!text.Contains('{') || text.IndexOf('{') > endIndex))
+                        if (endIndex > 0 && (!text.Substring(1).Contains('{') || text.IndexOf('{') > endIndex))
                         {
                             int index = text.IndexOf('}') + 1;
                             Pre += text.Substring(0, index);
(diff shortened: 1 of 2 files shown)


src/Logic/StripableText.cs
```diff
@@ -42,7 +42,7 @@ public StripableText(string text, string stripStartCharacters, string stripEndCh
                     if (text.StartsWith("{\\", StringComparison.Ordinal))
                     {
                         int endIndex = text.IndexOf('}');
-                        if (endIndex > 0 && (!text.Contains('{') || text.IndexOf('{') > endIndex))
+                        if (endIndex > 0 && (!text.Substring(1).Contains('{') || text.IndexOf('{') > endIndex))
```
xylographe: Or, perhaps, slightly more efficient,
```
if (text.StartsWith("{\\", StringComparison.Ordinal))
{
    int endIndex = text.IndexOf('}', 2);
    if (endIndex > 0)
    {
        int index = text.IndexOf('{', 2);
        if (index < 0 || index > endIndex)
        {
            Pre += text.Substring(0, ++endIndex);
            text = text.Remove(0, endIndex);
        }
    }
}
```

ivandrofly: yeah but I always try to avoid the nested ```if```

xylographe: This one is also correct,
```
if (text.StartsWith("{\\", StringComparison.Ordinal))
{
    int endIndex = text.IndexOf('}', 2);
    if (endIndex > 0 && text.LastIndexOf('{', endIndex - 1) == 0)
    {
        Pre += text.Substring(0, ++endIndex);
        text = text.Remove(0, endIndex);
    }
}
```

niksedk:
Committed in 6a48fa74cf09ff3e8c9a82d2e088e4f72c0bc3e

src/Logic/StripableText.cs
ivandrofly: :+1:

PR #566, opened 2015-02-23
Empty lines removed

ivandrofly:


niksedk:
I do like many of the empty lines... double new lines after a method is okay to remove, but I do like the rest...

PR #567, opened 2015-02-23
Refact Only

ivandrofly:


src/Core/HtmlUtil.cs
```diff
@@ -356,13 +356,10 @@ public static string EncodeNumeric(string source)
 
         public static string RemoveHtmlTags(string s)
         {
-            if (s == null)
-                return null;
-
-            if (s.Length < 3 || s.IndexOf('<') < 0)
+            if (s == null || s.Length < 3 || s.IndexOf('<') < 0)
                 return s;
```
ivandrofly: <code>if s == null</code> then null will still be returned

src/Core/HtmlUtil.cs
```diff
@@ -356,13 +356,10 @@ public static string EncodeNumeric(string source)
...
 
-            if (s.IndexOf("< ", StringComparison.Ordinal) >= 0)
+            if (s.IndexOf("< ", StringComparison.Ordinal) >= 0 || s.IndexOf(" >", StringComparison.Ordinal) >= 2)
```
ivandrofly: could be something like: ```<i >

src/Core/HtmlUtil.cs
```diff
@@ -374,26 +371,24 @@ public static string RemoveHtmlTags(string s, bool alsoSsaTags)
                 return null;
 
             s = RemoveHtmlTags(s);
-
-            if (alsoSsaTags)
-                s = Utilities.RemoveSsaTags(s);
-
-            return s;
+            return alsoSsaTags ? Utilities.RemoveSsaTags(s) : s;
```
ivandrofly: Shorter

src/Core/HtmlUtil.cs
```diff
@@ -374,26 +371,24 @@ public static string RemoveHtmlTags(string s, bool alsoSsaTags)
...
         }
 
         public static bool IsUrl(string text)
         {
-            if (string.IsNullOrWhiteSpace(text) || text.Length < 6 || !text.Contains(".") || text.Contains(" "))
+            if (string.IsNullOrWhiteSpace(text) || text.Length < 6 || !text.Contains('.') || text.Contains(' '))
```
ivandrofly: Use ```char``` instead

src/Core/HtmlUtil.cs
```diff
@@ -374,26 +371,24 @@ public static string RemoveHtmlTags(string s, bool alsoSsaTags)
...
                 return false;
 
             var allLower = text.ToLower();
-            if (allLower.StartsWith("http://") || allLower.StartsWith("https://") || allLower.StartsWith("www.") || allLower.EndsWith(".org") || allLower.EndsWith(".com") || allLower.EndsWith(".net"))
-                return true;
-
-            if (allLower.Contains(".org/") || allLower.Contains(".com/") || allLower.Contains(".net/"))
+            if (allLower.StartsWith("http://", StringComparison.Ordinal) ||
```
ivandrofly: Check with ```Ordinal```

src/Core/HtmlUtil.cs
```diff
@@ -374,26 +371,24 @@ public static string RemoveHtmlTags(string s, bool alsoSsaTags)
...
+                allLower.StartsWith("https://", StringComparison.Ordinal) ||
+                allLower.StartsWith("www.", StringComparison.Ordinal) ||
+                allLower.EndsWith(".org", StringComparison.Ordinal) ||
+                allLower.EndsWith(".com", StringComparison.Ordinal) ||
+                allLower.EndsWith(".net", StringComparison.Ordinal))
                 return true;
 
-            return false;
+            return allLower.Contains(".org/", StringComparison.Ordinal) || allLower.Contains(".com/", StringComparison.Ordinal) || allLower.Contains(".net/", StringComparison.Ordinal);
```
ivandrofly: Shorter

src/Core/HtmlUtil.cs
xylographe: `text.ToLower(CultureInfo.InvariantCulture)` might be slightly more efficient too. (not tested)

PR #568, opened 2015-02-23
Possible bug fixed + Minor refact

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -3211,7 +3210,7 @@ private static string RemoveSpaceBeforeAfterTag(string text, string openTag)
                     break;
             }
 
-            if (closeTag.Length == 0 && openTag.Contains("<font ", StringComparison.Ordinal))
+            if (closeTag.Length == 0 && openTag.Contains("<font", StringComparison.Ordinal))
```
ivandrofly: Could be ```<font>```

src/Logic/Utilities.cs
```diff
@@ -3185,18 +3185,17 @@ private static string RemoveSpaceBeforeAfterTag(string text, string openTag)
                 .Replace("</U>", "</u>")
                 .Replace("</FONT>", "</font>");
 
-            if (text.Contains("<FONT", StringComparison.Ordinal))
```
ivandrofly: IndexOf is already checking

src/Forms/DurationsBridgeGaps.cs
```diff
@@ -145,6 +145,5 @@ private void DurationsBridgeGaps_KeyDown(object sender, KeyEventArgs e)
             if (e.KeyCode == Keys.Escape)
                 DialogResult = DialogResult.Cancel;
         }
-
```
niksedk: I like a space here... makes it easy to add new methods...

ivandrofly: :+1:

PR #569, opened 2015-02-24
Refact

xylographe:
[2] Loop with arbitrary number of iterations replaced by loop with end-condition.

### DIFF
--- a/src/Forms/MultipleReplace.cs
+++ b/src/Forms/MultipleReplace.cs
@@ -193,11 +193,7 @@ namespace Nikse.SubtitleEdit.Forms
                             int index = newText.IndexOf(findWhat, StringComparison.OrdinalIgnoreCase);
                             while (index >= 0)
                             {
-                                if (index < newText.Length)
-                                    newText = newText.Substring(0, index) + replaceWith + newText.Substring(index + findWhat.Length);
-                                else
-                                    newText = newText.Substring(0, index) + replaceWith;
-
+                                newText = newText.Substring(0, index) + replaceWith + newText.Remove(0, index);
                                 hit = true;
                                 index = newText.IndexOf(findWhat, index + replaceWith.Length, StringComparison.OrdinalIgnoreCase);
                             }
--- a/src/Logic/StripableText.cs
+++ b/src/Logic/StripableText.cs
@@ -30,8 +30,11 @@ namespace Nikse.SubtitleEdit.Logic
             Pre = string.Empty;
             if (text.Length > 0 && !Utilities.AllLettersAndNumbers.Contains(text[0]))
             {
-                for (int i = 0; i < 5; i++)
+                int beginLength;
+                do
                 {
+                    beginLength = text.Length;
+
                     while (text.Length > 0 && stripStartCharacters.Contains(text[0]))
                     {
                         Pre += text[0];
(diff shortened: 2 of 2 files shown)


src/Forms/MultipleReplace.cs
```diff
@@ -193,11 +193,7 @@ private void GeneratePreview()
                             int index = newText.IndexOf(findWhat, StringComparison.OrdinalIgnoreCase);
                             while (index >= 0)
                             {
-                                if (index < newText.Length)
-                                    newText = newText.Substring(0, index) + replaceWith + newText.Substring(index + findWhat.Length);
-                                else
-                                    newText = newText.Substring(0, index) + replaceWith;
-
+                                newText = newText.Substring(0, index) + replaceWith + newText.Substring(index + findWhat.Length);
```
xylographe: I think, the intention was `if (index + findWhat.Length < newText.Length)`. But, that isn't necessary, because the potential maximum value is `newText.Length`.

src/Logic/StripableText.cs
```diff
@@ -60,20 +63,24 @@ public StripableText(string text, string stripStartCharacters, string stripEndCh
...
+                    beginLength = text.Length;
+
                     while (text.Length > 0 && stripEndCharacters.Contains(text[text.Length - 1]))
                     {
                         Post = text[text.Length - 1] + Post;
                         text = text.Substring(0, text.Length - 1);
                     }
 
-                    if (text.EndsWith('>'))
+                    if (text.Length >= 4 && text[text.Length - 1] == '>')
```
xylographe: Not faster than EndsWith(). But, requiring `Length >= 4` eliminates text comparisons that are bound to fail.

src/Forms/MultipleReplace.cs
```diff
@@ -193,11 +193,7 @@ private void GeneratePreview()
                             int index = newText.IndexOf(findWhat, StringComparison.OrdinalIgnoreCase);
                             while (index >= 0)
                             {
-                                if (index < newText.Length)
-                                    newText = newText.Substring(0, index) + replaceWith + newText.Substring(index + findWhat.Length);
-                                else
-                                    newText = newText.Substring(0, index) + replaceWith;
-
+                                newText = newText.Substring(0, index) + replaceWith + newText.Remove(0, index);
```
xylographe: I think, the intention was `if (index + findWhat.Length < newText.Length)`, which would also have been unnecessary.

xylographe:
Sorry, for the rubbish. First I'd forgotten that Remove() is faster than Substring(),
then I'd—apparently—forgotten how to cut&paste. :unamused:

Moved to PR #570

src/Logic/StripableText.cs
ivandrofly: I think we should be careful using Remove over Substring
http://www.dotnetperls.com/remove

xylographe: As usual I didn't get conclusive results ([from this test](https://gist.github.com/xylographe/561085cdea02c90ed0b5)).

ivandrofly: :+1: , You can also check this post in [stackoverflow](http://stackoverflow.com/questions/3222125/fastest-way-to-remove-first-char-in-a-string)

PR #570, opened 2015-02-24
Refact

xylographe:
[2] Loop with arbitrary number of iterations replaced by loop with end-condition.

### DIFF
--- a/src/Forms/MultipleReplace.cs
+++ b/src/Forms/MultipleReplace.cs
@@ -193,11 +193,7 @@ namespace Nikse.SubtitleEdit.Forms
                             int index = newText.IndexOf(findWhat, StringComparison.OrdinalIgnoreCase);
                             while (index >= 0)
                             {
-                                if (index < newText.Length)
-                                    newText = newText.Substring(0, index) + replaceWith + newText.Substring(index + findWhat.Length);
-                                else
-                                    newText = newText.Substring(0, index) + replaceWith;
-
+                                newText = newText.Substring(0, index) + replaceWith + newText.Remove(0, index + findWhat.Length);
                                 hit = true;
                                 index = newText.IndexOf(findWhat, index + replaceWith.Length, StringComparison.OrdinalIgnoreCase);
                             }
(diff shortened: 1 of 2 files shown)


src/Forms/MultipleReplace.cs
```diff
@@ -193,11 +193,7 @@ private void GeneratePreview()
                             int index = newText.IndexOf(findWhat, StringComparison.OrdinalIgnoreCase);
                             while (index >= 0)
                             {
-                                if (index < newText.Length)
-                                    newText = newText.Substring(0, index) + replaceWith + newText.Substring(index + findWhat.Length);
-                                else
-                                    newText = newText.Substring(0, index) + replaceWith;
-
+                                newText = newText.Substring(0, index) + replaceWith + newText.Remove(0, index + findWhat.Length);
```
xylographe: I think, the intention was `if (index + findWhat.Length < newText.Length)`. But, this shouldn't be necessary, as the potential maximum value equals `newText.Length`.

src/Logic/StripableText.cs
```diff
@@ -60,20 +63,24 @@ public StripableText(string text, string stripStartCharacters, string stripEndCh
...
+                    beginLength = text.Length;
+
                     while (text.Length > 0 && stripEndCharacters.Contains(text[text.Length - 1]))
                     {
                         Post = text[text.Length - 1] + Post;
                         text = text.Substring(0, text.Length - 1);
                     }
 
-                    if (text.EndsWith('>'))
+                    if (text.Length >= 4 && text[text.Length - 1] == '>')
```
xylographe: Not faster than EndsWith(). But, requiring `Length >= 4` eliminates text comparisons that are bound to fail.

xylographe: Actually, this was a bad idea! *Reverted*.

src/Forms/MultipleReplace.cs
niksedk: :)

PR #573, opened 2015-02-25
AutoBreakUnbreakLines optimized 10x faster

ivandrofly:
10x faster ;) /cc @niksedk, @xylographe

### DIFF
--- a/src/Forms/AutoBreakUnbreakLines.cs
+++ b/src/Forms/AutoBreakUnbreakLines.cs
@@ -216,7 +216,10 @@ namespace Nikse.SubtitleEdit.Forms
             {
                 Paragraph p = _paragraphs[i];
                 if (!IsFixAllowed(p))
+                {
                     _paragraphs.Remove(p);
+                    i--;
+                }
             }
             DialogResult = DialogResult.OK;
         }
--- a/src/Forms/AutoBreakUnbreakLines.cs
+++ b/src/Forms/AutoBreakUnbreakLines.cs
@@ -181,17 +181,38 @@ namespace Nikse.SubtitleEdit.Forms
 
         private bool IsFixAllowed(Paragraph p)
         {
-            foreach (ListViewItem item in listViewFixes.Items)
+            var count = listViewFixes.Items.Count;
+            int index = (count / 2) - 1;
+            index = index >= 0 ? index : 0;
+
+            var pr = listViewFixes.Items[index].Tag as Paragraph;
+            if (pr.ID == p.ID)
+                return listViewFixes.Items[index].Checked;
+
+            if (pr.Number > p.Number)
             {
-                if (item.Tag.ToString() == p.ToString())
-                    return item.Checked;
+                for (int i = 0; i < index; i++)
+                {
+                    var pp = listViewFixes.Items[i].Tag as Paragraph;
+                    if (pp.ID == p.ID)
+                        return listViewFixes.Items[i].Checked;
+                }
+            }
+            else
+            {
+                for (int i = index; i < count; i++)
+                {
+                    var pp = listViewFixes.Items[i].Tag as Paragraph;
+                    if (pp.ID == p.ID)
+                        return listViewFixes.Items[i].Checked;
+                }
             }
             return false;
         }
 
         private void ButtonOkClick(object sender, EventArgs e)
         {
-            for (int i = _paragraphs.Count - 1; i > 0; i--)
+            for (int i = 0; i < _paragraphs.Count; i++)
             {
                 Paragraph p = _paragraphs[i];
                 if (!IsFixAllowed(p))


src/Forms/AutoBreakUnbreakLines.cs
```diff
@@ -181,17 +181,38 @@ private void AddToListView(Paragraph p, string newText)
...
+                        return listViewFixes.Items[i].Checked;
+                }
             }
             return false;
         }
 
         private void ButtonOkClick(object sender, EventArgs e)
         {
-            for (int i = _paragraphs.Count - 1; i > 0; i--)
+            for (int i = 0; i < _paragraphs.Count; i++)
```
xylographe: This must be from end to begin, because items may be removed from the list (`_paragraphs.Remove()`).
I'm not sure why `_paragraphs[0]` is skipped; special marker, perhaps?

src/Forms/AutoBreakUnbreakLines.cs
```diff
@@ -181,17 +181,38 @@ private void AddToListView(Paragraph p, string newText)
...
+                }
+            }
+            else
+            {
+                for (int i = index; i < count; i++)
+                {
+                    var pp = listViewFixes.Items[i].Tag as Paragraph;
+                    if (pp.ID == p.ID)
+                        return listViewFixes.Items[i].Checked;
+                }
```
xylographe: Perhaps, a full binary search? Pseudo-code:
```
    first = 0
    last = Count - 1
    while (first <= last)
    {
        middle = (first + last) / 2
        if (ID(middle) == ID(wanted))
            return ...
        if (Number(middle) < Number(wanted))
            first = middle + 1
        else
            last = middle - 1
    }
    return false
```

src/Forms/AutoBreakUnbreakLines.cs
ivandrofly: yeah, and it's still not safe:
``` c#
            var removeList = new List<string>();
            for (int i = _paragraphs.Count - 1; i > 0; i--)
            //for (int i = 0; i < _paragraphs.Count; i++)
            {
                Paragraph p = _paragraphs[i];
                if (!IsFixAllowed(p))
                {
                    removeList.Add(p.ID);
                }
            }
            _paragraphs.RemoveAll(x => removeList.Contains(x.ID));
```

xylographe: I think, there are two possible optimizations,
`for (int i = _paragraphs.Count - 1; i > 0; i--)` ==> `for (var i = _paragraphs.Count; --i > 0; )`
and,
`_paragraphs.Remove(p)` ==> `_paragraphs.RemoveAt(i)`

then, of course, you wouldn't need the intermediate var,
`if (!IsFixAllowed(p))` ==> `if (!IsFixAllowed(_paragraphs[i]))`

ivandrofly: Note if item is removed previous will be moved to index of ```i``` and if you --i it will point the unwanted one

xylographe: Not if you work from end to begin. You remove items from the end, the items[0..(i-1)] will not be effected. Only items[(i+1)..(Count -1)] \(if any) will shift, and Count will decrease.

ivandrofly: yeah, starting from back should be better :+1:

src/Forms/AutoBreakUnbreakLines.cs
ivandrofly: @xylographe could you be more explicit here

xylographe: Assuming, of course, that listViewFixes.Items is sorted by Paragraph Number,
```
        private bool IsFixAllowed(Paragraph p)
        {
            var lastIndex = listViewFixes.Items.Count - 1;
            var firstIndex = 0;
            while (firstIndex <= lastIndex)
            {
                var index = (firstIndex + lastIndex) >> 1;
                var item = listViewFixes.Items[index];
                if (((Paragraph)item.Tag).ID == p.ID)
                    return item.Checked;
                if (((Paragraph)item.Tag).Number < p.Number)
                    firstIndex = index + 1;
                else
                    lastIndex = index - 1;
            }
            return false;
        }
```

ivandrofly: > private bool IsFixAllowed(Paragraph p)
>         {
>             var lastIndex = listViewFixes.Items.Count - 1;
>             var firstIndex = 0;
>             while (firstIndex <= lastIndex)
>             {
>                 var index = (firstIndex + lastIndex) >> 1;
>                 var item = listViewFixes.Items[index];
>                 if (((Paragraph)item.Tag).ID == p.ID)
>                     return item.Checked;
>                 if (((Paragraph)item.Tag).Number < p.Number)
>                     firstIndex = index + 1;
>                 else
>                     lastIndex = index - 1;
>             }
>             return false;
>         }

:clap:

ivandrofly: I think I found faster way:
We need to have another Paragraph list and _paragraphs2 which will be initialized and the _paragraphe list will serve only to add checked items from list view
``` C#
            _paragraphs = new List<Paragraph>();
            foreach (ListViewItem item in this.listViewFixes.Items)
            {
                if (item.Checked)
                    _paragraphs.Add((Paragraph)item.Tag);
            }
```

xylographe: :+1: 
The second list  (_paragraphs2) would not be necessary, if you put that loop in ButtonOk_Click().

PR #575, opened 2015-02-25
Minor (refact | translation | formatting)

xylographe:


xylographe:
In Forms/MultipleReplace.buttonCancel_Click() the Configuration.Settings.MultipleSearchAndReplaceList is restored from _oldMultipleSearchAndReplaceList, before returning DialogResult.Cancel. In MultipleReplace_KeyDown(), when Escape is pressed, the list is not restored.

Is this intentional, or should they both behave the same?

PR #579, opened 2015-02-26
AutoBreakUnbreakLines optimized 10x faster

ivandrofly:


src/Forms/AutoBreakUnbreakLines.cs
```diff
@@ -53,14 +53,10 @@ private void FixLargeFonts()
...
-        public void Initialize(Subtitle subtitle, bool autoBalance)
+        public void Initialize(List<Paragraph> paragraphs, bool autoBalance)
         {
             _modeAutoBalance = autoBalance;
-            _paragraphs = new List<Paragraph>();
-
-            foreach (Paragraph p in subtitle.Paragraphs)
-                _paragraphs.Add(p);
-
+            _paragraphs = paragraphs;
```
xylographe: As `Iinitialize()` is a *public* method, it might be better to not change it's signature in that way.
`_paragraphs = subtiltle.Paragraphs` is fast enough not to need optimisation. :smile:

ivandrofly: :+1: Fixed => https://github.com/SubtitleEdit/subtitleedit/pull/581

PR #589, opened 2015-02-28
Bugfix Part2

ivandrofly:
/cc @niksedk

### DIFF
--- a/src/Logic/Forms/RemoveTextForHI.cs
+++ b/src/Logic/Forms/RemoveTextForHI.cs
@@ -153,7 +153,7 @@ namespace Nikse.SubtitleEdit.Logic.Forms
 
                             if (remove && Settings.ColonSeparateLine)
                             {
-                                if (indexOfColon == line.Length - 1 || line.Substring(indexOfColon + 1).StartsWith(Environment.NewLine, StringComparison.Ordinal))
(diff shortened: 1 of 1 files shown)


src/Logic/Forms/RemoveTextForHI.cs
```diff
@@ -220,11 +220,8 @@ public string RemoveColon(string text)
                                     if (!Settings.RemoveTextBeforeColonOnlyUppercase || start == start.ToUpper())
                                     {
                                         int endIndex = start.LastIndexOfAny(new[] { '.', '!', '?' });
-                                        if (colonIndex > 0 && colonIndex < s2.Length - 1)
```
ivandrofly: colonIndex is always > 0

src/Logic/Forms/RemoveTextForHI.cs
```diff
@@ -153,7 +153,7 @@ public string RemoveColon(string text)
 
                             if (remove && Settings.ColonSeparateLine)
                             {
-                                if (indexOfColon == line.Length - 1 || line.Substring(indexOfColon + 1).StartsWith(Environment.NewLine, StringComparison.Ordinal))
```
ivandrofly: Will never starts with NewLine

PR #591, opened 2015-02-28
Use contains instead

ivandrofly:


niksedk:
Seems to be much slower in my tests...

PR #598, opened 2015-03-01
Keep compareForm open (Forms/Showhistory)

xylographe:


src/Forms/ShowHistory.cs
```diff
@@ -113,7 +113,7 @@ private void ButtonCompareClick(object sender, EventArgs e)
                 using (var compareForm = new Compare())
                 {
                     compareForm.Initialize(_subtitle, Configuration.Settings.Language.General.CurrentSubtitle, h2.Subtitle, descr2);
-                    compareForm.Show();
+                    compareForm.ShowDialog();
```
niksedk: I prefer compareForm.ShowDialog(this); (windows will be linked... works better if something is minimized)

xylographe:
Updated (ShowDialog(*this*)).

niksedk:
Good bug fix :)

src/Forms/ShowHistory.cs
xylographe: After re-reading docu I experimented a bit. Makes quite a difference.
Thank you, for the tip.

PR #600, opened 2015-03-01
zlib udpated to v1.3

ivandrofly:


src/SubtitleEdit.csproj
```diff
@@ -65,9 +65,9 @@
     <Reference Include="System.Web.Services" />
     <Reference Include="System.Windows.Forms" />
     <Reference Include="System.Xml" />
-    <Reference Include="zlib.net">
-      <HintPath>packages\zlib.net.1.0.4.0\lib\zlib.net.dll</HintPath>
-      <Private>False</Private>
+    <Reference Include="zlibnet, Version=1.3.0.0, Culture=neutral, processorArchitecture=MSIL">
```
ivandrofly: I think the problem is in this line

ivandrofly:
Ah okay we are getting error 'cause the dll file isn't include in PR, @niksedk should I send it to you or...?!

src/SubtitleEdit.csproj
```diff
@@ -65,9 +65,9 @@
     <Reference Include="System.Web.Services" />
     <Reference Include="System.Windows.Forms" />
     <Reference Include="System.Xml" />
-    <Reference Include="zlib.net">
-      <HintPath>packages\zlib.net.1.0.4.0\lib\zlib.net.dll</HintPath>
-      <Private>False</Private>
+    <Reference Include="zlibnet, Version=1.3.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>packages\zlibnet.dll</HintPath>
```
xylographe: HintPath should probably look something like `packages\zlib.net.1.3.0.0\lib\zlib.net.dll`.
And you might need a `zlib.net.1.3.0.0.nupkg`, or similar.

xylographe: Most likely, you'll need to chance `src\packages.config` to tell NuGet to 'get' something different.

ivandrofly: yeah but event if we "tell" NuGet to get from that location it still won't work 'cuase src/package is ignored in .gitignore

xylographe: NuGet downloads at compile time. Try it: remove src\packages\zlib.net.1.0.4.0, and run build.bat.
One of its messages will be "Installing 'zlib.net 1.0.4.0'.". That's why these packages are not controlled by git.

PR #602, opened 2015-03-01
Refact + Optimization

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -2963,7 +2965,7 @@ internal static string RemoveNonNumbers(string p)
             var sb = new StringBuilder();
             foreach (var c in p)
             {
-                if (char.IsDigit(c))
+                if (c >= 0x30 && c <= 0x39)
```
ivandrofly: TEST=> http://pastebin.com/737wU88U

PR #603, opened 2015-03-02
Avoid potential exception (Forms/Compare)

xylographe:


xylographe:
2nd commit: I think, the intention was to set fgcolor. At least, it looks better…

src/Forms/Compare.cs
```diff
@@ -577,7 +585,7 @@ private void ShowTextDifference()
                 {
                     richTextBox1.SelectionStart = i;
                     richTextBox1.SelectionLength = 1;
-                    richTextBox1.SelectionBackColor = Color.Red;
+                    richTextBox1.SelectionColor = Color.Red;
```
ivandrofly: I aelectionColor isn't going to have same meaning if text if: ' ' ...  as SelectionBackColor does

PR #613, opened 2015-03-04
Added SwapSides button (Forms/Compare)

xylographe:
Swaps left-right sides of the subtitle compare form.


ivandrofly:
Nice feat :+1:

src/Forms/Compare.cs
```diff
@@ -664,19 +665,31 @@ private void SubtitleListView2SelectedIndexChanged(object sender, EventArgs e)
 
         private void Compare_Resize(object sender, EventArgs e)
         {
-            subtitleListView1.Width = (Width / 2) - 20;
+            SuspendLayout();
 
-            subtitleListView2.Left = (Width / 2) - 3;
+            subtitleListView1.Width = (Width / 2) - 20;
```
ivandrofly: You could have store (Width /2) - 20 and just assign on both Width

PR #616, opened 2015-03-06
Fixed Bug in Remove Text From Hearing Impaired

ivandrofly:
/cc @niksedk

### DIFF
--- a/src/Forms/RemoveTextFromHearImpaired.cs
+++ b/src/Forms/RemoveTextFromHearImpaired.cs
@@ -158,6 +158,7 @@ namespace Nikse.SubtitleEdit.Forms
 
         public int RemoveTextFromHearImpaired()
         {
+            _removeTextForHILib.Settings = GetSettings();
             int count = 0;
             for (int i = listViewFixes.Items.Count - 1; i >= 0; i--)
             {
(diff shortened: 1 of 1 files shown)


src/Forms/RemoveTextFromHearImpaired.cs
```diff
@@ -165,11 +166,10 @@ public int RemoveTextFromHearImpaired()
                 if (item.Checked)
                 {
                     Paragraph p = (Paragraph)item.Tag;
-                    _removeTextForHILib.Settings = GetSettings();
                     string newText = _removeTextForHILib.RemoveTextFromHearImpaired(p.Text);
                     if (string.IsNullOrEmpty(newText))
                     {
-                        _subtitle.Paragraphs.RemoveAt(i);
```
ivandrofly: We are not looping through Paragraph anymore

src/Forms/RemoveTextFromHearImpaired.cs
```diff
@@ -165,11 +166,10 @@ public int RemoveTextFromHearImpaired()
                 if (item.Checked)
                 {
                     Paragraph p = (Paragraph)item.Tag;
-                    _removeTextForHILib.Settings = GetSettings();
```
ivandrofly: Setting is being loaded inside loop

PR #619, opened 2015-03-06
Bug Fixed Remove Text from Hearing Impaired

ivandrofly:
Bug fixed in Remove text from hearing impaired

### DIFF
--- a/src/Forms/RemoveTextFromHearImpaired.cs
+++ b/src/Forms/RemoveTextFromHearImpaired.cs
@@ -158,7 +158,6 @@ namespace Nikse.SubtitleEdit.Forms
 
         public int RemoveTextFromHearImpaired()
         {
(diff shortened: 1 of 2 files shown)


src/Forms/RemoveTextFromHearImpaired.cs
```diff
@@ -165,11 +165,10 @@ public int RemoveTextFromHearImpaired()
                 if (item.Checked)
                 {
                     Paragraph p = (Paragraph)item.Tag;
-                    _removeTextForHILib.Settings = GetSettings();
-                    string newText = _removeTextForHILib.RemoveTextFromHearImpaired(p.Text);
+                    string newText = columnHeaderAfter.Text;
```
ivandrofly: Just use the the fixed text from listView instead of recalling ```_removeTextForHILib.RemoveTextFromHearImpaired(p.Text);```

PR #620, opened 2015-03-06
Fixed problem from previous patch ;) #619

ivandrofly:
Here @niksedk, sorry for that ;)

### DIFF
--- a/src/Forms/RemoveTextFromHearImpaired.cs
+++ b/src/Forms/RemoveTextFromHearImpaired.cs
@@ -165,7 +165,7 @@ namespace Nikse.SubtitleEdit.Forms
                 if (item.Checked)
                 {
                     Paragraph p = (Paragraph)item.Tag;
-                    string newText = columnHeaderAfter.Text;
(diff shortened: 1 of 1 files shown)


ivandrofly:
AppVeyor fails 'cause patch didn't finished building up

src/Forms/RemoveTextFromHearImpaired.cs
```diff
@@ -165,7 +165,7 @@ public int RemoveTextFromHearImpaired()
                 if (item.Checked)
                 {
                     Paragraph p = (Paragraph)item.Tag;
-                    string newText = columnHeaderAfter.Text;
+                    string newText = item.SubItems[3].Text.Replace(Configuration.Settings.General.ListViewLineSeparatorString, Environment.NewLine);
```
niksedk: Very close :)
We're not a 100% sure what the ListViewLineSeparator is... could be space...

ivandrofly: yeah... I think neither "<br />" is the right solution

ivandrofly:
I think something can be done...

PR #624, opened 2015-03-08
Refact only | Optimization

ivandrofly:


src/Logic/CommandLineConvert.cs
```diff
@@ -610,10 +614,13 @@ internal static bool BatchConvertSave(string toFormat, string offset, Encoding t
                                         s += "_" + className + format.Extension;
                                     outputFileName = FormatOutputFileNameForBatchConvert(s, sf.Extension, outputFolder, overwrite);
                                     File.WriteAllText(outputFileName, newSub.ToText(sf), targetEncoding);
+                                    exitLoop = true;
                                 }
                             }
                         }
                         Console.WriteLine(" done.");
+                        if (exitLoop)
+                            break;
```
ivandrofly: Don't keep loop once File is already written on disk

niksedk: Cool :) 
I don't think you need the "exitLoop" variable, right?

ivandrofly: yeah, the reason I use a variable is that there is one that is located inside another for so... what you think?!

niksedk: A "break;" after 'Console.WriteLine(" done.");' should be fine.

ivandrofly: :+1:

ivandrofly: :+1:

PR #625, opened 2015-03-08
More refacts | Optimization

ivandrofly:


src/Core/HtmlUtil.cs
```diff
@@ -356,15 +356,9 @@ public static string EncodeNumeric(string source)
...
         public static string RemoveHtmlTags(string s)
         {
-            if (s == null)
-                return null;
-
-            if (s.Length < 3 || s.IndexOf('<') < 0)
+            if (string.IsNullOrEmpty(s) || s.IndexOf('<') < 0)
                 return s;
 
-            if (s.IndexOf("< ", StringComparison.Ordinal) >= 0)
```
ivandrofly: Not needed anymore HtmlUtil is already taking care of that *< i>* | *< i>*...

niksedk: Hm, not really the same. FixInvalidItalicTags does more, like removing "hanging" tags...

PR #631, opened 2015-03-10
Refact + Optimization

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -2122,9 +2122,10 @@ public static int CountTagInText(string text, string tag)
             while (index >= 0)
             {
                 count++;
-                if ((index + 1) == text.Length)
+                index = index + tag.Length;
```
ivandrofly: Should be faster :wink:

PR #640, opened 2015-03-14
Removed 'hit' variable from GeneratePreview() (Forms/MultipleReplace)

xylographe:
Removing the hit variable from GeneratePreview() will be slightly more efficient for paragraph texts that generate a hit, as the string comparison `newText != p.Text` is going to happen anyway. But, what about texts that don't generate a hit? Because `newText` and `p.Text` would still be the same reference pointer, the comparison should be fast, hence, the difference with using 'hit' might be negligible. I ran a test, and to my surprise, the NoHit version of GeneratePreview() appears to be slightly more efficient. However, my results might be a false positive, you should probably run your own test.

The test files I used [are in mrhittest.7z](https://www.dropbox.com/s/g6e898pg9cd5a5r/mrhittest.7z?dl=0),

- mrhittest.srt (60000 subtitle lines)
- mrhittest.template (6000 random replacement rules, not generating a hit)
- MultipleReplace.cs (with GeneratePreviewHit/GeneratePreviewNoHit)

Clicking OK will alternately invoke the Hit/NoHit preview, showing the result where normally the number of fixes is displayed. Cancel works as expected.

### DIFF
--- a/src/Forms/MultipleReplace.cs
+++ b/src/Forms/MultipleReplace.cs
@@ -192,7 +192,6 @@ namespace Nikse.SubtitleEdit.Forms
             
             foreach (Paragraph p in _subtitle.Paragraphs)
             {
-                bool hit = false;
                 string newText = p.Text;
                 foreach (ReplaceExpression item in replaceExpressions)
                 {
@@ -200,7 +199,6 @@ namespace Nikse.SubtitleEdit.Forms
                     {
                         if (newText.Contains(item.FindWhat))
                         {
-                            hit = true;
                             newText = newText.Replace(item.FindWhat, item.ReplaceWith);
                         }
                     }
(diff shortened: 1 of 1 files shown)


niksedk:
Interesting!
I also ran some tests. When a hit occurs the results was the same, but for no-hits the bool actually seems a lot faster. I ran tests in a loop, in release mode, outside VS:
```
WITH-HIT: 00:00:00.51 (using the bool)
NO-HIT  : 00:00:02.73 (using string != string)
WITH-HIT: 00:00:00.51
NO-HIT  : 00:00:02.73
WITH-HIT: 00:00:00.51
NO-HIT  : 00:00:02.73
WITH-HIT: 00:00:00.51
NO-HIT  : 00:00:02.73
WITH-HIT: 00:00:00.51
NO-HIT  : 00:00:02.73
WITH-HIT: 00:00:00.51
```
Even "ReferenceEquals" seems to be slower than the bool (about 0.65), so string compare does not use reference comparison. It seems a bool can speed things up in loops.

xylographe:
Excellent! That is more in line with expectations. Although, I didn't expect the difference to be that big—it's certainly *not* negligible. :smile:

Thank you, for testing.

PR #643, opened 2015-03-14
Patch 15

ivandrofly:


src/Core/StringExtensions.cs
```diff
@@ -137,5 +137,28 @@ public static int IndexOfAny(this string s, string[] words, StringComparison com
             }
             return -1;
         }
+
+        public static string FixExtraEspaces(this string s)
+        {
+            if (string.IsNullOrEmpty(s) || s.Trim().Length == 0)
```
xylographe: Perhaps, `if (string.IsNullOrWhiteSpace(s))` ?

ivandrofly: You should be :+1:
Trim() calls: 
```
private string TrimHelper(int trimType)
{
    int end = this.Length - 1;
    int start = 0;
    if (trimType != 1)
    {
        start = 0;
        while (start < this.Length)
        {
            if (!char.IsWhiteSpace(this[start]) && !IsBOMWhitespace(this[start]))
            {
                break;
            }
            start++;
        }
    }
    if (trimType != 0)
    {
        end = this.Length - 1;
        while (end >= start)
        {
            if (!char.IsWhiteSpace(this[end]) && !IsBOMWhitespace(this[start]))
            {
                break;
            }
            end--;
        }
    }
    return this.CreateTrimmedString(start, end);
}
```
[__DynamicallyInvokable]
```public static bool IsNullOrWhiteSpace(string value)
{
    if (value != null)
    {
        for (int i = 0; i < value.Length; i++)
        {
            if (!char.IsWhiteSpace(value[i]))
            {
                return false;
            }
        }
    }
    return true;
}```

PR #647, opened 2015-03-14
Patch 14

ivandrofly:


src/Forms/ChangeCasing.cs
```diff
@@ -130,14 +130,7 @@ private string FixCasing(string text, string lastLine, List<string> namesEtc)
             {
                 var st = new StripableText(text);
                 text = st.Pre + st.StrippedText.ToUpper() + st.Post;
-                text = text.Replace("<I>", "<i>");
-                text = text.Replace("</I>", "</i>");
-                text = text.Replace("<B>", "<b>");
-                text = text.Replace("</B>", "</b>");
-                text = text.Replace("<U>", "<u>");
-                text = text.Replace("<U>", "</u>");
```
ivandrofly: Bug fixed ```<U> => </u>```

niksedk: nice :)
could you make smaller patches? I think you sometimes change a bit much... which makes it easy to introduce new bugs...

PR #655, opened 2015-03-16
Patch 11

ivandrofly:


src/Logic/StripableText.cs
```diff
@@ -127,44 +127,30 @@ private void ReplaceNames1Remove(List<string> namesEtc, List<string> replaceIds,
                 Post = Post.Remove(0, 1);
             }
 
-            string lower = StrippedText.ToLower();
-
+            string strippedLower = StrippedText.ToLower();
```
ivandrofly: ***lower*** is a little bit confusing

PR #656, opened 2015-03-16
Minor Refact

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -2976,10 +2976,8 @@ public static string RemoveUnneededSpaces(string text, string language)
             while (text.Contains("  "))
                 text = text.Replace("  ", " ");
 
-            if (text.Contains(" " + Environment.NewLine, StringComparison.Ordinal))
-                text = text.Replace(" " + Environment.NewLine, Environment.NewLine);
-            if (text.Contains(Environment.NewLine + " ", StringComparison.Ordinal))
-                text = text.Replace(Environment.NewLine + " ", Environment.NewLine);
+            text = text.Replace(" " + Environment.NewLine, Environment.NewLine);
+            text = text.Replace(Environment.NewLine + " ", Environment.NewLine);
```
niksedk: nice, I thought "Contains" was faster... but it's not so (in my tests)

PR #657, opened 2015-03-16
Patch 11

ivandrofly:
/cc @niksedk

### DIFF
--- a/src/Logic/Utilities.cs
+++ b/src/Logic/Utilities.cs
@@ -3437,5 +3437,23 @@ namespace Nikse.SubtitleEdit.Logic
             return lines;
         }
 
+        internal static bool StartEndOkay(string text, string name, int startIdx)
+        {
+            if (string.IsNullOrWhiteSpace(text) || string.IsNullOrWhiteSpace(name) || startIdx < 0)
+                return false;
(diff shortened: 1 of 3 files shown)


src/Logic/Utilities.cs
```diff
@@ -3437,5 +3437,23 @@ public static int GetNumberOfLines(string text)
...
+            var postIdx = startIdx - 1;
+            bool startOk = (startIdx == 0) || (text[postIdx] == ' ') || (text[postIdx] == '-') || (text[postIdx] == '(') || (text[postIdx] == '[') ||
+                           (text[postIdx] == '"') || (text[postIdx] == '\'') || (text[postIdx] == '>') ||
+                           (Environment.NewLine.EndsWith(text[postIdx].ToString(CultureInfo.InvariantCulture)));
+
+            if (startOk)
+            {
+                int end = startIdx + name.Length;
+                return (end == text.Length || (end < text.Length && (@" ,.!?:;')]-<""" + Environment.NewLine).Contains(text[end])));
+            }
```
ivandrofly: ```)``` and ```]``` to allow fix if name is like: ```(Name)```

PR #661, opened 2015-03-17
Validate VLC path if pasted from clipboard

ivandrofly:
Portable VLC

### DIFF
--- a/src/Forms/Settings.cs
+++ b/src/Forms/Settings.cs
@@ -2521,11 +2521,16 @@ namespace Nikse.SubtitleEdit.Forms
             }
             if (openFileDialogFFmpeg.ShowDialog(this) == DialogResult.OK)
             {
-                textBoxVlcPath.Text = Path.GetDirectoryName(openFileDialogFFmpeg.FileName);
-                Configuration.Settings.General.VlcLocation = textBoxVlcPath.Text;
-                Configuration.Settings.General.VlcLocationRelative = GetRelativePath(textBoxVlcPath.Text);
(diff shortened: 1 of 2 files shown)


src/Forms/Settings.cs
```diff
@@ -2553,5 +2558,18 @@ private void linkLabelOpenDictionaryFolder_LinkClicked(object sender, LinkLabelL
             System.Diagnostics.Process.Start(dictionaryFolder);
         }
 
+        private void textBoxVlcPath_MouseLeave(object sender, EventArgs e)
+        {
+            try
+            {
+                var path = textBoxVlcPath.Text.Trim('\"');
+                if (path.Length > 3 && Path.IsPathRooted(path) && Path.GetFileName(path).Equals("vlc.exe", StringComparison.OrdinalIgnoreCase) && File.Exists(path))
+                {
```
xylographe: The .exe extension is an MS Windows peculiarity. This test might fail needlessly on a different OS.

ivandrofly: So why open file dialogue is filtering only ```*.vlc.exe``` :confused:

xylographe: I would assume, that might be the best guide for clueless users?
It wouldn't stop you, though, from entering something completely different. :smile:

ivandrofly: @niksedk

niksedk: I guess the portable stuff is only relevant for windows users, right?
Also, I've not succeeded in making VLC work as video player on linux (or osx - don't have a mac).

PR #671, opened 2015-03-26
Use constant for subtitle format name + Refact

ivandrofly:


src/Logic/SubtitleFormats/Spt.cs
```diff
@@ -13,9 +13,11 @@ public override string Extension
             get { return ".spt"; }
         }
 
+        public const string NameOfFormat = "NCI Caption";
```
niksedk: should probably be "spt"... otherwise nice :)

ivandrofly: :+1:

ivandrofly:
Done @niksedk

PR #686, opened 2015-04-02
Refact + Optimization

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -2105,9 +2105,10 @@ public static int CountTagInText(string text, string tag)
             while (index >= 0)
             {
                 count++;
-                if ((index + 1) == text.Length)
+                index = index + tag.Length;
```
ivandrofly: This is going to be faster

niksedk: in some situations... but the result is wrong ;)

niksedk: ups, sorry - the result is correct - my testing was buggy...

PR #698, opened 2015-04-06
Fix/patch 33

ivandrofly:
@niksedk


src/Logic/Dictionaries/OcrFixReplaceList.cs
```diff
@@ -324,64 +325,26 @@ public string FixCommonWordErrors(string word)
                 word = word.Substring(2);
             }
 
-            while (word.Length > 1 && word[0] == '-')
```
ivandrofly: @niksedk isn't better checking if length is greater than 0?

PR #703, opened 2015-04-09
Fix/patch

ivandrofly:


xylographe:
I think, `first.Append(' ').Append(word)` is more efficient than `first.Append(' ' + word)`.
See [simple test here.](https://gist.github.com/xylographe/30091b257e029f29373b)

PR #707, opened 2015-04-14
Fix/patch

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -2985,21 +2987,24 @@ public static string RemoveUnneededSpaces(string text, string language)
...
+            while (text.Contains("...."))
+                text = text.Replace("....", Ellipses);
+
+            text = text.Replace(" ..." + Environment.NewLine, Ellipses + Environment.NewLine);
             text = text.Replace(Environment.NewLine + "... ", Environment.NewLine + "...");
             text = text.Replace(Environment.NewLine + "<i>... ", Environment.NewLine + "<i>...");
             text = text.Replace(Environment.NewLine + "- ... ", Environment.NewLine + "- ...");
             text = text.Replace(Environment.NewLine + "<i>- ... ", Environment.NewLine + "<i>- ...");
             text = text.Replace(Environment.NewLine + "- ... ", Environment.NewLine + "- ...");
+            text = text.Replace(" -" + Environment.NewLine, "-" + Environment.NewLine);
```
ivandrofly: Was missing

src/Forms/Statistics.cs
```diff
@@ -177,21 +177,8 @@ private static void MostUsedWordsAdd(Dictionary<string, string> hashtable, strin
         {
             if (lastLine.Contains("< "))
                 lastLine = Utilities.FixInvalidItalicTags(lastLine);
-            lastLine = lastLine.Trim('\'');
```
ivandrofly: All done in new method

PR #718, opened 2015-04-25
Optimization + Refact

ivandrofly:


src/Forms/Main.cs
```diff
@@ -4707,11 +4707,6 @@ private void FixCommonErrors(bool onlySelectedLines)
                                     }
                                     i++;
                                 }
-                                deletes.Reverse();
```
ivandrofly: Not needed since we can remove from reversed for loop

Issue #313, opened 2014-10-16
customising languages during installation


diomed:
it would be useful if users could select their own language during installation, by having dropdown menu with list of languages; where they'd select just their own, which they need, and not others, so people would save some space on hard drive by disabling installer to install so much languages, which are unnecessary. so english + native should do.

PR #730, opened 2015-05-05
SE-installer: individually selectable translations (custom install)

xylographe:
Issue #313


XhmikosR:
1. So much bloat, well done.
2. Changing the file encoding without any reason. Seriously, just use a proper text editor.

PR #738, opened 2015-05-07
Develop/patch 1

ivandrofly:


src/Logic/Forms/FixCommonErrorsHelper.cs
```diff
@@ -443,15 +443,14 @@ public static string FixDoubleGreaterThanHelper(string text)
...
                 return text;
 
             string s = HtmlUtil.RemoveHtmlTags(text, true);
             if (s.Contains(Environment.NewLine) && s.Replace(Environment.NewLine, " ").Replace("  ", " ").Length < Configuration.Settings.Tools.MergeLinesShorterThan)
             {
-                s = s.TrimEnd().TrimEnd('.', '?', '!', ':', ';');
-                s = s.TrimStart('-');
-                if (s.IndexOfAny(new[] { '.', '?', '!', ':', ';', '-', '♪', '♫' }) < 0 &&
+                s = s.TrimEnd().TrimEnd('.', '?', '!', ':', ';').TrimStart('-');
+                if (s.ContainsAny(new[] { '.', '?', '!', ':', ';', '-', '♪', '♫' }) &&
```
xylographe: If  I'm not mistaken, `(s.IndexOfAny(v) < 0)` means that none of `v` are in `s`.
The equivalent would probably be `!s.ContainsAny(v)` ?

PR #745, opened 2015-05-10
Fix/patch 0.1

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -1386,9 +1385,7 @@ public static string AutoDetectLanguageName(string languageName, Subtitle subtit
             bool containsEnUs = false;
             foreach (string name in dictionaryNames)
             {
-                if (name.Contains("[en_GB]"))
-                    containsEnGb = true;
-                if (name.Contains("[en_US]"))
+                if (name.Contains("[en_GB]") || name.Contains("[en_US]"))
                     containsEnUs = true;
```
xylographe: No more support for British English?

ivandrofly: MB, thx ;) @xylographe

PR #751, opened 2015-05-12
Fix Uneeded spaces if like: <i> foobar </i>

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -3212,6 +3212,23 @@ private static string RemoveSpaceBeforeAfterTag(string text, string openTag)
...
+                idx = text.IndexOf(leftRightSpaceTag, StringComparison.Ordinal);
+            }
+            leftRightSpaceTag = " " + closeTag;
+            idx = text.LastIndexOf(leftRightSpaceTag, StringComparison.Ordinal);
+            while (idx >= 0 && ((text.Length == idx + leftRightSpaceTag.Length) || (text.Length > idx + leftRightSpaceTag.Length && text[idx + leftRightSpaceTag.Length] == '.' || text[idx + leftRightSpaceTag.Length] == ' ' ||
+                text[idx + leftRightSpaceTag.Length] == '?' || text[idx + leftRightSpaceTag.Length] == '!')))
+            {
+                text = text.Remove(idx, 1);
+                idx = text.LastIndexOf(leftRightSpaceTag, idx, StringComparison.Ordinal);
+            }
```
xylographe: I think, when `LastIndexOf()` returns `idx >= 0`, then `text.Length` will always be greater than or equal to `idx + leftRightSpaceTag.Length`, or there would be no match. Hence, the second test (`text.Length > idx + ...`) would not be necessary :
```
while (idx >= 0 &&
  (text.Length == idx + leftRightSpaceTag.Length ||
  " .?!,;:".Contains(text[idx + leftRightSpaceTag.Length])))
```

ivandrofly:
:+1: @xylographe

PR #753, opened 2015-05-13
Fixed Croatian subtitle auto-detection - thx diomed

xylographe:


xylographe:
[2]
Added `RegexOptions.CultureInvariant` to guarantee culture independent language detection.
Added `RegexOptions.ExplicitCapture` for efficiency.
[MSDN](https://msdn.microsoft.com/en-us/library/e7sf90t3): “Accessing members of this collection such as MatchCollection.Count and MatchCollection.CopyTo causes the collection to be populated immediately.”

[4]
Fixed `count` => `romanianCount`

PR #757, opened 2015-05-17
Bug fixed while generating Effect Karaoke

ivandrofly:
```<i>Eu sou o Flash.</i>```
==
Bug:
```<font color="#ff0000"><i>Eu sou o Flash..</i</font>```

Correct:
```<font color="#ff0000"><i>Eu sou o Flash.</i></font>```

I think the same bug is residing in Effect TypeWriter :wink:

### DIFF
--- a/src/Forms/EffectKaraoke.cs
+++ b/src/Forms/EffectKaraoke.cs
@@ -229,6 +229,7 @@ namespace Nikse.SubtitleEdit.Forms
             double duration = _paragraph.Duration.TotalMilliseconds - ((double)numericUpDownDelay.Value * 1000.0);
             double stepsLength = CalculateStepLength(_paragraph.Text, duration);
 
+            double totalMilliseconds = _paragraph.StartTime.TotalMilliseconds;
             double startMilliseconds;
             double endMilliseconds;
             TimeCode start;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/Forms/EffectKaraoke.cs [hunk 1]
```diff
@@ -290,8 +289,8 @@ private void MakeAnimation()
             if (numericUpDownDelay.Value > 0)
             {
                 startMilliseconds = index * stepsLength;
-                startMilliseconds += _paragraph.StartTime.TotalMilliseconds;
-                endMilliseconds = _paragraph.EndTime.TotalMilliseconds;
+                startMilliseconds += totalMilliseconds;
+                endMilliseconds = totalMilliseconds;
```
xylographe: `totalMilliseconds = _paragraph.StartTime.TotalMilliseconds`
and
 `endMilliseconds = totalMilliseconds`
==>
 `endMilliseconds = _paragraph.StartTime.TotalMilliseconds`

That would not be the same as the original
 `endMilliseconds = _paragraph.EndTime.TotalMilliseconds`

src/Forms/EffectKaraoke.cs
```diff
@@ -263,11 +264,11 @@ private void MakeAnimation()
...
-                        {
-                            tag += _paragraph.Text[i];
-                            i++;
-                        }
+                        var tagStart = i + 1;
+                        var tagEnd = _paragraph.Text.IndexOf('>', tagStart + 2);
+
+                        tag += tagEnd > tagStart ? _paragraph.Text.Substring(tagStart, tagEnd - tagStart + 1) : _paragraph.Text.Substring(tagStart);
+                        i += tag.Length;
                         text += tag;
```
xylographe: Or, perhaps, enforce the closing `>` ?
```
                        while (++i < _paragraph.Text.Length && _paragraph.Text[i] != '>')
                        {
                            tag += _paragraph.Text[i];
                        }
                        text += tag + '>';
```

src/Forms/EffectKaraoke.cs
ivandrofly: :+1: should be cause CTRL+. rename :confused:

PR #758, opened 2015-05-17
Bug fixed when generating Effect TypeWriter

ivandrofly:
From:
`<i><b>E`

To:
`<i><b>E</b></i>`


src/Forms/EffectTypewriter.cs
```diff
@@ -111,7 +113,42 @@ private void MakeAnimation()
...
+                            var reversed = string.Empty;
+                            while (tag.Length > 0)
+                            {
+                                var endIdx = tag.IndexOf('>') + 1;
+                                reversed = tag.Substring(0, endIdx) + reversed;
+                                tag = tag.Remove(0, endIdx);
+                            }
+                            tag = reversed;
+                        }
+                        listTag.Add(tag.Replace("<", "</"));
```
ivandrofly: Make it close tag

PR #766, opened 2015-05-21
Added new method to Fix Large Font + Dispose Graphics after usage

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -3435,5 +3435,18 @@ public static int GetNumberOfLines(string text)
             return lines;
         }
 
+        public static void FixLargeFonts(Control mainCtrl, Control ctrl)
+        {
+            using (Graphics graphics = mainCtrl.CreateGraphics())
```
ivandrofly: Should always be disposed

niksedk:
nice fix :)

PR #768, opened 2015-05-22
Refact + Bug fix

ivandrofly:


src/Logic/Subtitle.cs
```diff
@@ -291,7 +291,7 @@ private void FixEqualOrJustOverlappingFrameNumbers()
             for (int i = 0; i < Paragraphs.Count - 1; i++)
             {
                 Paragraph p = Paragraphs[i];
-                Paragraph next = Paragraphs[i + 1];
```
ivandrofly: Bug?

src/Logic/Subtitle.cs
```diff
@@ -465,9 +465,7 @@ public int RemoveEmptyLines()
                 for (int i = _paragraphs.Count - 1; i >= 0; i--)
                 {
                     Paragraph p = _paragraphs[i];
-                    string s = p.Text.Trim();
-
-                    if (s.Length == 0)
+                    if (p.Text.Trim().Length == 0)
                         _paragraphs.RemoveAt(i);
                 }
```
xylographe: I think, `string.IsNullOrWhiteSpace()` might be slightly more efficient,
```
for (int i = _paragraphs.Count - 1; i >= 0; i--)
{
    if (string.IsNullOrWhiteSpace(_paragraphs[i].Text))
        _paragraphs.RemoveAt(i);
}
```

ivandrofly: @xylographe :+1:

PR #770, opened 2015-05-23
Minor refact (Logic/Utilities)

xylographe:
Matching `[Hh]et` seems to be more efficient than matching `Het` and `het` separately.
Same for `[Cc]'est` vs. `C'est` and `c'est`.


ivandrofly:
Nice :+1:

PR #771, opened 2015-05-23
More Refact in StartNumberingForm

ivandrofly:
Use `e.KeyValue` instead repeating `e.KeyCode`


ivandrofly:
@niksedk take a look here

niksedk:
I think it makes the readability worse...

PR #777, opened 2015-05-24
Fixed Cyrillic _codePage init (Logic/SubtitleFormats/Pac)

xylographe:
Perhaps, it might be more convenient to use named codePage constants instead of integer literals?


niksedk:
Yeah, good idea with constants (I'll add this - I'm looking a bit at writing Chinese)

PR #789, opened 2015-05-27
Unneeded loop removed inside LoadPlugins

ivandrofly:
@niksedk Do you think it's really necessary to have these loops inside LoadPlugins?
Isn't it going to affect the launch time only


niksedk:
This is not unused... is (at least) used after showing the "plugin window".

PR #791, opened 2015-05-28
Do not entry FixInvalidItalicTags if text doesn't contain '<'

ivandrofly:
@niksedk This patch is safe and will provide some optimization... instead of running 10+ step without fixing anything why not exit the method if there is no `<`


ivandrofly:
@xylographe Anything to say about this patch?!

ivandrofly:
Will conflict with https://github.com/SubtitleEdit/subtitleedit/commit/98b9215d89e4b3f0737be74a64a48ea9217fdea9

PR #794, opened 2015-05-29
Use Utilities.FixLargeFonts in DCinema forms

xylographe:
[1]
Although the abstract class `DCinemaPropertiesForm` is now empty, it is still used in Forms/Main.

[2]
I think, the idea is to limit the `SetButtonHeight()` recursion depth.


niksedk:
1) yeah... I guess the base form should be removed - and I kinda dislike inheritance for visual components (on a side note it's annoying ms does not have a clear windows desktop/frontend strategy - wpf or winrt... so for new code I guess separation of UI and code in nice)

PR #799, opened 2015-05-29
Regex: Two more regex added to context menu + multi-lang support

ivandrofly:


LanguageMaster.xml
```diff
@@ -1421,8 +1421,10 @@ can edit in same subtitle file (collaboration)</Information>
     <NewLine>New line (\r\n)</NewLine>
     <NewLineShort>New line (\n)</NewLineShort>
     <AnyDigit>Any digit (\d)</AnyDigit>
+    <NonDigit>Non digit (\D)</NonDigit>
```
xylographe: I would suggest `<NonDigit>Not a digit (\D)</NonDigit>`.

Also, (line 1420)
 `<NonWordBoundary>Non word boundary (\B)</NonWordBoundary>`
 ==> `<NonWordBoundary>Not a word boundary (\B)</NonWordBoundary>`.

LanguageMaster.xml
```diff
@@ -1421,8 +1421,10 @@ can edit in same subtitle file (collaboration)</Information>
...
     <AnyCharacter>Any character (.)</AnyCharacter>
     <AnyWhitespace>Any whitespace (\s)</AnyWhitespace>
+    <NonSpaceCharacter>Non space character (\S)</NonSpaceCharacter>
```
xylographe: `<NonWhitespace>Non-whitespace (\S)</NonWhitepace>`

ivandrofly: Note that \s won't only match ` ` \r\n are also included

xylographe: Space is used for ASCII 32 (U+0020).
Whitespace[1] is used for a _set_ of characters, including space, nbsp, tab, linefeed, and others.
`\s` represents the set of all white space characters, `\S` is its complement (the set of characters that are not in `\s`).

[1] Actually, two words "white space", but "whitespace" seems to be more commonly used these days.

ivandrofly: Well, let's hear also what @niksedk will say about it :confused:

niksedk: English is not my native language, but "non-digit character" / "non-whitespace character" looks good to me - like from http://www.w3schools.com/jsref/jsref_regexp_digit_non.asp

PR #801, opened 2015-05-30
Store SE.ico only once, not thrice

xylographe:
Currently, SE.ico is stored in the resources of three Forms: Main, FindSubtitleLine, and AdjustDisplayDuration. This patch adds SE.ico to Properties.Resources (Nikse.SubtitleEdit.Properties.Resources.SubtitleEditFormIcon) instead.


niksedk:
nice fix :)

PR #804, opened 2015-05-31
Store SE.ico only once (Forms/GoToLine.cs)

ivandrofly:


niksedk:
Nice catch, for once I think the exe file will be smaller ;)

ivandrofly:
Indeed :smile:

PR #806, opened 2015-05-31
Improved: AutoDetectGoogleLanguage

ivandrofly:


ivandrofly:
Anything wrong? @niksedk

niksedk:
I think it's meant to be case sensitive... otherwise all possible conflicts with names etc. must be considered

ivandrofly:
Okay, I'll RP only new words added :+1:

PR #807, opened 2015-05-31
Minor refact

xylographe:
[1]
Release XmlWriter resources.
Using StringBuilder increases efficiency.

[2]
formSubOcr was used after being disposed.

[3]
A FindType property seems more natural than a GetFindType method, and consistent with the existing FindText property.
Some event handler names changed for consistency (ControlName_EventName).

### DIFF
--- a/src/Forms/DvdSubRipChooseLanguage.cs
+++ b/src/Forms/DvdSubRipChooseLanguage.cs
@@ -172,15 +172,17 @@ namespace Nikse.SubtitleEdit.Forms
                 subs.Add((x as SubListBoxItem).SubPack);
             }
 
-            var formSubOcr = new VobSubOcr();
-            formSubOcr.InitializeQuick(subs, _palette, Configuration.Settings.VobSubOcr, SelectedLanguageString);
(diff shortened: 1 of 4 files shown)


src/Logic/SubtitleFormats/SubtitleFormat.cs
```diff
@@ -363,35 +369,27 @@ public virtual bool HasStyleSupport
 
         public bool BatchMode { get; set; }
 
-        public static string ToUtf8XmlString(XmlDocument xml, bool omitXmlDeclaration)
+        public static string ToUtf8XmlString(XmlDocument xml, bool omitXmlDeclaration = false)
```
niksedk: I don't this works as intended as no xml declaration is generated...

xylographe: The behaviour of `ToUtf8XmlString()` should be exactly the same as before. I will investigate this later.
For now, I shall remove this commit from the PR.

xylographe:
[Removed [1]](https://github.com/SubtitleEdit/subtitleedit/pull/807#discussion-diff-31453273) pending investigation.

PR #814, opened 2015-06-01
Improved: AutoDetectGoogleLanguage [pt & fr]

ivandrofly:
This update is safe ;) @niksedk

### DIFF
--- a/src/Logic/Utilities.cs
+++ b/src/Logic/Utilities.cs
@@ -1204,7 +1204,7 @@ namespace Nikse.SubtitleEdit.Logic
                     return "es";
             }
 
-            count = GetCount(text, "un", "vous", "avec", "pas", "ce", "une");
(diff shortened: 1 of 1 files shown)


src/Logic/Utilities.cs
```diff
@@ -1230,13 +1230,13 @@ public static string AutoDetectGoogleLanguage(string text, int bestCount)
                                    "giorno", "faccio", "davvero", "negativo", "essere", "vuole", "sensitivo", "venire");
             if (count > bestCount)
             {
-                int frenchCount = GetCount(text, "[Cc]'est", "pas", "vous", "pour", "suis", "Pourquoi", "maison", "souviens", "quelque"); // not spanish words
+                int frenchCount = GetCount(text, "([Cc]')?est", "pas", "vous", "pour", "suis", "Pourquoi", "maison", "souviens", "quelque"); // not spanish words
```
ivandrofly: Matches `est` as swell

PR #822, opened 2015-06-03
Refact ToUtf8XmlString (Logic/SubtitleFormats/SubtitleFormat)

xylographe:
I spoke too soon. This is a _clean_ `ToUtf8XmlString()` implementation, and it happens to be more efficient than its predecessors.

See [ToUtf8XmlString-Efficiency-Test](https://gist.github.com/xylographe/b1b65f1a5f53c5440c99) and [ToUtf8XmlString-Validity-Test](https://gist.github.com/xylographe/dbb8eab90877b8f2c96f) on Gist.


niksedk:
Hm, my tests show that the new commit is slower... (running i a loop, in release mode, out of VS)

PR #832, opened 2015-06-06
Show friendly msg when user try to drop JPG file

ivandrofly:


niksedk:
Just checked a few jpg files - many of mine has DB i index 03 (fourth byte)... so perhaps it should be skipped or be between C0 and FE (both included) !?
Check "Common jpeg markers" here: http://en.wikipedia.org/wiki/JPEG

PR #838, opened 2015-06-08
Added: New context menu to add word to (NamesEtcList)

ivandrofly:
This patch will allow adding word by selecting & right click directly from `Subtitle Edit` TextBox `TESTED`


ivandrofly:
@niksedk

PR #840, opened 2015-06-08
Also remove SSA tags when checking text length

ivandrofly:


src/Forms/Styles/TimedTextStyles.cs
```diff
@@ -64,6 +64,7 @@ protected override void GeneratePreviewReal()
                 pictureBoxPreview.Image.Dispose();
             var bmp = new Bitmap(pictureBoxPreview.Width, pictureBoxPreview.Height);
 
+            var path = new GraphicsPath();
             using (Graphics g = Graphics.FromImage(bmp))
             {
```
xylographe:
```
            using (var g = Graphics.FromImage(bmp))
            using (var path = new GraphicsPath())
            {
                 ...
            }
```
Both `g` and `path` will be auto-disposed.

PR #843, opened 2015-06-09
Do CTRL+BACK moved to Utilities to allow sharing func with othersTextBoxes

ivandrofly:
There will  be more pushes regarding to this patch


ivandrofly:
@niksedk take a look here

niksedk:
Hm, perhaps "DoNotBreakAfterListEdit" should use the SETextBox ?

ivandrofly:
:+1: , I'll try implementing it

PR #845, opened 2015-06-09
Filter popular video format

ivandrofly:
In most of the case there will the a video file in same directory as subtitle, so why waste time checking if it's subtitle file or not


ivandrofly:
@niksedk

PR #850, opened 2015-06-10
Store the result in variable instead calc every time loop run

ivandrofly:


src/Logic/Subtitle.cs
```diff
@@ -299,12 +299,13 @@ private void FixEqualOrJustOverlappingFrameNumbers()
...
             {
                 double startFrame = p.StartTime.TotalMilliseconds / 1000.0 * oldFrameRate;
                 double endFrame = p.EndTime.TotalMilliseconds / 1000.0 * oldFrameRate;
-                p.StartTime.TotalMilliseconds = startFrame * (1000.0 / newFrameRate);
-                p.EndTime.TotalMilliseconds = endFrame * (1000.0 / newFrameRate);
+                p.StartTime.TotalMilliseconds = startFrame * st;
+                p.EndTime.TotalMilliseconds = endFrame * st;
                 p.CalculateFrameNumbersFromTimeCodes(newFrameRate);
             }
         }
```
xylographe: To me, it looks like a complicated way to write,
```
p.StartTime.TotalMilliseconds = (p.StartTime.TotalMilliseconds * oldFramerate) / newFrameRate;
p.EndTime.TotalMilliseconds = (p.EndTime.TotalMilliseconds * oldFramerate) / newFrameRate;
```
The 1000.0 in both numerator and denominator would seem to cancel out each other.

PR #859, opened 2015-06-14
Minor fixes

xylographe:
[1]
`s.Contains("")` returns true for any string `s != null`.


src/Logic/SubtitleFormats/Cavena890.cs
```diff
@@ -216,8 +215,7 @@ public void Save(string fileName, Subtitle subtitle)
                         p.Text.Contains('爱') ||
                         p.Text.Contains('的') ||
                         p.Text.Contains('早') ||
-                        p.Text.Contains('上') ||
-                        p.Text.Contains(""))
```
niksedk: nice catch :)

PR #867, opened 2015-06-15
Minor refact

xylographe:
[1] "new XmlTextWriter" ==> "XmlWriter.Create"

MSDN: “Starting with the .NET Framework 2.0, we recommend that you create XmlWriter instances by using the XmlWriter.Create method and the XmlWriterSettings class to take advantage of new functionality.”

Using a StringBuilder as output medium avoids the undisposed StringWriter issue. The StringWriter will now be created and disposed internally.

[2] "s.Text.Trim().Length == 0" ==> "string.IsNullOrWhiteSpace(s)"

IsNullOrWhiteSpace is slightly more efficient.

### DIFF
--- a/src/Core/StringExtensions.cs
+++ b/src/Core/StringExtensions.cs
@@ -145,8 +145,8 @@ namespace Nikse.SubtitleEdit.Core
 
         public static string FixExtraSpaces(this string s)
         {
-            if (string.IsNullOrEmpty(s) || s.Trim().Length == 0)
-                return s;
+            if (string.IsNullOrWhiteSpace(s))
+                return string.Empty;
 
             while (s.Contains("  "))
                 s = s.Replace("  ", " ");
--- a/src/Forms/Compare.cs
+++ b/src/Forms/Compare.cs
@@ -904,7 +904,7 @@ namespace Nikse.SubtitleEdit.Forms
 
         private static void CopyTextToClipboard(RichTextBox sender)
         {
-            if (sender.Text.Trim().Length == 0)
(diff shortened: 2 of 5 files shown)


src/Core/StringExtensions.cs
```diff
@@ -145,7 +145,7 @@ public static int IndexOfAny(this string s, string[] words, StringComparison com
 
         public static string FixExtraSpaces(this string s)
         {
-            if (string.IsNullOrEmpty(s) || s.Trim().Length == 0)
+            if (string.IsNullOrWhiteSpace(s))
```
ivandrofly: what is s == "   "

xylographe:
```
PS> [string]::IsNullOrEmpty('')
True
```
[MSDN](https://msdn.microsoft.com/en-us/library/system.string.isnullorwhitespace): “true if the value parameter is null or String.Empty, or if value consists exclusively of white-space characters.”

xylographe:
```
PS> $s = ' ' * 3; $s.Length; [string]::IsNullOrWhiteSpace($s)
3
True
```

ivandrofly: yeah, s.Trim().Length is a bug I think

xylographe: Changed `return s` to `return string.Empty`. That should fix it.

xylographe:
\+ [minor fix](https://github.com/SubtitleEdit/subtitleedit/pull/867#discussion_r32490440) in `FixExtraSpaces()`

PR #877, opened 2015-06-18
More Refact in (src/Test/Logic/Forms/RemoveTextForHearImpairedTest.cs)

ivandrofly:


src/Test/Logic/Forms/RemoveTextForHearImpairedTest.cs
```diff
@@ -31,16 +32,15 @@ private static RemoveTextForHI GetRemoveTextForHiLib()
...
-            target.Settings.OnlyIfInSeparateLine = false;
-            target.Settings.OnlyIfInSeparateLine = false;
-            target.Settings.ColonSeparateLine = false;
-            target.Settings.RemoveTextBeforeColonOnlyUppercase = false;
+            Target.Settings.RemoveIfAllUppercase = false;
+            Target.Settings.RemoveTextBeforeColon = true;
+            Target.Settings.OnlyIfInSeparateLine = false;
+            Target.Settings.OnlyIfInSeparateLine = false;
+            Target.Settings.ColonSeparateLine = false;
+            Target.Settings.RemoveTextBeforeColonOnlyUppercase = false;
```
ivandrofly: Actually these assignment can be used in departed method to avoid duplication

PR #880, opened 2015-06-18
Batch renumber

ivandrofly:


niksedk:
Nice - but I wonder if there could be any reason not to renumber?

ivandrofly:
Can't find any... was requested by @miniop

ivandrofly:
What you think about making plugin instead?

niksedk:
OK, then perhaps the batch convert should just always renumber? (no GUI stuff needed)
@xylographe ?

xylographe:
I can't think of _any_ reason (good or pedantic :wink: ) not to renumber.

I think, 'no GUI' is the best way to go. If someone were to complain later on, the GUI stuff could still be added.

src/Forms/BatchConvert.cs
```diff
@@ -942,6 +949,10 @@ private static void DoThreadWork(object sender, DoWorkEventArgs e)
                     }
                 }
             }
+            if (p.Renumber)
+            {
+                p.Subtitle.Renumber(_startRenumFrom);
```
niksedk: Could you start with just this line?

ivandrofly: :+1:

niksedk:
We start with the simple version: https://github.com/SubtitleEdit/subtitleedit/commit/8c90482b5f59aa9caf726cf152964941f6a606f4

PR #887, opened 2015-06-20
Fixed: Last column won't resize (src/Forms/AutoBreakUnbreakLines.cs)

ivandrofly:


src/Forms/AutoBreakUnbreakLines.cs
```diff
@@ -213,5 +213,11 @@ private void listViewFixes_ItemChecked(object sender, ItemCheckedEventArgs e)
                 _notAllowedFixes.Add(p.ID);
         }
 
+        private void listViewFixes_Resize(object sender, EventArgs e)
+        {
+            var newWidth = (listViewFixes.Width - (listViewFixes.Columns[0].Width + listViewFixes.Columns[1].Width)) / 2;
+            listViewFixes.Columns[3].Width = listViewFixes.Columns[2].Width = newWidth;
```
niksedk: I also think you can use something like: listViewFixes.Columns[3].Width = -2; (perhaps introduce a constant instead of -2)

ivandrofly: > I also think you can use something like: listViewFixes.Columns[3].Width = -2; (perhaps introduce a constant instead of -2)

Won't have same meaning because onlyl listViewFixes.Columns[3] will resize not listViewFixes.Columns[2].Width

niksedk: ah - I had not seen that it was a double assignment...

ivandrofly: > ah - I had not seen that it was a double assignment...

:smile:

PR #888, opened 2015-06-20
Fix + Refact

ivandrofly:


src/Forms/AutoBreakUnbreakLines.cs
```diff
@@ -138,12 +139,9 @@ private void Unbreak()
             listViewFixes.Items.Clear();
             foreach (Paragraph p in _paragraphs)
             {
-                if (p.Text != null && p.Text.Contains(Environment.NewLine) && p.Text.Length > minLength)
+                if (p.Text != null && p.Text.Contains(Environment.NewLine) && HtmlUtil.RemoveHtmlTags(p.Text).Length > minLength)
```
ivandrofly: Shouldn't count with HTML tags

src/Forms/AutoBreakUnbreakLines.cs
```diff
@@ -171,7 +169,7 @@ private void AddToListView(Paragraph p, string newText)
             item.SubItems.Add(subItem);
             subItem = new ListViewItem.ListViewSubItem(item, p.Text.Replace(Environment.NewLine, Configuration.Settings.General.ListViewLineSeparatorString));
             item.SubItems.Add(subItem);
-            subItem = new ListViewItem.ListViewSubItem(item, newText.Replace(Environment.NewLine, Configuration.Settings.General.ListViewLineSeparatorString));
+            subItem = new ListViewItem.ListViewSubItem(item, newText);
```
ivandrofly: `\r\n` will be removed before getting here

src/Logic/Utilities.cs
```diff
@@ -792,14 +792,25 @@ private static string ReInsertHtmlTags(string s, Dictionary<int, string> htmlTag
...
-                return newText;
+                singleLine = singleLine.Replace("</i> <i>", " ");
+                singleLine = singleLine.Replace("</i><i>", " ");
+
+                singleLine = singleLine.Replace("</b> <b>", " ");
+                singleLine = singleLine.Replace("</b><b>", " ");
+
+                singleLine = singleLine.Replace("</u> <u>", " ");
+                singleLine = singleLine.Replace("</u><u>", " ");
             }
```
ivandrofly: Useful when *user* just want to *unbreak* line without running the **Fix Common Errors**

src/Forms/AutoBreakUnbreakLines.cs
niksedk: I don't think so... also tested it ;)

ivandrofly: :+1:, test I did was only for UnbreakLines, Reverted

ivandrofly: > I don't think so... also tested it ;)

:+1:, test I did was only for UnbreakLines, Reverted

ivandrofly:
Should

ivandrofly:
https://github.com/SubtitleEdit/subtitleedit/pull/895

PR #897, opened 2015-06-20
Dispose implemented to dispose Hunspell (class SusggestionParameter)

ivandrofly:


ivandrofly:
`Hunspell` doesn't contain **manged/unmanaged** type

PR #899, opened 2015-06-20
Move below so if '—' replaced with '-' we won't miss multiple '—'

ivandrofly:


src/Logic/Dictionaries/OcrFixReplaceList.cs
```diff
@@ -292,8 +289,11 @@ public string FixCommonWordErrors(string word)
                 if (word.Contains('—'))
                     word = word.Replace('—', '-');
 
+                while (word.Contains("--"))
+                    word = word.Replace("--", "-");
```
ivandrofly: `—` will be replace with `-` so there may be more than one `-`

PR #904, opened 2015-06-21
Use copy constructor instead re-calling GetSsaStyle()

ivandrofly:


src/Forms/Styles/SubStationAlphaStyles.cs
```diff
@@ -625,7 +625,7 @@ private void buttonCopy_Click(object sender, EventArgs e)
             {
                 string styleName = listViewStyles.SelectedItems[0].Text;
                 SsaStyle oldStyle = GetSsaStyle(styleName);
-                SsaStyle style = GetSsaStyle(styleName);
```
ivandrofly: Note: `GetSsaStyle(...)` will call `AdvancedSubStationAlpha.GetSsaStyle(styleName, _header)` which is very expensive

niksedk: nice :)

PR #907, opened 2015-06-21
Minor fix + minor refact

xylographe:
[1]
[MSDN](https://msdn.microsoft.com/en-us/library/fs2xkftw%28v=vs.110%29.aspx)
[Forms.VobSubOcr](https://github.com/SubtitleEdit/subtitleedit/blob/master/src/Forms/VobSubOcr.cs#L8501)


ivandrofly:
:+1: @xylographe

PR #909, opened 2015-06-21
Cleanup

ivandrofly:


src/Logic/TransportStream/RegionCompositionSegment.cs
```diff
@@ -49,5 +49,4 @@ public RegionCompositionSegment(byte[] buffer, int index, int regionLength)
             }
         }
     }
-
```
ivandrofly: Outside class

niksedk: What do you mean by "Outside class" ?

ivandrofly: You said  before tab after method close is useful when adding new method

niksedk: Ah, I thought it was something like "inner class" I did not know about ;)

ivandrofly: > Ah, I thought it was something like "inner class" I did not know about ;)

:smile:

PR #915, opened 2015-06-22
Refactor

ivandrofly:


niksedk:
Nice fix with "Clear()" - about twice as fast in my tests :)

PR #928, opened 2015-06-24
Use TimeSpan's static method to get total milliseconds

ivandrofly:


src/Logic/TimeCode.cs
```diff
@@ -95,8 +93,7 @@ public int Minutes
             }
             set
             {
-                var ts = TimeSpan;
-                TimeSpan = new TimeSpan(0, ts.Hours, value, ts.Seconds, ts.Milliseconds);
+                _totalMilliseconds = TimeSpan.FromMinutes(value).TotalMilliseconds;
```
niksedk: I think this is wrong... it's not total Minutes...

ivandrofly: > I think this is wrong... it's not total Minutes...

Total minutes :confused:

niksedk: I think the above code will remove hours, seconds and milliseconds... right?
(I've also added some unit tests regarding this)

ivandrofly: I see, I thought by re-assigning we should 0 out every existing time.
:+1:

niksedk: only for "TotalMinutes" (if that was there) - but it's covered by unit tests now, so that's nice :)

PR #937, opened 2015-06-26
Updated Dutch OCRFixReplaceList

xylographe:
Generalised transformation to fix typical OCR error: lower case `l` instead of upper case `I`.
`\bln(?=\p{Ll}{2})` ==> `In`

### DIFF
--- a/Dictionaries/nld_OCRFixReplaceList.xml
+++ b/Dictionaries/nld_OCRFixReplaceList.xml
@@ -3,14 +3,10 @@
     <Word from="ls" to="Is" />
     <Word from="ln" to="In" />
     <Word from="lk" to="Ik" />
-    <Word from="lndien" to="Indien" />
     <Word from="ledereen" to="Iedereen" />
     <Word from="ledere" to="Iedere" />
     <Word from="lemand" to="Iemand" />
-    <Word from="lnderdaad" to="Inderdaad" />
-    <Word from="lnformatie" to="Informatie" />
     <Word from="lsolement" to="Isolement" />
-    <Word from="lnstructie" to="Instructie" />
     <Word from="ler" to="Ier" />
     <Word from="lerland" to="Ierland" />
     <Word from="lers" to="Iers" />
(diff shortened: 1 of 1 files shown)


Dictionaries/nld_OCRFixReplaceList.xml
```diff
@@ -26,5 +22,7 @@
   <BeginLines />
   <EndLines />
   <WholeLines />
-  <RegularExpressions />
+  <RegularExpressions>
+    <RegEx find="\bln(?=\p{Ll}{2})" replaceWith="In" />
```
ivandrofly: I think is safer `\bln(?=\p{Ll}{4,8}\b)`

xylographe: No, not safer, just incorrect:
```
PS> [regex]::Replace('lnkt','\bln(?=\p{Ll}{4,8}\b)','In')
lnkt
PS> [regex]::Replace('lnkt','\bln(?=\p{Ll}{2})','In')
Inkt
```
```
PS> [regex]::Replace('lnacceptabel','\bln(?=\p{Ll}{2,8}\b)','In')
lnacceptabel
PS> [regex]::Replace('lnacceptabel','\bln(?=\p{Ll}{2})','In')
Inacceptabel
```

ivandrofly: Example I shown above will match according to what was removed... I know it won't match for `lnacceptabel` cause it's > than 8...

xylographe: There are well over 2000 Dutch words starting with "in", not including the (countless) combined words where "in" acts as an amplifier, like "ingelukkig" and "inslecht".

The regex pattern `\bln(?=\p{Ll}{2})` is already quite restrictive, *van Dale* lists exactly two words starting with "ln": "lno" and "ln". Hence, one might argue that the whole-word-transformation `<Word from="ln" to="In" />` is unsafe, but, obviously, that would be “spĳkers op laag water zoeken.”  :smile:

PR #941, opened 2015-06-26
Refactor (src/Logic/Utilities.cs)

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -2734,24 +2734,23 @@ public static string RemoveUnneededSpaces(string text, string language)
             text = text.Replace(operatingSystemCommand, string.Empty);
             text = text.Replace('\t', ' ');
 
-            while (text.Contains("  "))
-                text = text.Replace("  ", " ");
-
-            text = text.Replace(" " + Environment.NewLine, Environment.NewLine);
-            text = text.Replace(Environment.NewLine + " ", Environment.NewLine);
```
ivandrofly: Already defined

PR #959, opened 2015-06-29
Refactor (LoadSpellingDictionaries())

ivandrofly:


src/Logic/OCR/OcrFixEngine.cs
```diff
@@ -86,29 +86,28 @@ private void LoadSpellingDictionaries(string threeLetterIsoLanguageName, string
...
+                }
+                else if (hunspellName.Equals("en_za", StringComparison.OrdinalIgnoreCase) && File.Exists(Path.Combine(dictionaryFolder, "en_ZA.dic")))
+                {
+                    LoadSpellingDictionariesViaDictionaryFileName("eng", new CultureInfo("en-ZA"), "en_ZA.dic", true);
+                }
+                else
+                {
+                    LoadSpellingDictionariesViaDictionaryFileName("eng", new CultureInfo("en-US"), "en_US.dic", true);
+                }
                 return;
```
ivandrofly: We don't need to check more than once if
`!string.IsNullOrEmpty(hunspellName) && threeLetterIsoLanguageName == "eng"`

PR #960, opened 2015-06-30
Add new variable to cache file path + Possible bug fixed

ivandrofly:
- [ ] More test

### DIFF
--- a/src/Logic/Dictionaries/NamesList.cs
+++ b/src/Logic/Dictionaries/NamesList.cs
@@ -13,6 +13,8 @@ namespace Nikse.SubtitleEdit.Logic.Dictionaries
         private readonly HashSet<string> _namesMultiList;
         private readonly string _languageName;
 
+        private string _userFile;
+
         public NamesList(string dictionaryFolder, string languageName, bool useOnlineNamesEtc, string namesEtcUrl)
         {
             _dictionaryFolder = dictionaryFolder;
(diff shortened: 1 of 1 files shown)


src/Logic/Dictionaries/NamesList.cs
```diff
@@ -159,13 +162,13 @@ public bool Remove(string name)
                     _namesMultiList.Remove(name);
 
                 var userList = new HashSet<string>();
-                var fileName = GetLocalNamesUserFileName();
-                LoadNamesList(fileName, userList, userList);
```
ivandrofly: Shouldn't third parameter be `_namesMultiList`? /cc @niksedk

ivandrofly: Shouldn't third parameter be `_namesMultiList`? /cc @niksedk

src/Logic/Dictionaries/NamesList.cs
```diff
@@ -159,13 +162,13 @@ public bool Remove(string name)
...
+                _userFile = _userFile ?? GetLocalNamesUserFileName(); // Try get file name if _userFile if null
```
ivandrofly: No loop needed for every add/remove name :wink:

src/Logic/Dictionaries/NamesList.cs
niksedk: Normally yes... but in "Remove" it's intended to load all entries in just one list

ivandrofly: :+1: Will do quick update then

ivandrofly: Reverted /cc @niksedk

PR #967, opened 2015-07-01
Refactor + Fix

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -1065,10 +1065,9 @@ public static string DictionaryFolder
         public static List<string> GetDictionaryLanguages()
         {
             var list = new List<string>();
-            string dictionaryFolder = DictionaryFolder;
```
ivandrofly: We don't need to store `DictionaryFolder` in a new variable

PR #968, opened 2015-07-01
Reduce memory allocation

ivandrofly:


ivandrofly:
Does Flash Xml as support to SSA tags? @niksedk

PR #977, opened 2015-07-02
Updated Serbian OCRFixReplaceList

xylographe:
Minor fix.

### DIFF
--- a/Dictionaries/srp_OCRFixReplaceList.xml
+++ b/Dictionaries/srp_OCRFixReplaceList.xml
@@ -82,7 +82,6 @@
     <RegEx find="uspijeh" replaceWith="uspjeh" />
     <RegEx find="(u|U)spiješan" replaceWith="$1spješan" />
     <RegEx find="(u|U)vjek" replaceWith="$1vijek" />
-    
     <RegEx find="(?&lt;=[a-zčđšž])Ij(?=[a-zčđšž])" replaceWith="lj" />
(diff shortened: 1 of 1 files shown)


niksedk:
What does '(?&lt;=[0-9])\$' do

ivandrofly:
> What does '(?<=[0-9])\$' do ?
> 
> Edit: Anything wrong with "$1" ?

E.g: it changes `100$` to `100 $`

niksedk:
OK, and that's correct in Serbian I guess.. what does the double $ do?

ivandrofly:
> OK, and that's correct in Serbian I guess.. what does the double $ do?

I did quick test seems like a bug... https://regex101.com/r/aD8oL7/1

xylographe:
```
PS> [regex]::Replace('500$', '(?<=[0-9])\$', ' $$')
500 $
```
[MSDN](https://msdn.microsoft.com/en-us/library/ewy2t5e0%28v=vs.110%29.aspx): “The [**$$**](https://msdn.microsoft.com/en-us/library/ewy2t5e0%28v=vs.110%29.aspx#DollarSign) substitution inserts a literal "$" character in the replaced string.”

Although a single `$` at the end of a replacement pattern (or followed by a space character) will also insert a literal "$" in the replaced string, this behaviour is undocumented by Microsoft. IMO, exploiting undocumented behaviour should be avoided, as the behaviour may change without notice in a future .NET release, while `$$` is guaranteed to work as expected.

niksedk:
thx for the info - always something new to learn :)

PR #978, opened 2015-07-03
[WIP] "Reduce more memory allocation"

ivandrofly:
# Do not merge yet!

### DIFF
--- a/src/Forms/FixCommonErrors.cs
+++ b/src/Forms/FixCommonErrors.cs
@@ -58,7 +58,7 @@ namespace Nikse.SubtitleEdit.Forms
         private bool _hasFixesBeenMade;
 
         private static readonly Regex FixMissingSpacesReComma = new Regex(@"[^\s\d],[^\s]", RegexOptions.Compiled);
-        private static readonly Regex FixMissingSpacesRePeriod = new Regex(@"[a-z][a-z][.][a-zA-Z]", RegexOptions.Compiled);
+        private static readonly Regex FixMissingSpacesRePeriod = new Regex(@"[a-z][a-z]\.[a-zA-Z]", RegexOptions.Compiled);
         private static readonly Regex FixMissingSpacesReQuestionMark = new Regex(@"[^\s\d]\?[a-zA-Z]", RegexOptions.Compiled);
         private static readonly Regex FixMissingSpacesReExclamation = new Regex(@"[^\s\d]\![a-zA-Z]", RegexOptions.Compiled);
         private static readonly Regex FixMissingSpacesReColon = new Regex(@"[^\s\d]\:[a-zA-Z]", RegexOptions.Compiled);
--- a/src/Forms/FixCommonErrors.cs
+++ b/src/Forms/FixCommonErrors.cs
@@ -3722,14 +3722,6 @@ namespace Nikse.SubtitleEdit.Forms
 
                              };
 
-            var regExIDag = new Regex(@"\bidag\b", RegexOptions.Compiled);
-            var regExIGaar = new Regex(@"\bigår\b", RegexOptions.Compiled);
-            var regExIMorgen = new Regex(@"\bimorgen\b", RegexOptions.Compiled);
-            var regExIAlt = new Regex(@"\bialt\b", RegexOptions.Compiled);
-            var regExIGang = new Regex(@"\bigang\b", RegexOptions.Compiled);
-            var regExIStand = new Regex(@"\bistand\b", RegexOptions.Compiled);
-            var regExIOevrigt = new Regex(@"\biøvrigt\b", RegexOptions.Compiled);
-
             for (int i = 0; i < Subtitle.Paragraphs.Count; i++)
             {
                 string text = Subtitle.Paragraphs[i].Text;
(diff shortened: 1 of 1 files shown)


src/Forms/FixCommonErrors.cs
```diff
@@ -3758,26 +3750,13 @@ private void FixDanishLetterI()
...
-
-                if (regExIGang.IsMatch(text))
-                    text = regExIGang.Replace(text, "i gang");
-
-                if (regExIStand.IsMatch(text))
-                    text = regExIStand.Replace(text, "i stand");
-
-                if (regExIOevrigt.IsMatch(text))
-                    text = regExIOevrigt.Replace(text, "i øvrigt");
+                text = Regex.Replace(text, @"\bidag\b", "i dag", RegexOptions.Compiled);
```
ivandrofly: Won't replace if doesn't `Match`

StevenThuriot:
I would personally create a regex instance and cache it instead. `Regex.Replace` will internally just call the following code:

```csharp
return new Regex(pattern, options, matchTimeout, true).Replace(input, evaluator);
```

Thus creating a new instance every run. 

While `RegexOptions.Compiled` will decrease regex execution time severely (**up to 60%**), it will also heavily increase instance creation (**5ish times slower**) as it will internally compile the regex to IL using Lightweight code generation (e.g. [DynamicMethod](https://msdn.microsoft.com/en-us/library/system.reflection.emit.dynamicmethod(v=vs.110).aspx)).

The generated method will be cached, but by default only the last 15 ones used (see [Regex.CacheSize](https://msdn.microsoft.com/en-us/library/system.text.regularexpressions.regex.cachesize.aspx)). Since the cache is a simple linked list, increasing it too much might have a backwards effect as the linkedlist will be traversed each time, which is a slow process in its own. (each time you create a new instance -- thus also when you do a static regex invocation)

Since there are plenty of regex's going around, it will be hard to predict what's in the cache so I would heavily advise against using a static invocation on the regex class.

ivandrofly:
:+1: @StevenThuriot

ivandrofly:
Hmm... I think something can be done by adding new class which extends regex... any other suggestion? /cc @StevenThuriot

StevenThuriot:
You could indeed inherit the `Regex` class since it isn't sealed, but I'm not completely sure if I follow what you would try to accomplish doing so? You won't be able to override the ctor, merely call it from your own code, so you'd basically have to re-implement all the static callers you use. The ctor where all the magic happens is a private one and which is also the endpoint for all the others (they call :this(...) )

Care to explain further what you want to achieve?

------------

Personally, I'd suggest the following:
- Pull the `Regex.Replace` instances back outside of the for-loop so the `linkedlist` doesn't have to be traversed each iteration of the `for-loop`. (assuming they'll still be in the cache, worst case they'll be recompiled again)
- Optimize the regexes created in `iList` to reduce memory instead :arrow_right: **huge gain**.

For instance, the following block of code 

```csharp
MyRegEx(@"\bat i aldrig\b"),
MyRegEx(@"\bat i alle bliver\b"),
MyRegEx(@"\bat i alle er\b"),
MyRegEx(@"\bat i alle forventer\b"),
MyRegEx(@"\bat i alle gør\b"),
MyRegEx(@"\bat i alle har\b"),
MyRegEx(@"\bat i alle ved\b"),
MyRegEx(@"\bat i alle vil\b"),
MyRegEx(@"\bat i bare\b"),
MyRegEx(@"\bat i bager\b"),
MyRegEx(@"\bat i bruger\b"),
MyRegEx(@"\bat i dræber\b"),
MyRegEx(@"\bat i dræbte\b"),
MyRegEx(@"\bat i fandt\b"),
MyRegEx(@"\bat i fik\b"),
MyRegEx(@"\bat i finder\b"),
MyRegEx(@"\bat i forstår\b"),
MyRegEx(@"\bat i får\b"),
MyRegEx(@"\bat i ikke\b"),
MyRegEx(@"\bat i kom\b"),
MyRegEx(@"\bat i kommer\b"),
MyRegEx(@"\bat i næsten er\b"),
MyRegEx(@"\bat i næsten fik\b"),
MyRegEx(@"\bat i næsten har\b"),
MyRegEx(@"\bat i næsten skulle\b"),
MyRegEx(@"\bat i næsten var\b"),
MyRegEx(@"\bat i også får\b"),
MyRegEx(@"\bat i også gør\b"),
MyRegEx(@"\bat i også mener\b"),
MyRegEx(@"\bat i også siger\b"),
MyRegEx(@"\bat i også tror\b"),
MyRegEx(@"\bat i rev\b"),
MyRegEx(@"\bat i river\b"),
MyRegEx(@"\bat i samarbejder\b"),
MyRegEx(@"\bat i snakkede\b"),
MyRegEx(@"\bat i scorer\b"),
MyRegEx(@"\bat i siger\b"),
MyRegEx(@"\bat i skal\b"),
MyRegEx(@"\bat i skulle\b"),
MyRegEx(@"\bat i to ikke\b"),
MyRegEx(@"\bat i to siger\b"),
MyRegEx(@"\bat i to har\b"),
MyRegEx(@"\bat i to er\b"),
MyRegEx(@"\bat i to bager\b"),
MyRegEx(@"\bat i to skal\b"),
MyRegEx(@"\bat i to gør\b"),
MyRegEx(@"\bat i to får\b"),
MyRegEx(@"\bat i udnyttede\b"),
MyRegEx(@"\bat i udnytter\b"),
MyRegEx(@"\bat i vil\b"),
MyRegEx(@"\bat i ville\b"),
```

has a lot of reoccurring pieces, so it can easily be replaced with just one single regex (off the cuff, sorry if I made a typo)

```csharp
MyRegEx(@"\bat i [aldrig|alle bliver|alle er|alle forventer|alle gør|alle har|alle ved|alle vil|bare|bager|bruger|dræber|dræbte|fandt|fik|finder|forstår|får|ikke|kom|kommer|næsten er|næsten fik|næsten har|næsten skulle|næsten var|også får|også gør|også mener|også siger|også tror|rev|river|samarbejder|snakkede|scorer|siger|skal|skulle|to ikke|to siger|to har|to er|to bager|to skal|to gør|to får|udnyttede|udnytter|vil|ville|ikke]\b")
```

That is 51 regexes replaced with one. I did a quick count: the fixes list seems to contain almost 600 regexes and most of them can be grouped in the same way. You can probably easily cut that list from 600 to 10-15. Most of your memory allocation (and quite a bit actually) will be gained here. You could get it all into one single regex, but that might end up hard to read and difficult to maintain. Might be worth it, though, but I will leave that up to you to decide. In that case, the inner for-loop will be gone, too!

As an extra, I personally like to keep them in a separate class like this:

```csharp
    static class SubtitleEditRegex
    {
        public static class DanishLetterI
        {
            public static readonly IReadOnlyCollection<Regex> Fixes;

            static DanishLetterI()
            {
                Fixes = new []
                        {
                            // not a complete list, more phrases will come
                            new Regex(......, RegexOptions.Compiled)
                            ...............
                        };
            }
        }

        public static class AnotherRegexFixingClass
        {
            public static readonly IReadOnlyCollection<Regex> Fixes;
            .........
        }
    }
```

The method would then just call

```csharp
                if (littleIRegex.IsMatch(text))
                {
                    foreach (var regex in SERegex.DanishLetterI.Fixes)
                    {
...
```

The neat thing about the static class with subclasses is that they'll be easy to maintain *(wether using multiple files and `partial` or not)*, serve as a cache and the static ctor's of the subclasses will only be called when the subclass is accessed the first time, not the main class *(unlike a method on the main class)*, so only the ones you need at that time will be compiled. The subclass also makes it somewhat descriptive, like a method would. :smile:

PR #989, opened 2015-07-06
Refactor - #1

ivandrofly:


src/Logic/SubtitleFormats/Json.cs
```diff
@@ -96,13 +96,15 @@ public override string ToText(Subtitle subtitle, string title)
 
         public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string fileName)
         {
+            // Already a json file
+            if (lines.Count > 0 && !lines[0].TrimStart().StartsWith("[{\"start", StringComparison.Ordinal))
+                return;
```
ivandrofly: this will prevent creating new instance of `StringBuilder` as well as entering the `for each loop`

ivandrofly:
Commit removed /cc @niksedk

PR #997, opened 2015-07-11
Refactor (src/Logic/SubtitleFormats/OpenDvt.cs)

ivandrofly:


src/Logic/SubtitleFormats/OpenDvt.cs
```diff
@@ -125,7 +125,7 @@ public override string ToText(Subtitle subtitle, string title)
                 line.AppendChild(qa);
 
                 XmlNode text = xml.CreateElement("Text");
-                text.InnerText = p.Text;
+                text.InnerText = HtmlUtil.RemoveHtmlTags(p.Text, true);
```
ivandrofly: Support `HTML & SSA` tags?

src/Logic/SubtitleFormats/OpenDvt.cs
```diff
@@ -151,25 +151,25 @@ public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string
...
                 try
                 {
-                    Paragraph p = new Paragraph();
+                    var p = new Paragraph();
 
                     XmlNode text = node.SelectSingleNode("Text");
                     if (text != null)
                     {
-                        StringBuilder pText = new StringBuilder();
+                        sb.Clear();
```
ivandrofly: Can be reused without allocated a new one :wink:

PR #999, opened 2015-07-11
Refactor (src/Logic/SubtitleFormats/MidwayInscriberCGX.cs)

ivandrofly:


src/Logic/SubtitleFormats/MidwayInscriberCGX.cs
```diff
@@ -65,7 +65,8 @@ public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string
             //eller bare mere heldige? <12:03:29:03> <12:03:35:06>
 
             subtitle.Paragraphs.Clear();
-            sb = new StringBuilder();
+            sb.Clear();
```
ivandrofly: To avoid re-alloc

PR #1002, opened 2015-07-11
Refactor (src/Logic/Utilities.cs)

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -1125,8 +1113,7 @@ private static int GetCountContains(string text, params string[] words)
             int count = 0;
             for (int i = 0; i < words.Length; i++)
             {
-                var regEx = new Regex(words[i]);
-                count += regEx.Matches(text).Count;
+                count += CountTagInText(text, words[i]);
```
ivandrofly: This will also reduce memory alloc

src/Logic/Utilities.cs
```diff
@@ -63,37 +63,32 @@ private static VideoInfo TryReadVideoInfoViaMatroskaHeader(string fileName)
...
-
-                    info.Width = width;
-                    info.Height = height;
-                    info.FramesPerSecond = frameRate;
-                    info.Success = true;
-                    info.TotalMilliseconds = milliseconds;
-                    info.TotalSeconds = milliseconds / TimeCode.BaseUnit;
-                    info.TotalFrames = info.TotalSeconds * frameRate;
-                    info.VideoCodec = videoCodec;
+                using (matroska = new MatroskaFile(fileName))
```
ivandrofly: Even if exception is thrown `matroska` will still be disposed.

PR #1019, opened 2015-07-15
Fix + Refactor

ivandrofly:


src/Logic/Utilities.cs
```diff
@@ -1099,6 +1099,9 @@ public static double GetOptimalDisplayMilliseconds(string text, double character
             double optimalCharactersPerSecond = charactersPerSecond;
             if (optimalCharactersPerSecond < 2 || optimalCharactersPerSecond > 100)
                 optimalCharactersPerSecond = 14.7;
+
+            // Newline characters (\r\n) shouldn't be counted
+            text = text.Replace(Environment.NewLine, string.Empty);
```
ivandrofly: We shouldn't be counting `\r\n`

PR #1022, opened 2015-07-16
Update RemoveSsaTags

ivandrofly:
@niksedk Anything to say about this?


src/Logic/Utilities.cs
```diff
@@ -838,22 +838,15 @@ public static void InitializeSubtitleFont(Control control)
 
         public static string RemoveSsaTags(string s)
         {
-            int k = s.IndexOf('{');
+            const string tagStart = "{\\";
```
ivandrofly: @niksedk  shouldn't we filter text by "{\\"?

PR #1038, opened 2015-07-22
Updated zh-CHS.xml for 3.4.7

LeonCheung:


LeonCheung:
Just One question: when using it in SE, why the language name is followed by two Chinese characters "旧版" which acually means "old version"?

oldv

xylographe:
See [ChooseLanguage.cs](https://github.com/SubtitleEdit/subtitleedit/blob/master/src/Forms/ChooseLanguage.cs#L77):
```
if (cultureName.Equals("zh-CHS", StringComparison.OrdinalIgnoreCase))
     comboBoxLanguages.Items.Add(new CultureListItem(new CultureInfo(0x4)));
```
Perhaps, it would be better to use the `CultureInfo.CreateSpecificCulture` method instead?
```
PS> $ci = new-object -type System.Globalization.CultureInfo -args 4
PS> $ci.EnglishName; $ci.NativeName
Chinese (Simplified) Legacy
中文(简体) 旧版
```
```
PS> $ci = [System.Globalization.CultureInfo]::CreateSpecificCulture('zh-CHS')
PS> $ci.EnglishName; $ci.NativeName
Chinese (Simplified, PRC)
中文(中华人民共和国)
```

LeonCheung:
Imho, it is highly recommended to claim **zh_CN** as "**简体中文**" (Chinese Simplified), and **zh_TW** as "**繁体中文**" (Chinese Traditional)。Because:

1. It is the general rule among softwares.
2. Not only Taiwan, regions like HongKong and Macao use traditional Chinese all the time as well.
3. PRC mainland also recognizes Traditional Chinese (though not so popular as the simplified version).
4. PAC and ROC are actually belong to "One China", just have individual definitions ([Ref](https://en.wikipedia.org/wiki/1992_Consensus)).

FYI .Thanks.

xylographe:
The `CultureInfo` objects used by SE are defined by Microsoft as part of .NET (see [National Language Support (NLS) API Reference](https://msdn.microsoft.com/en-us/goglobal/bb896001.aspx)). **zh-CHS** is an alias for **zh-CN** (`LCID 2052 [0x0804]`).

It would seem the current behaviour is what you want: **zh-CHS** (from the file name `zh-CHS.xml`) is redirected to **zh-Hans** (`LCID 4`— **中文(简体)**).

Perhaps, it might be cleaner to rename the translation files to `zh-Hans.xml` and `zh-Hant.xml`?

LeonCheung:
Well clear, that's informative, thanks. And yeah, it is acceptable I suppose.

xylographe:
@LeonCheung Please, tell me if these Chinese names are correct:
- zh-Hans (Simplified Chinese)  : 中文(简体)
- zh-Hant (Traditional Chinese) : 中文(繁體)

LeonCheung:
@xylographe  That is corrent. :)

PR #1042, opened 2015-07-22
:octocat: +screenshot+download button

diomed:


diomed:
@niksedk IMHO we should have some kind of screenshot and download button on front page, 
but if you don't like the idea & don't intend to merge it I shall close the pull request.

PR #1048, opened 2015-07-23
Cache Regex context menu

ivandrofly:


src/Forms/MultipleReplace.cs
```diff
@@ -89,6 +90,8 @@ public MultipleReplace()
 
             radioButtonCaseSensitive.Left = radioButtonNormal.Left + radioButtonNormal.Width + 40;
             radioButtonRegEx.Left = radioButtonCaseSensitive.Left + radioButtonCaseSensitive.Width + 40;
+
+            _regexContextMenu = FindReplaceDialogHelper.GetRegExContextMenu(textBoxFind);
```
ivandrofly: @niksedk this is going to optimize a bit, as you can see each time user right clicked `FindReplaceDialogHelper.GetRegExContextMenu(textBoxFind);` is going to be called so why not cache

PR #1049, opened 2015-07-24
Improve (src/Forms/MultipleReplace.cs) a bit

ivandrofly:
/cc @niksedk

### DIFF
--- a/src/Forms/MultipleReplace.cs
+++ b/src/Forms/MultipleReplace.cs
@@ -11,7 +11,6 @@ namespace Nikse.SubtitleEdit.Forms
 {
     public sealed partial class MultipleReplace : PositionAndSizeForm
     {
-
         internal class ReplaceExpression
         {
             internal const int SearchNormal = 0;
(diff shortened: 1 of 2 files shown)


src/Forms/MultipleReplace.cs
```diff
@@ -186,43 +185,32 @@ private void GeneratePreview()
...
-                                newText = newText.Remove(index, item.FindWhat.Length).Insert(index, item.ReplaceWith);
-                                index = newText.IndexOf(item.FindWhat, index + item.ReplaceWith.Length, StringComparison.OrdinalIgnoreCase);
-                            }
-                            while (index >= 0);
+                            newText = newText.Remove(index, item.FindWhat.Length).Insert(index, item.ReplaceWith);
+                            index = newText.IndexOf(item.FindWhat, index + item.ReplaceWith.Length, StringComparison.OrdinalIgnoreCase);
                         }
                     }
                 }
-                if (hit && newText != p.Text)
```
ivandrofly: If not hit `newText` won't be equals to `p.Text`

xylographe: See PR #640

ivandrofly: > See PR #640

:+1:

src/Forms/MultipleReplace.cs
```diff
@@ -210,15 +209,11 @@ private void GeneratePreview()
                     else
                     {
                         int index = newText.IndexOf(item.FindWhat, StringComparison.OrdinalIgnoreCase);
-                        if (index >= 0)
+                        hit = index >= 0;
```
ivandrofly: I think this should also fix a bug

PR #1050, opened 2015-07-24
Allow resizing 'Text' Column instead 'Gap to next in seconds'

ivandrofly:
This is really useful, most of the form are using SE's list view this patch will make SE Listview more flexible

### DIFF
--- a/src/Controls/SubtitleListView.cs
+++ b/src/Controls/SubtitleListView.cs
@@ -1108,5 +1108,12 @@ namespace Nikse.SubtitleEdit.Controls
             }
         }
 
+        public void SetCustomResize(EventHandler handler)
+        {
+            if (handler == null)
+                return;
+            Resize -= SubtitleListViewResize;
+            Resize += handler;
+        }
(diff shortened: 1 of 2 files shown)


src/Controls/SubtitleListView.cs
```diff
@@ -1104,5 +1104,12 @@ public void ShowAllColumns()
             }
         }
 
+        public void SetCustomResize(EventHandler handler)
+        {
+            if (handler == null)
+                return;
+            Resize -= SubtitleListViewResize;
+            Resize += handler;
```
ivandrofly: Remove default re-size method and add our new one

ivandrofly:
Any thoughts /cc @niksedk

niksedk:
Looks good :)

ivandrofly:
:wink:  Seems like we are having merge conflict, let me rebase first

niksedk:
super

ivandrofly:
Done!

PR #1053, opened 2015-07-25
Refactor subtitle not loaded message

ivandrofly:


src/Forms/Main.cs
```diff
@@ -7968,7 +7972,12 @@ private void UpdateListSyntaxColoring()
             if (_loading)
                 return;
 
-            if (!IsSubtitleLoaded || _subtitleListViewIndex < 0 || _subtitleListViewIndex >= _subtitle.Paragraphs.Count)
```
niksedk: I don't think this is correct...

src/Forms/Main.cs
```diff
@@ -7965,10 +7969,7 @@ private void StartUpdateListSyntaxColoring()
 
         private void UpdateListSyntaxColoring()
         {
-            if (_loading)
-                return;
-
-            if (!IsSubtitleLoaded || _subtitleListViewIndex < 0 || _subtitleListViewIndex >= _subtitle.Paragraphs.Count)
+            if (_loading || !IsSubtitleLoaded || _subtitleListViewIndex < 0 || _subtitleListViewIndex >= _subtitle.Paragraphs.Count)
```
ivandrofly: @niksedk :+1: ?

PR #1062, opened 2015-07-26
Remove native Subrip formatting tags

ivandrofly:
https://en.wikipedia.org/wiki/SubRip

### DIFF
--- a/src/Logic/SubtitleFormats/SubRip.cs
+++ b/src/Logic/SubtitleFormats/SubRip.cs
@@ -275,5 +275,17 @@ namespace Nikse.SubtitleEdit.Logic.SubtitleFormats
                 return new List<string> { ".wsrt" };
             }
         }
+
+        public override void RemoveNativeFormatting(Subtitle subtitle, SubtitleFormat newFormat)
+        {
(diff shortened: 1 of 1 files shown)


src/Logic/SubtitleFormats/SubRip.cs
```diff
@@ -275,5 +275,19 @@ public override List<string> AlternateExtensions
...
+            var count = subtitle.Paragraphs.Count;
+            for (int i = 0; i < count; i++)
+            {
+                var p = subtitle.Paragraphs[i];
+                var text = regex.Replace(p.Text, string.Empty);
+                if (text != p.Text)
+                    p.Text = text;
+            }
+        }
     }
```
xylographe: Or, perhaps,
```
foreach (var p in subtitle.Paragraphs)
{
    p.Text = regex.Replace(p.Text, string.Empty);
}
```
[MSDN](https://msdn.microsoft.com/en-us/library/xwewhkd1%28v=vs.110%29.aspx): *If the regular expression pattern is not matched in the current instance, the method returns the current instance unchanged.*

PR #1089, opened 2015-07-31
Return if text doesn't contain '<'

ivandrofly:
`text.IndexOfAny(tags, StringComparison.Ordinal);` is little bit expensive

### DIFF
--- a/src/Core/HtmlUtil.cs
+++ b/src/Core/HtmlUtil.cs
@@ -402,7 +402,7 @@ namespace Nikse.SubtitleEdit.Core
 
         internal static string FixUpperTags(string text)
         {
-            if (string.IsNullOrEmpty(text))
+            if (string.IsNullOrEmpty(text) || !text.Contains('<'))
                 return text;
             var tags = new string[] { "<I>", "<U>", "<B>", "<FONT", "</I>", "</U>", "</B>", "</FONT>" };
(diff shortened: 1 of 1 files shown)


src/Core/HtmlUtil.cs
```diff
@@ -402,7 +402,7 @@ public static bool StartsWithUrl(string text)
 
         internal static string FixUpperTags(string text)
         {
-            if (string.IsNullOrEmpty(text))
+            if (string.IsNullOrEmpty(text) && !text.Contains('<'))
```
niksedk: should that have been "|| !text.Contains('<')" ?

ivandrofly:
Fixed /cc @niksedk

ivandrofly:
Safe /cc @niksedk

niksedk:
I just could not see any performance difference - even at 10.000.000 tries...
Moving the string array out actually gave a larger improvement - even if the input string does not contain '<':
private static readonly string[] Tags = { "<I>", "<U>", "<B>", "<FONT", "</I>", "</U>", "</B>", "</FONT>" };

ivandrofly:
Okay, I thought it wouldn't bee necessary to waste time checking for all the tags

niksedk:
It's no waste of time... but don't optimize without reason :)

ivandrofly:
:+1:

PR #1095, opened 2015-08-01
Fix show early closes when ESC is pressed

ivandrofly:
# Status
- [x] Ready


src/Forms/ShowEarlierLater.cs
```diff
@@ -38,7 +38,7 @@ private void ShowEarlierLater_KeyDown(object sender, KeyEventArgs e)
         {
             if (e.KeyCode == Keys.Escape)
             {
-                DialogResult = DialogResult.Cancel;
```
ivandrofly: All the instance was called with `.Show(This)`

PR #1096, opened 2015-08-02
BUGFIXED + Optimization

ivandrofly:
@niksedk check this out

### DIFF
--- a/src/Logic/Forms/RemoveTextForHI.cs
+++ b/src/Logic/Forms/RemoveTextForHI.cs
@@ -97,7 +97,7 @@ namespace Nikse.SubtitleEdit.Logic.Forms
             foreach (string line in lines)
             {
                 int indexOfColon = line.IndexOf(':');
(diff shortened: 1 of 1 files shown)


src/Logic/Forms/RemoveTextForHI.cs
```diff
@@ -97,7 +97,7 @@ public string RemoveColon(string text)
             foreach (string line in lines)
             {
                 int indexOfColon = line.IndexOf(':');
-                if (indexOfColon > 0 && IsNotInsideBrackets(text, indexOfColon))
+                if (indexOfColon > 0 && IsNotInsideBrackets(line, indexOfColon))
```
ivandrofly: @niksedk you need to look at this.... isn't this a BUG?
`indexOfColon` as 'nothing' to do with `text`

xylographe: Yes, this is a bug. `indexOfColon` is an index into `line`, not into `text`.

niksedk:
Nice catch :)

PR #1100, opened 2015-08-02
Fix continuous exception while loading subtitle formats + remove redu…

ivandrofly:
…ndant TryCatch


ivandrofly:
@niskedk they can be seen by running Visual Studio 2015 Enterprise (Diagnostic Tool)

PR #1114, opened 2015-08-06
Some simple Dutch RegEx additions.

Betsy25:
Some simple Dutch RegEx additions to nld_OCRFixReplaceList.xml

### DIFF
--- a/Dictionaries/nld_OCRFixReplaceList.xml
+++ b/Dictionaries/nld_OCRFixReplaceList.xml
@@ -6,14 +6,6 @@
     <Word from="ledereen" to="Iedereen" />
     <Word from="ledere" to="Iedere" />
     <Word from="lemand" to="Iemand" />
-    <Word from="lsolement" to="Isolement" />
-    <Word from="ler" to="Ier" />
-    <Word from="lerland" to="Ierland" />
-    <Word from="lers" to="Iers" />
-    <Word from="lerse" to="Ierse" />
-    <Word from="eIk" to="elk" />
-    <Word from="eIkaar" to="elkaar" />
-    <Word from="eIke" to="elke" />
   </WholeWords>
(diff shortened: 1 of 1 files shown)


Dictionaries/nld_OCRFixReplaceList.xml
```diff
@@ -23,6 +15,11 @@
   <EndLines />
   <WholeLines />
   <RegularExpressions>
+    <RegEx find="\blk(?=\p{Ll}{2})" replaceWith="Ik" />
     <RegEx find="\bln(?=\p{Ll}{2})" replaceWith="In" />
+    <RegEx find="\bls(?=\p{Ll}{2})" replaceWith="Is" />
+    <RegEx find="\beIk(aar|e|)" replaceWith="elk\1" />
+    <RegEx find="\blsol(atie|eerde|eert|eer|eren|ement)" replaceWith="Isol\1" />
+    <RegEx find="\bler(land|se|s|)\b" replaceWith="Ier\1" />
```
xylographe: Replace `\1` with `$1`. See [MSDN](https://msdn.microsoft.com/en-us/library/az24scfc#substitutions)

Betsy25: Sorry, I'm really struggling with all those different RegEx flavors :blush:

xylographe: The .NET regex engine understands both, but the former is undocumented. I suppose Microsoft wanted backward compatibility, yet, keep the option to drop it open.

Just, a minor optimisation:
```
<RegEx find="\beIk" replaceWith="elk" />
<RegEx find="\beIk(aar|e|)" replaceWith="elk$1" />
```
Both perform the  same transformation, but the former is more efficient (no capture).

PR #1128, opened 2015-08-09
Refactor subtitle formats part 2

ivandrofly:


src/Logic/SubtitleFormats/TMPlayer.cs
```diff
@@ -59,7 +59,7 @@ public override string ToText(Subtitle subtitle, string title)
             var sb = new StringBuilder();
             foreach (Paragraph p in subtitle.Paragraphs)
             {
-                string text = HtmlUtil.RemoveHtmlTags(p.Text);
+                var text = HtmlUtil.RemoveHtmlTags(p.Text, true);
```
ivandrofly: Has support for Ssa?

PR #1144, opened 2015-08-12
Refactor (src/Core/HtmlUtil.cs)

ivandrofly:


ivandrofly:
This is safe /cc @niksedk

PR #1149, opened 2015-08-14
Refact uncaught exception handlers (Program)

xylographe:
Direct to SE GitHub issues page in error message.

Perhaps, instead of a MessageBox, use a custom Form with Report button that opens issue page in  browser after copying error+stacktrace to clipboard? (and only Abort/Ignore buttons, as Retry is meaningless in this context)

### DIFF
--- a/src/Program.cs
+++ b/src/Program.cs
@@ -28,24 +28,45 @@ namespace Nikse.SubtitleEdit
             Application.Run(new Main());
         }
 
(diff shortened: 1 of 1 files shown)


niksedk:
Good idea with a custom form :)
The issue page actually accepts parameters - like this: https://github.com/SubtitleEdit/subtitleedit/issues/new?title=crash&body=test-dummy

This could potentially create a large number of issues on the same bug...

PR #1151, opened 2015-08-15
Updated Croatian/Serbian language detection (word frequency)

aaaxx:
Replaced the word lists for Croatian and Serbian with new lists consisting of the most frequent words that have both Croatian and Serbian variant. This should hopefully stop wrong detections of one as the other.

I used the frequency list created from opensubtitles.org dump (2012), available here: https://invokeit.wordpress.com/frequency-word-lists/

### DIFF
--- a/src/Logic/Utilities.cs
+++ b/src/Logic/Utilities.cs
@@ -1189,10 +1189,21 @@ namespace Nikse.SubtitleEdit.Logic
         public static readonly string[] AutoDetectWordsRomanian = { "Какво", "тук", "може", "Как", "Ваше", "какво" };
         public static readonly string[] AutoDetectWordsArabic = { "Какво", "тук", "може", "Как", "Ваше", "какво" };
         public static readonly string[] AutoDetectWordsHebrew = { "אתה", "אולי", "הוא", "בסדר", "יודע", "טוב" };
-        public static readonly string[] AutoDetectWordsSerbian = { "sam", "što", "öto", "äto", "ovo", "vas", "nije", "Šta", "ovde", "za" };
+        public static readonly string[] AutoDetectWordsCroatianAndSerbian = { "sam", "ali", "nije", "samo", "ovo", "kako", "dobro", "sve", "tako", "će", "mogu", "ću", "zašto", "nešto", "za" };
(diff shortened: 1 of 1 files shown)


aaaxx:
Nikse, you'll just have to add Serbian detection to `public static string AutoDetectLanguageName(string languageName, Subtitle subtitle)` in the same way as British vs. American English works. From what I've figured out that is used specifically for detection of spelling dictionaries, so use "sr-Latn".

I would have done it myself but was afraid of messing something up.

src/Logic/Utilities.cs
```diff
@@ -1559,7 +1566,7 @@ public static string AutoDetectLanguageName(string languageName, Subtitle subtit
                         }
                         break;
                     case "hr_HR": // Croatian
-                        count = GetCount(text, AutoDetectWordsCroatian);
+                        count = GetCount(text, AutoDetectWordsCroatianAndSerbian);
                         if (count > bestCount)
```
xylographe: Won't this auto-detect Serbian Latin subtitles as "hr-HR" (Croatian)?

xylographe:
The specific culture name is "sr-Latn-RS",
```
PS> $ci = [System.Globalization.CultureInfo]::CreateSpecificCulture('sr-Latn-RS')
PS> $ci.EnglishName; $ci.NativeName
Serbian (Latin, Serbia)
srpski (Srbija)
```

src/Logic/Utilities.cs
aaaxx: It will but it's only temporary until Serbian detection is added as I said in my first comment. I don't have the knowledge to implement it myself. Besides, the current list also contains words from both languages so the behaviour is still the same.

aaaxx:
But the dictionary files that get downloaded from OpenOffice are named:
* `sr.dic` (list of words in Cyrillic script)
* `sr.aff` (affix rules in Cyrillic script)
* `sr-Latn.dic` (list of words in Latin script)
* `sr-Latn.aff` (affix rules in Latin script)

xylographe:
I think, the conversion from OO name to IETF tag should be tackled at the root, i.e. after downloading. Then the rest of SE can use CultureInfo.Name without need for exceptions.
```
PS> $ci = [System.Globalization.CultureInfo]::CreateSpecificCulture('sr')
PS> $ci.Name; $ci.EnglishName; $ci.NativeName
sr-Latn-RS
Serbian (Latin, Serbia)
srpski (Srbija)
```

PR #1157, opened 2015-08-17
Updated Croatian/Serbian language detection (Logic/Utilities)

xylographe:
[1]
Added "sr-Latn" detection
Fixed "en_GB" detection
[2]
Changed Croatian/Serbian detection words (PR #1151)


aaaxx:
Thanks.

xylographe:
One of them detects the locally available dictionaries (`Dictionaries\*.dic`), then tests the subtitle text only for the corresponding languages. The other one doesn't care if dictionaries or word/fix lists exist. The latter is used in combination with a detection based on subtitle encoding.

Dictionaries and NamesEtc lists are tied to a specific culture (language-script-region/country), other word/fix lists can be identified by a neutral culture (language-script w/o region/country).

The [CultureInfo Class docu at MSDN](https://msdn.microsoft.com/en-us/library/system.globalization.cultureinfo(v=vs.110).aspx) has some info in the **Remarks** section (search for *Invariant, neutral, and specific cultures*).

Issue #1159, opened 2015-08-18
Uninstallation


ivandrofly:
When uninstalling Subtitle Edit even after accepting to delete all the file present in **"%APPDATA%\Subtitle Edit\"** **Plugins** folder won't be delete.

I know it's okay to keep the plugin files if you plan to re-install Subtitle Edit

PR #1160, opened 2015-08-18
[Installer] Remove Plugins folder before Subtitle Edit's

ivandrofly:
Patch for PR ==> https://github.com/SubtitleEdit/subtitleedit/issues/1159

Subtitle Edit folder won't be removed unless its empty


niksedk:
nice :)

PR #1161, opened 2015-08-18
[Internal] Set and break the loop if selection choice is SelectionOnly

ivandrofly:


src/Forms/Main.cs
```diff
@@ -13533,16 +13533,18 @@ public void ShowEarlierOrLater(double adjustMilliseconds, SelectionChoice select
 
             for (int i = startFrom; i < _subtitle.Paragraphs.Count; i++)
             {
-                switch (selection)
+                if (selection == SelectionChoice.SelectionOnly)
                 {
-                    case SelectionChoice.SelectionOnly:
-                        if (SubtitleListview1.Items[i].Selected)
-                            ShowEarlierOrLaterParagraph(adjustMilliseconds, i);
-                        break;
```
ivandrofly: This break will only exit the switch statement /cc @niksedk

PR #1163, opened 2015-08-18
[internal] Remove duplicated assignment

ivandrofly:
Already done inside if statement


niksedk:
Yeah, even looks like a small bug

PR #1166, opened 2015-08-19
[Internal] - Fix translate from/to same language

ivandrofly:


ivandrofly:
@niksedk can you test this?

PR #1172, opened 2015-08-20
Update Ukrainian localization

Maximaximum:


ivandrofly:
@Maximaximum I think you might need to squash the commits

xylographe:
But if he had squashed them we could not have seen diffs without fetching the branch.
Together they are above the GitHub limit (maximum size of raw diff per file).

PR #1173, opened 2015-08-21
Fix for (src/Forms/Main.cs)

ivandrofly:


src/Forms/Main.cs
```diff
@@ -7748,7 +7749,6 @@ private void MergeSelectedLines()
                             return;
                         next++;
                     }
-                    first = false;
```
ivandrofly: will be re-assigned every-time loop run

src/Forms/Main.cs
```diff
@@ -10049,8 +10054,7 @@ private void ImportAndOcrVobSubSubtitleNew(string fileName, bool showInTaskbar)
 
         private void ToolStripMenuItemMergeLinesClick(object sender, EventArgs e)
         {
-            if (_subtitle.Paragraphs.Count > 0 && SubtitleListview1.SelectedItems.Count >= 1)
-                MergeAfterToolStripMenuItemClick(null, null);
+            MergeAfterToolStripMenuItemClick(null, null);
```
ivandrofly: Check: `if (_subtitle.Paragraphs.Count > 0 && SubtitleListview1.SelectedItems.Count >= 1)` already done inside
`MergeAfterToolStripMenuItemClick(null, null);`

PR #1180, opened 2015-08-23
[feature] add new feature to Find Dialog: "Count Matches"

ivandrofly:
# Status
- [x] Regex texted
     - [x] Valid regex.
     - [x] Invalid regex.
- [x] Large subtitle file (10k lines)


src/Forms/FindDialog.cs
```diff
@@ -178,9 +180,16 @@ internal void SetIcon(Bitmap bitmap)
             if (bitmap != null)
             {
                 IntPtr Hicon = bitmap.GetHicon();
-                this.Icon = Icon.FromHandle(Hicon);
```
ivandrofly: Couldn't leave this ;)

niksedk:
Rebase?

Have you tested with really large subtitle + regex?

ivandrofly:
> Rebase?
> 
> Have you tested with really large subtitle + regex?

:+1: Check status

ivandrofly:
@niksedk  can be merged

niksedk:
Does this work when just hitting ctrl+f and typing something? (I could not see anything)

ivandrofly:
> Does this work when just hitting ctrl+f and typing something? (I could not see anything)

:+1:

niksedk:
I could not see anything...

ivandrofly:
> I could not see anything...

What do you mean? o.O

ivandrofly:
@niksedk can you try again?

niksedk:
What steps do I do to see anything? Still cannot see any info / text!?

ivandrofly:
> What steps do I do to see anything? Still cannot see any info / text!?

You will have enter text and click count it will display how many repetition exist in current subtitle.

niksedk:
Ah, works nicely :)
Could you add the two localization texts also?

niksedk:
and the button text...

ivandrofly:
> and the button text...

Done

xylographe:
Multi-lang `FindDialog` texts:
```
    WholeWord = "Whole word",
    Count = "Count",
    XNumberOfMatches = "{0} matches"
```

ivandrofly:
:+1:  @xylographe

xylographe:
**UpdateLanguageFiles** is still generating `src\Logic\LanguageDeserializer.cs` instead of `libse\LanguageDeserializer.cs`.

I fixed `libse\LanguageDeserializer.cs` manually, but this needs to be fixed in `build_helpers.bat` and `src\UpdateLanguageFiles\LanguageDeserializerGenerator.cs`.

xylographe:
I think, the order of the buttons could be

- Find
- Count
- Cancel

Or, better, only **Find** and **Count**, because the close box in this small dialogue is near the buttons.

niksedk:
cool, some minor stuff was missing which should be fixed now - also the UpdateLanguageFiles should be fixed now, thx :)

PR #1181, opened 2015-08-23
Update (src/Logic/Forms/RemoveTextForHI.cs)

ivandrofly:
;) /cc @niksedk


src/Logic/Forms/RemoveTextForHI.cs
```diff
@@ -697,7 +697,7 @@ public string RemoveInterjections(string text)
                         if (!_interjectionList.Contains(upper))
                             _interjectionList.Add(upper);
 
-                        string pascalCasing = char.ToUpper(s[0]) + s.Substring(1);
+                        string pascalCasing = char.ToUpper(s[0]) + lower.Substring(1);
```
ivandrofly: What if `s` is already upper case

PR #1183, opened 2015-08-23
Complete Ukrainian localization

Maximaximum:
This update contains a complete Ukrainian UI localization, which is also up to date with the current version of LanguageMaster.xml


niksedk:
Great  :)

PR #1187, opened 2015-08-24
Update LanguageMaster.xml

LeonCheung:
small fix


xylographe:
LanguageMaster.xml is an auto-generated file, based on [Logic/Language.cs](https://github.com/SubtitleEdit/subtitleedit/blob/master/src/Logic/Language.cs#L1002).

PR #1190, opened 2015-08-24
Let ChooseLanguage not depend on installed Windows cultures

xylographe:
Because not every Windows system has all cultures installed, it would be better not to depend on them just to select a language. The information is also available in the xml translation files.

[1]
Fix CultureName and Language Name attribute values in translations, where necessary.
[2]
Language.Name is a special case (attribute) in the LanguageDeserializer.
[3]
Replace unnecessary use of (possibly missing) CultureInfo in Main.
[4]
Use custom TranslationInfo class instead of CultureInfo in ChooseLanguage.

### DIFF
--- a/src/Languages/pt-BR.xml
+++ b/src/Languages/pt-BR.xml
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Language Name="Brazilian Portuguese">
+<Language Name="Português (Brasil)">
   <General>
     <Title>Subtitle Edit</Title>
     <Version>3.4.8</Version>
--- a/src/Forms/ChooseLanguage.cs
+++ b/src/Forms/ChooseLanguage.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Windows.Forms;
 using Nikse.SubtitleEdit.Logic;
 using System.Xml;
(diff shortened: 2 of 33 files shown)


src/Languages/pt-BR.xml
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Language Name="Brazilian Portuguese">
+<Language Name="Português (Brasil)">
```
ivandrofly: Isn't this supposed to be lowercase `p`? /cc @xylographe

xylographe: I used the native names provided by Windows:
```
PS> [System.Globalization.CultureInfo]::GetCultureInfo('pt-BR').NativeName
Português (Brasil)
PS> [System.Globalization.CultureInfo]::GetCultureInfo('pt-PT').NativeName
português (Portugal)
```
You might get better data from [CLDR.](http://cldr.unicode.org/)
Download core.zip from [CLDR Releases/Downloads/latest-data.](http://unicode.org/Public/cldr/latest)

I was surprised to see
```
PS> [System.Globalization.CultureInfo]::GetCultureInfo('nl-NL').NativeName
Nederlands (Nederland)
```
 because we (Dutch) definitely do *not* capitalise language names when used in a sentence.

ivandrofly:
I did get different result:
``` powershell
```

xylographe:
Interesting! Win10, I presume?

Perhaps, the capital **P** in Win7 was incorrect, and has now been fixed.

Anyway, I think, the Language Name attribute value in `pt-BR.xml` can be changed to "português (Brasil)". I'll put that new value in my SE-repo maintenance script.

ivandrofly:
Sounds good

PR #1191, opened 2015-08-25
Fix (src/Forms/FixCommonErrors.cs)

ivandrofly:
# Status
- [x] Ready
- [x] Test

### DIFF
--- a/src/Forms/FixCommonErrors.cs
+++ b/src/Forms/FixCommonErrors.cs
@@ -4799,10 +4799,13 @@ namespace Nikse.SubtitleEdit.Forms
                 Next();
 
                 // restore de-selected fixes
-                foreach (ListViewItem item in listViewFixes.Items)
+                if (deSelectedFixes.Count > 0)
                 {
-                    if (deSelectedFixes.Contains(item.SubItems[1].Text + item.SubItems[2].Text + item.SubItems[3].Text))
-                        item.Checked = false;
+                    foreach (ListViewItem item in listViewFixes.Items)
+                    {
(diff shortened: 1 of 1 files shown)


src/Forms/FixCommonErrors.cs
```diff
@@ -4776,7 +4772,8 @@ private void ToolStripMenuItemDeleteClick(object sender, EventArgs e)
                         int number = Convert.ToInt32(item.SubItems[1].Text);
                         if (number > firstIndex)
                             number -= subtitleListView1.SelectedItems.Count;
-                        deSelectedFixes.Add(number + item.SubItems[2].Text + item.SubItems[3].Text);
```
ivandrofly: Note that `number` can be `>` than `firstIndex` but we are not sure if `subtitleListView1.SelectedItems.Count` if going to be `<` than number so  `number -= subtitleListView1.SelectedItems.Count` can be negative

src/Forms/FixCommonErrors.cs
```diff
@@ -4802,10 +4799,13 @@ private void ToolStripMenuItemDeleteClick(object sender, EventArgs e)
                 Next();
 
                 // restore de-selected fixes
-                foreach (ListViewItem item in listViewFixes.Items)
+                if (deSelectedFixes.Count > 0)
```
ivandrofly: Can be *0* if `listViewFixes.SelectedItem` was included in deletion, so there is no need to run the *0(n)*

niksedk:
I've also tested some basic stuff (testing is most important:)

ivandrofly:
Indeed :)

Issue #675, opened 2015-03-28
Replacing default icon?


diomed:
I noticed how gray and outdated icon that represents program looks. 

So, I took the liberty of asking icon designer **Philippe Hong** make new, better one.
He agreed and only asked to credit him.

I hope you'll accept this, and replace that old gray icon with new, modern one

![subtiteeditor](https://cloud.githubusercontent.com/assets/72777/6882505/637f9d32-d587-11e4-9c30-b2e4c8516fe8.png)

My apologies, I don't know how to add .ico through github, so I hosted it on box
.ico file is here:
https://app.box.com/s/ookg1eksrx8q95yyd4q7jm2498qzlqws

diomed:
@niksedk  What are your thoughts on this suggestion?

niksedk:
Sorry, I'm not really too impressed... does not really look like an icon. Also tried to add it to an exe file - did not look pretty...

diomed:
@niksedk 

How about this one:
![Imgur](http://i.imgur.com/YNV7Onm.png)

https://app.box.com/s/5k9y6swf9lqoknp9brmpuffosftr7dhd

Betsy25:
I like it.

diomed:
@niksedk , please.

niksedk:
Much better than the previous one... still the "SE" text could be a little bigger and the bottom corners looks weird.

diomed:
corners look weird only on png, not on small icon. you can't notice them on smaller sizes on screen.

I doubt anyone uses extra large icon size. on large size it looks okay.

niksedk:
Also just got this one:
![Imgur](SubtitleEdit Icon-1024.png)

LeonCheung:
@niksedk Nice. BTW, your avatar image can be another nice option. LOL

niksedk:
He, yes - just took the worst image I could find - I think I'm eating :)

diomed:
How about this one

![se ico](https://cloud.githubusercontent.com/assets/72777/11014539/4f09d8f4-853b-11e5-9489-5c494ded63e7.png)

[icon file](https://app.box.com/s/6ckef7bh6aeq4md540jep9thumdnxxh2)

diomed:
3 years have passed since my suggestion, latest version being this one, sent to you via gitter in January:
![SE icon](https://files.gitter.im/SubtitleEdit/subtitleedit/nfDe/s.e.png)

[...]

PR #1204, opened 2015-08-28
Replace SE Application & Form icon

diomed:
lets have new icon in 3.4.9. and if there are some serious changes that need to be done with it,  
we can improve it in 3.5. 

p.s. upon testing it displays no visible shadows, since it's never displayed in that big size in real work environment.
SE
(https://github.com/SubtitleEdit/subtitleedit/issues/675)


niksedk:
It's not bad, but still it could be improved a bit as I said in the issue https://github.com/SubtitleEdit/subtitleedit/issues/675

PR #1205, opened 2015-08-29
Move packages.config ==> packages/packages.config

xylographe:
Testing...

Something like this should do it. (*AppVeyor build succeeded*)


xylographe:
[2]
Not sure why `packages.config` is in the MSBuild project file—can't harm to change that too.

niksedk:
Is this OK?

xylographe:
Yes, it is (after updating `appveyor.yml` [cache dependencies](http://www.appveyor.com/docs/build-cache#cache-dependencies) and squashing)
Note: `src/packages/repositories.config` is NuGet auto-generated and not under Git control.

However, vstest reports a warning:
```
Test Run deployment issue: The assembly or module 'zlib.net' directly or indirectly
 referenced by the test container  src\test\bin\release\test.dll  was not found.
```
This happens in both master and this branch. Hence, it wouldn't be related to moving `packages.config`, but you might want to have a closer look first.

PR #1206, opened 2015-08-29
[internal] - Use array init instead .ToArray()

ivandrofly:


libse/Utilities.cs
```diff
@@ -24,7 +24,7 @@ public static class Utilities
         /// <summary>
         /// Cached environment new line characters for faster lookup.
         /// </summary>
-        public static readonly char[] NewLineChars = Environment.NewLine.ToCharArray();
+        public static readonly char[] NewLineChars = { '\r', '\n' };
```
xylographe: The original is platform independent.

ivandrofly: I know, but the thing is even for **unix** user **NewLineChars* isn't going to 'cause problem

PR #1208, opened 2015-08-30
[BUGFIX] - Validate regex to avoid unexpected exception

ivandrofly:
To see to problem add:
``` xml
  <RegularExpressions>
    <RegEx find="(" replaceWith="I" />
  </RegularExpressions>
```

Filename: **eng_OCRFixReplaceList.xml**

### DIFF
--- a/libse/Dictionaries/OcrFixReplaceList.cs
+++ b/libse/Dictionaries/OcrFixReplaceList.cs
@@ -85,21 +85,16 @@ namespace Nikse.SubtitleEdit.Core.Dictionaries
         private static Dictionary<string, string> LoadReplaceList(XmlDocument doc, string name)
         {
             var list = new Dictionary<string, string>();
-            if (doc.DocumentElement != null)
+            if (!IsValidXmlDocument(doc, name))
+                return list;
+            foreach (XmlNode item in doc.SelectSingleNode(name).ChildNodes)
             {
-                XmlNode node = doc.DocumentElement.SelectSingleNode(name);
-                if (node != null)
+                if (item.Attributes != null && item.Attributes["to"] != null && item.Attributes["from"] != null)
                 {
-                    foreach (XmlNode item in node.ChildNodes)
-                    {
-                        if (item.Attributes != null && item.Attributes["to"] != null && item.Attributes["from"] != null)
-                        {
-                            string to = item.Attributes["to"].InnerText;
-                            string from = item.Attributes["from"].InnerText;
-                            if (!list.ContainsKey(from))
-                                list.Add(from, to);
-                        }
-                    }
+                    string to = item.Attributes["to"].InnerText;
+                    string from = item.Attributes["from"].InnerText;
+                    if (!list.ContainsKey(from))
+                        list.Add(from, to);
                 }
             }
             return list;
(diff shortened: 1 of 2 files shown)


libse/Dictionaries/OcrFixReplaceList.cs
```diff
@@ -113,14 +113,19 @@ public static OcrFixReplaceList FromLanguageId(string languageId)
...
                         {
                             string to = item.Attributes["replaceWith"].InnerText;
                             string from = item.Attributes["find"].InnerText;
-                            if (!list.ContainsKey(from))
-                                list.Add(from, to);
+                            if (Utilities.IsValidRegex(to) && Utilities.IsValidRegex(from))
+                            {
+                                if (!list.ContainsKey(from))
+                                    list.Add(from, to);
+                            }
```
xylographe: `to` contains a replacement pattern (see [Substitution Elements and Replacement Patterns](https://msdn.microsoft.com/en-us/library/ewy2t5e0(v=vs.110).aspx)).

xylographe: Also, **from** cannot be zero Length, but for **to** that's OK.

ivandrofly: :+1:  @xylographe

ivandrofly:
:+1:  @xylographe, what do you mean from **to** its ok?

xylographe:
The **to** string can be empty (Length == 0). It might seem weird, but if the regex pattern in **from** contains look-ahead and/or look-behind sub-expressions, it could happen.

An empty **from** string OTOH *must* be a mistake. For example:
```
PS> $re = [regex][string]::Empty
PS> $re.Replace('abc', 'x')
xaxbxcx
```
The empty string as a regex pattern matches everywhere.

ivandrofly:
Nice :)

ivandrofly:
:+1: /cc @niksedk

libse/Dictionaries/OcrFixReplaceList.cs
```diff
@@ -113,14 +113,19 @@ public static OcrFixReplaceList FromLanguageId(string languageId)
...
                     {
-                        if (item.Attributes != null && item.Attributes["replaceWith"] != null && item.Attributes["find"] != null)
+                        if (item.Attributes != null && item.Attributes[ReplaceWith] != null && item.Attributes[Find] != null)
                         {
-                            string to = item.Attributes["replaceWith"].InnerText;
-                            string from = item.Attributes["find"].InnerText;
-                            if (!list.ContainsKey(from))
-                                list.Add(from, to);
+                            string to = item.Attributes[ReplaceWith].InnerText;
+                            string from = item.Attributes[Find].InnerText;
```
xylographe: Is this OK?
According to [MSDN](https://msdn.microsoft.com/en-us/library/system.xml.xmlattribute.innertext%28v=vs.110%29.aspx) the `XmlAttribute.InnerText` property would be **set** only.
Perhaps, better use the [`XmlAttribute.Value` property](https://msdn.microsoft.com/en-us/library/system.xml.xmlattribute.value%28v=vs.110%29.aspx) ?

ivandrofly: > According to MSDN the XmlAttribute.InnerText property would be set only.
> Perhaps, better use the XmlAttribute.Value property ?

:+1:

ivandrofly: I'm learning **PowerShell** any book you recommend? /cc @xylographe  :smile:

ivandrofly:
I `force push` wrong branch :angry:

libse/Dictionaries/OcrFixReplaceList.cs
xylographe: [Windows PowerShell in Action, Second Edition](https://www.manning.com/books/windows-powershell-in-action-second-edition)

It's about V2, hence, it doesn't address V3/V4 additions, but it's the best posh reference I know. It starts every subject simple (no previous knowledge required), before going into more complex details, until everything is revealed. :smile: 

It's miles ahead of other posh books I read, only downside is that there have been no upgrades for V3 and V4. Of course, MSDN contains info on changes between V2-V3, and V3-V4. And for the latest *usage* of any cmdlet, you should use `get-help -online` anyway.

ivandrofly: Nice, :smile: thanks ;)

PR #1210, opened 2015-08-30
Update - (src/Forms/AutoBreakUnbreakLines.cs)

ivandrofly:


src/Forms/AutoBreakUnbreakLines.cs
```diff
@@ -19,11 +18,6 @@ public partial class AutoBreakUnbreakLines : PositionAndSizeForm
             get { return _fixedText; }
         }
 
-        public int Changes
```
ivandrofly: Never used

src/Forms/AutoBreakUnbreakLines.cs
```diff
@@ -113,12 +107,11 @@ private void AutoBalance()
             {
                 if (HtmlUtil.RemoveHtmlTags(p.Text, true).Length > minLength || p.Text.Contains(Environment.NewLine))
                 {
-                    var text = Utilities.AutoBreakLine(p.Text, 5, MergeLinesShorterThan, language);
+                    var text = Utilities.AutoBreakLine(p.Text, 5, minLength, language);
```
ivandrofly: Why go through reg since it's already in `minLength`

PR #1217, opened 2015-09-01
Spectrogram optimization

jdpurcell:
Optimizations to the spectrogram code.  On a 10 minute video this reduced the time to generate the spectrogram from ~9.7 seconds to ~2.7 seconds on my computer.

* Use FastBitmap to speed up drawing.
* Use better optimized FFT routines.
* Other optimizations to reduce CPU/memory overhead.

About 67% of the speed-up came simply from using FastBitmap (~4 line code change), so I encourage you to do that even if you don't accept all the other changes.


niksedk:
Impressive :)
I look forward to testing this (a hardcore place you started).
The original code was c? Looks obfuscated ;)

jdpurcell:
If you mean the FFT code, yes it was originally C.  I ported it about 5 years ago for one of my programs, Center Cut GUI.  I found that particular library by looking through the benchmarks on fftw.org and finding one that was a good balance of performance, simplicity (i.e. self-contained, one file), and permissive license.

jdpurcell:
By the way, if you'd prefer a version of the FFT code without the 'unsafe' keywords, let me know. I just tried it out and it's not too much slower than the unsafe version (say, ~20 ms slower for 10 minutes of audio). It still ends up being about 4 times faster than the original FFT code.

niksedk:
Between 4 and 6 times faster in my tests and that's really nice :)
Do go for the 'safe' code if it only adds about 20ms per 10 min.

jdpurcell:
Alright, I got rid of the unsafe stuff.

PR #1221, opened 2015-09-02
[bugfix] - fix bug (invalid regex pattern) in 'find' + better element validation

ivandrofly:
another bug-fix /cc @niksedk 
Ref to: #1208

### DIFF
--- a/libse/Dictionaries/OcrFixReplaceList.cs
+++ b/libse/Dictionaries/OcrFixReplaceList.cs
@@ -17,8 +17,8 @@ namespace Nikse.SubtitleEdit.Core.Dictionaries
         private static readonly Regex HexNumber = new Regex(@"^#?[\dABDEFabcdef]+$", RegexOptions.Compiled);
         private static readonly Regex StartEndEndsWithNumber = new Regex(@"^\d+.+\d$", RegexOptions.Compiled);
 
-        public Dictionary<string, string> WordReplaceList;
-        public Dictionary<string, string> PartialLineWordBoundaryReplaceList;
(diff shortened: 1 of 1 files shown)


libse/Dictionaries/OcrFixReplaceList.cs
```diff
@@ -85,21 +85,16 @@ public static OcrFixReplaceList FromLanguageId(string languageId)
...
-                        if (item.Attributes != null && item.Attributes["to"] != null && item.Attributes["from"] != null)
-                        {
-                            string to = item.Attributes["to"].InnerText;
-                            string from = item.Attributes["from"].InnerText;
-                            if (!list.ContainsKey(from))
-                                list.Add(from, to);
-                        }
-                    }
+                    string to = item.Attributes["to"].Value;
+                    string from = item.Attributes["from"].Value;
```
ivandrofly: thanks to @xylographe for this one

libse/Dictionaries/OcrFixReplaceList.cs
```diff
@@ -108,26 +103,49 @@ public static OcrFixReplaceList FromLanguageId(string languageId)
...
+                if (node.Attributes["find"] != null && node.Attributes["replaceWith"] != null)
+                {
+                    return Utilities.IsValidRegex(node.Attributes["find"].Value);
+                }
+            }
+            else
+            {
+                if (node.Attributes["from"] != null && node.Attributes["to"] != null)
+                {
+                    return (node.Attributes["from"].Value != node.Attributes["to"].Value);
```
ivandrofly: We probably don't want to matching value to be `from` and `to`, right?

PR #1227, opened 2015-09-03
Spectrogram optimization, round 2

jdpurcell:
Using my same test (~10 minute video) that took 2.7 seconds after the first round of optimization, I now have it down to 1.8 seconds.

This should be my last pull request for spectrogram optimizations; it's about as good as it's going to get.

The timing breakdown is now 54% in Bitmap.Save (saving the GIF), 35% in SpectrogramDrawer.Draw (including the FFT stuff), and 11% for other overhead (e.g. converting the audio bytes to double).

### DIFF
--- a/libse/FastBitmap.cs
+++ b/libse/FastBitmap.cs
@@ -8,12 +8,20 @@ namespace Nikse.SubtitleEdit.Core
 {
     unsafe public class FastBitmap
     {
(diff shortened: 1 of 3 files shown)


niksedk:
I think I will use spectrograms more often now that it takes so little time to generate :)

PR #1234, opened 2015-09-03
Update + Fix (src/Forms/FixCommonErrors.cs)

ivandrofly:
To see the issues (Uncheck):
- [ ] - Remove line break in short text with only one sentence

In **Fix Common Error**

@niksedk 
Here is the text sample:

{\an4}
hey


niksedk:
Nice :)

PR #1241, opened 2015-09-04
Update + Fix (src/Forms/ChooseEncoding.cs)

ivandrofly:


src/Forms/ChooseEncoding.cs
```diff
@@ -31,23 +36,22 @@ public void Initialize(string fileName)
...
-                _fileBuffer = new byte[length];
-                file.Read(_fileBuffer, 0, length);
-
-                for (int i = 0; i < length; i++)
+                using (var file = new FileStream(fileName, FileMode.Open, FileAccess.Read, FileShare.ReadWrite))
                 {
-                    if (_fileBuffer[i] == 0)
-                        _fileBuffer[i] = 32;
+                    int length = (int)file.Length;
+                    if (length > textBoxPreview.MaxLength)
```
ivandrofly: TextBox's default MaxLength is **32767** which means we don't need to buffer size above that

src/Forms/ChooseEncoding.cs
```diff
@@ -12,6 +12,11 @@ public sealed partial class ChooseEncoding : Form
         private Encoding _encoding;
         private byte[] _fileBuffer;
 
+        public Encoding GetEncoding
+        {
+            get { return _encoding; }
+        }
+
```
xylographe: As it is no longer a method, perhaps, you could rename `GetEncoding` to `Encoding`?

ivandrofly: :+1:

src/Forms/ChooseEncoding.cs
xylographe: [MSDN:](https://msdn.microsoft.com/en-us/library/system.windows.forms.textboxbase.maxlength)
“In code, you can set the value of the Text property to a value that has
 a length greater than the value specified by the MaxLength property.
 This property only affects text entered into the control at run time.”

ivandrofly: > “In code, you can set the value of the Text property to a value that has
> a length greater than the value specified by the MaxLength property.
> This property only affects text entered into the control at run time.”

yeah I know, even in design mode you can, but I think **32767** is alright since it's used just for preview

ivandrofly:
@niksedk

PR #1242, opened 2015-09-05
[internal] - check if dropped file is directory as soon as the file is dropped

ivandrofly:


src/Forms/Main.cs
```diff
@@ -9668,18 +9668,18 @@ private void DoSubtitleListview1Drop(object sender, EventArgs e)
...
                 string fileName = _dragAndDropFiles[0];
 
-                var dirName = Path.GetDirectoryName(fileName);
-                saveFileDialog1.InitialDirectory = dirName;
-                openFileDialog1.InitialDirectory = dirName;
-
-                var file = new FileInfo(fileName);
-
                 // Do not allow directory drop
                 if (FileUtil.IsDirectory(fileName))
```
ivandrofly: Move above

PR #1246, opened 2015-09-05
Improve accuracy of spectrogram

jdpurcell:
The original spectrogram code applied a window function to each segment, but didn't do any overlapping to compensate for the signal loss near the edges. I've fixed that and the result is more peasant/accurate looking spectrogram images.

Also includes:
* Clean up some unused fields in Info.xml (I did leave one unnecessary field however so that older versions can still read files generated by the new version).
* Fixes to allow for FFT sizes other than 256 (not that it will actually be used, but just for the sake of correctness).
* Save out GIF in a separate thread. For this reason, the code is yet again faster even though it's doing a bit more work with the overlapping.

### DIFF
--- a/libse/WaveToVisualizer.cs
+++ b/libse/WaveToVisualizer.cs
@@ -493,11 +493,10 @@ namespace Nikse.SubtitleEdit.Core
 
                 // save image
                 string imagePath = Path.Combine(spectrogramDirectory, iChunk + ".gif");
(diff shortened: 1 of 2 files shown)


niksedk:
Sorry, I've changed a bit in AudioVisualizer too - could you rebase?

jdpurcell:
Okay fixed.

niksedk:
thx :)

PR #1254, opened 2015-09-06
[internal] - minor update

ivandrofly:


src/Forms/BatchConvert.cs
```diff
@@ -912,7 +912,6 @@ private static void DoThreadWork(object sender, DoWorkEventArgs e)
                         {
                             p.Subtitle.Paragraphs.RemoveAt(deleteIndex);
                         }
-                        p.Subtitle.Renumber();
```
ivandrofly: Will be done in
```
            // always re-number
            p.Subtitle.Renumber();
```

PR #1256, opened 2015-09-07
WaveformPaint cleanup

jdpurcell:
This is mostly just cleanup to the WaveformPaint code to make it simpler and easier to follow.

Aside from that, there are a few user-visible changes:
* Fix crash when using the mouse wheel over the waveform display if a waveform isn't loaded.
* Make the vertical zooming progress more naturally (i.e. it doesn't make such massive changes near the end of the zoom range now).

Other thoughts:
* The original code had a special case for ZoomFactor > 0.9999 && ZoomFactor < 1.00001. Although I handled everything in one bit of code with no special case, I ensured that floating point rounding errors will never affect the 100% zoom view by rounding to 2 decimal places in the ZoomFactor setter. (1.0 can be represented exactly in floating point but it depends how you get there; setting directly to 1.0 is fine for example, but setting by way of 0.9 + 0.1 will result in something like 0.9999999... if you don't round afterwards)
* There is new code for determining which pixels are inside the time ranges of selected paragraphs. This was partly an attempted optimization to avoid iterating over the selected paragraph list as much as possible, but it ended up not mattering too much (this area is kind of a drop in the bucket in terms of total execution time). I was very careful with the logic and tested it well; it handles various overlapping situations properly.
* I noticed the original code for drawing scene changes had some logic to only draw the markers within 20 seconds of the start position. I removed that logic as it seemed unnecessary, but if it was there for a reason let me know.
* I did test the overall performance of WaveformPaint before and after my changes. It's at least not slower, I promise :smiley:. It's about the same, or perhaps very slightly faster.

### DIFF
--- a/libse/TimeCode.cs
+++ b/libse/TimeCode.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Globalization;
 using Nikse.SubtitleEdit.Core.SubtitleFormats;
 
 namespace Nikse.SubtitleEdit.Core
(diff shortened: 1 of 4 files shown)


niksedk:
Hi, 
nice crash fix :)
Some minor issues... 
1) No border on the text/time codes makes it hard to read (please use old "poor mans outline or better" but do note that I've already tested multiple different draw routines that do not work well with all fonts etc.)
2) Text is lower (it's better higher as it conflicts lesser with the waveform - please use old y)
3) Text should be cut of with "..." if it longer than the waveform (please use old code or better)
4) Please use old duration text logic
5) Coding style: Please don't use "one line if" - like " if (value < ZoomMinimum) value = ZoomMinimum"...

jdpurcell:
1/2/3) I don't see any differences compared to the original code. I didn't change any of the text rendering logic. The outputted image looks identical to me, with the exception of an occasional paragraph marker's start/end line differing in X position by 1 pixel due to rounding differences. Did you use the same settings when comparing?
4) Do you mean the logic to determine which pixels fall inside a selection, i.e. the IsSelectedIndex method?

jdpurcell:
I think you had the HH:MM:SS:FF time code mode turned on. For some reason (even in the original code) there was a different code path for drawing the paragraph text depending on the time code mode. I went ahead and fixed that so it all uses the same text rendering logic.

I also noticed that with the default time code mode (HH:MM:SS.MS) the decimal separator was not localized so I fixed that as well. I accomplished this with a new "localize" parameter on some of the TimeCode's ToString methods, which is opt-in (disabled by default) to avoid changing/breaking stuff anywhere else.

niksedk:
Yes, you're of course right :)
Huge improvement to the visualizer - especially in frame mode ;)

PR #1264, opened 2015-09-09
Add support for 8-bit, 20-bit, and 24-bit WAVs

jdpurcell:
When I was working on the spectrogram stuff, I noticed the code for bit depths other than 16 wouldn't work properly.  I didn't bother fixing it then because when I tried opening a 24 bit file, it got "extracted" and converted to 16 bit.  A commit earlier today changed that behavior, so now seems like a good time to fix this.

I tested with 8-bit, 20-bit, and 24-bit WAVs and all are working fine now.  As for 32-bit, I don't think it's worth the trouble (not even sure how common 32-bit integer WAVs are, I think a lot of them are floating point instead).

### DIFF
--- a/libse/WaveToVisualizer.cs
+++ b/libse/WaveToVisualizer.cs
@@ -15,7 +15,6 @@ namespace Nikse.SubtitleEdit.Core
     public class WaveHeader
     {
         private const int ConstantHeaderSize = 20;
-        private readonly byte[] _headerData;
 
         public string ChunkId { get; private set; }
         public uint ChunkSize { get; private set; }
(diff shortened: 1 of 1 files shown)


niksedk:
Very cool :)

My commit to auto-use wave files is much better now, but it still probably needs some validation (e.g. non 'PCM' or 32-bit wave files)

niksedk:
Does this look okay: https://github.com/SubtitleEdit/subtitleedit/commit/fc296789ef31f58afbb911e98756816cedb96666#diff-dc310822c0a855cb5f53a37a73ece408R239

jdpurcell:
Yes, looks good!

PR #1265, opened 2015-09-09
Update + Fix

ivandrofly:
/cc @niksedk


ivandrofly:
> It's not already visible but I think it's correct anyway as subtitle2 is not loaded...

:+1:

PR #1267, opened 2015-09-09
[HI] - Refactor

ivandrofly:


libse/Forms/RemoveTextForHI.cs
```diff
@@ -175,16 +177,17 @@ public string RemoveColon(string text)
                                     var content = line.Substring(indexOfColon + 1).Trim();
                                     if (content.Length > 0)
                                     {
+                                        newText += Environment.NewLine;
```
ivandrofly: Since it's going to be added anyway

PR #1273, opened 2015-09-12
[Internal] - Minor update

ivandrofly:


libse/Utilities.cs
```diff
@@ -2925,10 +2923,7 @@ public static Subtitle LoadMatroskaSSA(MatroskaTrackInfo matroskaSubtitleInfo, s
                             arr = arr[1].Split(new[] { ':', '.' });
                             if (arr.Length == 4)
                             {
-                                int hour;
-                                int min;
-                                int sec;
-                                int ms;
+                                int hour, min, sec, ms;
```
niksedk: I'm not too found of multi var initialization in one line (can make it harder to read IMO - especially if ppl start adding initialization values too)

ivandrofly: Ah :+1: let me remove it then ;)

ivandrofly:
Done! /cc @niksedk

PR #1274, opened 2015-09-13
Some more wave visualizer enhancements

jdpurcell:
* WaveToVisualizer: Fix peak generation skipping the last fractional second of audio.
* Add waveform batch: Generate spectrogram (if respective setting is enabled).
* Add waveform batch: Fix status not showing the "calculating..." part.
* Small code cleanups (e.g. fix typo, make constant match naming convention, factor out some redundant code).

EDIT: Added some more:
* Audio visualizer: Fix waveform height calculation being 2x too high (but set default vertical zoom to 2.0 to keep the same effective look). Now you can zoom out vertically to 1.0 to see the correct height.
* Audio visualizer: Allow vertical zoom to be changed with control + shift + scroll wheel.

### DIFF
--- a/libse/WaveToVisualizer.cs
+++ b/libse/WaveToVisualizer.cs
@@ -249,9 +249,8 @@ namespace Nikse.SubtitleEdit.Core
         /// <summary>
         /// Generate peaks (samples with some interval) for an uncompressed wave file
         /// </summary>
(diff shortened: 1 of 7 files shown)


niksedk:
Cool, gotta add the short cuts to the help page

PR #1290, opened 2015-09-18
Move "creation of AssemblyInfo.cs" from post-build to pre-build

xylographe:
Consistent with `libse\SubtitleEdit.csproj`.


xylographe:
Unfortunately, it still doesn't work for a blank clone.

After copying `AssemblyInfo.cs.template` to `AssemblyInfo.cs` there will be a (loosely translated from Dutch) "invalid version 3.4.8.\[REVNO]" error.

If you remove "\[REVNO]" from the template, you will have a problem with both UpdateLanguageFiles and build.bat.

Perhaps, there might be some way to let VS update the templates instead of `AssemblyInfo.cs`?
It would certainly make matters easier.

Issue #1285, opened 2015-09-17
[Linux] - SE throws expection


ivandrofly:
Was running some test on using Linux (SE portable): 

![image](http://i.imgur.com/JxiLux9.png)

niksedk:
When running the pre-compiled SubtitleEdit.exe or after compiling with Xamarin? or both?

ivandrofly:
> When running the pre-compiled SubtitleEdit.exe or after compiling with Xamarin? or both?

Pre-compiled

niksedk:
Nice to get this working :)

PR #1291, opened 2015-09-18
Work-around for issue #1285

xylographe:
Apparently, `Encoding.GetEncodings()` can return `EncodingInfo` objects that throw an exception when invoking `EncodingInfo.GetEncoding()`. Hence, accessing the `EncodingInfo.Name` property might throw an exception.


ivandrofly:
yeah I think that is the case :+1:

PR #1304, opened 2015-09-20
Added Import sub-menu

### LINKED ISSUE #1225: [UI] File menu strip
Hey @niksedk I think we should move all the **Import subtitle** option in **File** menu strip into **Import** context menu just like **Export**

what do you think?

menu

suvjunmd:
Fixes #1225


ivandrofly:
Any reason this didn't merge? @niksedk

PR #1305, opened 2015-09-21
Fix not all paragraphs showing in the audio visualizer when zoomed out

jdpurcell:
The previous code always loaded about 9 paragraphs before/after the "current" paragraph (i.e. the one at the current video position), which was usually sufficient except when zoomed out too much.

The new code figures out which paragraphs are needed to fill the whole displayed time range.

The logic for determining the "current" paragraph has been cleaned up but is pretty much the same as before. The only difference is that when no more subtitles exist after the current video position, the old code would default to the first displayed one, if present, before defaulting to the last subtitle. This didn't seem to serve any purpose so I just default to the last one. In fact, as far as I can tell, the rest of the code in the visualizer makes no distinction between current and previous/next paragraphs. It could probably all just be combined into one "displayable paragraphs" list, unless it was separated out for some other reason (optimization?).

### DIFF
--- a/libse/WaveToVisualizer.cs
+++ b/libse/WaveToVisualizer.cs
@@ -260,7 +260,7 @@ namespace Nikse.SubtitleEdit.Core
 
             PeaksPerSecond = Math.Min(Configuration.Settings.VideoControls.WaveformMinimumSampleRate, Header.SampleRate);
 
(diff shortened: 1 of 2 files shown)


niksedk:
Yes, works nicely when zoomed out :)
Previous/next paragraphs was made for performance - but there don't seem to be much difference.

PR #1311, opened 2015-09-21
[Cleanup] - Remove redundant 'This' qualification

ivandrofly:
Is it okay to do this in design forms? /cc @niksedk 

I'm planing to do some cleanup like remove redundant cast, 'This' qualification...


niksedk:
The design forms are auto-generated so I don't think we should change stuff there.

PR #1316, opened 2015-09-22
Resolve problems with UpdateAssemblyInfo

xylographe:
- new release: update template files when SE version has been bumped in VS
- building from source tarball: set build number to 9999 (no commit hash)

Auto-updating template files is necessary to allow building from tarball—the version number must be correct to be able to select translations.

Note: libSE version follows SE version, but not the other way around.


niksedk:
Almost rewritten ;)
Looks good and very close, but still fails for a blank solution for me - I think you can fix it by adding a project dependency in solution properties: libse depends on updateassembly... right?

xylographe:
\[ Fixed SE/libSE dependencies + re-based ]

Oops, I'd forgotten the dependencies.  Thank you!

However, the three scenarios I've tested also worked before, with the incorrect dependencies:
* Download ZIP-archive of this branch: new AssemblyInfo files are created with build number 9999.
* Clone my SE fork, and checkout this branch: new AssemblyInfo files are created with build/guid from repository.
* Create new release 3.4.9: templates and AssemblyInfo files are updated (correct build/guid after running `git commit -a` and `git tag 3.4.9`).

xylographe:
@niksedk If the dependency fix did not solve the problem, perhaps, the problem (SE/libSE pre-build tasks executing concurrently) can be resolved with a Mutex, as in branch [xylographe:xfix3](https://github.com/xylographe/subtitleedit/tree/xfix3)

niksedk:
Cool, now it works here too, both with and without git :)

PR #1318, opened 2015-09-23
Experimental - only re-draw when something changes

jdpurcell:
Please do not merge this! I only want to ask for your feedback.

I noticed idle CPU usage (4% on my laptop, 10% on my [old] desktop) which is caused by the forced redraws every 1/10 of a second. I guess it's not a big deal for desktop users but it could cause some extra heat / lower battery life for laptop users. So I played around with the code a bit to see if I could get rid of the forced redraws and this is a prototype that I came up with. It reduces the idle CPU usage to just about 0%.

My question is: do you think this is worth pursuing further (needs some cleanup / more testing for sure), or is it too risky with not enough benefit?

### DIFF
--- a/src/Controls/AudioVisualizer.cs
+++ b/src/Controls/AudioVisualizer.cs
@@ -80,6 +80,7 @@ namespace Nikse.SubtitleEdit.Controls
         private double _currentVideoPositionSeconds = -1;
         private WavePeakGenerator _wavePeaks;
         private Subtitle _subtitle;
+        private string _subtitleHash;
         private IEnumerable<int> _selectedIndices = Enumerable.Empty<int>();
         private bool _noClear;
         private double _gapAtStart = -1;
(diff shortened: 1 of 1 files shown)


niksedk:
If it's really easy, do it... otherwise let it be (does the current patch make working in audio wiz take a bit more cpu?)

What do you think about always showing the audio wiz - even if no waveform/spectrogram?

jdpurcell:
I will let it be for now :)

As for always showing... I guess you mean drawing all the paragraphs and stuff even if the waveform isn't loaded?  Well the first problem is how to trigger loading of the waveform.  Right now any click triggers the load, so there'd need to be a different way to do it to allow the user to interact with the paragraphs without triggering waveform load.  Also there are places in the visualizer code that need to know the length of the audio; passing in the length of the video from the media player would probably work fine as well.  I don't see any major reasons why it wouldn't work, but personally I'm not sure how useful it would be.

PR #1323, opened 2015-09-25
SubtitleListView: Enable double buffering to eliminate flickering

jdpurcell:
Demo of before/after: https://youtu.be/tP_1n0i9iR4


niksedk:
Wow, that's nice :)
The flickering has annoyed me for years...

Issue #1108, opened 2015-08-05
Alt Drag Delay in Waveform (still occur in 3.4.8?)


LeonCheung:
Regarding to my last comments in #238, which has been closed, I think it would be more appropriate to open a new issue since it is still happening, but the cause has been targeted fortunately (I hope).

NOTE:
1. This issue only seems to be more obvious **after working for some minutes**.
2. I believe I got a very high performance PC (24 thread, 32GB RAM, dual GTX680), so I think there should be nothing to do with the harware.

Please see my last comments there for more info, thank you so so so much!

niksedk:
I still cannot re-create this on any of my PCs!
Could you send me your complete settings.xml file?

Can anybody else re-create this?

LeonCheung:
@niksedk [Here](https://www.youtube.com/watch?v=fF62gAHellY) is another video (a bit long, key part is from 15:00), to show from initial setting to working on it and then targeting it. 

BTW,  at 0:15, I cannot customize shortcut for Save operator other than the default setting. Is this a finshy bug?

LeonCheung:
[Here](https://drive.google.com/file/d/0B1ojv5B6Lb-3dUp6dkJSLVpzXzQ/view?usp=sharing) is the current xml file.

LeonCheung:
P.S. I should stress that the problem will disappear after restarting SE, but it will show up again after working for a while.

Now I wonder if there is something to do with the cache or something iterated? Or, it probably about my workflow? (You see how I set S and D shortcut, which are the most commonly way I used to work on subs)

niksedk:
I've fixed a minor memory leak in the waveform (fb1710992ab27d4500b6361347aae3a00b26a638)... could you test if the problem is still there: http://www.nikse.dk/SubtitleEdit.zip (or github source)

LeonCheung:
Hi, First, thank you very much for not forgetting this issue! :P 

After a hard testing on a 30 minutes sub work, I didn't see any obvious fix on this particular one.

Well, there still might be kinda memory leak things existing (since this issue becomes obvious ONLY after working for quite a while), I still believe the key problem appears fairly related to the Syntax coloring rule(s), especially the last checkbox **"Color tme code overlap"**. Because things will go extremely smooth after toggling it off. Would you please take more look into that aspect? I think the suspect was hiding there. :)

Steps to repeat the issue:

1. Use a long-enough sub file, make sure the minimum gap between lines is **1 milisecond**.
[...]

PR #1330, opened 2015-09-26
Fix for issue #1108

jdpurcell:
The main fix is just removing one line that causes the NumericUpDownDurationValueChanged event to get wired up multiple times.

Before I found that problem, I noticed that it seemed to be masked by avoiding changing the text in the warning labels so frequently.  Basically it was an optimization that reduced the symptoms of the problem.  Since it is more efficient, I decided to include it in the patch as well. (EDIT: I figured out why it works. In Control's Text property setter there's logic like "if (value == Text) return;". By setting the text to string.Empty and then the actual value later, it effectively bypassed that logic.)


niksedk:
Good fix :)

PR #1343, opened 2015-10-01
Waveform - show actual peak values

jdpurcell:
I was playing with another subtitle editor program and noticed that in its waveform display it was much easier to notice where the quiet/loud parts were. Curious how it worked, I read through the source code and found that for each "column" (1 pixel wide area) it displays, it scans through a range of audio samples corresponding to that time slice, finds the minimum and maximum samples, and then draws a vertical line from a point representing the minimum to a point representing the maximum.

And then of course I had to implement it in Subtitle Edit :smile:.

Notes:
* Although it is backwards compatible with the peak files written by previous versions, you will not get the full benefit of the new code unless you re-generate the peaks. **So please clear your old waveforms folder before trying it out!**
* There is some performance impact when generating the peaks, since it now has to scan through all the samples. That said, it's still quite fast, and I think it's worth it for the improved usefulness of the waveform. To give you an example, it took ~450 ms to scan through 10 minutes of audio, vs. the old code's ~40 ms. That's still not much compared to the ~10 seconds it took to extract the audio or the spectrogram generation time.

Other changes:
* Rounding fix to the spectrogram drawing code (results in slightly better alignment).
* Ctrl+0 shortcut key in the audio visualizer to reset to 100% zoom.
* Some code cleanups (e.g. spectrogram drawing / loading).

Before:
before2

After:
after2

### DIFF
--- a/libse/WaveToVisualizer.cs
+++ b/libse/WaveToVisualizer.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Drawing;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 using System.Xml;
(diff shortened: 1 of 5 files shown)


jdpurcell:
During testing I also noticed that "seek silence" and "guess time codes" didn't work. These were both pre-existing bugs from before I changed the code. I've now fixed them both; the commit is included in this pull request.

I've also tested my other changes thoroughly and am confident that everything is safe.

niksedk:
Some very nice fixes :)  

I tested drawing the waveform like this earlier (in two ways - use higher sample rate or just draw vertical line for each x). The reason it did not use it was performance - but I'll test your version tomorrow (Git for windows is troublesome atm). It does look better...

niksedk:
Does take a while to generate... but the drawing/working performance is the same :)
Perhaps the default green and red are a bit too intense now - what do you think?

jdpurcell:
The red especially is a bit intense... it doesn't look terrible though. I can play around with it later and see if I can figure out some more pleasant colors.

Issue #1317, opened 2015-09-22
Fix Common Errors wrong language detection


marb99:
Hi... 
I have noticed that in recent beta builds of SE 3.4.8 it has detected in (Fix common errors) the wrong subtitle language.

Detects subtitles pt-pt as Spanish. 

And lately happens a lot in several pt-pt subs.

![capturar](https://cloud.githubusercontent.com/assets/7856897/10031312/64b48608-6174-11e5-9098-93469401af30.PNG)

I leave here the sub in pt-pt as the SE (Fix common errors) detects as Spanish language.

Subtitle:
http://www93.zippyshare.com/v/Kdnd8q0d/file.html

-----------------------------------
I have a take a look in code.

And I think the problem can be in AutoDetectWordsSpanish = { "el", "bien", "Vamos", "Hola", "casa", "con" };

and in code:

int spanishCount = GetCount(text, "Hola", "nada", "Vamos", "pasa", "los", "como"); // not french words

"Vamos" and "casa" and "nada" and "como" are equal and very common words in Portuguese too.

perhaps because of it detects subtitles in pt-pt as Spanish.

aaaxx:
If you want to help, get the word frequency lists for Spanish and Portuguese from [here](https://invokeit.wordpress.com/frequency-word-lists/), compile new lists with the most used words for both languages and post them here so the code can be updated.

Twenty or so unique words per language should be enough.

xylographe:
@marb99  What about Portuguese detection words:
```
AutoDetectWordsPortuguese = { "não", "Não", "Estás", "Então", "isso", "com" };
```
Are (some of) these words common in another language (or languages)?


BTW Word lists to differentiate between pt-PT and pt-BR would also be nice.

marb99:
@xylographe 

I think it is not common.

the differences between pt-PT and pt-BR are many differences.

orthographic, syntactical, prosodic and semantics

for some lexical differences.

try this list: https://pt.wikipedia.org/wiki/Lista_de_diferen%C3%A7as_lexicais_entre_vers%C3%B5es_da_l%C3%ADngua_portuguesa

xylographe:
I was thinking of changing the detection order: try Portuguese before Spanish. However, "Estás" is also common in Spanish, "com" is common in Catalan, "não" in Vietnamese, and "isso" might be Italian.

I think this might resolve the issue for now: (with the currently detectable languages)
```
[...]

PR #1346, opened 2015-10-02
Fix Portuguese language detection (Core/Utilities)

xylographe:
Avoid detection of Portuguese subtitles as Spanish—issue #1317


niksedk:
Cool :)
I'll try to add some unit tests for this in the future...

PR #1360, opened 2015-10-06
Update AutoDetectWordsRussian

Elheym:
Advanced avto detection list of words for Russian.
I would also like to have been added for Ukrainian, which is not represented here. Add the necessary data, similar to the Russian Ukrainian heading ... Such as a count and case, but it did no good, they were identified as Russian. Add please: "public static readonly string[] AutoDetectWordsUkrainian = { "Ні", "ні", "Привіт", "привіт", "Це", "це", "Що", "що", "Його", "його", "Він", "він", "Як", "як", "Гаразд", "гаразд", "Якщо", "якщо", "Мені", "мені", "Твій", "твій", "Їх", "їх", "Вітаю", "вітаю", "Дякую", "дякую", "вже", "було", "був", "цього", "нічого", "немає", "може", "знову", "бо", "щось", "щоб", "цим", "тобі", "хотів", "твоїх", "мої", "мій", "має", "їм", "йому", "дуже" };"

### DIFF
--- a/libse/Utilities.cs
+++ b/libse/Utilities.cs
@@ -1074,7 +1074,7 @@ namespace Nikse.SubtitleEdit.Core
         public static readonly string[] AutoDetectWordsItalian = { "Cosa", "sono", "Grazie", "Buongiorno", "bene", "questo", "ragazzi", "propriamente", "numero", "hanno", "giorno", "faccio", "davvero", "negativo", "essere", "vuole", "sensitivo", "venire" };
         public static readonly string[] AutoDetectWordsPortuguese = { "não", "Não", "Estás", "Então", "isso", "com" };
         public static readonly string[] AutoDetectWordsGreek = { "μου", "είναι", "Είναι", "αυτό", "Τόμπυ", "καλά", "Ενταξει", "Ενταξει", "πρεπει", "Λοιπον", "τιποτα", "ξερεις" };
-        public static readonly string[] AutoDetectWordsRussian = { "все", "это", "как", "Воробей", "сюда", "Давай" };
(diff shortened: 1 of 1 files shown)


xylographe:
Both Russian and Ukrainian lists have (too?) many words. The AutoDetectWords* lists should be kept as small as possible for efficiency (the number of supported languages will only increase). When two or more languages have similar words, extra lists can be used to differentiate. An extreme example is Croatian/Serbian: only one AutoDetectWords list is used to recognize both, two separate lists are used to keep them apart.

I think, primary lists should be 5‒10 words, secondary lists can be longer (10‒25, perhaps).

Elheym:
In both languages, use a unique word, that is between them, they are not repeated. If this argument is misplaced, they can be shortened as you said. The main thing - that they were correctly identified with each other. Russian as Russian, Ukrainian as the Ukrainian, but not both at the same. :) 
The rest is not so important. ;)

niksedk:
I don't see a problem in have longer lists (performance needs to be checked), but shorter words like 2 letters might be problematic as it has greater possibility to clash with other languages (and for ansi subtitles this is rather complicated if wrong encoding is used - or auto-detection is tried).
I would like to extract these functions to their own class and add many unit tests :)

Elheym:
Determination of Russian works fine, it is checked. With the complexity of the Ukrainian - Russian is defined as in any encoding. The words of both languages have been specially selected to each other could not interfere, they are totally different. By default, for comparisons, use UTF-8, ANSI does not apply.
  If you will need any tests - I will be happy to help.

PR #1363, opened 2015-10-07
[Refactor] Minor refactor

ivandrofly:


niksedk:
nice, could you also move ExpectedChars out into a static for improved performance?

ivandrofly:
> nice, could you also move ExpectedChars out into a static for improved performance?

Done

PR #1372, opened 2015-10-10
Updated SE-installer to resolve issue #1369

### LINKED ISSUE #1369: Impossible to do a visual sync
OS : Win 10
The error occurs each time I clic on the button Sync, with every subtitles (srt) or videos (mkv).

untitled

However no problem with the portable version ^^

xylographe:
Resolves #1369

[1]
I'm not a big fan of hard-coded, host-dependent paths. :)

[2]
The `CanRunNGen()` function can be refined when more info becomes available.

### DIFF
--- a/installer/Subtitle_Edit_installer.iss
+++ b/installer/Subtitle_Edit_installer.iss
@@ -369,13 +369,15 @@ Type: dirifempty; Name: {app}\Languages;                Check: not IsComponentSe
 
 
 [Run]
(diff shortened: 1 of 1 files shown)


jdpurcell:
Is there any concern about PowerShell not being installed (e.g. it didn't ship with Vista)?

What's the problem with ngen on Windows 10? I am running Windows 10 and the official 3.4.10 installer version works fine for me.

xylographe:
I think, PoSh is included in Framework 4?
Anyway, I re-added abs path as fall-back.

Not clear why ngen generates bad executable on OP's Win10.
Hence: “The CanRunNGen() function can be refined when more info becomes available.”

PR #1379, opened 2015-10-12
Fixes for OS X (still not usable really)

jdpurcell:
These are the fixes from when I got Subtitle Edit running on OS X. I didn't get video to display but it played the audio fine. Unfortunately Mono's WinForms is so bad (at least on OS X) that it's not really usable in my opinion. But I wanted to contribute these fixes in case someone else wants to play around with it or in hopes of Mono's WinForms improving in the future.

### DIFF
--- a/libse/Configuration.cs
+++ b/libse/Configuration.cs
@@ -47,12 +47,17 @@ namespace Nikse.SubtitleEdit.Core
 
         public static bool IsRunningOnLinux()
         {
-            return Environment.OSVersion.Platform == PlatformID.Unix;
+            return Environment.OSVersion.Platform == PlatformID.Unix && !IsRunningOnMac();
         }
 
         public static bool IsRunningOnMac()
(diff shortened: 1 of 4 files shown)


niksedk:
@jdpurcell: I guess you're using Xamarin?
I've actually got a mac now (still a PC guy though), and I will take a look at Xamarin and Cocoa as WinForms is not really usable or feel mac-like. However it seems that the new Xamarin framework does not support System.Drawing which is used a lot by SE. Xamarin can still use the old Mono framework, but then it cannot compile to 64-bit (and use latest VLC)...

jdpurcell:
I've only used the standard Xamarin Mono stuff, not like Xamarin Platform that provides access to platform-specific APIs.

Another project I've contributed to (BizHawk) is starting to use Eto.Forms to build cross-platform UIs. That may be worth looking into as well.

PR #1384, opened 2015-10-17
SE-installer: added [Registry] section with .srt association

xylographe:
Works as expected on Win7.
Is expected to work on XP, Vista and 8.1, too.

**`SE-3.4.10.60-xreg-setup.exe`** available for testing in [this folder](https://www.dropbox.com/sh/thgk5507tteg6ze/AABgb8jgWBnz9LLgDHsECfika?dl=0).

Note: SE-installer won't “steal” .srt from a previously registered application.

### DIFF
--- a/installer/Subtitle_Edit_installer.iss
+++ b/installer/Subtitle_Edit_installer.iss
@@ -98,6 +98,7 @@ SolidCompression=yes
 ShowTasksTreeLines=yes
 DisableReadyPage=yes
 PrivilegesRequired=admin
+ChangesAssociations=yes
 ShowLanguageDialog=yes
 DisableDirPage=auto
 DisableProgramGroupPage=auto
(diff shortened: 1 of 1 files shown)


niksedk:
Nice, what about an icon for the srt file type?
Something like http://d1dpm1gwwpbdrv.cloudfront.net/images/icons/lg/srt.png ?

niksedk:
Or like this: http://www.nikse.dk/srt.ico

niksedk:
Does anybody know how to show the .srt files as a document icon and not as an application icon?

ukdtom:
https://msdn.microsoft.com/en-us/library/windows/desktop/hh127427(v=vs.85).aspx

So add the desired icon to your exe, and follow above during installation

/T

xylographe:
@niksedk  See PR #1388

niksedk:
Thx :)

PR #1388, opened 2015-10-21
Show .srt files as a document icon instead of an application icon

xylographe:
The new `Application.res` Win32Resource file contains the SE application icon (at index 0) as well as the .srt document icon (at index 1). Other icons (for .ssa, .ass, .sup, etc.) could be added as well.

Because Win32Resource files are rather primitive, I've added the VS generated icon identifier to the icon file names for clarity:
- `SE-101.exe.ico` : the application icon
- `SE-102.srt.ico` : the .srt document icon


xylographe:
Sorry! Apparently, a Win32Resource file in .NET project is no good, as explained in [this post by Einar Egilsson.](http://einaregilsson.com/add-multiple-icons-to-a-dotnet-application/)  Einar's alternative, to add the document icons after the executable is created, would be better. Unfortunately, we cannot use that (MIT licensed) lib. :-(

niksedk:
We could use libse.dll for icon(s)...

xylographe:
1. libse doesn't seem the appropriate assembly for GUI stuff?
2. ATM libse is included (ILRepack) in the SE executable, not a separate assembly (dll).

There's a simple alternative in PR #1389. Disadvantage is a separate icon file for each document type.

PR #1389, opened 2015-10-21
SE-installer: use .srt document icon

xylographe:


niksedk:
For other file associations a plugin could be made...

PR #1324, opened 2015-09-25
[Internal] - Use built-in helper method to copy stream

ivandrofly:

PR #1396, opened 2015-10-24
Revert "[Internal] - Use built-in helper method to copy stream"

niksedk:
Reverts SubtitleEdit/subtitleedit#1324


niksedk:
code crashed...

PR #1401, opened 2015-10-29
SE-installer: use localized instead of fixed strings in Windows Registry

xylographe:
In the interest of localization, I added a resource-only DLL, containing localizable application and
document type descriptions, alongside the document type icon(s). By using a separate DLL, the size of
SE-portable is not needlessly increased: the DLL is only included with SE-installer.

Also, the registration of document types is now a user-selectable task.


niksedk:
I was unable to make .srt files open in SE even though I ticked the check box... 

Perhaps the text should be something like:
 Associate .srt files with Subtitle Edit
 Associate .srt file type with Subtitle Edit
 Always open SubRip (.srt) files with subtitle Edit
 ...

xylographe:
If you select a .srt file in explorer and open the context menu, you should see SE in the *Open with* sub-menu. To make SE the .srt default program you would open the *Properties* dialogue and click the **Change** button to the right of *Open with*. This will make SE the .srt default program for the current user (HKCU registry hive) as indicated by the icon (very nice BTW, I would like visually matching icons for .sup and .sub). The installer will only make SE the system-wide (HKLM registry hive) .srt default program if the .srt extension has not been claimed by another program. That is what I meant when I wrote: *SE-installer won't “steal” .srt from a previously registered application.*

Executing `get-item HKLM:\Software\Classes\.srt; get-item HKCU:\Software\Classes\.srt` in PoSh shows the default values. If the HKCU default value exists, it supersedes the HKLM default value.

In hindsight, I think, I may have been to cautious. So, I've changed the SE-installer behaviour:
- Original SE-installer behaviour:
  - file-association option is **unchecked:** only the SE registry value under "Classes\\Applications\\App Paths" will be created (used by explorer to locate SubtitleEdit.exe w/o path). The other SE registry entries (keys+values) will be deleted.
  - file-association option is **checked:** if the "Classes\\.srt" default value does not exist it will be set to "SubtitleEdit.srt", making SE the system-wide default program for .srt files. Any other SE-related registry entries will be created/overwritten unconditionally, marking SE as a program that supports .srt files (*Open with* sub-menu in the explorer context menu).
- New SE-installer behaviour:
  - file-association option is **unchecked:** create all SE registry entries except for the "Classes\\.srt" default value. SE will show in *Open with* and can be made the current user's .srt default program in *Properties*.
  - file-association option is **checked:** create all SE registry entries including the "Classes\\.srt" default value (overwriting the existing value, if any). Apart from showing in *Open with*, SE will also be the system-wide .srt default program, which users can overwrite individually.

This is consistent with many other installers (only Microsoft installers change the file type default program w/o asking).


Changed option text to: *Associate SubRip Text (.srt) files with Subtitle Edit*
Consistent with Microsoft terminology: *Associate a file type with a program*


As the name “DocumentIcons” was a poor choice, I changed that too:
- Project name: `Win32Resources`
- DLL name: `SubtitleEdit.resources.dll` (consistent with Microsoft file names)

niksedk:
Nice :)

.sub/.sup icons: http://www.nikse.dk/SE-Icons.zip

xylographe:
Thank you for the icons, @niksedk.

Showing these dedicated document icons looks so much better than using (as I've done until now) the SE application icon for subtitle files too.

PR #1413, opened 2015-11-08
Fix Autobreak fuction

ivandrofly:
If you try to Autobreak:
`<i>30 years ago I'd found</i> The Book of the Dead.`
it ends up like:

*\<i>30 years ago I'd found
\</i> The Book of the Dead.*


niksedk:
Nice catch :)
could you add the unit test also?

ivandrofly:
> Nice catch :)
> could you add the unit test also?


Sure :)

ivandrofly:
:+1:  /cc @niksedk

niksedk:
thx :)

PR #1415, opened 2015-11-12
Optimize ChangeCasingNames

ivandrofly:


src/Forms/ChangeCasingNames.cs
```diff
@@ -230,10 +227,12 @@ internal void FixCasing()
             {
                 if (item.Checked)
                 {
-                    _noOfLinesChanged++;
```
ivandrofly: I think this is probably a 'bug', **_noOfLinesChanged** should only be incremented if **paragraph** is not **null**, right?

PR #1425, opened 2015-12-02
Fix minor bug - thanks to @xylograph

ivandrofly:


ivandrofly:
I'll add more test ;)

PR #1440, opened 2015-12-08
Show fixed names in Change Casing without HTML tags

ivandrofly:


src/Forms/ChangeCasingNames.cs
```diff
@@ -76,12 +76,11 @@ private void GeneratePreview()
...
             foreach (Paragraph p in _subtitle.Paragraphs)
             {
-                string text = p.Text;
+                var text = p.Text;
+                var textNoTags = HtmlUtil.RemoveHtmlTags(text);
                 foreach (ListViewItem item in listViewNames.Items)
                 {
-                    string name = item.SubItems[1].Text;
-
-                    string textNoTags = HtmlUtil.RemoveHtmlTags(text);
```
ivandrofly: This should also be outside listViewNames's for-each

PR #1457, opened 2015-12-20
Update for OcrFixReplaceList.cs

ivandrofly:


niksedk:
Perhaps static array list variables for better performance?

ivandrofly:
> Perhaps static array list variables for better performance?

yeah, I wanted to added it but since the 2 arrays contains different items I didn't want to add 2 separated arrays. I'll push UPDATE soon

PR #1473, opened 2016-01-05
Remove redundant function call

ivandrofly:
I think we don't need to call `FixInvalidItalicTags` anymore `RemoveOpenCloseTags` will handle if tag contains spaces inside


niksedk:
'FixInvalidItalicTags' does stuff that 'RemoveOpenCloseTags' does not... (not often though)

PR #1489, opened 2016-01-17
Merge ToHHMMSSFF and ToHHMMSSPeriodFF

ivandrofly:
Merge ToHHMMSSFF and ToHHMMSSPeriodFF since the difference is only  between format
**colon** & **period**: 
`{0:00}:{1:00}:{2:00}`**:**`{3:00}` - `{0:00}:{1:00}:{2:00}`**.**`{3:00}`


libse/SubtitleFormats/SubtitleFormat.cs
```diff
@@ -342,6 +342,11 @@ public static int MillisecondsToFrames(double milliseconds)
             return (int)Math.Round(milliseconds / (TimeCode.BaseUnit / Configuration.Settings.General.CurrentFrameRate));
         }
 
+        public static double MillisecondsToFramesOutDouble(double milliseconds)
```
ivandrofly: Or maybe a generic?

PR #1490, opened 2016-01-18
[Refact] - Call helper method instead redefining ms/frame

ivandrofly:


libse/Paragraph.cs
```diff
@@ -114,12 +114,8 @@ public void Adjust(double factor, double adjustmentInSeconds)
         {
             if (StartTime.IsMaxTime)
                 return;
-
-            double seconds = StartTime.TimeSpan.TotalSeconds * factor + adjustmentInSeconds;
-            StartTime.TimeSpan = TimeSpan.FromSeconds(seconds);
-
-            seconds = EndTime.TimeSpan.TotalSeconds * factor + adjustmentInSeconds;
-            EndTime.TimeSpan = TimeSpan.FromSeconds(seconds);
+            StartTime.TotalSeconds *= factor + adjustmentInSeconds;
```
niksedk: Gives a wrong result... I've added some unit tests :)

ivandrofly: > Gives a wrong result... I've added some unit tests :)

Hmmm interesting, any idea what causes the wrong result?

niksedk: *= just multiplies, so it gives StartTime.TotalSeconds = StartTime.TotalSeconds * (actor + adjustmentInSeconds)
Did you test it?

ivandrofly: > *= just multiplies, so it gives StartTime.TotalSeconds = StartTime.TotalSeconds * (actor + adjustmentInSeconds)
> Did you test it?

Aaah okay, So *= will distribute `StartTime.TotalSeconds` :+1:

PR #1497, opened 2016-01-22
Add Apply feature to RemoveTextFromHearImpaired Close #1496

### LINKED ISSUE #1496: [Feature Request] Apply button to HI removal window
When I am using the hearing impaired removal feature, I tend to be careful not to remove anything important. This means that I remove lines over several steps, checking and un-checking the different options available.

Right now, your can confirm the changes by hitting the "OK" button, but this closes the window. Which means I have to re-open the window for every change I make.

I'd like either an Apply button, which will do the same as OK, but keep the window open. Or just leave the window open after hitting OK.

ivandrofly:
* Note: after user click <kbd>Apply</kbd> the whole listview items will be rechecked and if you click <kbd>OK</kbd> all the HI notation will be removed. Is it okay to leave it like this or we should remember the unchecked indexes? /cc @niksedk

### DIFF
--- a/src/Forms/Main.cs
+++ b/src/Forms/Main.cs
@@ -4961,10 +4961,10 @@ namespace Nikse.SubtitleEdit.Forms
             ReloadFromSourceView();
             using (var removeTextFromHearImpaired = new FormRemoveTextForHearImpaired())
             {
+                MakeHistoryForUndo(_language.BeforeRemovalOfTextingForHearingImpaired);
                 removeTextFromHearImpaired.Initialize(_subtitle);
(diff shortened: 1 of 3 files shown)


niksedk:
I'll do some testing...

niksedk:
Changing check boxes in the UI does not remember the unchecked indexes so that okay that "Apply" don't do that either - there's a minor issue with tab stop (I'll fix that). The undo could probably also be improved to only add history if anything is change...

ivandrofly:
> Changing check boxes in the UI does not remember the unchecked indexes so that okay that "Apply" don't do that either - there's a minor issue with tab stop (I'll fix that). The undo could probably also be improved to only add history if anything is change...

Nice :)

PR #1499, opened 2016-01-23
Cleanup + Refact

ivandrofly:


libse/RichTextToPlainText.cs
```diff
@@ -243,9 +243,9 @@ internal static string ConvertToRtf(string value)
 
             // special RTF chars
             var backslashed = new StringBuilder(value);
-            backslashed.Replace(@"\", @"\\");
-            backslashed.Replace(@"{", @"\{");
-            backslashed.Replace(@"}", @"\}");
+            backslashed.Replace('\\', '\\');
+            backslashed.Replace('{', '{');
+            backslashed.Replace('}', '}');
```
xylographe: `@"\\"` == `"\\\\"` (a string with **two** backslashes)
`@"\{"` == `"\\{"` (also a two character string)
`@"\}"` == `"\\}"`

PR #1500, opened 2016-01-23
Update + Clean up 2

ivandrofly:


niksedk:
Paragraphs is more an internal word... I'm however not sure to use the word "subtitle" or "line"... 
@xylographe: What do you think?

xylographe:
I think the the word **subtitle** is just fine in the dialogue caption: one subtitle can have one or more lines of text.  This is also consistent with the text of the menu item in Tools.

The word **lines** in the description of the NumericUpDown control also makes sense, because in the preview the subtitles are shown as a table: one subtitle per table line.

PR #1502, opened 2016-01-23
Move ToText to an Interface

ivandrofly:
# DO NOT MERGE YET


libse/SubtitleFormats/AdvancedSubStationAlpha.cs
```diff
@@ -38,7 +38,7 @@ public static string DefaultHeader
             {
                 SubtitleFormat format = new AdvancedSubStationAlpha();
                 var sub = new Subtitle();
-                string text = format.ToText(sub, string.Empty);
+                string text = ((IText)format).ToText(sub, string.Empty);
```
ivandrofly: Will find way to cast this since `SubtitleFormat` doesn't implementes `IText`

PR #1503, opened 2016-01-24
Move ToText to an Interface

ivandrofly:


niksedk:
I like a solution with a virtual method in the SubtitleFormat class better...

PR #1513, opened 2016-01-29
Refact ChangeFrameRate

ivandrofly:
Inspired from PR #850


niksedk:
Simpler and faster :)

(also better to do only one thing in a pull request)

ivandrofly:
> Simpler and faster :)
> 
> (also better to do only one thing in a pull request)

:+1:

PR #1518, opened 2016-01-30
[Bugfix] - Search will always start from 0 index

ivandrofly:


src/Logic/FindReplaceDialogHelper.cs
```diff
@@ -68,9 +68,9 @@ public FindReplaceDialogHelper(FindType findType, bool matchWholeWord, string fi
             MatchWholeWord = matchWholeWord;
         }
 
-        public bool Find(Subtitle subtitle, Subtitle originalSubtitle, int startIndex)
+        public bool Find(Subtitle subtitle, Subtitle originalSubtitle, int startIndex, int position = 0)
         {
-            return FindNext(subtitle, originalSubtitle, startIndex, 0, Configuration.Settings.General.AllowEditOfOriginalSubtitle);
```
ivandrofly: Note that the search will always start from TEXT position 0 which will cause the previously matched text to match again.

ivandrofly:
Have you tested this? /cc @niksedk

PR #1521, opened 2016-01-31
[LineNumber] - Call already defined tag counter

ivandrofly:


niksedk:
Made that way on purpose (optimization)

PR #1526, opened 2016-02-02
Update SpellCheckWordLists

ivandrofly:
* Move variable outside loop (Optimization)
* Use dictionary to store **id** & **wordWithDashesOrPeriods** instead of two lists
* Variable name consistency


ivandrofly:
Nice https://github.com/SubtitleEdit/subtitleedit/commit/768059359de7a10eb3ae71c4927fd6fe814017f3...
What do you think about using [Dictionary](https://github.com/SubtitleEdit/subtitleedit/pull/1526/files#diff-9e4e72215c5f592ba3b97997bfed63feR197) instead 2 lists? /cc @niksedk

Issue #1527, opened 2016-02-03
[Smells] - SpellCheckWordLists


ivandrofly:
Bug? https://github.com/SubtitleEdit/subtitleedit/blob/master/libse/SpellCheck/SpellCheckWordLists.cs#L323

``` c#
            var s = "foobar";
            var sb = new StringBuilder();
            for (int i = 0; i < s.Length; i++)
            {
                sb.Append(s[i]);
            }
            if (sb.Length > 0)
                Console.WriteLine(s.Length - 1 - sb.Length);
            Console.ReadLine();
```
**Output:** -1
I think that -1 shouldn't be there :confused:, right?

xylographe:
Logically it would be `Index = ( (s.Length - 1) - sb.Length ) + 1`, I believe, which could be simplified to `Index = s.Length - sb.Length`

PR #1530, opened 2016-02-03
[Bugfix] - Fix possible bug last 'word' wrong index

ivandrofly:
Ref: https://github.com/SubtitleEdit/subtitleedit/issues/1527


niksedk:
Good catch :)

PR #1539, opened 2016-02-05
[bugfix] - only tries to add hyphen if one of the lines doesn't have one

ivandrofly:
This will also fix bug e.g:

\{\an5\}\- At least I was going back to Hawaii.
\- Woody.

if you run FCE which text above and enable **Fix (add dash)** the text becomes:

{\an5}- At least I was going back to Hawaii.
**\- \-** Woody. _(add one more hyphen)_

Note: For two lines paragraphs only... PS: since the fix will only happen if paragraph contains 2 lines this is :+1:


niksedk:
Have you tested this?

ivandrofly:
> Have you tested this?

I did with text I posted here

PR #1544, opened 2016-02-07
Feat + Fix (Main.cs)

ivandrofly:


src/Forms/Main.cs
```diff
@@ -15601,34 +15603,44 @@ private void ColorToolStripMenuItem1Click(object sender, EventArgs e)
...
                             var start = s.IndexOf(" face=", StringComparison.Ordinal);
-                            s = s.Insert(start, string.Format(" color=\"{0}\"", color));
-                            text = s;
+                            text = s.Insert(start, string.Format(" color=\"{0}\"", color));
                             done = true;
                         }
                         else if (f.Contains(" color="))
                         {
-                            int colorStart = f.IndexOf(" color=", StringComparison.Ordinal);
-                            if (s.IndexOf('"', colorStart + " color=".Length + 1) > 0)
```
ivandrofly: `" color=".Length + 1` is a known constant which is equal to 8

src/Forms/Main.cs
```diff
@@ -15601,34 +15603,44 @@ private void ColorToolStripMenuItem1Click(object sender, EventArgs e)
                 string color = Utilities.ColorToHex(colorDialog1.Color);
                 bool done = false;
                 string s = text;
-                if (s.StartsWith("<font "))
+                if (s.StartsWith("<font ", StringComparison.OrdinalIgnoreCase))
```
ivandrofly: Colors aren't always lowercase so...

src/Forms/Main.cs
```diff
@@ -15399,10 +15399,11 @@ private void DeleteToolStripMenuItemClick(object sender, EventArgs e)
         private void NormalToolStripMenuItem1Click(object sender, EventArgs e)
         {
             var tb = GetFocusedTextBox();
-
             string text = tb.SelectedText;
+            if (text.Length == 0)
+                return;
             int selectionStart = tb.SelectionStart;
-            text = HtmlUtil.RemoveHtmlTags(text);
+            text = HtmlUtil.RemoveHtmlTags(text, true);
```
ivandrofly: Right Click + Normal (remove formatting) will not strip SSA tag

niksedk: That is on purpose...

ivandrofly: > That is on purpose...

Ah :+1:

PR #1570, opened 2016-02-16
[Optimization] - Do renumbering only if no paragraphs were removed

ivandrofly:
Once a paragraph is removed the RemoveEmptyLines will call Renumber so its redundant to re-run Renumber again


ivandrofly:
Did you get change to review this? /cc @niksedk

PR #1587, opened 2016-02-23
[Bugfix] - Utx won't load text if starts with '#'

ivandrofly:
Sorry for including the refact. ;)

### DIFF
--- a/libse/SubtitleFormats/Utx.cs
+++ b/libse/SubtitleFormats/Utx.cs
@@ -1,11 +1,14 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Text;
+using System.Text.RegularExpressions;
 
 namespace Nikse.SubtitleEdit.Core.SubtitleFormats
 {
     public class Utx : SubtitleFormat
     {
+        // #0:03:03.23,0:03:08.05
+        private static readonly Regex RegexTimeCode = new Regex(@"^#\d\d?:\d\d:\d\d\.\d\d,\d:\d\d:\d\d\.\d\d$", RegexOptions.Compiled);
 
         public override string Extension
         {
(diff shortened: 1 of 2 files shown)


libse/SubtitleFormats/Utx.cs
```diff
@@ -53,37 +56,37 @@ public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string
...
+            char[] splitChars2 = { ':', '.' };
             for (int i = 0; i < lines.Count; i++)
             {
                 string line = lines[i].Trim();
-
-                if (line.StartsWith('#'))
+                if (line.StartsWith('#') && RegexTimeCode.IsMatch(line))
                 {
-                    var timeParts = line.Split(new[] { '#', ',' }, StringSplitOptions.RemoveEmptyEntries);
-                    if (timeParts.Length == 2)
```
ivandrofly: Will be handled inside `DecodeTimeCodeFramesFourParts` since `DecodeTimeCodeFramesFourParts` throws exception if parts aren't equal to 4 this will improve detecting error count (_errorCount)

libse/SubtitleFormats/Utx.cs
```diff
@@ -53,37 +56,37 @@ public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string
...
+                        subtitle.Paragraphs.Add(new Paragraph(start, end, sb.ToString().Trim()));
+                    }
+                    catch
+                    {
+                        _errorCount++;
                     }
                 }
                 else if (line.Length > 0)
                 {
-                    text.AppendLine(line.Trim());
```
ivandrofly: line.Trim() is redundant here, line is already trimmed https://github.com/SubtitleEdit/subtitleedit/pull/1587/files#diff-a1759a18bf40f7264007c4b7775a2160R64

libse/SubtitleFormats/Utx.cs
```diff
@@ -53,37 +56,37 @@ public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string
...
             subtitle.Paragraphs.Clear();
-            var text = new StringBuilder();
+            var sb = new StringBuilder();
+            char[] splitChars = { '#', ',' };
+            char[] splitChars2 = { ':', '.' };
             for (int i = 0; i < lines.Count; i++)
             {
                 string line = lines[i].Trim();
-
-                if (line.StartsWith('#'))
```
ivandrofly: Here is where the bug was present. Since subtitle text can start with `#` we should always check if it's time-code or not, if we assume that it's time-code and time-code parsing process fails the text will be ignored.

niksedk:
Nice :)
(still try to remember separating refact/bugfix... I also forget sometimes, but it's just much clearer)

PR #1596, opened 2016-02-26
[SubtitleFormat] - Update (UnknownSubtitle36.cs)

ivandrofly:


libse/SubtitleFormats/UnknownSubtitle36.cs
```diff
@@ -28,7 +28,7 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
 
     public class UnknownSubtitle36 : SubtitleFormat
     {
-        private static readonly Regex RegexTimeCodes = new Regex(@"^\d+:\d+:\d+:\d+           ,           \d+:\d+:\d+:\d+           ,.*$", RegexOptions.Compiled);
```
ivandrofly: I think `.*` is not a good idea.
we only need to filter time-code with regex and the rest will be treated as string

libse/SubtitleFormats/UnknownSubtitle36.cs
```diff
@@ -57,7 +57,7 @@ public override bool IsMine(List<string> lines, string fileName)
 
         public override string ToText(Subtitle subtitle, string title)
         {
-            const string paragraphWriteFormat = "{0}           ,           {1}           ,{2}\r\n";
+            string paragraphWriteFormat = "{0}           ,           {1}           ,{2}" + Environment.NewLine;
```
ivandrofly: Platform independent

libse/SubtitleFormats/UnknownSubtitle36.cs
```diff
@@ -80,20 +82,21 @@ public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string
                     continue;
                 }
 
-                if (RegexTimeCodes.Match(line).Success)
+                if (line.Length > 57 && (match = RegexTimeCodes.Match(line)).Success)
```
ivandrofly: First check if line has probability to match the RegexTimeCodes before trying to match

PR #1603, opened 2016-02-27
Update (UnknownSubtitle3.cs)

ivandrofly:


libse/SubtitleFormats/UnknownSubtitle3.cs
```diff
@@ -64,14 +64,14 @@ private void ReadLine(Subtitle subtitle, string line)
         {
             // 150583||3968723||Rythme standard quatre-par-quatre.\~- Sûr... Accord d'entrée, D majeur?||
             // 155822||160350||Rob n'y connait rien en claviers. Il\~commence chaque chanson en D majeur||
-            string[] parts = line.Split(new char[] { '|' }, StringSplitOptions.RemoveEmptyEntries);
```
ivandrofly: Inside a method which is being called by a loop

PR #1600, opened 2016-02-27
[SubtitleFormat] - Update (SubViewer20.cs)

ivandrofly:

PR #1605, opened 2016-02-28
Revert "[SubtitleFormat] - Update (SubViewer20.cs)"

niksedk:
Reverts SubtitleEdit/subtitleedit#1600


niksedk:
Always reports more errors than good lines...

ivandrofly:
> Always reports more errors than good lines...

That is because the metadata + CRLN were include in List lines. Should be fixed in next PR

PR #1608, opened 2016-02-29
[Optimization] - Minor optimization for (UnknownSubtitle37.cs)

ivandrofly:


libse/SubtitleFormats/UnknownSubtitle37.cs
```diff
@@ -39,15 +39,18 @@ public override string ToText(Subtitle subtitle, string title)
...
         {
+            if (lines.Count > 0 && !lines[0].TrimStart().StartsWith("{\\rtf", StringComparison.Ordinal))
+            {
+                _errorCount++;
+                return;
+            }
+
             _errorCount = 0;
             var sb = new StringBuilder();
             foreach (string line in lines)
```
ivandrofly: We don't need to append entire lines and then do the check

PR #1612, opened 2016-03-02
Update (UnknownSubtitle41.cs)

ivandrofly:
This patch will also drop Elapsed tick from 42312863 to 9313 there is no ambiguity with others subtitle formats.


niksedk:
I think you're doing too much... the combine regex is a good one :)
I'll pick some of it...

PR #1614, opened 2016-03-03
Minor fix/Update

ivandrofly:


libse/VobSub/VobSubParser.cs
```diff
@@ -215,7 +215,7 @@ public static bool IsMpeg2PackHeader(byte[] buffer)
 
         public static bool IsPrivateStream1(byte[] buffer, int index)
         {
-            return buffer.Length >= index + 3 &&
+            return buffer.Length >= index + 4 &&
```
ivandrofly: This is actually a bug fix, right? /cc @niksedk

xylographe: Yes, if `buffer.Length` equals `index + 3`, then the last comparison (`buffer[index + 3] == 0xbd`) will throw an exception, assuming the three previous comparisons are true (byte buffer ends with \[0, 0, 1], there is no fourth byte ).

PR #1622, opened 2016-03-06
Corrected misspelled word in Language.cs

xylographe:
[Oops,](https://www.oxforddictionaries.com/definition/learner/oops) "ups" is the plural of "up" – as in ["ups and downs"](https://www.oxforddictionaries.com/definition/american_english/ups-and-downs?q=ups+and+downs). :smile:


niksedk:
Good catch - actually "Ups" is the danish word for "Oops", so "oops" ;)

PR #1630, opened 2016-03-13
[SubtitleListView] - Remove extra column index from all lvi subitem

ivandrofly:
`

    public void HideExtraColumn()
        {
            if (IsExtraColumnVisible)
            {
                IsExtraColumnVisible = false;

                if (IsAlternateTextColumnVisible)
                    ColumnIndexExtra = ColumnIndexTextAlternate + 1;
                else
                    ColumnIndexExtra = ColumnIndexTextAlternate;

                for (int i = 0; i < Items.Count; i++)
                {
                    if (Items[i].SubItems.Count == ColumnIndexExtra + 1)
                    {
                        Items[i].SubItems.RemoveAt(ColumnIndexExtra);
                    }
                }
                Columns.RemoveAt(ColumnIndexExtra);
                SubtitleListViewResize(null, null);
            }
        }
`
So far it's not causing any problem but let say in future we decide to use "Items[i].SubItems.Count" after exiting the loop it will return wrong value

### DIFF
--- a/src/Controls/SubtitleListView.cs
+++ b/src/Controls/SubtitleListView.cs
@@ -765,9 +765,7 @@ namespace Nikse.SubtitleEdit.Controls
                 {
                     if (Items[i].SubItems.Count == ColumnIndexExtra + 1)
                     {
-                        Items[i].SubItems[ColumnIndexExtra].Text = string.Empty;
-                        Items[i].SubItems[ColumnIndexExtra].BackColor = BackColor;
-                        Items[i].SubItems[ColumnIndexExtra].ForeColor = ForeColor;
+                        Items[i].SubItems.RemoveAt(ColumnIndexExtra);
                     }
                 }
                 Columns.RemoveAt(ColumnIndexExtra);


src/Controls/SubtitleListView.cs
```diff
@@ -765,9 +765,7 @@ public void HideExtraColumn()
                 {
                     if (Items[i].SubItems.Count == ColumnIndexExtra + 1)
                     {
-                        Items[i].SubItems[ColumnIndexExtra].Text = string.Empty;
-                        Items[i].SubItems[ColumnIndexExtra].BackColor = BackColor;
-                        Items[i].SubItems[ColumnIndexExtra].ForeColor = ForeColor;
+                        Items[i].SubItems.RemoveAt(ColumnIndexExtra);
```
niksedk: Hm, is this if statement ever entered?

ivandrofly: > Hm, is this if statement ever entered?

yeah, just change the subtitle format for example (Sami) and add a break point

niksedk: :)

PR #1632, opened 2016-03-14
[UiUtils] - Fix autobreak in SE textbox

ivandrofly:


ivandrofly:
This patch fixes autobreak in SE's textbox

src/Logic/UiUtil.cs
```diff
@@ -228,35 +228,36 @@ public static void InitializeVideoPlayerAndContainer(string fileName, VideoInfo
 
         public static void CheckAutoWrap(TextBox textBox, KeyEventArgs e, int numberOfNewLines)
         {
-            int length = HtmlUtil.RemoveHtmlTags(textBox.Text).Length;
-            if (e.Modifiers == Keys.None && e.KeyCode != Keys.Enter && numberOfNewLines < 1 && length > Configuration.Settings.General.SubtitleLineMaximumLength)
```
ivandrofly: Issue was here numberOfNewLines would never be less then 1.

niksedk:
Nice find :)

PR #1636, opened 2016-03-16
[Language] - Stage language deserializer/master language

ivandrofly:
Every time I build the solution these two files are automatically changed


xylographe:
Looks like your `libse\LanguageStructure.cs` and `libse\Language.cs` are outdated?

ivandrofly:
> Looks like your libse\LanguageStructure.cs and libse\Language.cs are outdated?

yeah, It's been generated since lang file was changed :confused:

xylographe:
```
git fetch upstream master
git diff master upstream/master
```
The tags you remove  in this commit do exist in the current [Language.cs](https://github.com/SubtitleEdit/subtitleedit/blob/master/libse/Language.cs#L2025) from upstream/master.

ivandrofly:
> The tags you remove in this commit do exist in the current Language.cs from upstream/master.

AH! the problem is that I'm using a custom Solution (*.sln are excluded locally) I after built using SubtitleEdit.sln seems to work fine :)

PR #1639, opened 2016-03-16
Update ru-RU.xml

Elheym:
Latest updates


niksedk:
thx for the update - good timing :)

PR #1641, opened 2016-03-17
[Bugfix] - Weird behavior (Wrong line #) when navigating using UP/Down Keys

ivandrofly:
To produce bug switch to source view mode and try navigating using e.g UP/Down arrow and see how wrong line numbers are displayed.

### DIFF
--- a/src/Forms/Main.cs
+++ b/src/Forms/Main.cs
@@ -3875,9 +3875,13 @@ namespace Nikse.SubtitleEdit.Forms
             ShowSourceLineNumber();
         }
 
-        private void TextBoxSourceKeyDown(object sender, KeyEventArgs e)
+        private void textBoxSource_KeyUp(object sender, KeyEventArgs e)
         {
             ShowSourceLineNumber();
+        }
+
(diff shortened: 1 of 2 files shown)


src/Forms/Main.cs
```diff
@@ -3875,9 +3875,13 @@ private void TextBoxSourceClick(object sender, EventArgs e)
             ShowSourceLineNumber();
         }
 
-        private void TextBoxSourceKeyDown(object sender, KeyEventArgs e)
+        private void textBoxSource_KeyUp(object sender, KeyEventArgs e)
         {
             ShowSourceLineNumber();
+        }
+
+        private void TextBoxSourceKeyDown(object sender, KeyEventArgs e)
```
ivandrofly: Since the caret will only change position after the process is out of this method so ShowSourceLineNumber() will be called and **textBoxSource.GetLineFromCharIndex(textBoxSource.SelectionStart)** will return line of caret's current position.

niksedk:
thx - nice catch :)
Also committed in 1a198985abae0b0524bd61fb83bf6adadf857ffc

PR #1645, opened 2016-03-19
[EffectKaraoke] - More fix/update

ivandrofly:


src/Forms/EffectKaraoke.cs
```diff
@@ -50,7 +50,7 @@ internal void Initialize(Paragraph paragraph)
             AddToPreview(richTextBoxPreview, paragraph.Text);
             RefreshPreview();
             labelTotalMilliseconds.Text = string.Format("{0:#,##0.000}", paragraph.Duration.TotalMilliseconds / TimeCode.BaseUnit);
-            numericUpDownDelay.Maximum = (int)((paragraph.Duration.TotalMilliseconds - 500) / TimeCode.BaseUnit);
+            numericUpDownDelay.Maximum = (decimal)((paragraph.Duration.TotalMilliseconds - 500) / TimeCode.BaseUnit);
```
ivandrofly: Casting to `int` will trim floating point. (Found while fixing EffectTypeWritter)

src/Forms/EffectKaraoke.cs
```diff
@@ -222,8 +222,6 @@ private void ButtonPreviewClick(object sender, EventArgs e)
 
         private void PlayAnimation()
         {
-            _colorList = new List<ColorEntry>();
-            _fontList = new List<FontEntry>();
```
ivandrofly: Will be called when timer rise tick rises up.

PR #1647, opened 2016-03-20
Update UiUtil.cs

ivandrofly:


src/Logic/UiUtil.cs
```diff
@@ -339,11 +339,10 @@ public static void FixLargeFonts(Control mainCtrl, Control ctrl)
         {
             using (Graphics graphics = mainCtrl.CreateGraphics())
             {
-                SizeF textSize = graphics.MeasureString(ctrl.Text, mainCtrl.Font);
+                SizeF textSize = graphics.MeasureString(ctrl.Text, ctrl.Font);
```
ivandrofly: We shouldn't always assume that `mainCtrl` and `ctrl` will have same font.

PR #1648, opened 2016-03-20
[UiUtil] - Fix wrong coloring label info (line > maxline)

ivandrofly:


src/Logic/UiUtil.cs
```diff
@@ -389,10 +389,14 @@ public static void GetLineLengths(Label label, string text)
                 }
 
                 sb.Append(line.Length);
-                if (line.Length > Configuration.Settings.General.SubtitleLineMaximumLength)
-                    label.ForeColor = Color.Red;
-                else if (line.Length > Configuration.Settings.General.SubtitleLineMaximumLength - 5)
-                    label.ForeColor = ColorDarkOrange;
```
ivandrofly: label info his fore color will always be dark orange as long as last ling is not greater than `Configuration.Settings.General.SubtitleLineMaximumLength`

niksedk:
How can I see this bug in the UI?

ivandrofly:
# Before
Before

# After
After

PR #1649, opened 2016-03-21
[UpdateAssemblyInfo] Update General/Version in XML translation files

xylographe:
In addition to updating the short version (major.minor.maintenance) in `AssemblyInfo.cs.template`, also update `<General><Version>` in the XML translation files when preparing a new release.

Convenient side effect: ensures that all translation files contain syntactically valid XML, because a malformed translation file would throw an exception (`XmlDocument.Load()`).

New release would go smth like this:
- Update `Changelog.txt`
- Increase version to 3.4.13 in SubtitleEdit project properties "Assembly Information" in Visual Studio
- Rebuild solution (updates AssemblyInfo templates, translations, and LanguageMaster)
- `git commit -am "Release 3.4.13"`
- `git tag 3.4.13`
- Rebuild solution (set full version and hash from new git tag)
- `git push --tags origin master`

### DIFF
--- a/src/UpdateAssemblyInfo/Program.cs
+++ b/src/UpdateAssemblyInfo/Program.cs
@@ -1,8 +1,10 @@
 ﻿using System;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
+using System.Xml;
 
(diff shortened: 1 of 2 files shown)


xylographe:
Improved file name selection: regex i/o glob pattern.

PR #1652, opened 2016-03-23

diomed:


Dictionaries/hrv_OCRFixReplaceList.xml
```diff
@@ -573,6 +574,7 @@
     <RegEx find="([bB])ožij([aeiu]|e[mg]|om|ima)?\b" replaceWith="$1ožj$2" />
     <RegEx find="boži[čć]([aeiu]|em|ima)?\b" replaceWith="Božić$1" />
     <RegEx find="(?&lt;!\A|[.!?][&quot;”’]?\s+)\bBoži[čć]n([aeiou]|om|im)\b" replaceWith="božićn$1" />
+    <RegEx find="([bB])ubašvab" replaceWith="žohar" />
```
xylographe: Perhaps, `$1` is missing in `replaceWith`? (`<RegEx find="([bB])ubašvab" replaceWith="$1žohar" />`)

diomed: No, it doesn't. It's supposed to be with big Ž and little ž, but I left it like this instead. It's not that often in beginning of the sentence anyway.

please check out gitter when you find time, I left you a couple of pm's there. :/

PR #1653, opened 2016-03-23
[Subtitle] - Add helper method to get all text from Subtitle

ivandrofly:


libse/Subtitle.cs
```diff
@@ -580,5 +580,21 @@ public string GetFastHashCode()
...
+            var sb = new StringBuilder();
+            foreach (Paragraph p in Paragraphs)
+            {
+                if (appendNewLine)
+                {
+                    sb.AppendLine(p.Text);
+                }
+                else
+                {
+                    sb.Append(p.Text.TrimEnd() + ' ');
```
xylographe: Perhaps, `sb.Append(p.Text.TrimEnd()).Append(' ')` would be more efficient?

xylographe: And perhaps, to be somewhat consistent with `File.ReadAllText` (which preserves line endings),
 `public string GetAllText(bool joinLines = false)`

ivandrofly: What do you suggest should we just append without removing lines break (\r\n) from p.Text? in [sb.Append(p.Text...)]

xylographe: I would probably go with:
```
        public string GetAllText(bool joinLines = false)
        {
            var sb = new StringBuilder();
            foreach (Paragraph p in Paragraphs)
            {
                if (joinLines)
                {
                    sb.Append(p.Text.TrimEnd()).Append(' ');
                }
                else
                {
                    sb.AppendLine(p.Text);
                }
            }
            return joinLines ? sb.ToString().TrimEnd() : sb.ToString();
        }
```
Or, if joining lines is not necessary, use a property:
```
        public string AllText
        {
            get
            {
                var sb = new StringBuilder();
                foreach (Paragraph p in Paragraphs)
                {
                    sb.AppendLine(p.Text);
                }
                return sb.ToString();
            }
        }
```
Of course, joined lines **could** even be a separate property:
```
        public string AllTextJoined
        {
            get
            {
                var sb = new StringBuilder();
                foreach (Paragraph p in Paragraphs)
                {
                    sb.Append(p.Text.TrimEnd()).Append(' ');
                }
                return sb.ToString().TrimEnd();
            }
        }
```
However, the question here is “What does Nikolaj prefer?” :wink:

xylographe: Oops, obviously, I wasn't thinking clearly. `Paragraph.Text` may contain line breaks. Hence, the return statement should probably be something like:
```
sb.Replace("\r\n", " ").Replace('\r', ' ').Replace('\n', ' ').ToString().TrimEnd()
```

niksedk:
Perhaps it could be named "GetTextOnly" / "GetAllTexts" or something similar to avoid confusing it with getting the raw text of a subtitle format which includes time codes and more... perhaps a comment telling about the return value.

Just always returning the new line seems simpler to me.

ivandrofly:
Updated /cc @niksedk

PR #1656, opened 2016-03-24
[SplitLongLinesHelper] - Fix wrong single line max character.

ivandrofly:


libse/Forms/SplitLongLinesHelper.cs
```diff
@@ -8,25 +8,25 @@ public static class SplitLongLinesHelper
...
+            foreach (string line in s.SplitToLines())
             {
                 if (line.Length > singleLineMaxCharacters)
                     return true;
             }
 
-            var tempText = Utilities.UnbreakLine(text);
-            if (Utilities.CountTagInText(tempText, '-') == 2 && (text.StartsWith('-') || text.StartsWith("<i>-")))
+            s = Utilities.UnbreakLine(s);
+            if ((s.StartsWith('-') || s.StartsWith("<i>-", StringComparison.Ordinal)) && Utilities.CountTagInText(s, '-') == 2)
```
ivandrofly: Check if *s* start with '-' before Counting '-'

libse/Forms/SplitLongLinesHelper.cs
```diff
@@ -8,25 +8,25 @@ public static class SplitLongLinesHelper
...
 
-            var arr = s.SplitToLines();
-            foreach (string line in arr)
+            foreach (string line in s.SplitToLines())
             {
                 if (line.Length > singleLineMaxCharacters)
                     return true;
             }
 
-            var tempText = Utilities.UnbreakLine(text);
```
ivandrofly: tempText will still contain html tags if present which will give us unwanted result in line 31

libse/Forms/SplitLongLinesHelper.cs
```diff
@@ -8,25 +8,25 @@ public static class SplitLongLinesHelper
...
-            if (Utilities.CountTagInText(tempText, '-') == 2 && (text.StartsWith('-') || text.StartsWith("<i>-")))
+            s = Utilities.UnbreakLine(s);
+            if (s.StartsWith('-') && Utilities.CountTagInText(s, '-') == 2)
```
ivandrofly: s doesn't contain any tag so we can just check if it starts with '-'

libse/Forms/SplitLongLinesHelper.cs
```diff
@@ -8,25 +8,25 @@ public static class SplitLongLinesHelper
         public static bool QualifiesForSplit(string text, int singleLineMaxCharacters, int totalLineMaxCharacters)
         {
             string s = HtmlUtil.RemoveHtmlTags(text.Trim(), true);
+            // TODO: Exclude \r\n.
```
ivandrofly: Will add optimized method for this

PR #1661, opened 2016-03-27
[NamesList] - Combine two collection using AddRange.

ivandrofly:
Take a look: http://stackoverflow.com/questions/4488054/merge-two-or-more-lists-into-one-in-c-sharp-net


niksedk:
Nice :)

PR #1666, opened 2016-03-30
Replaced duplicated code with ApplyControlBackspace method

xylographe:
- Fixes input `<i>foo`, result `<i`, expected `<i>`.
- Supports French punctuation spacing.
- Treats HTML tag as a unit.

### DIFF
--- a/src/Controls/SETextBox.cs
+++ b/src/Controls/SETextBox.cs
@@ -1,4 +1,5 @@
 ﻿using Nikse.SubtitleEdit.Core;
+using Nikse.SubtitleEdit.Logic;
 using System;
 using System.Drawing;
 using System.Windows.Forms;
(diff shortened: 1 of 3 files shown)


niksedk:
Seems to work fine - how do I see the issue with  `<i>foo` and `Treats HTML tag as a unit`?

xylographe:
Just enter **`mortadella e <i>pâté`** in SETextBox, and press **Ctl+BS**.
- Without this commit the result is **`mortadella e <i`** (closing bracket of HTML tag is deleted too).
- With this commit the result is **`mortadella e <i>`** (only `pâté` is deleted). Press **Ctl+BS** again, and the result is **`mortadella e `** (whole HTML tag, including the opening bracket, will be deleted).

Another example of  *"Treats HTML tag as a unit"* :
**`mortadella e <font face="Segoe Print" colour="#ff0000"> `** followed by **Ctl+BS** 
results in **`mortadella e `**.
However, if you were to position the caret inside the tag before pressing **Ctl+BS**, deletion will be normal (for example, with the caret at the first `=` only `face` will be deleted).

niksedk:
OK, thx for explaining - it works very nice :)

PR #1670, opened 2016-04-01
[AdjustSelectedLines] - Always adjust selection if greater than 1.

ivandrofly:
If user select more than one item from subtitle-listview and click `Tools\Adjust durations` the intention is always to adjust selected line right?

(I know it can also be done by selecting lines from subtitle-listview and right click `Adjust duration for selected lines..`) but it's also useful to do it from Tools.
I also think it's quicker to do adjustment from `Tools` (Mouse).


niksedk:

PR #1682, opened 2016-04-05
[Subtitle] - Do not try reloading same format twice

ivandrofly:


libse/Subtitle.cs
```diff
@@ -96,21 +96,24 @@ public Paragraph GetParagraphOrDefaultById(string id)
...
+                {
+                    if (subtitleFormat.IsMine(lines, fileName))
+                    {
+                        subtitleFormat.LoadSubtitle(this, lines, fileName);
+                        _format = subtitleFormat;
+                        return subtitleFormat;
+                    }
+                }
+            }
+            else if (format.IsMine(lines, fileName))
```
ivandrofly: Will prevent same format to run twice!

If given format is null and **IsMine** returns false just return null.

PR #1690, opened 2016-04-07
Minor fix + refact (SubtitleFormats/Son)

xylographe:
The index and start-timecode are separated by one **or more** HTs (`regexTimeCodes` regex in `LoadSubtitle`). Therefore it is not possible to use `Substring` with a fixed index to extract the start/end timecodes from the input line.

### DIFF
--- a/libse/SubtitleFormats/Son.cs
+++ b/libse/SubtitleFormats/Son.cs
@@ -25,24 +25,19 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
         public override bool IsMine(List<string> lines, string fileName)
         {
             var subtitle = new Subtitle();
-
-            var sb = new StringBuilder();
-            foreach (string line in lines)
-                sb.AppendLine(line);
-
             LoadSubtitle(subtitle, lines, fileName);
             return subtitle.Paragraphs.Count > _errorCount;
         }
 
         public override string ToText(Subtitle subtitle, string title)
         {
+            const string writeFormat = "{0:0000}\t{1}\t{2}\t{3}";
             var sb = new StringBuilder();
             int index = 0;
-            const string writeFormat = "{0:0000}\t{1}\t{2}\t{3}";
             foreach (Paragraph p in subtitle.Paragraphs)
             {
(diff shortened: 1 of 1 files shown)


libse/SubtitleFormats/Son.cs
```diff
@@ -56,32 +51,24 @@ private static string EncodeTimeCode(TimeCode time)
...
             int index = 0;
             foreach (string line in lines)
             {
-                if (regexTimeCodes.IsMatch(line))
+                var match = regexTimeCodes.Match(line);
+                if (match.Success)
                 {
-                    string temp = line.Substring(0, regexTimeCodes.Match(line).Length);
-                    string start = temp.Substring(5, 11);
-                    string end = temp.Substring(12 + 5, 11);
```
xylographe: Fixed start indices (`5` and `12+5`)  will be incorrect if index and start time-code are separated by more than one HT.

libse/SubtitleFormats/Son.cs
```diff
@@ -56,32 +51,24 @@ private static string EncodeTimeCode(TimeCode time)
...
-
-                    string[] startParts = start.Split(SplitCharColon, StringSplitOptions.RemoveEmptyEntries);
-                    string[] endParts = end.Split(SplitCharColon, StringSplitOptions.RemoveEmptyEntries);
-                    if (startParts.Length == 4 && endParts.Length == 4)
```
xylographe: Always true. Otherwise there would be no match.

libse/SubtitleFormats/Son.cs
```diff
@@ -56,32 +51,24 @@ private static string EncodeTimeCode(TimeCode time)
...
-                    {
-                        int lastIndexOfTab = line.LastIndexOf('\t');
-                        string text = line.Remove(0, lastIndexOfTab + 1).Trim();
-                        if (!text.Contains(Environment.NewLine))
-                            text = text.Replace("\t", Environment.NewLine);
```
xylographe: Impossible. `text` is the part of `line` after the last HT.

PR #1694, opened 2016-04-09
[FixShortDisplayTimes] - Calculate time without loop.

ivandrofly:


niksedk:

PR #1696, opened 2016-04-09
Replaced IfoParser language lists with DvdSubtitleLanguage class

xylographe:
Provides localized language names for UI elements (ComboBox).


xylographe:
Some refactoring:
- Replace `GetLanguage` method (which was only invoked with literal string `"en"`) with `English` property.
- Use `KeyedCollection` to improve look-up performance from O(n) to O(1).
- Remove `GetHashCode` override, and other surplus code.

libse/DvdSubtitleLanguage.cs
```diff
@@ -0,0 +1,206 @@
...
+        private static string GetNonCompliantLocalLanguageName(string code)
+        {
+            string name = null;
+            try
+            {
+                var seCulture = CultureInfo.GetCultureInfo(Configuration.Settings.Language.General.CultureName);
+                var uiCulture = Thread.CurrentThread.CurrentUICulture;
+                if (seCulture.Name != uiCulture.Name)
+                {
+                    Thread.CurrentThread.CurrentUICulture = seCulture;
```
niksedk: Could you explain why you're setting the current ui culture?

xylographe:
Because SE can use a translation that is different from `CurrentUICulture`.

For example:
```
PS> [System.Globalization.CultureInfo]::CurrentUICulture

LCID             Name             DisplayName
----             ----             -----------
1043             nl-NL            Nederlands (Nederland)
```
But in SE I was using the `fr-FR` translation. So, I want the localized language name in SE to be in French, not in Dutch.

AFAIK .Net does not offer a `GetCultureInfo` overload that takes two culture names.
There doesn't seem to be a simple method to get the `DisplayName` (localized name) of culture xx in the language of culture yy. Setting `CurrentUICulture` to yy first and then asking for xx, would seem to be the only way to achieve this?

If culture yy has not been installed falling back to the language ot the Windows default culture (i.e., `CurrentUICulture`) would seem to be ‘reasonable’ (the alternative is using `EnglishName`).

If culture xx has not been installed the name will be "Unknown (xx)" (where the word "Unknown" is, of course, localized). The alternative would be to use the internet (e.g., CLDR core), but that seems a tad excessive :) for something that will happen very rarely.

niksedk:
Do you think it a good idea  to change the UI culture just because a user loads a subtitle.?

xylographe:
The change of UI culture is limited to the executing thread, other threads are not affected.

xylographe:
Would you rather skip this step and immediately fall back?
If so, which fall back?
1. language name in language of the UI culture
2. language name in English

niksedk:
What is the reason for setting the UI culture?

xylographe:
Getting the language name (for some two letter language code) in the same language as the active translation. Only if the translation language is not the same as the current UI culture language.

Of course, this only happens if the language code is not DVD compliant.

xylographe:
Yes, it will be the SE UI thread. But the change of UI culture is restored immediately after invoking `GetCultureInfo`.

xylographe:
Removed setting the UI culture.
English name is returned if SE culture and UI culture are different.

PR #1702, opened 2016-04-12
[Utilities] - Unbreak each line-breaks chars individually.

ivandrofly:


libse/Utilities.cs
```diff
@@ -684,7 +684,9 @@ public static string UnbreakLine(string text)
             if (!text.Contains(Environment.NewLine))
                 return text;
 
-            var singleLine = text.Replace(Environment.NewLine, " ");
+            text = text.Replace('\r', ' ');
+            text = text.Replace('\n', ' ');
+            var singleLine = text;
```
ivandrofly: We shouldn't assume that text will ways contains use \r\n (Carriage-Return+Line-Feed) as line break character.

niksedk: Perhaps it's possible to use `SplitToLines` + join the strings?

PR #1704, opened 2016-04-12
[Utilities] - Exit the loop once *allOk is false.

ivandrofly:


ivandrofly:
Not much to review :)... /cc @niksedk

niksedk:
The array will normally be of 1-2 elements, so this not needed...

PR #1712, opened 2016-04-16
Added /multiplereplace to command line convert

xylographe:
Support for importing replace rule files will be added later without changing this cli.
`"/multiplereplace:., mr1.xml, mr2.xml"` will then use default rules (Settings) + rules from two files.
`/multiplereplace` will remain as a convenient shortcut for `/multiplereplace:.` (default rules only).

Some other small changes:
- Improved accepted pattern—`*.ssa,*.pac,sub\*,sup\*,F:/st/ebu/*` for example.
- Improved usage message—the one liner was getting ridiculously long. :)
- Accept `/?` as first argument.
- Several tiny fixes.


niksedk:
Looks good :)

PR #1723, opened 2016-04-24
fixed broken link to German (Germany) dictionary

be-sc:
I’m not a .NET or Visual Studio developer, so I couldn’t actually test Subtitle Edit with the new link. But the change seemed simple enough and the new link does work.

The .gz was created with 7-Zip and default settings.


niksedk:
thx :)

PR #1738, opened 2016-05-01
[Bugfix] - Do not insert dash if 1st line length is equal to 1.

ivandrofly:
Try this:

1
00:38:02,313 --> 00:38:03,792
TRACY & BRIAN: I
like the sleeves.

**Becomes:**
\- I
\- like the sleeves.


niksedk:

PR #1740, opened 2016-05-02
Fix miss replace '#' with music symbol.

ivandrofly:
/cc @niksedk


niksedk:

PR #1760, opened 2016-05-17
[Refact] - Minor refact.

ivandrofly:


src/Forms/FixCommonErrors.cs
```diff
@@ -495,7 +495,7 @@ public bool AllowFix(Paragraph p, string action)
         public void ShowStatus(string message)
         {
             message = message.Replace(Environment.NewLine, "  ");
-            if (message.Length > 83)
+            if (message.Length >= 83)
                 message = message.Substring(0, 80) + "...";
```
xylographe: Why replace the last three characters with `...`? (if `message.Length == 83`)

ivandrofly: > Why replace the last three characters with ...? (if message.Length == 83)

Only 80 char 

> Why replace the last three characters with ...? (if message.Length == 83)

Ah! you are right, the intention was only to allow 80. :+1:

xylographe: Perhaps,
```
if (message.Length > 80)
    message = message.Remove(79) + '\u2026'; // … U+2026 'HORIZONTAL ELLIPSIS'
```

PR #1761, opened 2016-05-20
[ModifySelection] - Do not fire event when initializing control.

ivandrofly:


src/Forms/ModifySelection.Designer.cs
```diff
@@ -133,7 +133,6 @@ private void InitializeComponent()
             this.checkBoxCaseSensitive.TabIndex = 2;
             this.checkBoxCaseSensitive.Text = "Case sensitive";
             this.checkBoxCaseSensitive.UseVisualStyleBackColor = true;
-            this.checkBoxCaseSensitive.CheckedChanged += new System.EventHandler(this.checkBoxCaseSensitive_CheckedChanged);
```
ivandrofly: Will cause <<code below>> to fire up.
`         private void checkBoxCaseSensitive_CheckedChanged(object sender, EventArgs e)
         {
             Preview();
         }
`

niksedk:

PR #1762, opened 2016-05-20
Refact Text.Encoding and encodingComboBox related stuff

xylographe:
- Replaces replicated `encodingComboBox` code with shared UiUtil methods
- Sets `DropDownWith` appropriately while initialising the `encodingComboBox`
- Sets `Settings.General.DefaultEncoding` consistently to platform and locale independent IANA registered encoding name

### DIFF
--- a/libse/Configuration.cs
+++ b/libse/Configuration.cs
@@ -1,5 +1,7 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Text;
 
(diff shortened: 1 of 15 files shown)


niksedk:
Nice :)
( The first thing I tested was utf32 encoding... and that did not work but it was an old bug which should be fixed with https://github.com/SubtitleEdit/subtitleedit/commit/78a7154945414a48e1393d16b6664db701057f13 )

PR #1765, opened 2016-05-22
Statistics: Bugfix/Update.

ivandrofly:
Also do not include _HTML_ tags when displaying most used lines.


src/Forms/Statistics.cs
```diff
@@ -179,43 +180,52 @@ private static void MostUsedWordsAdd(Dictionary<string, int> hashtable, string t
                 text = text.Remove(idx, endIdx - idx);
                 idx = text.IndexOf("<font", idx, StringComparison.OrdinalIgnoreCase);
             }
-            if (!error)
+            if (!error && idx >= 0)
```
ivandrofly: Only try to replace "\</font> if text contains "\<font

niksedk:

PR #1769, opened 2016-05-28
[Main] - Minor refact.

ivandrofly:


src/Forms/Main.cs
```diff
@@ -3203,8 +3203,11 @@ private DialogResult SaveSubtitle(SubtitleFormat format)
                     foreach (var p in _subtitleAlternate.Paragraphs)
                         s.Paragraphs.Add(p);
                     allText = s.ToText(format);
+                } // Seungki end
+                else
+                {
+                    allText = sub.ToText(format);
```
ivandrofly: SubtitleFormat.ToText() will be called only once in all cases

niksedk:

PR #1776, opened 2016-06-03
[FixLongDisplayTime] - Refact.

ivandrofly:
Some refact for FixLongDisplayTime /cc @niksedk


niksedk:

PR #1788, opened 2016-06-13
Fixed and improved build process

xylographe:
- Replace unsuitable PreBuildEvents with "BeforeCompile" Targets
- Replace PostBuildEvents with more potent "AfterBuild" Targets
- Fix issue with missing zlib.net.dll while running Unit Tests

While trying to figure out how to integrate ILRepack in the (Release) build, I ran MSBuild without the `/m` switch. To my surprise the PreBuildEvent (`build_helpers.bat`) was run **before** the UpdateLanguageFiles project was build. After re-examining [MSDN](https://msdn.microsoft.com/en-us/library/dd293582.aspx) it became clear that the PreBuildEvent is unsuitable for this task, hence I replaced them with *BeforeCompile* Targets. Now the build is processed as intended.


xylographe:
AppVeyor (which also [runs MSBuild without parallelism](https://ci.appveyor.com/project/SubtitleEdit/subtitleedit/build/4793#L7)) can finally build SE without the usual [**"Compile Subtitle Edit first!"** message](https://ci.appveyor.com/project/SubtitleEdit/subtitleedit/build/4722#L15) from `build_helpers.bat`! :)

niksedk:
I get this error now (the branch worked fine):

Unable to copy file "packages\zlib.net.1.0.4.0\lib\zlib.net.dll" to "bin\Debug\zlib.net.dll". The process cannot access the file 'bin\Debug\zlib.net.dll' because it is being used by another process.	SubtitleEdit	C:\Data\SubtitleEdit\subtitleedit\src\SubtitleEdit.csproj	1669	

Any hints?

xylographe:
VS (vshost.exe) keeps zlib.net.dll open (loaded) after the first Build/Rebuild.
But the "unable to copy file" should not happen, because of the `SkipUnchangedFiles="True"` attribute. When I tested in VS this went fine. Only when upgrading the zlib.net package (which will probably never happen) there will be a problem.

Not sure why it doesn't work for you. I'll look into it.

xylographe:
The Copy task uses only file size and modification time to decide if a file is unchanged. So, if NuGet re-installs the zlib package, the Copy task thinks the assembly has changed.

Preliminary fix in [csproj branch](https://github.com/SubtitleEdit/subtitleedit/tree/csproj)

xylographe:
@niksedk  Please, try the ['csproj'](https://github.com/SubtitleEdit/subtitleedit/tree/csproj) branch.

Apparently, the real culprit was `<Private>False</Private>` in the zlib.net Reference item.

niksedk:
@xylographe: I've built SE from csproj multiple time without problems :)

PR #1796, opened 2016-06-15
[Main] - Only add file to recent-files if ShowRecentFiles is enabled.

ivandrofly:


xylographe:
This is slightly confusing. In Language.cs : `RememberRecentFiles = "Show recent files (for reopen)"`.
This implies that the recent files will be remembered, but not shown (it doesn't say "Remember recent files"). However, most translations have incorrectly translated the tag name instead of the content.

ivandrofly:
Yeah, there is something else if you uncheck `Show recents files (for reopen)` from options menu it will clear all the recent files. I think the intention is to only 'not showing the **Reopen** from File\.

ivandrofly:
@niksedk what is yours suggestion about this, should we always store recent files even when the options from Option/Menu is unchecked? :confused:

niksedk:
This was meant to always store recent files, so yeah, the translation tag is badly named...

This reason for not storing the files would be... incognito mode or what?

ivandrofly:
Y-yeah, what I'm trying to say is if we decide to always store recent files we should change the behavior in `Options/Settings/Show recent files (for reopen)` because every time we un-check it and click OK it SE clears recent files list.

ivandrofly:
> This was meant to always store recent files,

This patch can be merged then it will prevent adding recent list to list on form-closing if `Configuration.Settings.General.ShowRecentFiles` if `false`

PR #1797, opened 2016-06-16

diomed:


Dictionaries/hrv_OCRFixReplaceList.xml
```diff
@@ -1605,7 +1612,7 @@
     <!-- mijenja u korist češće riječi -->
     <RegEx find="([vV])rača" replaceWith="$1raća" />
     <RegEx find="zleči" replaceWith="zliječi" />
-    <RegEx find="(?&lt;!obra)zovati" replaceWith="zirati" />
+    <RegEx find="(?&lt;!([Oo])bra)zova([lt])" replaceWith="zira$2" />
```
aaaxx: No need for the brackets around "[Oo]": `<RegEx find="(?&lt;![Oo]bra)zova([lt])" replaceWith="zira$1" />`

aaaxx: Maybe also "n" and "o", `zova([lnot])`, for e.g. "organizovan" and "organizovao".

diomed: okay. I'll add this in next pull request, with some other fixes, since this one is merged.
I don't mind you changing anything else, if you think it could be better, after all, that's why we're here.

PR #1799, opened 2016-06-17
Integrate ILRepack into build process

xylographe:


niksedk:
Any reason for this? Running SE from VS will be a bit slower, right?

xylographe:
This does not affect Debug configuration build (Target Condition is "Configuration==Release").

Building Release configuration will, of course, be slower. The presumption is that users who build the most recent SE Release configuration, will want to replace their current SE assembly, and therefore need a repacked assembly anyway. For those who use VS to build SE it is more convenient if ILRepack is run automatically as part of the build (i/o having to open a shell and run it manually).

But, if this PR conflicts with your general work flow, or the way you prepare official SE releases, let it be. Personally, I don't need this, as I'm using a script that takes care of everything from updating my local SE Git repository up to and including installation of the new assemblies/translations/etc. :)

PR #1800, opened 2016-06-17
[OCR] - Minor optimizations.

ivandrofly:
http://stackoverflow.com/questions/6709022/is-it-more-or-less-efficent-to-perform-a-check-before-performing-a-replace-in-c

### DIFF
--- a/libse/Dictionaries/OcrFixReplaceList.cs
+++ b/libse/Dictionaries/OcrFixReplaceList.cs
@@ -182,27 +182,21 @@ namespace Nikse.SubtitleEdit.Core.Dictionaries
                 {
                     if (s.Contains(from))
                     {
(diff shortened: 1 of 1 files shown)


niksedk:
Almost good... do keep the the old line 316-320 or SE will freeze ;) 
Did you try it on a real subtitle?

ivandrofly:
> Did you try it on a real subtitle?

I did and it didn't produce any problem... can you please link the sub you tried with? Maybe it's another bug.

niksedk:
Okay it's not freezing... but it's way slower. Here's the result of my performance test:
AVERAGE NEW: 10084 milliseconds
AVERAGE OLD: 43 milliseconds

I'll merge and fix it...

PR #1804, opened 2016-06-19
ExportCustomTextFormat enhancement - thx aaaxx :)

### LINKED ISSUE #1802: Newline in “Export custom text format...”
Is there some other way to specify a newline  apart from leaving the box set to `[Do not modify]`? The problem with it is that it turns to literally "[Do not modify]" if you include anything else around it.

I'm trying to make a custom HTML format  and would like to set it to `<br/>[newline][tab]`, so that it displays like this:

```html
<p>
	First subtitle line,<br/>
	second subtitle line.
</p>
```

xylographe:
Accept placeholders "{tab}", "{nl}", "{lf}" and "{cr}" in newline template.

Closes #1802

### DIFF
--- a/src/Forms/ExportCustomTextFormat.cs
+++ b/src/Forms/ExportCustomTextFormat.cs
@@ -8,9 +8,8 @@ namespace Nikse.SubtitleEdit.Forms
 {
     public partial class ExportCustomTextFormat : Form
     {
-
+        public const string EnglishDoNotModify = "[Do not modify]";
         public string FormatOk { get; set; }
-        public const string EnglishDoNoModify = "[Do not modify]";
 
         public ExportCustomTextFormat(string format)
         {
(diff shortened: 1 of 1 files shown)


src/Forms/ExportCustomTextFormat.cs
```diff
@@ -20,6 +19,8 @@ public ExportCustomTextFormat(string format)
             comboBoxNewLine.Items.Add(l.DoNotModify);
             comboBoxNewLine.Items.Add("||");
             comboBoxNewLine.Items.Add(" ");
+            comboBoxNewLine.Items.Add("{nl}{tab}");
+            comboBoxNewLine.Items.Add("{cr}{lf}");
```
aaaxx: Hmm, why preset `{nl}{tab}` when it's useful only for my particular issue? Wouldn't it be better simply to list all the supported templates, just so that people are aware of them and can mix and match on their own? Like:

- `[Do not modify]`
- `||`
- ` `
- `{nl}`
- `{lf}`
- `{cr}{lf}` (this one's a combo to make it less likely for someone to use only `\r`)
- `{tab}`

And it might prevent some issues being opened in the future asking what `{nl}` is if you explicitly use `{environment newline}` instead.

xylographe: The combobox items only serve as a reminder of the existence of these placeholders. I thought it might be better than showing a tooltip. I started with putting all 4 in one item (`{tab}{nl}{lf}{cr}`), but thought dividing them over two items looks slightly better. The `{cr}{lf}` is sort of a 'natural' combination, which left me with either `{tab}{nl}` or `{nl}{tab}` in the other item, which is not a particularly difficult choice. :)  I never considered adding a separate item for each placeholder.

The combobox text is editable, any combination of placeholders and literal characters is possible. Only users can decide which combination makes sense in their context (including a single `{cr}`).

Not many people are familiar with .Net. I doubt very much that  `{environment.newline}` will be clearer than `{nl}`. But, perhaps, `{newline}` might be considered preferable to `{nl}`.

aaaxx:
:+1:

PR #1806, opened 2016-06-20
Added volatile bookmarking to main subtitle list view - thx diomed :)

xylographe:


niksedk:
@diomed / @xylographe: How volatile is the bookmarks meant to be? They disappears after pressing "OK" in settings and several other menus...

xylographe:
Yes, only the SubtitleListView *items* are marked. Hence `SubtitleListView1.Items.Clear()` will remove all marks.

Do you think it would be better to mark paragraphs i/o items?

niksedk:
Yes, I think it would be better to save bookmarks in a dictionary or in the paragraphs or something else that can survive a `SubtitleListView1.Items.Clear()` - any good ideas?

xylographe:
Added a (preliminary\[1]) commit to remember bookmarked paragraphs. However, there are many occasions where the list of paragraphs is re-generated. For example, undoing the deletion of a paragraph changes all paragraph IDs.

\[1] Obviously, the `SubtitleListView._bookmarks` collection should be cleared when the subtitle paragraph list is cleared, or at least in `ResetHistory()`.

xylographe:
I'm considering modifying the Subtitle class, (concept only, not all members will be useful)
```
public event EventHandler ClearingParagraphs;

public IEnumerable<Paragraph> Paragraphs { get { return _paragraphs.AsEnumerable<Paragraph>(); } }
public int ParagraphCount { get { return _paragraphs.Count; } }
public void ClearParagraphs() { /* triggers ClearingParagraphs before _paragraphs.Clear() */  }
public void Add(Paragraph paragraph) { ... }
public void Add(IEnumerable<Paragraph> range) { ... }
public void Insert(int index, Paragraph paragraph) { ... }
public void Insert(int index, IEnumerable<Paragraph> range) { ... }
public void Remove(string paragraphID) { ... }
public void Remove(Paragraph paragraph) { ... }
public void Remove(IEnumerable<Paragraph> range) { ... }
public void RemoveAt(int index) { ... }
public void RemoveAt(int index, int length) { ... }
```
This would hide the Paragraphs implementation from Subtitle consumers, and allow additional events if required.

As it won't be a "minor refact" ;) prior consultation seems in order.
Any thoughts?

niksedk:
I think it would be nice to modify the Paragraphs list to contain events (should also be readonly then) - perhaps something like:
`    public class ObservableList<T> : List<T>
    {

        public event EventHandler OnClear;

        public new void Clear()
        {
            if (null != OnClear)
            {
                OnClear(this, null);
            }
            base.Clear();
        }

    }
`

Or?

PR #1811, opened 2016-06-25
Added support for Czech and Slovak languages in LanguageAutoDetect

mm6502:
Added support for Czech and Slovak languages in LanguageAutoDetect.


niksedk:
thx :)

PR #1813, opened 2016-06-28
[VideoPlayerContainer] - Thread-safe invoke.

ivandrofly:


niksedk:
I had not thought about thread-safety - less code and better - cool :)

PR #1821, opened 2016-06-30
[SettingsMpv] - Bugfix - Prevent main-thread to change cursor to default style.

ivandrofly:


src/Forms/SettingsMpv.cs
```diff
@@ -75,7 +75,6 @@ private void buttonDownload_Click_1(object sender, EventArgs e)
                     labelPleaseWait.Text = Configuration.Settings.Language.General.PleaseWait + "  " + args.ProgressPercentage + "%";
                 };
                 wc.DownloadDataAsync(new Uri(url));
```
ivandrofly: Download is going to be started asynchronously so the main thread will move to the next line and restore the cursor to its default style. TESTED

PR #1828, opened 2016-07-04
[ImportText ] - Update/Refact

ivandrofly:


src/Forms/ImportText.cs
```diff
@@ -260,8 +254,7 @@ private void MakePseudoStartTime()
             {
                 p.EndTime.TotalMilliseconds = millisecondsIndex + p.Duration.TotalMilliseconds;
                 p.StartTime.TotalMilliseconds = millisecondsIndex;
-
-                millisecondsIndex += p.Duration.TotalMilliseconds + millisecondsInterval;
```
ivandrofly: Referencing to p.Duration will create new instance of TimeCode.

PR #1839, opened 2016-07-10
[FSWULAPIP] - Allow fixing one character after the punctuation.

ivandrofly:
**All right, Mitchell, be cool. r.**
After calling `StrippableText` there will be no dot (.) after lowercase **r** which means **start + 3** will be false.


niksedk:

PR #1841, opened 2016-07-10
[FCE] - Remove duplicated event signing.

ivandrofly:


src/Forms/FixCommonErrors.cs
```diff
@@ -1031,10 +1031,7 @@ private void SubtitleListView1SelectedIndexChanged(object sender, EventArgs e)
                 Paragraph p = GetParagraphOrDefault(firstSelectedIndex);
                 if (p != null)
                 {
-                    textBoxListViewText.TextChanged -= TextBoxListViewTextTextChanged;
                     InitializeListViewEditBox(p);
-                    textBoxListViewText.TextChanged += TextBoxListViewTextTextChanged;
-
```
ivandrofly: Both are handled inside **InitializeListViewEditBox()**

PR #1847, opened 2016-07-15
[Subrip] - Improve reading text.

ivandrofly:
WIP

### DIFF
--- a/libse/SubtitleFormats/SubRip.cs
+++ b/libse/SubtitleFormats/SubRip.cs
@@ -167,9 +167,9 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
                     }
                     break;
                 case ExpectingLine.Text:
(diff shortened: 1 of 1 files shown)


libse/SubtitleFormats/SubRip.cs
```diff
@@ -167,34 +167,25 @@ private void ReadLine(Subtitle subtitle, string line, string next, string nextNe
                     }
                     break;
                 case ExpectingLine.Text:
-                    if (!string.IsNullOrWhiteSpace(line) || IsText(next))
+                    if (!string.IsNullOrWhiteSpace(line))
                     {
-                        if (_isWsrt && !string.IsNullOrEmpty(line))
+                        if (_isWsrt && line.Contains('<'))
```
ivandrofly: Do not run the loop if line doesn't contain '<'.

PR #1860, opened 2016-07-24
[Main] - Minor refact.

ivandrofly:


src/Forms/Main.cs
```diff
@@ -2598,13 +2598,13 @@ private void OpenSubtitle(string fileName, Encoding encoding, string videoFileNa
 
                 _fileDateTime = File.GetLastWriteTime(fileName);
 
-                if (format != null && format.IsFrameBased)
-                    _subtitle.CalculateTimeCodesFromFrameNumbers(CurrentFrameRate);
-                else
-                    _subtitle.CalculateFrameNumbersFromTimeCodes(CurrentFrameRate);
```
ivandrofly: If `format == null` will return false without changing anything

PR #1861, opened 2016-07-24
[Icons] - Do not 'Always copy' to build folder.

ivandrofly:


xylographe:
Causes Inno setup compile error,
```
Error on line 223 in installer\Subtitle_Edit_installer.iss: Source file "installer\..\src\bin\Release\Icons\Find.png" does not exist.
Compile aborted.
```

ivandrofly:
Try delete from line 223-235 and see if it still causes compile error. (I don't have Inno installed) /cc @xylographe

niksedk:
Installer works fine now :)

PR #1865, opened 2016-07-26

diomed:


Dictionaries/hrv_OCRFixReplaceList.xml
```diff
@@ -1518,6 +1524,7 @@
     <RegEx find="dranj" replaceWith="deranj" />
     <RegEx find="dsek" replaceWith="dsjek" />
     <RegEx find="[dt]cj?eni" replaceWith="dcijeni" />
+    <RegEx find="cj?enim" replaceWith="cijenim" />
```
xylographe: I don't know if `tcjenim` and/or `tcenim` can be (part of) a Croatian word. But, if they can, the result after conversion is unpredictable, because the `<RegEx />` conversions are not necessarily applied in the order of definition.

`tcjenim` (and also `tcenim`) could become either `dcijenim` (line 1526) or `tcijenim` (line 1527).

diomed: yes they can. so, should I put them like this (?&lt;![dt])cjenim  ???

xylographe: Yes, assuming `tcjenim` ==> `dcijenim` is the correct conversion, line 1527 should be,
```xml
    <RegEx find="(?&lt;![dt])cj?enim" replaceWith="cijenim" />
```

PR #1868, opened 2016-07-27
[UnknownSubtitle7] - Fix/Update.

ivandrofly:
- Remove unnecessary regex.
- Make compiled regex static & readonly.
- Delete redundant method to 'encode' time code (when serializing to text).
- Simplify 'deserialization' + optimization.


niksedk:
Please do not mix fix and refact...

ivandrofly:
> Please do not mix fix and refact...

But some of them are just refact/optimization like this one.

For the rest I'll try to send 'em in separated PRs.

niksedk:

PR #1870, opened 2016-07-28
[SubtitleFormat] - Rename UnknownSubtitle16 to CavenaCIP.

ivandrofly:
Source: http://screens.tv/wp-content/uploads/2015/06/Subtitle-File-Formats.pdf

I even found some 'new' subtitle format in there :point_up:


libse/SubtitleFormats/SubtitleFormat.cs
```diff
@@ -174,7 +174,7 @@ public static IEnumerable<SubtitleFormat> AllSubtitleFormats
                     new UnknownSubtitle13(),
                     new UnknownSubtitle14(),
                     new UnknownSubtitle15(),
-                    new UnknownSubtitle16(),
+                    new CavenaCIP(),
```
niksedk: This is a sorted list... and perhaps "UnknowSubtitle83" (the last unknown) should be renamed to UnknownSubtitle16 ?

ivandrofly: > This is a sorted list... and perhaps "UnknowSubtitle83" (the last unknown) should be renamed to UnknownSubtitle16 ?

Sounds good

niksedk:

PR #1871, opened 2016-07-28
[SwiftInterchagne2] - Update/Fix.

ivandrofly:
- Use built-in method to get filename without extension.
- Check for if `fileNameNoExt` is null before entering the loop.


niksedk:
Please do not mix fix and refact...

ivandrofly:
Updated /cc @niksedk 

- I'll try to send them in  smaller patch and also try to separate refacts with fixes.

PR #1876, opened 2016-08-01
[SwiftInterchange2] - Remove unneeded loop to load italics.

ivandrofly:


niksedk:
Will miss italic for last line...

ivandrofly:
> Will miss italic for last line...

Oh yeah, isn't the better solution to use for loop instead two foreach?

PR #1883, opened 2016-08-06
[Main] - Only store recent files if ShowRecentFiles == true.

ivandrofly:
Same thing as https://github.com/SubtitleEdit/subtitleedit/commit/5c20c883875814d6ce73c4d4f385c9921173864d


niksedk:
The intention was to always save recent files... e.g. like Visual Studio.

PR #1887, opened 2016-08-08
Descenders in PNGs, simplified line height calculation.

ischeriad:
For an upcoming project with Romanian subtitles, I needed to improve the baseline calculation for PNG creation.

Instead of keeping a list with all possible characters and comparing line height to a line with these characters removed (textNoBelow), I just compare with a simple line "a,".

There is still an occasional difference of 1px in height for generated PNGs, this seems to be a rounding error somewhere. (Changing border width in 1px increments shows the offset on different lines).

### DIFF
--- a/src/Forms/ExportPngXml.cs
+++ b/src/Forms/ExportPngXml.cs
@@ -2317,7 +2317,7 @@ $DROP=[DROPVALUE]" + Environment.NewLine + Environment.NewLine +
                     bmp = new Bitmap(sizeX, sizeY);
                 }
 
-                // align lines with gjpqy, a bit lower
+                // align lines with descenders, a bit lower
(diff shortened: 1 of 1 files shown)


ischeriad:
I just discovered that I have broken some other things with line heigt calculation.

Trying to fix it first.

niksedk:
OK, I've changed the default "baseLinePadding" in https://github.com/SubtitleEdit/subtitleedit/commit/b62b157b254be47a395b8fa27141fa90d1ba3274

I more precise (and without hardcoded letters) baseline calc is very welcome  :)
Do let us know your findings.

aaaxx:
Have you thought about rendering multiline subs as a single line then chopping them into separate lines  and positioning them?

PR #1902, opened 2016-08-14
Add search feature to combobox subtitle format.

ivandrofly:
- Helps finding subtitle formats quicker.


niksedk:

PR #1903, opened 2016-08-14
[Main] - Remove redundant call to calculate (time/frame).

ivandrofly:
`CalculateTimeCodesFromFrameNumbers` and/or `CalculateFrameNumbersFromTimeCodesNoCheck` 
is going to be called inside `ShowSource()`.


ivandrofly:
Did I miss something? o.O /cc @niksedk

PR #1904, opened 2016-08-14
[Main] - Remove useless statement.

ivandrofly:
# Refact only
- Remove fruitless if condition.


src/Forms/Main.cs
```diff
@@ -4407,59 +4407,55 @@ private void ReplaceSourceView(ReplaceDialog replaceDialog)
                     return;
                 }
                 int replaceCount = 0;
-                bool searchStringFound = true;
-                if (searchStringFound)
```
ivandrofly: redundant

niksedk:

PR #1914, opened 2016-08-21
[Compare] - Minor refact.

ivandrofly:


src/Forms/Compare.cs
```diff
@@ -415,7 +413,7 @@ private void CompareSubtitles()
                 subtitleListView2.BeginUpdate();
                 if (max != _differences.Count)
                 {
-                    for (index = Math.Max(subtitleListView1.Items.Count, subtitleListView2.Items.Count); index >= 0; index--)
+                    for (index = max - 1; index >= 0; index--)
```
ivandrofly: Minor fix here.

PR #1916, opened 2016-08-23
[Compare] - Assign verification method directly to both LVs.

ivandrofly:


src/Forms/Compare.cs
```diff
@@ -840,22 +840,9 @@ private void ShowTip(Control control)
...
-            VerifyDragEnter(e);
-        }
-
-        private static void VerifyDragEnter(DragEventArgs e)
-        {
-            if (e.Data.GetDataPresent(DataFormats.FileDrop))
-                e.Effect = DragDropEffects.Copy;
-            else
-                e.Effect = DragDropEffects.None;
+            e.Effect = e.Data.GetDataPresent(DataFormats.FileDrop) ? DragDropEffects.Copy : e.Effect = DragDropEffects.None;
```
xylographe: Minor refact (the second assignment is a little confusing):
```
e.Effect = e.Data.GetDataPresent(DataFormats.FileDrop) ? DragDropEffects.Copy : DragDropEffects.None;
```

However, the original with if/else might be easier to read (understand) ?

ivandrofly: > However, the original with if/else might be easier to read (understand) ?

No problem can be changed to standard if/else style. :)

PR #1917, opened 2016-08-23
[Compare] - Minor fix sse/ße.

ivandrofly:
To produce:

Enter `sse` and replace it with `ße` open compare and you will see that entire text from textbox2 gets colored red.

Since last chars are equal they should be keep black.

compare

### DIFF
--- a/src/Forms/Compare.cs
+++ b/src/Forms/Compare.cs
@@ -588,7 +588,7 @@ namespace Nikse.SubtitleEdit.Forms
             }
 
             int maxLength = Math.Max(richTextBox1.Text.Length, richTextBox2.Text.Length);
-            for (int i = startCharactersOk; i <= maxLength; i++)
+            for (int i = startCharactersOk; i < maxLength; i++)
             {
                 if (i < richTextBox1.Text.Length)
                 {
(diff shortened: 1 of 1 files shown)


src/Forms/Compare.cs
```diff
@@ -641,7 +641,7 @@ private void ShowTextDifference()
             }
             else if (richTextBox2.Text.Length > richTextBox1.Text.Length)
             {
-                if (richTextBox2.Text.StartsWith(richTextBox1.Text))
+                if (richTextBox2.Text.StartsWith(richTextBox1.Text, StringComparison.Ordinal))
```
ivandrofly: Llast 'e' don't be colored red from textbox2

src/Forms/Compare.cs
```diff
@@ -588,7 +588,7 @@ private void ShowTextDifference()
             }
 
             int maxLength = Math.Max(richTextBox1.Text.Length, richTextBox2.Text.Length);
-            for (int i = startCharactersOk; i <= maxLength; i++)
```
ivandrofly: We are working with indices. :P

PR #1924, opened 2016-08-25
[DurationsBridgeGaps] - Update/Refact.

ivandrofly:


src/Forms/DurationsBridgeGaps.cs
```diff
@@ -150,7 +146,7 @@ private void GeneratePreviewReal()
             }
 
             foreach (var index in fixedIndexes)
-                SubtitleListview1.SetBackgroundColor(index, Color.Green);
+                SubtitleListview1.SetBackgroundColor(index, Color.LightGreen);
```
ivandrofly: More visible than `Green`.

PR #1925, opened 2016-08-25
[Compare] - Move syntaxcoloring to constructor.

ivandrofly:
Move `UseSyntaxColoring = false` to constructor so both Initialize method can rely on it.


src/Forms/Compare.cs
```diff
@@ -41,13 +41,12 @@ public Compare()
...
             UiUtil.InitializeSubtitleFont(subtitleListView2);
             UiUtil.FixLargeFonts(this, buttonOK);
+            subtitleListView1.UseSyntaxColoring = false;
+            subtitleListView2.UseSyntaxColoring = false;
         }
 
         public void Initialize(Subtitle subtitle1, string subtitleFileName1, string title)
         {
-            subtitleListView1.UseSyntaxColoring = false;
-            subtitleListView2.UseSyntaxColoring = false;
```
ivandrofly: **Should've been declared in both Initialize methods....**

PR #1935, opened 2016-08-31
[StringExtensions] - Minor optimization for ContainsLetter().

ivandrofly:
This is how `String.IsNullOrWhiteSpace` looks like:
```C#
public static bool IsNullOrWhiteSpace(string value)
{
    if (value != null)
    {
        for (int i = 0; i < value.Length; i++)
        {
            if (!char.IsWhiteSpace(value[i]))
            {
                return false;
            }
        }
    }
    return true;
}
```

We were running: `2n`
Optimized to: `n`

### DIFF
--- a/libse/StringExtensions.cs
+++ b/libse/StringExtensions.cs
@@ -153,13 +153,14 @@ namespace Nikse.SubtitleEdit.Core
 
         public static bool ContainsLetter(this string s)
         {
-            if (string.IsNullOrWhiteSpace(s))
-                return false;
-
-            foreach (var c in s)
+            if (s != null)
             {
-                if (char.IsLetter(c))
-                    return true;
+                foreach (var index in StringInfo.ParseCombiningCharacters(s))
+                {
+                    var uc = CharUnicodeInfo.GetUnicodeCategory(s, index);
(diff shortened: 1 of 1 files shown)


libse/StringExtensions.cs
```diff
@@ -153,13 +153,13 @@ public static string FixExtraSpaces(this string s)
...
-            foreach (var c in s)
+            if (s != null)
             {
-                if (char.IsLetter(c))
-                    return true;
+                foreach (var c in s)
+                {
+                    if (char.IsLetter(c))
+                        return true;
+                }
```
xylographe: Perhaps, to provide real Unicode support (including Asian languages),
```C#
if (s != null)
{
    foreach (var index in StringInfo.ParseCombiningCharacters(s))
    {
        var uc = CharUnicodeInfo.GetUnicodeCategory(s, index);
        if (uc == UnicodeCategory.LowercaseLetter || uc == UnicodeCategory.UppercaseLetter || uc == UnicodeCategory.TitlecaseLetter || uc == UnicodeCategory.ModifierLetter || uc == UnicodeCategory.OtherLetter)
            return true;
    }
    return false;
}
```

ivandrofly:
@niksedk should we use xylographes?

niksedk:
Yes, seems to be more correct :)

ivandrofly:
@niksedk 

Updated :)

niksedk:
I normally prefer to have a method return in the beginning if something is null - rather than to have code nested inside an if-statement, but it's no biggie (and I've not tested performance either)

PR #1936, opened 2016-08-31
[Program] - Only catch Application/AppDomain execptions if not in DEBUG Config-mode.

ivandrofly:
Will help to quickly find the line when the exception has been thrown while in Visual-Studio


niksedk:
Nice, why didn't I think of that...

ivandrofly:
> Nice, why didn't I think of that...

:P

PR #1940, opened 2016-09-04
[Compare] - Minor refact.

ivandrofly:


niksedk:
Nice, could you remove the comment though (not really any need for it)

ivandrofly:
Updated
@niksedk

PR #1941, opened 2016-09-05
[Main/Plugin] - Use EnumerateFiles for more efficiency.

ivandrofly:
When working with many files **EnumerateFiles** can be more efficient than **GetFiles()**. (Stated in [MSDN](https://msdn.microsoft.com/en-us/library/dd383458(v=vs.110).aspx))

### DIFF
--- a/src/Forms/Main.cs
+++ b/src/Forms/Main.cs
@@ -15062,7 +15062,6 @@ namespace Nikse.SubtitleEdit.Forms
             var path = Configuration.PluginsDirectory;
             if (!Directory.Exists(path))
                 return;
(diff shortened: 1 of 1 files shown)


niksedk:
Did not know about this... interesting. Just tried some tests with this and loading the plugins is slightly slower with this method, but I guess it might work better in something like "Batch convert"

ivandrofly:
> Did not know about this... interesting. Just tried some tests with this and loading the plugins is slightly slower with this method, but I guess it might work better in something like "Batch convert"

I think the performance is only noticed with many.

PR #1948, opened 2016-09-10
[FixMissingSpaces] - Add space if tag was found after 1st char.

ivandrofly:
For example if you try to fix: `I<i>will</i>` **SE** won't add space after **I**.


niksedk:

PR #1950, opened 2016-09-11
Minor fixes

ivandrofly:


src/Forms/FixCommonErrors.cs
```diff
@@ -565,7 +565,7 @@ public void FixOcrErrorsViaReplaceList(string threeLetterIsoLanguageName)
                     var p = Subtitle.Paragraphs[i];
                     string text = ocrFixEngine.FixOcrErrors(p.Text, i, lastLine, false, OcrFixEngine.AutoGuessLevel.Cautious);
                     lastLine = text;
-                    if (AllowFix(p, fixAction) && p.Text != text)
+                    if (AllowFix(p, fixAction) && p.Text.Trim() != text.Trim())
```
ivandrofly: Because we called text.Trim() inside `ocrFixEngine.FixOcrErrors(...)` even if there are no fixes made **p.Text** won't be equal to `text`.

PR #1952, opened 2016-09-12
[SEListView] - Closes #1408

### LINKED ISSUE #1408: inserted first line?
why this program doesn't start up with already open first line, and ready for editing, but we have to insert it all the time we open it? is there any good reason for that?

ivandrofly:
REBASED


niksedk:

PR #1956, opened 2016-09-14
[FCE]- Check for Irish name. closes #1955. [WIP]

### LINKED ISSUE #1955: “Fix uppercase ‘i’ inside lowercase words” needs some exceptions
Irish name prefixes:
- Mc
- Mac
- O'

For example, I've just run into "McInturff".

There might be similar cases in other languages, but I can't think of anything right now.

ivandrofly:


niksedk:
Could not find the unit test!? ;)

ivandrofly:
> Could not find the unit test!? ;)

Oh sure. 😄

ivandrofly:
Updated. @niksedk  :)

libse/Forms/FixCommonErrors/FixUppercaseIInsideWords.cs
```diff
@@ -17,25 +17,25 @@ public void Fix(Subtitle subtitle, IFixCallbacks callbacks)
...
-
+                var st = new StripableText(p.Text);
                 Match match = ReAfterLowercaseLetter.Match(p.Text);
                 while (match.Success)
                 {
-                    if (!(match.Index > 1 && p.Text.Substring(match.Index - 1, 2) == "Mc") // irish names, McDonalds etc.
-                        && p.Text[match.Index + 1] == 'I'
-                        && callbacks.AllowFix(p, fixAction))
+                    string word = GetWholeWord(st.StrippedText, match.Index);
+                    if (!callbacks.IsName(word) && callbacks.AllowFix(p, fixAction))
```
ivandrofly: Now will goes through callbacks to check if word is name too.

libse/Forms/FixCommonErrors/EmptyFixCallback.cs
```diff
@@ -68,4 +69,19 @@ public HashSet<string> GetAbbreviations()
         }
 
     }
+
+    public class CustomFixCallback : EmptyFixCallback
```
ivandrofly: Add new new fix-callbacks which extends **EmptyFixllCallback** because we won't be able to check if word is name if we use **EmptyFixCallback** 'cause it always return **false** if we call **IsName(candidate)**

aaaxx:
Why do you check if the language is English? The names could be present in translations too.

aaaxx:
@ivandrofly, one more thing: `"O'I"` should also work with curly apostrophe (`’`).

aaaxx:
I've just run into this case again. What's the status here?

ivandrofly:
> I've just run into this case again. What's the status here?

the branch seems to have merge conflicts now... will rebase and add the remaining part soon :)

niksedk:

PR #1958, opened 2016-09-16
[Ebu] - Remove redundant cases in switch.

ivandrofly:
Simplify color switch.


niksedk:

PR #1961, opened 2016-09-19
Updated OcrFixReplaceList.cs

aaaxx:
Added other common Latin ligatures present in Unicode.

Also added the acute accent, which I've often seen used instead of the
apostrophe, either as an OCR error or because people mistake it for the
curly apostrophe.


niksedk:
@aaaxx: could you resolve the conflicts?

aaaxx:
Thanks, xylographe.

PR #1963, opened 2016-09-19
[LICENSE] - Rename gpl => LICENSE.

ivandrofly:
Will add new column **MIT** https://help.github.com/articles/adding-a-license-to-a-repository/


xylographe:
```
PS> git grep -lF gpl.txt
build.bat
installer/Subtitle_Edit_installer.iss
src/SubtitleEdit.csproj
```

src/SubtitleEdit.csproj
```diff
@@ -1355,6 +1355,9 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="Resources\Transparent Background.png" />
+    <Content Include="..\LICENSE.txt">
+      <Link>LICENSE.txt</Link>
```
niksedk: Is the "Link" really needed?

ivandrofly: > Is the "Link" really needed?

Yeah, I think so. It was added by VS2015 automatically.
Maybe it's a new tag in VS2015

ivandrofly: Or maybe exclude LICENSE.txt from VS since it's not going to be *changed?

xylographe: IIRC the `<Link>` item metadata is used when the file is located outside the project directory. But it is not really necessary here. I would suggest item `<Content Include="..\LICENSE.txt" />` without metadata (i.e., replace `..\gpl.txt` with `..\LICENSE.txt`).

ivandrofly: Updated. @niksedk

PR #1964, opened 2016-09-21
[PhoenixSubtitle] - Add new subtitle format.

ivandrofly:
Implements http://devel.aegisub.org/wiki/SubtitleFormats/PJS

- [x] Ready.

### DIFF
--- a/libse/SubtitleFormats/PhoenixSubtitle.cs
+++ b/libse/SubtitleFormats/PhoenixSubtitle.cs
@@ -6,14 +6,16 @@ using System.Text.RegularExpressions;
 
 namespace Nikse.SubtitleEdit.Core.SubtitleFormats
 {
-    // Specs: http://devel.aegisub.org/wiki/SubtitleFormats/PJS
+    /// <summary>
+    /// Implements https://web.archive.org/web/20080213213927/http://trac.annodex.net/wiki/PJS
+    /// https://github.com/SubtitleEdit/subtitleedit/pull/1964#issuecomment-249379407
+    /// </summary>
     public class PhoenixSubtitle : SubtitleFormat
     {
         //2447,   2513, "You should come to the Drama Club, too."
         //2513,   2594, "Yeah. The Drama Club is worried|that you haven't been coming."
         //2603,   2675, "I see. Sorry, I'll drop by next time."
 
-        // TODO: Build optimized regex to match time-codes with non-whites spaces.
         private static readonly Regex RegexTimeCodes = new Regex(@"^(\d+),\s+(\d+),", RegexOptions.Compiled);
         private static readonly char[] TrimChars = { ' ', '"' };
 
@@ -104,18 +106,12 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
 
         private static TimeCode DecodeTimeCode(string encodedTime)
         {
-            // Format is timed in tenths of seconds. 2447
             int time;
             if (int.TryParse(encodedTime, out time))
             {
-                return new TimeCode(time * .1 * TimeCode.BaseUnit);
+                return new TimeCode(FramesToMilliseconds(time));
             }
             return new TimeCode(0);
-        }
-
-        private static string EncodeTimeCode(TimeCode tc)
-        {
-            return Convert.ToString((int)Math.Round(tc.TotalMilliseconds / 100.0));
         }
 
         public override string ToText(Subtitle subtitle, string title)
@@ -124,10 +120,11 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
             var sb = new StringBuilder();
             foreach (var p in subtitle.Paragraphs)
             {
-                string text = HtmlUtil.RemoveHtmlTags(p.Text, true).Trim(TrimChars);
+                string text = HtmlUtil.RemoveHtmlTags(p.Text, true);
                 // Pipe character for forced line breaks.
                 text = text.Replace(Environment.NewLine, "|");
-                sb.AppendFormat(writeFormat, EncodeTimeCode(p.StartTime), EncodeTimeCode(p.EndTime), text);
+                sb.AppendFormat(writeFormat, MillisecondsToFrames(p.StartTime.TotalMilliseconds),
+                    MillisecondsToFrames(p.EndTime.TotalMilliseconds), text);
             }
             return sb.ToString();
         }
--- a/libse/SubtitleFormats/PhoenixSubtitle.cs
+++ b/libse/SubtitleFormats/PhoenixSubtitle.cs
@@ -85,6 +85,7 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
                             line = line.Trim(TrimChars);
                         }
 
+                        paragraph.Number = i + 1;
                         paragraph.Text = string.Join(Environment.NewLine, line.Split('|'));
                         subtitle.Paragraphs.Add(paragraph);
                         paragraph = new Paragraph();
@@ -99,7 +100,6 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
                     _errorCount++;
                 }
             }
-            subtitle.Renumber();
         }
 
         private static TimeCode DecodeTimeCode(string encodedTime)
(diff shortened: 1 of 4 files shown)


ivandrofly:
Line counta test fails because **Phoenix Subtitle** doesn't use \r\n as new line separator.

niksedk:
@ivandrofly: the "Line count" unit test is about the number of subtitles (not the number of lines)... has also saved me several times :)

libse/SubtitleFormats/PhoenixSubtitle.cs
```diff
@@ -0,0 +1,123 @@
...
+            return Convert.ToString(tm);
+        }
+
+        public override string ToText(Subtitle subtitle, string title)
+        {
+            const string writeFormat = "{0},   {1},   \"{2}\"\r\n";
+            var sb = new StringBuilder();
+            foreach (var p in subtitle.Paragraphs)
+            {
+                string text = HtmlUtil.RemoveHtmlTags(p.Text, true);
```
niksedk: How should quotes (") inside strings be handled?

ivandrofly: > How should quotes (") inside strings be handled?

Now mentioned in "specs" (http://devel.aegisub.org/wiki/SubtitleFormats/PJS)

ivandrofly:
@niksedk updated.

I pushed all changes in a new commit so the changes can be visible.

aaaxx:
[Here's some conflicting info.](http://web.archive.org/web/20080213213927/http://trac.annodex.net/wiki/PJS) you should probably check how MPC-HC and VLC handle them because they're gonna be the ones to play them, if anyone happens to actually use the format.

ivandrofly:
> Here's some conflicting info. you should probably check how MPC-HC and VLC handle them because they're gonna be the ones to play them, if anyone happens to actually use the format.

I think it's all done now.
PS: MPC-HC doesn't seem to support it :(.

libse/SubtitleFormats/PhoenixSubtitle.cs
```diff
@@ -0,0 +1,137 @@
...
+                return new TimeCode(time * .1 * TimeCode.BaseUnit);
+            }
+            return new TimeCode(0);
+        }
+
+        private static string EncodeTimeCode(TimeCode tc)
+        {
+            int tm = (int)Math.Round(tc.TotalSeconds);
+            return Convert.ToString(tm * 10);
+        }
```
xylographe: Is rounding to whole seconds intended? Or should it be rounded to tenths of seconds?
```
    return Convert.ToString((int)Math.Round(tc.TotalMilliseconds / 100.0));
```

ivandrofly: It was intended, but your version seems more accurate. 👍 
will be implemented soon.

aaaxx:
Hmm, I tried to test a PJS file in the latest version of VLC and it loads and is shown in the menu, but nothing gets displayed. This is what used:

```text
1, 100,"if timecodes treated as 1/10 seconds, this lasts 10 s; if frames, 4 s"
150, 250,"Line 1|Line 2"
```

Also, I'm not really sure the timecodes should be in tenths of seconds. The [previous link](http://web.archive.org/web/20080213213927/http://trac.annodex.net/wiki/PJS), its [Python script](http://web.archive.org/web/20070830155543/http://svn.annodex.net/scripts/trunk/subtitles/script-phoenix2cmml.py), [Wikipedia](https://en.wikipedia.org/wiki/Subtitle_(captioning)#Subtitle_formats) and [VLC's Wiki](https://wiki.videolan.org/Subtitles/) all say it's frames, and, in addition, VLC's source has a [comment](http://git.videolan.org/?p=vlc.git;a=blob;f=modules/demux/subtitle.c;h=653c52943c160f79994850763796f429387901f0;hb=HEAD#l1546) `/* 1/10th of second ? Frame based ? FIXME */`, so they're not sure themselves. It just seems more likely to me it's supposed to be frames.

I was gonna file a [bug report](https://trac.videolan.org/vlc/) with VLC for not displaying the file, but it might be a good idea that you do it, mentioning that you're implementing the format, and ask them to think about the unit while they're looking into the issue.

xylographe:
@aaaxx It looks like `i_start` and `i_stop` members are expected to contain milliseconds.
Hence, the assignments should probably be,
```
p_subtitle->i_start = 100 * t1;
p_subtitle->i_stop = 100 * t2;
```
Alternatively (only for the purpose of testing), you could try to multiply all timestamps by ten,
```
10,1000,"if timecodes treated as 1/10 seconds, this lasts 10 s; if frames, 4 s"
1500,2500,"Line 1|Line 2"
```

xylographe:
Actually, looking at `subtitle_ParseSubRipTimingValue()`, it would appear `i_start` and `i_stop` are in microseconds?
```
p_subtitle->i_start = (int64_t)t1 * 100000;
p_subtitle->i_stop = (int64_t)t2 * 100000;
```

aaaxx:
Worse: it expects one hundred-thousandth of a second (10⁻⁵). I don't think there's even a named prefix for that one 😆. This is what worked:

```text
0, 100000,"Lasts 1 second"
100000, 1000000,"Lasts from 1st to 10th second of video"
```

So they're definitely handling it wrong. Can one of you coder guys file a bug report (and maybe get them to switch to frames, which I think was the originally used format)? There's a better chance they're gonna fix it if comes from a fellow developer.

xylographe:
I think you are right, the PJS format is frame based. The sentence [ *"Phoenix Japanimation Society (PJS) is a Phoenix, AZ based anime fan community that dates back to the 1980's."* ](https://web.archive.org/web/20080213213927/http://trac.annodex.net/wiki/PJS) is a strong indication that frames are much more likely than tenths of seconds. And also, Gnome Subtitles and FFmpeg use frames.

The fix in VLC would be quite simple: (names changed to be consistent with `ParseMicroDvd()`)

```diff
diff --git a/modules_demux_subtitle.c b/modules_demux_subtitle.c
index 653c529..cd90087 100644
--- a/modules_demux_subtitle.c
+++ b/modules_demux_subtitle.c
@@ -1531,7 +1531,7 @@ static int ParsePJS( demux_t *p_demux, subtitle_t *p_subtitle, int i_idx )
     for( ;; )
     {
         const char *s = TextGetLine( txt );
-        int t1, t2;
+        int i_start, i_stop;

         if( !s )
             return VLC_EGENERIC;
@@ -1541,11 +1541,11 @@ static int ParsePJS( demux_t *p_demux, subtitle_t *p_subtitle, int i_idx )
             return VLC_ENOMEM;

         /* Data Lines */
-        if( sscanf (s, "%d,%d,\"%[^\n\r]", &t1, &t2, psz_text ) == 3 )
+        if( sscanf (s, "%d,%d,\"%[^\n\r]", &i_start, &i_stop, psz_text ) == 3 )
         {
-            /* 1/10th of second ? Frame based ? FIXME */
-            p_subtitle->i_start = 10 * t1;
-            p_subtitle->i_stop = 10 * t2;
+            /* Frame based */
+            p_subtitle->i_start = i_start * p_sys->i_microsecperframe;
+            p_subtitle->i_stop = i_stop * p_sys->i_microsecperframe;
             /* Remove latest " */
```

See full edited file at [Gist](https://gist.github.com/xylographe/49323e40b80c48bff4e520b2fd7ec650)

ivandrofly:
Frame it's then! :)

xylographe:
> Frame it's then! :)

Perhaps wait for Nikolaj's opinion? :)

libse/SubtitleFormats/PhoenixSubtitle.cs
xylographe: From other PJS subtitle code I've seen, quotes inside a subtitle are not treated special (no escapes).

While parsing only the opening and closing quote should be removed. Missing quote at start and/or end of text could be ignored, and counted as an error.

niksedk: :)

niksedk:
Sounds like frames are slightly more likely, so let's go for that :)

Does the code handle texts correct (preserve quotes) that already starts/ends with quotes? (is only the first/last quote removed when loading a subtitle?)

(off topic: sorry, I don't have too much time atm)

ivandrofly:
- [x] - More tests added.
- [x] - Handle quotes inside/out text when saving to **PJS**.
- [x] - Preserve quotes if there are more then one/more quotes following delimiter (").

PR #1970, opened 2016-09-23
Fix issue-1969 cavena890 character mapping

ghoogland:
This is a proposed fix for missing characters in Cavena890 files:
I added mapping for more diacritics/special characters in cavena890.cs, back and forth.
Reverse engineered and tested with FAB.
I only added characters which are practically used in latin languages, according to wikipedia:
https://en.wikipedia.org/wiki/Latin_alphabets
.. although FAB seems to map everything, even characters which are not applied in any language..


niksedk:
thx :)
Works nicely with my Danish file.

ghoogland:
Thanks for the merge! Glad it works for danish :)

PR #1973, opened 2016-09-24
[Main] - Cache subtitle hash.

ivandrofly:


src/Forms/Main.cs
```diff
@@ -1929,13 +1929,12 @@ private void OpenSubtitle(string fileName, Encoding encoding, string videoFileNa
...
-                if (change)
-                    change = _lastDoNotPrompt != _subtitle.GetFastHashCode();
+                string _subtitleHash = _subtitle.GetFastHashCode();
+                bool hasChanged = (_changeSubtitleToString != _subtitleHash) && (_lastDoNotPrompt != _subtitleHash);
 
                 SubtitleFormat format = _subtitle.LoadSubtitle(fileName, out encoding, encoding);
-                if (!change)
-                    _changeSubtitleToString = _subtitle.GetFastHashCode();
+                if (!hasChanged)
+                    _changeSubtitleToString = _subtitleHash;
```
niksedk: Could you not use the cached hash here.? (it should be from the freshly loaded subtitle)

ivandrofly: Fixed :) , actually changed by mistake

PR #1974, opened 2016-09-25
[TarReader] - Fixes for GetTesseractDictionaries/TarReader.

ivandrofly:


libse/TarReader.cs
```diff
@@ -4,10 +4,9 @@
 
 namespace Nikse.SubtitleEdit.Core
 {
-    public class TarReader : IDisposable
+    public class TarReader
     {
         public List<TarHeader> Files { get; private set; }
-        private Stream _stream;
```
ivandrofly: It's unnecessary to use this as a field

niksedk:

PR #1984, opened 2016-09-29
[Utilities] - Change to native array init to avoid issue with unix platforms.

ivandrofly:
The return value of **NewLine** can differ depending on the environment.
See: https://msdn.microsoft.com/en-us/library/system.environment.newline(v=vs.110).aspx


niksedk:
:)

Some of the usages should probably be .SplitToLines...

PR #1992, opened 2016-10-07
[ExportPngXml] - Update/Fix.

ivandrofly:


src/Forms/ExportPngXml.cs
```diff
@@ -3611,6 +3606,7 @@ internal void InitializeFromVobSubOcr(Subtitle subtitle, SubtitleFormat format,
 
         private void subtitleListView1_SelectedIndexChanged(object sender, EventArgs e)
         {
+            if (_isLoading) return;
```
ivandrofly: Will prevent calling this method every time we set combobox to a value when initializing.

src/Forms/ExportPngXml.cs
```diff
@@ -3745,7 +3741,6 @@ private void comboBoxSubtitleFont_SelectedValueChanged(object sender, EventArgs
 
         private void comboBoxSubtitleFontSize_SelectedIndexChanged(object sender, EventArgs e)
         {
-            UpdateLineSpacing();
```
ivandrofly: Not necessary will be called inside `subtitleListView1_SelectedIndexChanged()`

src/Forms/ExportPngXml.cs
```diff
@@ -3236,8 +3233,6 @@ internal void Initialize(Subtitle subtitle, SubtitleFormat format, string export
             SubtitleListView1AutoSizeAllColumns();
 
             _subtitle = new Subtitle(subtitle);
-            _subtitle.Header = subtitle.Header;
-            _subtitle.Footer = subtitle.Footer;
```
ivandrofly: Already copied inside `new Subtitle(subtitle);`

PR #1998, opened 2016-10-10
Update es-MX.xml

paconaranjo:


src/Languages/es-MX.xml
```diff
@@ -1878,6 +1903,10 @@ puedan editar el mismo archivo de subtítulo (colaboración)</Information>
     <MainFileSaveAll>Guardar todo</MainFileSaveAll>
     <Miscellaneous>Otros</Miscellaneous>
     <UseDoNotBreakAfterList>Utilizar «no saltar después de la lista» (para salto automático)</UseDoNotBreakAfterList>
+    <MicrosoftBingTranslator>Traductor Microsoft Bing</MicrosoftBingTranslator>
+    <HowToSignUp>Cómo inscribirse</HowToSignUp>
+    <ClientId>Usuario</ClientId>
```
ivandrofly: Huh, Isn't it: *ID *del Usuario*?

paconaranjo: You are right, "ID del usuario" is correct.

PR #2005, opened 2016-10-13
[Settings] - Prefer using for disposing obj.

ivandrofly:


libse/Settings.cs
```diff
@@ -1263,10 +1263,10 @@ public static Settings GetSettings()
         private static Settings CustomDeserialize(string fileName)
         {
             var doc = new XmlDocument { PreserveWhitespace = true };
-
-            var stream = new FileStream(fileName, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
-            doc.Load(stream);
-            stream.Close();
+            using (var stream = new FileStream(fileName, FileMode.Open, FileAccess.Read, FileShare.ReadWrite))
```
ivandrofly: If somehow an exception occur, the object will still be disposed

PR #2006, opened 2016-10-14
[Main] - Remove duplicated BeginUpdate call.

ivandrofly:
- SubtitleListview.Fill(...) always calls BeginUpdate
- Exit the loop as soon the file is found in **Recent File Entry**. (will save us from 25 loops in case file is in position 1)
- Compare names using **OrdinalIgnoreCase** rule.


niksedk:

PR #2024, opened 2016-10-20
[Configurations] - Cache configuration directories.

ivandrofly:
Note: Calling Path.Combine is not that much efficient it will loop through all given paths twice + validation ... I have also noticed that `GetDataDirectory` in `Configuration.cs` has used it alot, I will optimize that one too.


niksedk:
Would it be possible to make the private strings public and delete the getters ?

ivandrofly:
Should we postfix everything with: **Folder** or **Directory**?  😕

ivandrofly:
This one is ready :) @niksedk

PR #2025, opened 2016-10-21
[TimeCode] - Fix critical issues in TimeCode class.

ivandrofly:
Anyone who have access to TimeCode class can easily modify the `MaxTime`, this patch will prevent them from doing that.

### DIFF
--- a/src/Forms/Main.cs
+++ b/src/Forms/Main.cs
@@ -7740,8 +7740,8 @@ namespace Nikse.SubtitleEdit.Forms
 
                 if (currentParagraph.StartTime.IsMaxTime && currentParagraph.EndTime.IsMaxTime)
                 {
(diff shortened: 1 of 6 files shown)


src/Forms/Main.cs
```diff
@@ -7740,8 +7740,8 @@ private void SplitSelectedParagraph(double? splitSeconds, int? textIndex)
 
                 if (currentParagraph.StartTime.IsMaxTime && currentParagraph.EndTime.IsMaxTime)
                 {
-                    newParagraph.StartTime = TimeCode.MaxTime;
-                    newParagraph.EndTime = TimeCode.MaxTime;
```
ivandrofly: If `newParagraph.StartTime` or `newParagraph.EndTime` changes wit will cause the `TimeCode.MaxTime` to change aswell.

niksedk:

PR #2028, opened 2016-10-24
[SubtitleFormat] - Filter by extension when checking subs.

ivandrofly:
Part 1

Will prevent loading xml/json/.txt/... using subtitle format which has nothing to do with them.
Let me know if it's okay to do this if not (FREE TO CLOSE).
Note: fileName param (required).


niksedk:
Sorry, it's on purpose that most formats don't check format name - subtitles are not always named correctly...

PR #2032, opened 2016-10-26
[SETextBox] - Use DateTime.UtcNow instead time-zoned one.

ivandrofly:
Interanally **DataTime.Now.Ticks** will create instance of **DateTime.UtcNow** and try to figure out the Time-zone using **DateTime.UtcNow** which is futile in this scenario.


niksedk:
Wow, really a huge speed factor speed improvement :)
Had no idea about that, thx !

PR #2049, opened 2016-10-30
[Tagging] - Fix/Update

ivandrofly:
@niksedk


libse/HtmlUtil.cs
```diff
@@ -675,8 +675,9 @@ public static string FixInvalidItalicTags(string text)
 
         public static string ToogleTag(string text, string tag)
         {
-            if (text.Contains("<" + tag + ">"))
+            if (text.LastIndexOf(tag + '>', StringComparison.OrdinalIgnoreCase) > 0)
             {
+                text = FixUpperTags(text);
```
ivandrofly: Is it okay to lower all tags aswell? @niksedk

PR #2050, opened 2016-10-30
[HtmlUtil] - Fix toggle/Untoggle if uppercase.

ivandrofly:
Fixes toggle/untoggle if uppercase while in listview
also fixies if tag contains only one tag (e.g: <i> or </i>)

_Note: I have also found this same behavior in textbox. (patch on its way)_

### DIFF
--- a/libse/HtmlUtil.cs
+++ b/libse/HtmlUtil.cs
@@ -675,8 +675,9 @@ namespace Nikse.SubtitleEdit.Core
 
         public static string ToogleTag(string text, string tag)
         {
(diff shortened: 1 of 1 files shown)


libse/HtmlUtil.cs
```diff
@@ -675,8 +675,9 @@ public static string FixInvalidItalicTags(string text)
 
         public static string ToogleTag(string text, string tag)
         {
-            if (text.Contains("<" + tag + ">"))
```
ivandrofly: Checking like this, will miss if text contains only </i>. :)

PR #2056, opened 2016-11-01
[Subtitle] - Fix bug when calculating display time. [WIP]

ivandrofly:
DO NOT REVIEW YET.

Via => Tool\Adjust (percent)

Compare both results (from this patch and latest release) and you will see :)

### DIFF
--- a/libse/Subtitle.cs
+++ b/libse/Subtitle.cs
@@ -319,21 +319,8 @@ namespace Nikse.SubtitleEdit.Core
 
         public void AdjustDisplayTimeUsingPercent(double percent, List<int> selectedIndexes)
         {
-            for (int i = 0; i < _paragraphs.Count; i++)
-            {
-                if (selectedIndexes == null || selectedIndexes.Contains(i))
-                {
-                    double nextStartMilliseconds = double.MaxValue;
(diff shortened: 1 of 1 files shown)


libse/Subtitle.cs
```diff
@@ -319,21 +319,8 @@ public bool WasLoadedWithFrameNumbers
...
-                        nextStartMilliseconds = _paragraphs[i + 1].StartTime.TotalMilliseconds;
-
-                    double newEndMilliseconds = _paragraphs[i].EndTime.TotalMilliseconds;
-                    newEndMilliseconds = _paragraphs[i].StartTime.TotalMilliseconds + (((newEndMilliseconds - _paragraphs[i].StartTime.TotalMilliseconds) * percent) / 100);
-                    if (newEndMilliseconds > nextStartMilliseconds)
-                        newEndMilliseconds = nextStartMilliseconds - 1;
-                    _paragraphs[i].EndTime.TotalMilliseconds = newEndMilliseconds;
-                }
-            }
+            double adjustFactorInSeconds = percent % 100.00 / 100.00;
```
ivandrofly: It's okay to do this if percent is going to be < 2% (or 200%)

PR #2075, opened 2016-11-10
[StringExtension] - Optimize FixExtraWhiteSpaces.

ivandrofly:
Fixed:
- [x] - New instance of string every time *a* white-space chars is removed.
- [x] - Running before:  ~ *n^2 + n + n*  => Now: *n*.

Note: *I don't know how*  `private extern string ReplaceInternal(string oldValue, string newValue);` *works so I assumed it's running* **n** *for every call before returning a new string.*


niksedk:
Works nicely :)
A little more than 3x faster in my tests!

PR #2076, opened 2016-11-12
[Utilities] - Minor fixes.

ivandrofly:


libse/Utilities.cs
```diff
@@ -429,24 +429,8 @@ public static string AutoBreakLine(string text, int maximumLength, int mergeLine
             }
 
             string s = RemoveLineBreaks(text);
-            string noTagText = HtmlUtil.RemoveHtmlTags(s, true);
-
-            if (noTagText.Length < mergeLinesShorterThan)
+            if (HtmlUtil.RemoveHtmlTags(s, true).Length < mergeLinesShorterThan)
             {
-                var noTagLines = noTagText.SplitToLines();
```
ivandrofly: noTagText will never contains \r\n because we unbreak var `s` before removing html tags.

PR #2080, opened 2016-11-14

diomed:


aaaxx:
@diomed, can you somehow take care of "veke vekova" -> "vijeke vjekova"?

diomed:
I can replace veke -> vijeke

But some words are simply like that - [vjekova / stoljeća] that I don't know what to do in this case.
Maybe it's better not to replace at all?

What do you think?
I go by quantity of word use - may God forgive me for _Očenaš_.

aaaxx:
The only options are word and regex search, right? I assume the word search looks for the string surrounded by whitespace or something like regex `\b`, right?

What's wrong with the regex `veke vekova`? Would it be a matter of which replacement is executed first?

PR #2082, opened 2016-11-15
[HtmlUtil] - Add optimized method to remove empty tags. [WIP]

ivandrofly:
TODO:
- [x] - Add unit-test.
- [ ] - Remove tag if like: `<i> </i>` (space between)
     - [ ] - Add Unit-test.

### DIFF
--- a/src/Test/Core/HtmlUtilTest.cs
+++ b/src/Test/Core/HtmlUtilTest.cs
@@ -70,5 +70,65 @@ namespace Test.Core
             Assert.AreEqual("<i>foobar?</i>" + Environment.NewLine + "<i>foobar?</i>", HtmlUtil.FixInvalidItalicTags(s));
         }
 
+        #region <<Remove Empty Tags>>
+
+        [TestMethod]
+        public void RemoveEmptyTagsTest1()
+        {
+            const string expected = "Foobar.";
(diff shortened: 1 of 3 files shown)


src/Test/Core/HtmlUtilTest.cs
```diff
@@ -70,5 +70,40 @@ public void FixInvalidItalicTags2()
...
+        [TestMethod]
+        public void RemoveEmptyTagsTest4()
+        {
+            const string expected = "Foo bar.";
+            string output = HtmlUtil.RemoveEmptyTags("Foo<font color=\"#000000\"> </font>bar.");
+            Assert.AreEqual(expected, output);
+        }
+
+        #endregion
     }
```
niksedk: Some more unit tests would be nice... with multiple tags and more evil constructs like "><Foobar<b></b>.><"

Have you tried it on a number of live subtitles?

ivandrofly:
I have made some changes can you please review and let me know 
if you agree/disagree with remove logics I added. @niksedk 


Ps: the failing test is for tags like: "> <". will be fixed when all is done.

niksedk:
I don't see any reason to check more tags than i, b, u and font - just keep it as simple and fast as possible.

PR #2083, opened 2016-11-15
Added MPV player support for mono

var1ap:
If libmpv.so is found in the system then it will be used instead of MPlayer. 
MPV player works many times better then MPlayer fallback.
tested with: direct3d, sdi, vaapi, vdpau.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ SubtitleEdit-*-setup.exe
 /src/SubtitleEdit.VC.opendb
 /src/SubtitleEdit.VC.VC.opendb
 /src/SubtitleEdit.VC.db
+/src/SubtitleEdit.userprefs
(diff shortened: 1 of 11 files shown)


niksedk:
Cool :)
Yes, MPlayer does not work well as an integrated player, so very nice you got this working :)
(I don't have Ubuntu installed atm)

Could you move native methods (DllImport) to NativeMethods.cs? (see https://github.com/SubtitleEdit/subtitleedit/blob/master/src/Logic/NativeMethods.cs )

Could you add some comment about why setlocale is needed for (was it only for libmpv?)

var1ap:
>>Could you add some comment about why setlocale is needed (was it only for libmpv?)
Yes, it is only required by libmpv. This is alternative way to setup LC_NUMERIC=C.

PR #2088, opened 2016-11-19
[Utilities] - Minor optimization for tag-counting.

ivandrofly:
Saves us from all irrelevants check when IndexOf is called.


niksedk:
Slower :(

ivandrofly:
Oh oh... yeah! you are right, I was comparing it the one that take *String* index of char. 👍

PR #2105, opened 2016-11-22
[TextDraw] - No instance of StringBuilder needed here.

ivandrofly:


libse/TextDraw.cs
```diff
@@ -100,11 +98,10 @@ public static float MeasureTextHeight(Font font, string text, bool bold)
             {
                 using (var path = new GraphicsPath())
                 {
-                    var sb = new StringBuilder(text);
```
ivandrofly: ☝️

libse/TextDraw.cs
```diff
@@ -64,11 +63,10 @@ public static float MeasureTextWidth(Font font, string text, bool bold)
...
```
ivandrofly: ☝️

libse/TextDraw.cs
```diff
@@ -6,7 +6,7 @@ namespace Nikse.SubtitleEdit.Core
 {
     public static class TextDraw
     {
-        public static void DrawText(Font font, StringFormat sf, GraphicsPath path, StringBuilder sb, bool isItalic, bool isBold, bool isUnderline, float left, float top, ref bool newLine, float leftMargin, ref int pathPointsStart)
+        public static void DrawText(Font font, StringFormat sf, GraphicsPath path, string s, bool isItalic, bool isBold, bool isUnderline, float left, float top, ref bool newLine, float leftMargin, ref int pathPointsStart)
```
ivandrofly: We can simply pass a string since no build operation is been done inside DrawText.

niksedk:

PR #2119, opened 2016-11-27
[NamesList] - Fix exception when specific dictionary is no installed.

ivandrofly:
Produce:
FCE -> Language (pt)  (make sure it's not installed) - > Fix


niksedk:

PR #2123, opened 2016-11-30
[JsonEncoding]  - Avoid creating *n instance of stringbuilder.

ivandrofly:
Do not merge yet... need to handle newline replacing!


niksedk:
Tested it some... both as is and with better new line replacement and it's much slower than the old code.
I've done some optimizations in latest commit: 704d7bbeeab028eb23c7a72f2ef1a66433aa459c

ivandrofly:
> it's much slower than the old code.

yeah, because if this: https://github.com/SubtitleEdit/subtitleedit/pull/2123/files#diff-82fc7fc6ff9d43685e3f12cdceadacedR48

We are looking through entire text (including timecodes) for \r\n...  I think I found a better solution. still reviewing it.

niksedk:

PR #2144, opened 2016-12-13
[Update] - Change casing.

ivandrofly:


niksedk:
What is the performance improvement? 1.5x, 3x or ?

src/Forms/ChangeCasing.cs
```diff
@@ -65,13 +67,15 @@ internal void FixCasing(Subtitle subtitle, string language)
...
+            bool fixLowercaseI = language.StartsWith("en", StringComparison.Ordinal) && radioButtonNormal.Checked;
+
             string lastLine = string.Empty;
             foreach (Paragraph p in subtitle.Paragraphs)
             {
                 p.Text = FixCasing(p.Text, lastLine, namesEtc);
 
                 // fix casing of English alone i to I
-                if (radioButtonNormal.Checked && language.StartsWith("en", StringComparison.Ordinal))
+                if (fixLowercaseI)
```
ivandrofly: Well, for every paragraph we are checking:
`(radioButtonNormal.Checked && language.StartsWith("en", StringComparison.Ordinal))`. there is always a impact on accessing a variable (ref type).

ivandrofly: Ah no-no... "heavy" may be wrong word in this case I just mean the whole operation can be reduced by 1...

niksedk:

PR #2154, opened 2016-12-15
[SpellCheckWordLists] - Update RemoveName.

ivandrofly:
- Give parameter more meaningful name.
- Fix redundant checkes (if names ends with "s").


niksedk:

PR #2156, opened 2016-12-16
[SpellCheck] - Minor fixes.

ivandrofly:
- Remove useless switch-case.
- Cache string when it's converted to upper-case to avoid recalling to uppercase.


niksedk:

PR #2162, opened 2016-12-23
[OcrFixEngine] - Optimize fixing ocr-hardocded.

ivandrofly:
Optimization for **OCR fix engine**:
Before: ~4n
After: n

Merry christmas @niksedk  | @diomed  | @xylographe 🎄

### DIFF
--- a/src/Logic/OCR/OcrFixEngine.cs
+++ b/src/Logic/OCR/OcrFixEngine.cs
@@ -355,15 +355,37 @@ namespace Nikse.SubtitleEdit.Logic.Ocr
                 text = text.Replace(Environment.NewLine + Environment.NewLine, Environment.NewLine);
             text = text.Trim();
 
-            var sb = new StringBuilder();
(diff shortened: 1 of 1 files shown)


diomed:
Merry Christmas @ivandrofly  & you all! 
Have a happy holidays 🎄 🎅 🎁

aaaxx:
Shouldn't these match the ones in [subtitleedit/libse/Dictionaries/OcrFixReplaceList.cs](https://github.com/SubtitleEdit/subtitleedit/blob/master/libse/Dictionaries/OcrFixReplaceList.cs#L282)?

niksedk:
Merry Christmas :)

ivandrofly:
> Shouldn't these match the ones in subtitleedit/libse/Dictionaries/OcrFixReplaceList.cs?

That is probably true, but I think they shouldn't be merged 'cause it looks like they doesn't share same context.

PR #2163, opened 2016-12-25
Minor fixes.

ivandrofly:


libse/Utilities.cs
```diff
@@ -1077,8 +1077,7 @@ public static bool StartsAndEndsWithTag(string text, string startTag, string end
             if (!text.Contains(startTag) || !text.Contains(endTag))
                 return false;
 
-            while (text.Contains("  "))
-                text = text.Replace("  ", " ");
+            text = text.FixExtraSpaces();
```
ivandrofly: Note: FixExtraSpaces also fixes spaces after/before new-line separator...

niksedk:

PR #2170, opened 2017-01-05
[Utilities] - Optimize remove white space.

ivandrofly:


libse/Utilities.cs
```diff
@@ -1702,19 +1702,31 @@ public static string RemoveSpaceBetweenNumbers(string text)
...
-
-            text = text.Replace(zeroWidthSpace, string.Empty);
-            text = text.Replace(zeroWidthNoBreakSpace, string.Empty);
-            text = text.Replace(noBreakSpace, " ");
-            text = text.Replace(operatingSystemCommand, string.Empty);
-            text = text.Replace('\t', ' ');
-
+            int len = text.Length;
+            int count = 0;
+            char[] textChars = new char[len];
```
niksedk: Nice with the "custom string builder" - that's really what gives the improved performance (could most likely be used other places too).

libse/Utilities.cs
```diff
@@ -1702,19 +1702,31 @@ public static string RemoveSpaceBetweenNumbers(string text)
...
+            for (int i = 0; i < len; i++)
+            {
+                char ch = text[i];
+                // Ignore: \u200B, \uFEFF and \u009D.
+                if (!(ch == zeroWidthSpace || ch == zeroWidthNoBreakSpace || ch == operatingSystemCommand))
```
niksedk: If you used a switch it will be a bit faster... and also more readable.

ivandrofly: Well, about the switch I'm not sure it it's a good idea, note that here we are filtering (skipping) these chars not **selecting** then which would be better to do which a switch-statement.

niksedk: It's about 13 percent faster:
```cs
                switch (ch)
                {
                    case zeroWidthSpace:
                        break;
                    case zeroWidthNoBreakSpace: 
                        break;
                    case noBreakSpace: 
                        textChars[count++] = ' ';
                        break;
                    case operatingSystemCommand:
                        break;
                    case '\t':
                        textChars[count++] = ' ';
                        break;
                    default:
                        textChars[count++] = ch;
                        break;
                }
```

And about 50% more readable ;)

ivandrofly:
Updated @niksedk 

The StringBuilderCache helper class can be use in many scenarios which makes it very useful :)

niksedk:
First, this second commit is much slower. Please please measure performance!

Secondly, this "StringBuilderCache" is from the .net framework work 4.x, right? 
See https://referencesource.microsoft.com/#mscorlib/system/text/stringbuildercache.cs or http://stackoverflow.com/questions/20029868/understanding-of-net-internal-stringbuildercache-class-configuration  - and I don't think it's open source...

First commit was nice... only needed the switch ;)

niksedk:
Nice :)

PR #2171, opened 2017-01-06
[Pac/Utilities] - Optimize checking for ascci letter/digit.

ivandrofly:
Reduced operation ~**2n+10** to **~2** where **n** is the length of English alphabet


niksedk:

PR #2194, opened 2017-01-29
[FixCommonErrors] - Minor fixes 

ivandrofly:


src/Forms/FixCommonErrors.cs
```diff
@@ -452,19 +452,19 @@ private void FixLargeFonts()
         {
             using (var graphics = CreateGraphics())
             {
-                var textSize = graphics.MeasureString(buttonCancel.Text, Font);
+                var textSize = graphics.MeasureString(buttonCancel.Text, buttonCancel.Font);
```
ivandrofly: We shouldn't assume that `Font` is going to be equal to `buttonCancel.Font`

niksedk:

PR #2202, opened 2017-02-06
[VobSubOcr] - Minor optimizations.

ivandrofly:


src/Forms/VobSubOcr.cs
```diff
@@ -6024,7 +6023,10 @@ private bool HasSingleLetters(string line)
             foreach (string s in arr)
             {
                 if (s.Length == 1 && !@"♪♫-:'”1234567890&aAI""".Contains(s))
+                {
                     count++;
```
ivandrofly: We only need to be > 0 os this method can return **true**

PR #2211, opened 2017-02-09
[VobSubOcr] - No need for keeping the loop ON.

ivandrofly:


src/Forms/VobSubOcr.cs
```diff
@@ -8844,6 +8844,10 @@ private void importNewTimeCodesToolStripMenuItem_Click(object sender, EventArgs
                         currentP.EndTime.TotalMilliseconds = p.EndTime.TotalMilliseconds;
                         subtitleListView1.SetStartTimeAndDuration(index, currentP);
                     }
+                    else
+                    {
+                        break;
+                    }
                     index++;
```
niksedk: Perhaps a while loop?

ivandrofly:
Updated :), it's cleaner actually. /cc @niksedk

PR #2213, opened 2017-02-09
[VobSubNOcrTrain] - No need for N instances of string + Bugfixed

ivandrofly:
Bug fixed, read the notation.

### DIFF
--- a/src/Forms/VobSubNOcrTrain.cs
+++ b/src/Forms/VobSubNOcrTrain.cs
@@ -203,19 +203,28 @@ namespace Nikse.SubtitleEdit.Forms
             const float leftMargin = left;
             int newLinePathPoint = -1;
             Color c = _subtitleColor;
-            while (i < text.Length)
+            int textLen = text.Length;
+            while (i < textLen)
             {
-                if (text.Substring(i).StartsWith(Environment.NewLine))
+                char ch = text[i];
+                if (ch == '\n' || ch == '\r')
                 {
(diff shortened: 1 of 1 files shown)


src/Forms/VobSubNOcrTrain.cs
```diff
@@ -203,19 +203,28 @@ private Bitmap GenerateImageFromTextWithStyle(string text, bool bold)
             const float leftMargin = left;
             int newLinePathPoint = -1;
             Color c = _subtitleColor;
-            while (i < text.Length)
+            int textLen = text.Length;
+            while (i < textLen)
             {
-                if (text.Substring(i).StartsWith(Environment.NewLine))
```
ivandrofly: Substring will return a new segment of **text**...

src/Forms/VobSubNOcrTrain.cs
```diff
@@ -203,19 +203,28 @@ private Bitmap GenerateImageFromTextWithStyle(string text, bool bold)
...
+                char ch = text[i];
+                if (ch == '\n' || ch == '\r')
                 {
                     TextDraw.DrawText(font, sf, path, sb, false, subtitleFontBold, false, left, top, ref newLine, leftMargin, ref newLinePathPoint);
 
                     top += lineHeight;
                     newLine = true;
-                    i += Environment.NewLine.Length - 1;
```
ivandrofly: Incrementing by 1 will only skip 1st char from \r\n... so the pointer will stop at '\n' then when we compare with ** if (text.Substring(i).StartsWith(Environment.NewLine))** it will be false and '\n' will be included in **sb**.

PR #2218, opened 2017-02-11
[RemoveHi] - Only renumber if paragraphs were removed.

ivandrofly:
This patch will prevent Remove-HI from renumbering subtitle when OK is clicked even when no HI is present in text. /cc @niksedk 

To produce:
Renumber subtitle with any number >=2 (make sure subtitle does not contain any HI tags/notation)
Open Remove-HI and click OK. Subtitle will be renumbered.


niksedk:

PR #2226, opened 2017-02-16
[Main] - Call helper method to takeout unicode chars.

ivandrofly:


src/Forms/Main.cs
```diff
@@ -6994,13 +6994,7 @@ private void UpdateListViewTextInfo(Label lineLengths, Label singleLine, Label l
...
 
             // remove unicode control characters
-            s = s.Replace("\u8207", string.Empty).
-                Replace("\u8206", string.Empty).
-                Replace("\u202A", string.Empty).
-                Replace("\u202B", string.Empty).
-                Replace("\u202D", string.Empty).
-                Replace("\u202E", string.Empty).
-                Replace("\u002B", string.Empty);
+            s = s.RemoveControlCharacters();
```
ivandrofly: Note: RemoveControlCharacters will probably remove more control-char then mentioned above...

PR #2228, opened 2017-02-17
[Main] - Fix issues regarding to multi-monitor.

ivandrofly:
I just run into this problem recently... 😄 

Ready /cc @niksedk 

PS: I just added one fix-commit, please do *squash merged* when you decide to do merge.

### DIFF
--- a/src/Forms/Main.cs
+++ b/src/Forms/Main.cs
@@ -1081,17 +1081,19 @@ namespace Nikse.SubtitleEdit.Forms
                     }
                 }
 
-                var screen = Screen.FromControl(this);
+                var ctrlScreen = Screen.FromControl(this);
 
-                if (screen.Bounds.Width < Width)
-                    Width = screen.Bounds.Width;
-                if (screen.Bounds.Height < Height)
-                    Height = screen.Bounds.Height;
+                if (ctrlScreen.Bounds.Width < Width)
+                    Width = ctrlScreen.Bounds.Width;
+                if (ctrlScreen.Bounds.Height < Height)
+                    Height = ctrlScreen.Bounds.Height;
 
-                if (screen.Bounds.X + screen.Bounds.Width - 200 < Left)
-                    Left = screen.Bounds.X + screen.Bounds.Width - Width;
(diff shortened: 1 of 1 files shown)


src/Forms/Main.cs
```diff
@@ -1088,9 +1088,14 @@ private void Main_Load(object sender, EventArgs e)
                 if (screen.Bounds.Height < Height)
                     Height = screen.Bounds.Height;
 
-                if (screen.Bounds.X + screen.Bounds.Width - 200 < Left)
```
ivandrofly: MSDN: [Control-Left](https://msdn.microsoft.com/en-us/library/system.windows.forms.control.left(v=vs.110).aspx)

**Control.Left:** _Gets or sets the distance, in pixels, between the left edge of the control and the left edge of its container's client area._

src/Forms/Main.cs
```diff
@@ -1088,10 +1088,16 @@ private void Main_Load(object sender, EventArgs e)
                 if (screen.Bounds.Height < Height)
                     Height = screen.Bounds.Height;
 
-                if (screen.Bounds.X + screen.Bounds.Width - 200 < Left)
-                    Left = screen.Bounds.X + screen.Bounds.Width - Width;
-                if (screen.Bounds.Y + screen.Bounds.Height - 100 < Top)
-                    Top = screen.Bounds.Y + screen.Bounds.Height - Height;
+                // Fix main window coordinate (Multi-Monitor issue)
+                if (screen.Bounds.Right < Left)
```
ivandrofly: Screen right == `screen.Bounds.X + screen.Bounds.Width`

niksedk: What was the problem exactly?
I'll test it next week (I can test multi screen setup then)

ivandrofly: This is how you produce it.
1. Start Subtitle Edit
2. Move it to secondary screen.
3. Close the entire app (while in secondary screen).
4. Unconnect HDMI cable (or just WinKey + P and choose PC Screen Only)
5. Re open subtitle edit.

After reopening Subtitle Edit  you won't be able to see the Main window, it will still be stuck in secondary screen boundary.

PR #2232, opened 2017-02-20
Organize Control into folder

ivandrofly:


niksedk:
Nice. Perhaps the "VobSub" folder should be called "Ocr"... OCR in SE started out with just the VobSub format, but now it's a lot more formats...

ivandrofly:
Updated @niksedk

PR #2234, opened 2017-02-21
[UIUtil] - Fix/Update.

ivandrofly:
1. Give more meaningful name for method `UiUtil.GetLineLengths`.
2. No need for creating instance in order to get extension.


niksedk:

PR #2245, opened 2017-02-25
[SubtitleListView] - Minor refact.

ivandrofly:


src/Controls/SubtitleListView.cs
```diff
@@ -50,23 +50,17 @@ public int GetColumnIndex(SubtitleColumn column)
 
         public override bool RightToLeftLayout
         {
-            get { return base.RightToLeftLayout; }
+            get
+            {
+                return base.RightToLeftLayout;
+            }
```
ivandrofly: ` get => base.RightToLeftLayout;` would also work (vs2k17) didn't test it with 2k15

niksedk: Are you running VS2017? Does SE compile okay with VS2017?

ivandrofly: It does :) @niksedk

PR #2251, opened 2017-02-25
[Main] - Minor optimization.

ivandrofly:


src/Forms/Main.cs
```diff
@@ -7282,9 +7282,8 @@ private void TextBoxListViewTextTextChanged(object sender, EventArgs e)
         {
             if (_subtitleListViewIndex >= 0)
             {
-                int numberOfNewLines = Utilities.GetNumberOfLines(textBoxListViewText.Text);
```
ivandrofly: Was being computed even when not necessary...

PR #2254, opened 2017-02-27
[Main] - Only run alternate timer when tb-alternate is visible.

ivandrofly:
Please do not merge yet, still looking for side effects.


niksedk:

PR #2269, opened 2017-03-06
[Main] - Minor update.

ivandrofly:


niksedk:
could be even shorter

                var selectedIndices = SubtitleListview1.SelectedIndices;
                for (int i = selectedIndices.Count - 1; i > 0; i--)
                {
                    SubtitleListview1.Items[selectedIndices[i]].Selected = false;
                }

ivandrofly:
Was what I wrote at first, but this way we are double checking which I think is unnecessary...
I can change it that way if you want :)  /cc @niksedk

ivandrofly:
I got a question... are we planning to switch .NET4.5 anytime soon?
there are bunch of apis I want to integrate on plugin system...

niksedk:
How is it double checking?

I think we should stay on 4.0 for a little while longer and jump to 4.6.2 at some point (also support long files names). I think SE still have some WinXP users...

ivandrofly:
https://github.com/SubtitleEdit/subtitleedit/pull/2269/files#diff-2ec591b6c56d41734fb35d871b4eca07L12459

If we ignore that line then `e.SuppressKeyPress = true;` will always to set to true whenever CTRL+D is pressed...

PR #2275, opened 2017-03-08
[PluginsGet] - Update/Fix.

ivandrofly:
- Restore cursor to default when download operation fails after setting cursor to wait.
- Remove `Cursor = Cursors` after a async operation.
- Add helper method to avoid code duplication.

### DIFF
--- a/src/Forms/PluginsGet.cs
+++ b/src/Forms/PluginsGet.cs
@@ -67,10 +67,7 @@ namespace Nikse.SubtitleEdit.Forms
             }
             catch (Exception exception)
             {
-                labelPleaseWait.Text = string.Empty;
-                buttonOK.Enabled = true;
-                buttonDownload.Enabled = true;
-                listViewGetPlugins.Enabled = true;
+                ChangeControlsState(true);
                 MessageBox.Show(exception.Message + Environment.NewLine + Environment.NewLine + exception.StackTrace);
             }
         }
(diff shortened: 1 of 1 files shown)


src/Forms/PluginsGet.cs
```diff
@@ -377,14 +375,10 @@ private void buttonUpdateAll_Click(object sender, EventArgs e)
                     wc.DownloadDataCompleted += wc_DownloadDataCompleted;
                     wc.DownloadDataAsync(new Uri(_updateAllListUrls[i]));
                 }
-                Cursor = Cursors.Default;
```
ivandrofly: Shouldn't be here... it will be called just after the line # 378 finishes creating async task.

ivandrofly:
Plugins update seems to have a bug... WIP.
(Plugins)[http://i.imgur.com/Mo1cIl7.gifv]

niksedk:
@ivandrofly: Ready?

ivandrofly:
yeah this one is ready. I'm currently on bug mentioned above @niksedk

niksedk:
I cannot re-create the issue above - works fine here...

ivandrofly:
that is weird, I tried from both VS and latest release (portable) it still not updating the plugins after clicking **Update All** button ... I'll try on another pc to see if problem still exist.

ivandrofly:
the issue was being caused due to un-updated (binary file) (LinesUnbreaker). fixed :)

PR #2278, opened 2017-03-09
[Name-List] - Update names_etc.xml

ivandrofly:
### TODO:
- [x] Rename \<ignore_list> to \<names>
- [x] Introduce new \<blacklist> element
    - [x] Add names like Male to en_names.xml and Rabat to da_names.xml \<blacklist> element.
- [x] Make file names culture insensitive (en_US_names_etc.xml and en_GB_name.xml to en_names.xml)
- [x] **Fix broken codes due to root changed in names \<ignore_list> => \<names>**
- [x] Prevent adding names that are already in blacklist
- [x] Ready!

**Note:** ~In order to avoid merge-conflicts I'd recommend to hold any updates regarding to name list, until this branch is finished.~

Note: **Remove and copy a brand new **Dictionaries** folder to your bin\Debug**


niksedk:
Nice, looks good :)
"Male" should probably be removed, "Rabat" moved to English names. 
The difficult part is that a general name must not collide with common words in other languages.

ivandrofly:
> "Rabat" moved to English names.

I'm a bit confused.. any reason why **Rabat** should be moved to English names?

ivandrofly:
@niksedk what do you think about making name-list culture insensitivity? like US-en and GB-en can share same name-list. I have found several common names in both cultures.

niksedk:
@ivandrofly: "Rabat" is a common Danish word. 
Yes, culture insensitive name lists makes sense.

ivandrofly:
another thing, The root element used in all names list I think is misspelled **<ignore_words>**  should be <names>

niksedk:
@ivandrofly: **\<names>** or **\<names_etc>** is better, and a **\blacklist** would be a nice solution the name-word clashes.

ivandrofly:
By weekend I think this one will be ready ;)

ivandrofly:
I think this one is ready... @niksedk

niksedk:
@ivandrofly: Is it ready or ? 
(check the big "DO NOT MERGE YET" in first post)

ivandrofly:
@niksedk ready

niksedk:
@ivandrofly: Great :)
Will look at this soon when I get enough time.

Have you thought of how to handle existing user dictionaries? Auto-convert or ?

niksedk:
Just had a quick look. The Danish blacklist does not work. Also the local names list xml file is loaded twice (both in `LoadNamesList` and `UnloadBlacklistNames`).
Do you want to fix it or should I look at it?

ivandrofly:
> Just had a quick look. The Danish blacklist does not work

Have you deleted the **Dictionaries** folder from bin\Debug\ and recopy the new modified dictionaries files?

ivandrofly:
> Have you thought of how to handle existing user dictionaries? Auto-convert or ?

User dictionaries? Do you mean "*.dic" files or name lists?

ivandrofly:
> Also the local names list xml file is loaded twice

Hmm... names.xml or {en}_names.xml?

niksedk:
The "en_names.xml" and "en_names_user.xml" xml files are loaded twice.
Does the Danish blacklist work now?

ivandrofly:
There was a spelling mistake in tag. Fixed in last commit :) @niksedk

ivandrofly:
> "en_names_user.xml"

There is  `en_names_user.xml`... did you mean en_US_names.xml?
where is it been loaded twice? 😕

ivandrofly:
There is one more thing I think needs to be handled, when adding names, we need to check if name that is been added isn't is black list before adding it!

niksedk:
The xml files are loaded both in `LoadNamesList` and `UnloadBlacklistNames`.
Let me know when this is ready :)

ivandrofly:
this one is ready... seems to have merge conflicts!

Dictionaries/en_names.xml
```diff
@@ -1175,4 +1175,9 @@
   <name>Zion</name>
   <name>Zoie</name>
   <name>Zuri</name>
+  <blacklist>
+    <name>Male</name>
+    <name>Bill</name>
+    <name>Black</name>
```
ivandrofly: The hard coded version can now be removed from code

niksedk:
Code looks better, but Tools -> "Change casing" does not work for me (no names).

ivandrofly:
> Code looks better, but Tools -> "Change casing" does not work for me (no names).

1. Have you copied the Dictionaries to your debug\bin?
2. If subtitle is too short SE won't be unable to detect the language, so english culture will be set as the default one... what I suggest to try using the UI/Thread culture if subtitle is too short. any suggestion?

niksedk:
1 + 2) I kinda know this, so yes

2) "en" is a nice default - but the language auto detector could be improved for smaller texts.

ivandrofly:
> "Change casing" does not work for me (no names).

😕 did you try casing using a danish subtitle or?

Does casing english subtitle works?

niksedk:
The Danish blacklist does not work... could you verify/fix?

niksedk:
Are you still working on this?

ivandrofly:
> Are you still working on this?

I changed AutoDetectLanguageName to AutoDetectGoogleLanguage which seems more accurate...

niksedk:
OK, change casing works for me now :)

Still some stuff needs to be fixed, like "add to black list" in UI, NamesListManager plugin, and find out how to handle old existing names xml files...

ivandrofly:
> "add to black list" in UI

when we remove the name via UI it will be moved to blacklist if it's in names.xml otherwise it will just be removed.

marb99:
Can you explain what it's for "black list"?

ivandrofly:
Let say we save three names lists

1. en_names.xml (English)
2. pt_names.xml (Portuguese)
3. da_names.xml (Danish)

so for every single one of the list we have: *Google*, in order to avoid duplication we introduce a new name list which will contain all the names that are common for all these languages, and we name that list **names** (names.xml), so we move **Google** to **names** (names.xml).

We find another name that are common in most of the language (not all) for example: Rabat. For both English/Portuguese it mean the capital of Morocco but for Danish it meas something different, we can't just add Rabat to general names because it will cause problem for danish user (according to google Rabat means discount) so that is were blacklist comes to play.

Blacklist: is a list that will contains all the names from general name list, that shouldn't be capitalized from a specific language. according to our example above Rabat will be moved to Danish name list (da_names.xml).

marb99:
@ivandrofly 

thanks for the explanation and makes sense :)

PR #2281, opened 2017-03-11
[SpellCheckWordLists] - Fix/Optimization.

ivandrofly:
### Optimization

- Use HashSet for faster lookup

### Fix

- Only add element with dash/period from **namesEtcMultiWordList** to **_wordsWithDashesOrPeriods**
- Only add element with dash/period from **_userPhraseList** to **_wordsWithDashesOrPeriods**

### DIFF
--- a/libse/SpellCheck/SpellCheckWordLists.cs
+++ b/libse/SpellCheck/SpellCheckWordLists.cs
@@ -19,11 +19,11 @@ namespace Nikse.SubtitleEdit.Core.SpellCheck
 
         private readonly NamesList _namesList;
         private readonly HashSet<string> _namesEtcList;
-        private readonly List<string> _namesEtcListUppercase = new List<string>();
-        private readonly List<string> _namesEtcListWithApostrophe = new List<string>();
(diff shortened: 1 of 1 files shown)


libse/SpellCheck/SpellCheckWordLists.cs
```diff
@@ -81,10 +79,12 @@ public SpellCheckWordLists(string dictionaryFolder, string languageName, IDoSpel
                     }
                 }
             }
-
             // Add names/userdic with "." or " " or "-"
-            _wordsWithDashesOrPeriods = new List<string>();
-            _wordsWithDashesOrPeriods.AddRange(namesEtcMultiWordList);
```
ivandrofly: Here we are just copying all the element s of namesEtcMultiWordList to _wordsWithDashesOrPeriods which is not correct, we need to make sure that the element being copied as **period** or **dash**.
See https://github.com/SubtitleEdit/subtitleedit/pull/2281/files#diff-9e4e72215c5f592ba3b97997bfed63feR95

PR #2286, opened 2017-03-14
[Settings] - Fix issues on word list tab.

ivandrofly:
Fixes/update:

- When name/word being added to list is length is equal to 1 do not displaying message saying it's already in list.
- Display warning message when user select/entered invalid language via combox language and tries to add a word/name.
- Display warning message user tries to add a name with lesser than two character to lists.


niksedk:

PR #2288, opened 2017-03-14
GetDictionaries] - Minor fix for cursor state.

ivandrofly:


src/Forms/GetDictionaries.cs
```diff
@@ -154,6 +153,7 @@ private void wc_DownloadDataCompleted(object sender, DownloadDataCompletedEventA
                                 Environment.NewLine +
                                 e.Error.Message);
                 DialogResult = DialogResult.Cancel;
+                Cursor = Cursors.Default;
```
niksedk: The `Cursor = Cursors.Default;` should probably up higher... clicking on a msgbox with a hourglass seems weird.

ivandrofly:
Updated @niksedk

PR #2290, opened 2017-03-15
[GetDictionary] - Fix resource leak.

ivandrofly:
This leak is even mentioned in *coverity.com*

### DIFF
--- a/src/Forms/GetDictionaries.Designer.cs
+++ b/src/Forms/GetDictionaries.Designer.cs
@@ -7,19 +7,6 @@
         /// </summary>
         private System.ComponentModel.IContainer components = null;
 
-        /// <summary>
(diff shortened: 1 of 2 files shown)


niksedk:
Does this really fix it? I thought the object should be disposed in `Dispose`?

ivandrofly:
yeah that is where it's disposed did you see this line => https://github.com/SubtitleEdit/subtitleedit/pull/2290/files#diff-2cea33a9c3d6f44ce30d83b540530a7fR20

niksedk:
Could you please not edit the designer file and do it as stackoverflow suggests?
(I really would not expect custom code in the designer cs file)

ivandrofly:
👍 ? @niksedk

PR #2304, opened 2017-03-22
Create hrv_diacriticOCRFixReplaceList.xml

diomed:
@ivandrofly @niksedk 

Croatian language has diacritic signs on letters [čćžšđ]
and sometimes people write subs without those. [cszd]

I've dedicated time to create this special version, which would allow conversion of such subs into one with proper letters. Please merge with name you see fit, and also add it properly. I think there is more that needs to be included in code except this file, so it can be connected to program itself.


aaaxx:
You're mixing Serbian-to-Croatian conversion with diacritic fixing. I'm not sure that's the best approach, because you'll have to keep the file updated in parallel with `hrv_OCRFixReplaceList.xml`, which also contains Serbian-to-Croatian conversion, contrary to its name.

I think the optimal thing here would be to have this diacritic fixing in a separate list (or maybe even in the OCR-fixes replacement list if it was refactored so that it actually contains OCR fixes instead of language conversion, but that would be a huge task, so for now it's better to keep diacritics in their own list). Then it can be applied before the language conversion step, which would then need only one set of replacement rules for accented words.

diomed:
@aaaxx I urge you to jump in on this and help me sorting it, in the future, if you will. honestly, 
I do some testing, and although it's not perfect, I think this approach brings more good than bad.

That's why I decided to include S2C conversion. Sure, no1 asked me, but on the other hand, no1 did 
ocr but me, and I've spent more than a year, just perfecting it to suit my needs - then sharing it with others, in hope it'd be of use for them.

diomed:
@niksedk can u please merge this too? thanks

niksedk:
Not really sure how it could be used... in Edit -> Multiple replace or a plugin or ?

diomed:
@niksedk can't it be included somehow as a dropdown in Fix Common Errors, like all the rest?

PR #2330, opened 2017-04-08
[LanguageAutoDetect] - Fix dash in language name "sr-Latn" => srt_Latn

ivandrofly:
`Utilities.GetDictionaryLanguages()` will return language name where the dash is replace with underscore...


niksedk:
Nice catch :)

(still do try to fix only one thing per pr:)

PR #2332, opened 2017-04-09
[LanguageAutoDetect] - Optimize language detection.

ivandrofly:
Optimize language detection

If you could take some time to review these PR regarding to language-detection a bit fast it would be nice, 'cause it somehow related to https://github.com/SubtitleEdit/subtitleedit/pull/2278

PS: there is more patch coming for LanguageAutoDetect I don't want to include them all in a single PR.

### DIFF
--- a/libse/LanguageAutoDetect.cs
+++ b/libse/LanguageAutoDetect.cs
@@ -384,8 +384,7 @@ namespace Nikse.SubtitleEdit.Core
                         if (count > bestCount)
                         {
                             int norwegianCount = GetCount(text, "ut", "deg", "meg", "merkelig", "mye", "spørre");
-                            int dutchCount = GetCount(text, AutoDetectWordsDutch);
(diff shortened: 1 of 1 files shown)


libse/LanguageAutoDetect.cs
```diff
@@ -394,9 +393,10 @@ public static string AutoDetectLanguageName(string languageName, Subtitle subtit
                         if (count > bestCount)
                         {
                             int danishCount = GetCount(text, "siger", "dig", "mig", "mærkelig", "tilbage", "spørge");
-                            int dutchCount = GetCount(text, AutoDetectWordsDutch);
-                            if (danishCount < 2 && dutchCount < count)
+                            if (danishCount < 2 && GetCount(text, AutoDetectWordsDutch) < count)
```
ivandrofly: we only need to evaluate dutch if danish's test condition passes...

PR #2336, opened 2017-04-10
Added autosync functionality.

DeveloperChrisT:
I added code that makes it possible to auto point sync with another subtitle.
I use google translate to compare one subtitle with the other.

The code is probably not in the correct file and should be in the logic layer but I don't know the project very well and didn't want to make a mess of it.

It works pretty well I was able to point sync half of the lines of my subtitle.

### DIFF
--- a/src/Forms/SyncPointsSync.cs
+++ b/src/Forms/SyncPointsSync.cs
@@ -5,7 +5,11 @@ using System.Collections.Generic;
 using System.Drawing;
 using System.Globalization;
 using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
 using System.Windows.Forms;
+using static Nikse.SubtitleEdit.Forms.GoogleTranslate;
 
 namespace Nikse.SubtitleEdit.Forms
 {
@@ -65,6 +69,12 @@ namespace Nikse.SubtitleEdit.Forms
             subtitleListView2.AutoSizeAllColumns(this);
             UiUtil.FixLargeFonts(this, buttonOK);
             labelAdjustFactor.Text = string.Empty;
+
+            GoogleTranslate gt = new GoogleTranslate();
+            gt.FillComboWithGoogleLanguages(comboBoxOtherSubLanguage);
+            gt.FillComboWithGoogleLanguages(comboBoxSubToSyncLanguage);
+            comboBoxOtherSubLanguage.SelectedItem = comboBoxOtherSubLanguage.Items[19];
+            comboBoxSubToSyncLanguage.SelectedItem = comboBoxSubToSyncLanguage.Items[18];
         }
 
         public void Initialize(Subtitle subtitle, string subtitleFileName, string videoFileName, int audioTrackNumber)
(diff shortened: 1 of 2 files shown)


marb99:
I tried this but freeze subtitleedit

subtitleedit Stop responding, I had to kill the process in taskmanager

DeveloperChrisT:
When you hit the button Auto Sync the program freezes until the code is finished. This might take a while.
In the subtitle I used, it takes about 2 minutes for 500 subtitle lines. Did you wait long enough for the code to finish?

I don't know how I can make the code faster setting the sync points takes takes a long time and somehow I can't get it working multi threaded.
I also don't know how I kan make the GUI continue to run while the code is running.

marb99:
I tested with 1452 lines and waited more than 2 minutes which I think is a lot of time for the task that is intended.

but...

src/Forms/SyncPointsSync.cs
```diff
@@ -466,5 +476,116 @@ private void ButtonFindTextOtherClick(object sender, EventArgs e)
...
+                //for each paragraph subtitle to sync
+                for (int i = 0; i < subtitleWrong.Paragraphs.Count; i++)
+                {
+                    subtitleWords = Regex.Matches(subtitleWrong.Paragraphs[i].Text.ToLower(), @"\w+").Cast<Match>().Select(x => x.Value).ToArray(); //_subtitle.Paragraphs[i].Text.Split(' ');
+                    //for each translated paragraph in range
+                    for (int j = nextSyncPoint; j < otherTranslated.Paragraphs.Count && (subtitleWrong.Paragraphs[i].StartTime.Minutes < otherTranslated.Paragraphs[j].StartTime.Minutes + timeMargin && subtitleWrong.Paragraphs[i].StartTime.Minutes > otherTranslated.Paragraphs[j].StartTime.Minutes - timeMargin); j++)
+                    {
+                        numberOfEqualWords = 0;
+                        numberOfWrongWords = 0;
+                        translatedSubs = Regex.Matches(otherTranslated.Paragraphs[j].Text.ToLower(), @"\w+").Cast<Match>().Select(x => x.Value).ToArray(); //otherTranslated.Paragraphs[j].Text.Split(' ');
```
DeveloperChrisT: If you look at the CPU usage I notice that the most time consuming operation is the foreach loop on line 570 when `SetSyncPointAuto(point.Key, point.Value);` is called. Also the translating takes a lot of time.

DeveloperChrisT:
@marb99 So you still weren't able to get a result? The more sync points you have the longer it will take for the code to complete. if it took 2 minutes for 500 lines it might take up to 6 minutes for 1500 lines. Dependent on you single core performance of your CPU (I have an i7 860 @ 3GHz). So in 10 minutes you really should get a result. Which I know is way to long but I am not sure how I can improve it. Any suggestions would be greatly appreciated.

src/Forms/SyncPointsSync.cs
ivandrofly: Well, I could try helping with some optimized codes, but is Nikse really interested in this patch?

hlcb91: Regex is not too slow? From what I've observed you're using Regex just to pick up the words and this can be done in much faster ways by returning the words separated by the "space" character and then converting all words to lowercase.

If you need help with this let me know, synchronization is a subject that I really like. I have ready codes related to this, but in other languages and maybe I can get by and then you adapt to the project.

DeveloperChrisT: I guess I need help implementing the code into the program. And other points below.

- Preventing the UI from freezing when the code is running.

- Adding a progress bar that is updated properly

- Making the code faster

- Maybe add a options menu for the auto translation

niksedk:

PR #2359, opened 2017-04-30
[utilities] - allow fixing spaces after quote.

ivandrofly:
Allow fixing text like:

" In five years the Corleone family
will be completely legitimate."

note there is a space after the 1st quote.

---
- [x] Add unit-test.
- [x] Ready.

### DIFF
--- a/libse/Utilities.cs
+++ b/libse/Utilities.cs
@@ -1930,6 +1930,26 @@ namespace Nikse.SubtitleEdit.Core
                     text = text.Remove(idx, 1);
                 }
             }
+
+            // Fix spaces after quotes
+            // e.g: Foobar. " Foobar" => Foobar. "Foobar"
+            string preText = string.Empty;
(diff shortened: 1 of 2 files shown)


libse/Utilities.cs
```diff
@@ -2004,6 +2004,26 @@ public static string RemoveSpaceBeforeAfterTag(string text, string openTag)
...
+            if (text.StartsWith('"'))
+            {
+                text = '"' + text.Substring(1).TrimStart();
+            }
+            text = preText + text;
+            // Foobar. " Foobar" => Foobar. "Foobar"
+            text = text.Replace(". \" ", ". \"");
+            text = text.Replace("? \" ", "? \"");
+            text = text.Replace("! \" ", "! \"");
+            text = text.Replace(") \" ", ") \"");
```
ivandrofly: This can actually be optimize to n instead of 3n, but lets keep it this way for now 😉 ...

PR #2371, opened 2017-05-05
[timecode] - more cleanup.

ivandrofly:
Remove unnecessary method in **TimeCode.cs**

### DIFF
--- a/libse/TimeCode.cs
+++ b/libse/TimeCode.cs
@@ -58,7 +58,7 @@ namespace Nikse.SubtitleEdit.Core
             return 0;
         }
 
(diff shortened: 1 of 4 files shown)


libse/TimeCode.cs
```diff
@@ -58,7 +58,7 @@ public static double ParseHHMMSSFFToMilliseconds(string text)
             return 0;
         }
 
-        public TimeCode() : this(0)
```
ivandrofly: there is no need for calling this(0) `_totalmilliseconds` will be initialized to zero when `new TimeCode()`

PR #2410, opened 2017-05-27
move ui stuff to SubtitleEdit project.

ivandrofly:
Move VideoInfo.cs and all methods inside Utilities.cs class that process it...

/cc @niksedk


niksedk:

PR #2414, opened 2017-05-29
[fixcommonerrors] - add reset default button.

ivandrofly:


darnn:
This looks good! Will there be a way to set the default?

ivandrofly:
> This looks good! Will there be a way to set the default?

not yet, probably in future it can be implemented 😉 

pst: actually it would be better to have groups just like multiple-replace...

src/Forms/FixCommonErrors.cs
```diff
@@ -1608,5 +1609,10 @@ private void comboBoxLanguage_Enter(object sender, EventArgs e)
             SaveConfiguration();
         }
 
+        private void buttonResetDefault_Click(object sender, EventArgs e)
+        {
+            Configuration.Settings.CommonErrors.LoadDeafultConfigs();
```
ivandrofly: oops... typo **LoadDeafultConfigs**

niksedk:
Could you move the new button to the right of "Invert selection"?
And perhaps call it "Select default" or "Select default fixes" or similar...

Also, don't know if the default settings are good - don't reset my settings often ;)

ivandrofly:
Updated @niksedk

PR #2415, opened 2017-05-29
[GetDictionaries] - allow switching dictionary provider.

ivandrofly:
⚠️  Work in progress... ⚠️ 

TODO:

- [x] Fix long text in label Description
- [x] Fix download button
- [x] Localize
- [ ] Make all dictionaries available in embedded xml lookup file
- [ ] Ready ✔️ 

Hope this patch solves the dictionary hosting problem with: libreoffice.org and  sourceforge.net


niksedk:
You're not really meant to be able to switch via UI (ppl don't know what is what and why...) but only as a last resort if the primary provider is down

diomed:
@ivandrofly where can I find those dictionaries on github? can u gimme link?

ivandrofly:
> @ivandrofly where can I find those dictionaries on github? can u gimme link?

https://github.com/einsiol/Dictionaries

ivandrofly:
> You're not really meant to be able to switch via UI (ppl don't know what is what and why...) but only as a last resort if the primary provider is down

💯

ivandrofly:
yeah by default libre-office will be selected, if anything goes wrong user-will be notified to switch provider....

diomed:
sadly, not even libreoffice per my experience brings all correct words. people add their own words, 
but no one checks if they're really correct. some really aren't.. still, it's better than nothing.

PR #2418, opened 2017-05-30
Fixes #44

### LINKED ISSUE #44: Unexpected line breaks after merging
After merging translation and original texts (Ctrl Shift Alt M), extra line break would appear in each line, which is unexpected.

PS: However, it looks fine in source view, quite strange.  :\ 

Before merging:
beofre

Aftrer merging:
after

ivandrofly:


niksedk:

PR #2420, opened 2017-05-31
Improved Arabic language handling in PAC files

Zehelein:
The current PAC implementation for the Arabic language was using only "basic" characters and the "accent" ones were added to the wrong side.

Original:

image

After the change the "accent" symbols are correctly assigned to the letters:

image

### DIFF
--- a/libse/SubtitleFormats/Pac.cs
+++ b/libse/SubtitleFormats/Pac.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text;
 
 namespace Nikse.SubtitleEdit.Core.SubtitleFormats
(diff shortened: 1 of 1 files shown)


niksedk:
Looks good, also nice with dictionaries :)

libse/SubtitleFormats/Pac.cs
```diff
@@ -1324,11 +951,11 @@ private int AutoDetectEncoding()
                     {
                         _codePage = CodePageLatin;
                         var sb = new StringBuilder("ABCDEFGHIJKLMNOPPQRSTUVWXYZÆØÅÄÖÜabcdefghijklmnopqrstuvwxyzæøäåü(1234567890, .!?-\r\n'\"):;&");
```
ivandrofly: Some of the characters here is already present in `LatinCodes.Values`. e.g: `ÆØÅÄÖÜ`...

niksedk: thx :)
Fixed here: c66cccfdd0bb3f5aa13438b4b1ee278a7ffa22e0

PR #2442, opened 2017-06-17
Exclude 'color' in html font tag (British vs American) english.

ivandrofly:
Makes auto-detecting British subtitle more accurate...


niksedk:

PR #2473, opened 2017-07-10
[IsPlainText ] - minor optimizations

ivandrofly:


libse/FileUtil.cs
```diff
@@ -321,35 +321,31 @@ public static bool IsPlainText(string fileName)
...
                 char ch = s[i];
                 if (char.IsLetter(ch) || " -,.!?[]()\r\n".Contains(ch))
                 {
                     letterCount++;
                 }
-                else if (char.IsControl(ch) && ch != '\t')
+                else if (char.IsControl(ch) && ch != '\t') // binary found
                 {
-                    binaryCount++;
+                    return false;
```
ivandrofly: we don't need to continue looking if a control char is found

PR #2478, opened 2017-07-18
[DvdSubtitleSystem] - Optimize parsing.

ivandrofly:


libse/SubtitleFormats/DvdSubtitleSystem.cs
```diff
@@ -64,25 +55,27 @@ public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string
...
+                if (match.Success)
                 {
-                    string temp = line.Substring(0, RegexTimeCodes.Match(line).Length);
+                    string temp = line.Substring(0, match.Length);
                     string start = temp.Substring(0, 11);
                     string end = temp.Substring(12, 11);
 
                     string[] startParts = start.Split(SplitCharColon, StringSplitOptions.RemoveEmptyEntries);
                     string[] endParts = end.Split(SplitCharColon, StringSplitOptions.RemoveEmptyEntries);
-                    if (startParts.Length == 4 && endParts.Length == 4)
```
ivandrofly: Unnecessary check the regex already confirmed the expected values...

libse/SubtitleFormats/DvdSubtitleSystem.cs
```diff
@@ -64,25 +55,27 @@ public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string
...
+                {
+                    _errorCount += 10;
+                    continue;
+                }
+
+                Match match = RegexTimeCodes.Match(line);
+
+                if (match.Success)
                 {
-                    string temp = line.Substring(0, RegexTimeCodes.Match(line).Length);
```
ivandrofly: Double checking `.Match()`

libse/SubtitleFormats/DvdSubtitleSystem.cs
```diff
@@ -64,25 +55,27 @@ public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string
             _errorCount = 0;
             foreach (string line in lines)
             {
-                if (RegexTimeCodes.IsMatch(line))
+                // line must contain atleast 24 characters (time-code)...
+                if (line.Length < 24)
```
ivandrofly: Regex pattern is equal to 24 to there is no need to force regex matching shorted text (for example others subtitle formats...)

PR #2482, opened 2017-07-22
[cheetahcaption] - better lookup for latin chars

ivandrofly:


niksedk:
I cannot make this work... does it work for you?

ivandrofly:
> I cannot make this work... does it work for you?

I used the C# 7 dictionary initialization style... are you using VS 2017?

PS: I can change it to old-version/style

niksedk:
Try to test it with special letters...

ivandrofly:
updated /cc @niksedk

niksedk:
@ivandrofly: Nice, now the code works (please always do some testing on pull requests).
Perhaps you can eliminate the duplication of the dictionary too?

PR #2499, opened 2017-08-01
[Main] - minor refact setting right-to-left alignment

ivandrofly:
More cleaner code to set right-to-left text format.


niksedk:

niksedk:

PR #2500, opened 2017-08-01
[Ultech130] - update/optimize

ivandrofly:


libse/SubtitleFormats/Ultech130.cs
```diff
@@ -189,26 +183,36 @@ private static void WriteTime(FileStream fs, TimeCode timeCode)
...
+            }
+            var buffer = new byte[UltechId.Length];
+            using (var fs = new FileStream(fileName, FileMode.Open, FileAccess.Read, FileShare.Read))
+            {
+                int read = fs.Read(buffer, 0, UltechId.Length);
+                if (read != UltechId.Length)
                 {
-                    if (fileName.EndsWith(".ult", StringComparison.OrdinalIgnoreCase) || fileName.EndsWith(".uld", StringComparison.OrdinalIgnoreCase)) //  drop frame is often named uld, and ult for non-drop
-                    {
-                        byte[] buffer = FileUtil.ReadAllBytesShared(fileName);
```
ivandrofly: We don't need to read the entire content just to  check ~12 bytes... the file size is also already checked!

libse/SubtitleFormats/Ultech130.cs
```diff
@@ -189,26 +183,36 @@ private static void WriteTime(FileStream fs, TimeCode timeCode)
 
         public override bool IsMine(List<string> lines, string fileName)
         {
-            if (!string.IsNullOrEmpty(fileName) && File.Exists(fileName))
+            // invalid extension
+            if (!(fileName.EndsWith(".ult", StringComparison.Ordinal) || fileName.EndsWith(".uld", StringComparison.Ordinal)))
```
ivandrofly: Less expensive first...

PR #2505, opened 2017-08-03
[DCSubtitle] - optimize GetColorStringFromDCinema

ivandrofly:
~Do not merge yet!~

Optimization => ~**16n => n**

ready /cc @niksedk

### DIFF
--- a/libse/SubtitleFormats/DCSubtitle.cs
+++ b/libse/SubtitleFormats/DCSubtitle.cs
@@ -670,32 +670,30 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
             subtitle.Renumber();
         }
 
-        private static string GetColorStringFromDCinema(string p)
+        private static string GetColorStringFromDCinema(string s)
         {
-            string s = p.ToLower().Trim();
-            if (s.Replace("#", string.Empty).
-                Replace("0", string.Empty).
-                Replace("1", string.Empty).
-                Replace("2", string.Empty).
-                Replace("3", string.Empty).
-                Replace("4", string.Empty).
-                Replace("5", string.Empty).
-                Replace("6", string.Empty).
-                Replace("7", string.Empty).
-                Replace("8", string.Empty).
-                Replace("9", string.Empty).
-                Replace("a", string.Empty).
-                Replace("b", string.Empty).
-                Replace("c", string.Empty).
-                Replace("d", string.Empty).
-                Replace("e", string.Empty).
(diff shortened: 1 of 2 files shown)


libse/SubtitleFormats/DCSubtitle.cs
```diff
@@ -670,45 +668,41 @@ public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string
...
-                Replace("d", string.Empty).
-                Replace("e", string.Empty).
-                Replace("f", string.Empty).Length == 0)
+            if (string.IsNullOrWhiteSpace(s))
+            {
+                return s;
+            }
+            if (s[0] == '#')
+            {
+                s = s.Substring(1);
```
niksedk: Remove(0, 1) is better as it will not crash with the string "#"

libse/SubtitleFormats/DCSubtitle.cs
```diff
@@ -670,45 +668,41 @@ public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string
...
         private static TimeCode GetTimeCode(string s)
         {
-            var parts = s.Split(':', '.', ',');
-
-            int milliseconds = int.Parse(parts[3]) * 4; // 000 to 249
-            if (s.Contains('.'))
-                milliseconds = int.Parse(parts[3].PadRight(3, '0'));
+            var tokens = s.Split(':', '.', ',');
+            // 000 to 249
+            int milliseconds = s.Contains('.') ? int.Parse(tokens[3].PadRight(3, '0')) : int.Parse(tokens[3]) * 4;
```
niksedk: please don't do multiple stuff per pr

ivandrofly: Okay, I'll send them in small packs

ivandrofly:
> Remove(0, 1) is better as it will not crash with the string "#"

Neither will `"#".Substring(1);` :)

ivandrofly:
Updated, remove all refacts / clean up... /cc @niksedk

niksedk:
thx, I thought`"#".Substring(1);` would crash... but I guess it does not crash as substring(1,0) could be valid? I do like Remove(0, 1) better as it more logical.

ivandrofly:
This actually be simplified, so no `char[] chars` is needed

```c#
for (int i = s.Length - 1; i >= 0; i--)
{
	if (!CharUtils.IsHexadecimal(s[i]))
	{
		return s;
	}
}
return "#" + s;
```

PR #2506, opened 2017-08-05
[configuration] - minor optimization

ivandrofly:
ready /cc @niksedk

### DIFF
--- a/libse/Configuration.cs
+++ b/libse/Configuration.cs
@@ -178,7 +178,7 @@ namespace Nikse.SubtitleEdit.Core
                     // though advertised, this code page is not supported
                 }
             }
-            return encodings.ToArray();
+            return encodings;
         }
 
     }


libse/Configuration.cs
```diff
@@ -171,14 +171,14 @@ private static IEnumerable<Encoding> GetAvailableEncodings()
...
                 {
-                    encodings.Add(Encoding.GetEncoding(ei.CodePage));
+                    encodings.Add(ei.GetEncoding());
                 }
                 catch
                 {
                     // though advertised, this code page is not supported
                 }
             }
-            return encodings.ToArray();
```
ivandrofly: **.ToArray()** is an O(n) operation

libse/Configuration.cs
```diff
@@ -171,14 +171,14 @@ private static IEnumerable<Encoding> GetAvailableEncodings()
             {
                 try
                 {
-                    encodings.Add(Encoding.GetEncoding(ei.CodePage));
```
niksedk: And why this?

PR #2521, opened 2017-08-14
Update es-ES.xml from 3.2 to 3.5.3

paconaranjo:


niksedk:
Thx for the updates :)
This one seems to conflict with your previous pr... what is correct?

paconaranjo:
I made the mistake of editing twice in a row. You can delete it. I correct it again based on the current one. Thank you.

PR #2539, opened 2017-09-01
[CommandLineConvert] - Display cwd after console is detached.

ivandrofly:


src/Logic/CommandLineConvert.cs
```diff
@@ -1263,17 +1263,11 @@ private static string FormatOutputFileNameForBatchConvert(string fileName, strin
         private static void DetachedConsole(string cwd)
         {
             _stdOutWriter?.Close();
-
             if (!IsWindows)
             {
                 return;
             }
-
-            if (_stdOutWriter == null)
```
ivandrofly: Will never be null...

PR #2644, opened 2017-11-25
[Namelist] - Update

ivandrofly:


ivandrofly:
Updated... Rename `Contains` => `TryAdd` to avoid confusion

PR #2655, opened 2017-11-29
Closes #2654

### LINKED ISSUE #2654: [FindReplaceDialogHelper] - Bug
counting words returns incorrect results, e.g; if text  = `a` then you try to count using regex with pattern (a\d?) it will return 0

ivandrofly:


src/Logic/FindReplaceDialogHelper.cs
```diff
@@ -253,31 +253,27 @@ public int FindCount(Subtitle subtitle, bool wholeWord)
...
                     return count;
                 }
                 _regEx = new Regex(_findText);
             }
-
-            // count matches
+            StringComparison comparison = FindReplaceType.FindType == FindType.Normal ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
             foreach (var p in subtitle.Paragraphs)
             {
-                if (p.Text.Length < FindText.Length)
```
ivandrofly: The problem was in this line :)

niksedk:
Nice :)

In future PRs do try not mix bug fixes with refactoring!

PR #2663, opened 2017-12-08
[main] - avoid redundant process when alternate subtitle contains no paragraphs

ivandrofly:


src/Forms/Main.cs
```diff
@@ -3589,7 +3589,7 @@ private DialogResult SaveSubtitle(SubtitleFormat format)
                 string allText = sub.ToText(format);
 
                 // Seungki begin
-                if (_splitDualSami && _subtitleAlternate != null)
+                if (_splitDualSami && _subtitleAlternate?.Paragraphs.Count > 0)
                 {
                     var s = new Subtitle(_subtitle);
```
ivandrofly: if _subtitleAlternate.Paragraph.Count = 0 new Subtitle(_subtitle) would still run ~n+ operations when copying paragraphs to the new instance of subtitlie "s"

PR #2666, opened 2017-12-10
[main] - use linq instead

ivandrofly:


niksedk:
A bit less code - but also much slower

PR #2723, opened 2018-01-26
[checkforupdateshelper] - minor update

ivandrofly:


ivandrofly:
I mailed you it's been a week already cloud you please check you spam box (could be there cause I forgot to specify mail title) /cc @niksedk

PR #2725, opened 2018-01-27
Adding Resync multiple paragraphs feature.

ShariatPanah:
it was missing some sort of "RESYNC MULTIPLE PARAGRAPHS AT ONCE", I added this, but had to comment out some line of code in SubtitleListView1SelectedIndexChange method, it was resetting the first selected item in SubtitleListView1 that just synced. anyway, here's some pics of this feature:

image

image

and the results:
image


aaaxx:
Unless I'm missing something here, the option already exists: Menu bar > Synchronization > Adjust all times (show earlier/later) > Selected lines only.

It even automatically picks "Selected lines only" option if you have more than one line selected.

ShariatPanah:
@aaaxx oh my .... yes you're right, it was my mistake... sorry :)

PR #2733, opened 2018-02-01
[commandlineconvert] - fix/update.

ivandrofly:


src/Logic/CommandLineConvert.cs
```diff
@@ -60,72 +38,78 @@ private static void Write(string s)
                 {
                     _stdOutWriter = new StreamWriter(stdout) { AutoFlush = true };
                 }
-                NativeMethods.AttachConsole(ATTACH_PARENT_PROCESS);
+                else
```
ivandrofly: we don't need to attach when output is being redirected

PR #2792, opened 2018-02-27
[Synchronisation] "One button" sync to video

GulBroz:
I use subtitleedit a lot to resync subtitles to video, I first started to do it manually:
- Load subtitle & video
- Play video
- Stop when you hear the start of an out-of-sync line
- Calculate time offset between video and subtitle
- Resync using "Show earlier/latter" buttons in "Adjust all times" dialog box
- Check, adjust, repeat...

As this is a rather painful process, I added a new button in "Adjust all times" dialog box, that does it all automatically by clicking "Video sync" button:
adjustalltimes_dialog_box
- It computes the time offset
- Takes the value in "HH:MM:SS:ms" as a "reaction time" for adjustment (see it as the time it takes you to react to what you hear, click the video pause button, player to pause...). I found that ½ second is perfect, but it can be lowered if you're faster than me...
- Adjust "All lines/Selected line only/Selected line and forward" to the exact position the video is paused, depending on what is selected.

If multiple lines are selected, the synchronization is made based on the 1st line of the selection.

As it's using the same internal mechanism as "Show earlier/later" buttons, operations are registered in history, so you can use Undo/Redo.


marb99:
this is good, thanks

PR #2815, opened 2018-03-08
Minor clean-up

ivandrofly:


niksedk:
nice :)

Perhaps you could delete some of your old branches?

PR #2816, opened 2018-03-09
Update DarkTheme

ivandrofly:


src/Logic/DarkTheme.cs
```diff
@@ -56,19 +57,30 @@ private static List<T> GetSubControls<T>(Control c)
...
-        public static void SetDarkTheme(Control form, int iterations = 5)
+        private static bool _isConfigUpdated;
+
+        public static void SetDarkTheme(Control ctrl, int iterations = 5)
         {
-            Configuration.Settings.General.SubtitleBackgroundColor = BackColor;
-            Configuration.Settings.General.SubtitleFontColor = ForeColor;
-            Configuration.Settings.VideoControls.WaveformBackgroundColor = BackColor;
-            Configuration.Settings.VideoControls.WaveformGridColor = Color.FromArgb(62, 62, 60);
+            if (!_isConfigUpdated)
```
ivandrofly: The **SetDarkTheme** will be called recursively, so add guard to avoid redundant re-assignment.

src/Logic/DarkTheme.cs
```diff
@@ -56,19 +57,30 @@ private static List<T> GetSubControls<T>(Control c)
             return menus.Cast<T>().ToList();
         }
 
-        public static void SetDarkTheme(Control form, int iterations = 5)
+        private static bool _isConfigUpdated;
+
+        public static void SetDarkTheme(Control ctrl, int iterations = 5)
```
ivandrofly: Rename **form** => **ctrl** see: https://github.com/SubtitleEdit/subtitleedit/pull/2816/files#diff-4e0c1aa2089eeedfb01eba69172efb07L69

it doesn't really make sense saying ` if (form is Form)` so **form** is now renamed to **ctrl** `if (ctrl is Form)`

src/Logic/DarkTheme.cs
```diff
@@ -155,36 +165,24 @@ public static void SetDarkTheme(Control form, int iterations = 5)
             }
         }
 
-        private static void C_Paint(object sender, PaintEventArgs e)
```
ivandrofly: Unused...

src/Logic/DarkTheme.cs
```diff
@@ -215,10 +212,15 @@ private static void lv_DrawColumnHeader(object sender, DrawListViewColumnHeaderE
                 e.Graphics.FillRectangle(b, e.Bounds);
 
             var strFormat = new StringFormat();
-            if (e.Header.TextAlign == HorizontalAlignment.Center)
-                strFormat.Alignment = StringAlignment.Center;
-            else if (e.Header.TextAlign == HorizontalAlignment.Right)
-                strFormat.Alignment = StringAlignment.Far;
+            switch (e.Header.TextAlign)
```
ivandrofly: More readable 😉

PR #2823, opened 2018-03-12
Check Some PVS-STUDIO fixes

LinqManJr:
pvs-studio fixes

### DIFF
--- a/libse/RegexUtils.cs
+++ b/libse/RegexUtils.cs
@@ -176,8 +176,8 @@ namespace Nikse.SubtitleEdit.Core
             }
             try
             {
-                return Regex.IsMatch(string.Empty, testPattern);
-                //return true;
+                Regex.Match(string.Empty, testPattern);
+                return true;
             }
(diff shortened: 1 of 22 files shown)


libse/RegexUtils.cs
```diff
@@ -176,8 +176,8 @@ public static bool IsValidRegex(string testPattern)
             }
             try
             {
-                Regex.Match(string.Empty, testPattern);
-                return true;
+                return Regex.IsMatch(string.Empty, testPattern);
+                //return true;
```
niksedk: The old code is correct - it does not need to be a match but only not throw an `ArgumentException`.
Luckily a unit test caught this: https://ci.appveyor.com/project/SubtitleEdit/subtitleedit/build/1.0.7288

niksedk:
@LinqManJr: Looks nice :)
The fixes in LambdaCap.cs seems to cause an extra blank line - could you fix that or revert changes in LambdaCap.cs?

PR #2931, opened 2018-05-14
Update Pac.cs

cedinu:
For PAC2000 flavour at least, PAC files with inserts extends numbering to 0x6F.
0xFE can appear elsewhere than at the end of header; 0xFE in that position and 0x03 at the end of the header is more likely to be correct.
This patch is only for reading PACs.


niksedk:
Sorry for the late reply... could you provide a sample file so I can verify?

niksedk:
Closed for now - please provide more/better info/reason if you want this re-opened

PR #3018, opened 2018-08-12
Update and fix es_MX.xml

paconaranjo:


src/Languages/es-MX.xml
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
niksedk: thx for the updates - could you keep the utf-bom header, so we are certain that SE will always read these files with correct encoding?

paconaranjo: Works fine, but you tell me. I did it with your XML translator tool.

PR #3089, opened 2018-09-09
Create hrv_NoBreakAfterList.xml

diomed:


diomed:
I don't see any good reason why other languages should have this, but my shouldn't. 

please merge this.

Issue #3112, opened 2018-09-19
Fix common errors - Break long line and ellipsis


SvenErik1968:
It looks like SE (v3.5.7 build 1) have some issues with a proper ellipsis "…" and with breaking long lines.

Example using ellipsis:
```
1
00:00:28,506 --> 00:00:33,171
- Come on, I…
- Here, let's try this one, right over here.
```
will end up like this:
```
1
00:00:28,506 --> 00:00:33,171
- Come on, I… - Here, let's try
this one, right over here.
```
while an identical line using 3 periods instead of the ellipsis will not be changed. This will also not be caught by the _Fix dialogs on one line_.

[Ellipsis_error_break_long_line.srt.txt](https://github.com/SubtitleEdit/subtitleedit/files/2398019/Ellipsis_error_break_long_line.srt.txt)

SvenErik1968:
I am afraid I still get the line break after the ellipsis and end up with with two dialog's on one line with the new beta (build 15).

![snip_20180919225429](https://user-images.githubusercontent.com/10983802/45780947-11b83500-bc5f-11e8-89ad-4a52104e7b77.png)

niksedk:
That's probably because the line is too long - go to Options -> Settings if you want to allow longer lines - or split to two lines.

SvenErik1968:
But why does it allows two dialogs on one line when the first dialog ends with an ellipsis, but handles the case where the first dialog ends with 3 periods correctly?

PR #3115, opened 2018-09-20
Fixes for #3112

ivandrofly:


ivandrofly:
> don't work, not detect long line

What do you mean?

marb99:
look to image above. "nothing to fix but a few...."

is a long line but don´t fix anything

ivandrofly:
> look to image above. "nothing to fix but a few...."
> is a long line but don´t fix anything
> I downloaded subtitleedit-master made the changes like your fix and tested the result is don´t fix anything, does not even detect that a long line

That is the purpose, a paragraph that is a dialog (as the one above) shouldn't be changed...

marb99:
> 
> 
> > look to image above. "nothing to fix but a few...."
> > is a long line but don´t fix anything
> > I downloaded subtitleedit-master made the changes like your fix and tested the result is don´t fix anything, does not even detect that a long line
> 
> That is the purpose, a paragraph that is a dialog (as the one above) shouldn't be changed...

Ohhh, oki, but this wrong in my opinion, the second paragraph is a long line cause trouble on display eats letters because its a long line.

PR #3336, opened 2019-02-03
Optimize backup-cleanup process

ivandrofly:


src/Forms/RestoreAutoBackup.cs
```diff
@@ -166,30 +168,26 @@ private void linkLabelOpenContainingFolder_LinkClicked(object sender, LinkLabelL
...
                     {
-                        var name = Path.GetFileName(fileName);
-                        if (RegexFileNamePattern.IsMatch(name) && Convert.ToDateTime(name.Substring(0, 10), CultureInfo.InvariantCulture) < targetDate)
+                        try
                         {
-                            File.Delete(fileName);
-                            filesDeleted++;
-                            if (filesDeleted > maxCount)
+                            var name = Path.GetFileName(fileName);
+                            if (RegexFileNamePattern.IsMatch(name) && Convert.ToDateTime(name.Substring(0, 10), CultureInfo.InvariantCulture) <= targetDate)
```
ivandrofly: Here we want to delete backups after a specific month:
`Convert.ToDateTime(name.Substring(0, 10), CultureInfo.InvariantCulture) <= targetDate`

The old code was `Convert.ToDateTime(name.Substring(0, 10), CultureInfo.InvariantCulture) == targetDate` which I think is wrong...

src/Forms/RestoreAutoBackup.cs
```diff
@@ -166,30 +168,26 @@ private void linkLabelOpenContainingFolder_LinkClicked(object sender, LinkLabelL
 
         public static void CleanAutoBackupFolder(string autoBackupFolder, int autoBackupDeleteAfterMonths)
         {
-            const int maxCount = 100; // to avoid locking computer
```
ivandrofly: Not needed 'cause we are running on a worker thread.

src/Forms/RestoreAutoBackup.cs
niksedk: The old code was ´<´... (not `==`)

ivandrofly: > The old code was ´<´... (not `==`)
Exactly, typo

PR #3350, opened 2019-02-09
Update cleanup

ivandrofly:


libse/SubtitleFormats/VocapiaSplit.cs
```diff
@@ -108,6 +108,7 @@ public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string
                     text = text.Replace("<s/>", Environment.NewLine);
                     text = text.Replace("  ", " ");
                     var p = new Paragraph(text, ParseTimeCode(start), ParseTimeCode(end));
+                    p.Number = subtitle.Paragraphs.Count + 1;
```
ivandrofly: @niksedk This would be a good optimization would avoid us calling `subtitle.Renumber()`

PR #3366, opened 2019-02-13
Don't force paragraphs copying

ivandrofly:


libse/Forms/SplitLongLinesHelper.cs
```diff
@@ -42,8 +42,7 @@ public static bool QualifiesForSplit(string text, int singleLineMaxCharacters, i
 
         public static Subtitle SplitLongLinesInSubtitle(Subtitle subtitle, int totalLineMaxCharacters, int singleLineMaxCharacters)
         {
-            var splittedSubtitle = new Subtitle(subtitle);
-            splittedSubtitle.Paragraphs.Clear();
```
ivandrofly: will save us from this

ivandrofly:
**Codacy/PR Quality Review** is complaining that named params should be replaced with constructor overload

PR #3370, opened 2019-02-15
Better SplitLongLinesInSubtitle

ivandrofly:


libse/Forms/SplitLongLinesHelper.cs
```diff
@@ -43,105 +43,113 @@ public static bool QualifiesForSplit(string text, int singleLineMaxCharacters, i
...
-                                {
-                                    var minMsBtwnLnBy2 = Configuration.Settings.General.MinimumMillisecondsBetweenLines / 2;
-                                    int spacing1 = minMsBtwnLnBy2;
-                                    int spacing2 = minMsBtwnLnBy2;
-                                    if (Configuration.Settings.General.MinimumMillisecondsBetweenLines % 2 == 1)
-                                    {
-                                        spacing2++;
-                                    }
-
-                                    double duration = p.Duration.TotalMilliseconds / 2.0;
```
ivandrofly: I think this is really wrong, we can't assume both line with have same about of characters... this problem is handle in this PR.

libse/Forms/SplitLongLinesHelper.cs
```diff
@@ -43,105 +43,113 @@ public static bool QualifiesForSplit(string text, int singleLineMaxCharacters, i
...
+            return text[closeIdx + 1] == '-';
+        }
+
+        private static bool IsTextClosed(string text)
+        {
+            if (string.IsNullOrEmpty(text) || text.Length == 0)
+            {
+                return false;
+            }
+            string textNoTags = HtmlUtil.RemoveHtmlTags(text);
```
ivandrofly: Handles if text ends with any of the supported html tags

niksedk:
It would be really nice if you could add some unit tests :)

ivandrofly:
> It would be really nice if you could add some unit tests :)

Done ✅

src/Test/Logic/SplitLongLinesHelperTest.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            var procSubtitle = SplitLongLinesHelper.SplitLongLinesInSubtitle(_subtitle, _maxLineLength * 2, _maxLineLength);
+
+            Assert.AreEqual("We have never been to Asia,\r\nnor have we visited Africa.", procSubtitle.Paragraphs[0].Text);
+            Assert.AreEqual("We have never been to Asia,\r\nnor have we visited Africa.", procSubtitle.Paragraphs[1].Text);
+            Assert.AreEqual(_subtitle.Paragraphs[2].Text, procSubtitle.Paragraphs[2].Text);
+
+            Assert.AreNotEqual(_subtitle.Paragraphs.Count, procSubtitle.Paragraphs.Count);
+
+            // too long (dialog)
```
niksedk: dialog?

src/Test/Logic/SplitLongLinesHelperTest.cs
```diff
@@ -0,0 +1,74 @@
...
+            Assert.AreNotEqual("Sometimes, all you need to do is completely make an ass?", procSubtitle.Paragraphs[3].Text);
+            Assert.AreNotEqual("Of yourself and laugh it off to realise that life isn’t so bad after all.", procSubtitle.Paragraphs[4].Text);
+
+            // too long
+            Assert.AreNotEqual("Sometimes, all you need to do is completely make an ass?", procSubtitle.Paragraphs[5].Text);
+            Assert.AreNotEqual("Of yourself and laugh it off to realise that life isn’t so bad after all.", procSubtitle.Paragraphs[5].Text);
+        }
+    }
+}
```
niksedk: Some tests of the new calculation milliseconds per char :)

src/Test/Logic/SplitLongLinesHelperTest.cs
ivandrofly: > dialog?

https://github.com/SubtitleEdit/subtitleedit/pull/3370/files#diff-80dfd8f5edf891672e3bdfef28befd5eR32

src/Test/Logic/SplitLongLinesHelperTest.cs
ivandrofly: > Some tests of the new calculation milliseconds per char :)

Test added

PR #3382, opened 2019-02-18
Update start with uppercase after a symbol

ivandrofly:
## TODO 👷 
- [x] FixStartWithUppercaseLetterAfterColon
- [x] FixStartWithUppercaseLetterAfterPeriodInsideParagraph
- [ ] FixStartWithUppercaseLetterAfterParagraph


ivandrofly:
What do you think above making all these classes extend a base class that will implement all the common logics they shared?  /cc @niksedk

PR #3399, opened 2019-02-22
Switching tabs: minor optimization

ivandrofly:


src/Forms/Main.cs
```diff
@@ -6087,29 +6087,29 @@ private void ReloadFromSourceView()
...
 
                     int index = 0;
-                    foreach (object obj in comboBoxSubtitleFormats.Items)
+                    foreach (string formatName in comboBoxSubtitleFormats.Items)
                     {
-                        if (obj.ToString() == format.FriendlyName)
+                        if (formatName == format.FriendlyName)
                         {
                             comboBoxSubtitleFormats.SelectedIndex = index;
+                            break;
```
ivandrofly: break as soon as the format is found

src/Forms/Main.cs
```diff
@@ -6087,29 +6087,29 @@ private void ReloadFromSourceView()
                 if (!string.IsNullOrWhiteSpace(textBoxSource.Text))
                 {
                     SubtitleFormat format = GetCurrentSubtitleFormat();
-                    var list = textBoxSource.Lines.ToList();
-                    format = new Subtitle().ReloadLoadSubtitle(list, null, format);
+                    var lines = new List<string>(textBoxSource.Lines);
+                    format = new Subtitle().ReloadLoadSubtitle(lines, null, format);
```
ivandrofly: 2n => n

src/Forms/Main.cs
```diff
@@ -22680,7 +22680,7 @@ private void TabControlSubtitleSelecting(object sender, TabControlCancelEventArg
                     return;
                 }
 
-                SubtitleFormat format = new Subtitle().ReloadLoadSubtitle(textBoxSource.Lines.ToList(), _fileName, currentFormat);
+                SubtitleFormat format = new Subtitle().ReloadLoadSubtitle(new List<string>(textBoxSource.Lines), _fileName, currentFormat);
```
ivandrofly: 2n => n

src/Forms/Main.cs
niksedk: Ran some tests (1000 iterations in release mode without VS) and `textBoxSource.Lines.ToList();` and `new List<string>(textBoxSource.Lines)` are *very* close - ToList seems a little faster with large subtitles and new List seems a little faster with small subtitles, so not really a clear winner IMO.

src/Forms/Main.cs
niksedk: :)

PR #3400, opened 2019-02-22
PhoenixSubtitle: Update + bugfix

ivandrofly:
The main bug can be produced by changing the format from default (**Subrip**) to **PhoenixSubtitle** in **combobox subtitle format** and go to **sourceview**, if you try to switch back to listview **SE** won't be able to find the correct format 'cause no **file-name** with extension .**pjs** is been supplied.

### DIFF
--- a/libse/SubtitleFormats/PhoenixSubtitle.cs
+++ b/libse/SubtitleFormats/PhoenixSubtitle.cs
@@ -27,33 +27,43 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
 
         public override bool IsMine(List<string> lines, string fileName)
         {
-            if (fileName?.EndsWith(".pjs", StringComparison.OrdinalIgnoreCase) == false)
+            if (fileName?.EndsWith(".pjs", StringComparison.OrdinalIgnoreCase) == true)
             {
-                return false;
+                return true;
             }
-            return base.IsMine(lines, fileName);
+
+            var subtitle = new Subtitle();
+            LoadSubtitle(subtitle, lines, fileName);
+            return subtitle.Paragraphs.Count > _errorCount;
         }
 
         public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string fileName)
         {
             subtitle.Paragraphs.Clear();
             _errorCount = 0;
-            var paragraph = new Paragraph();
             for (int i = 0; i < lines.Count; i++)
             {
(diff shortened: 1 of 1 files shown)


libse/SubtitleFormats/PhoenixSubtitle.cs
```diff
@@ -27,33 +27,43 @@ public class PhoenixSubtitle : SubtitleFormat
 
         public override bool IsMine(List<string> lines, string fileName)
         {
-            if (fileName?.EndsWith(".pjs", StringComparison.OrdinalIgnoreCase) == false)
```
ivandrofly: This was preventing detecting when there is no file supplied which mostly happen when the format is changed from Format combobox in main-window

libse/SubtitleFormats/PhoenixSubtitle.cs
```diff
@@ -67,10 +77,8 @@ public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string
                             line = line.Trim(TrimChars);
                         }
 
-                        paragraph.Number = i + 1;
```
ivandrofly: Don't assume all the lines will be parsed correctly (we now allow ignoring white-line), so the `i` is not a good candidate here. 😞

libse/SubtitleFormats/PhoenixSubtitle.cs
```diff
@@ -27,33 +27,43 @@ public class PhoenixSubtitle : SubtitleFormat
...
                 string line = lines[i].Trim();
-                Match match = null;
-                if (line.Length >= 4)
+
+                // too short line
+                if (line.Length < 4)
                 {
-                    match = RegexTimeCodes.Match(line);
+                    _errorCount++;
+                    continue;
```
ivandrofly: ignore white-line right away

niksedk:
Nice catch :)
The fileName parameter in the IsMine method was actually wrong after changing format in the toolbar dropdown... see Main.cs here: https://github.com/SubtitleEdit/subtitleedit/commit/bd52083f1f61efebf8eec138664d6e3727d4bdae#diff-2ec591b6c56d41734fb35d871b4eca07L22683

I've just set fileName parameter to ´null´ which is a simple fix - but perhaps we should use the `_fileName` if `_converted == false && File.Exists(_fileName)`... ?

PR #3406, opened 2019-02-24
namelist: optimization

ivandrofly:


libse/Dictionaries/NameList.cs
```diff
@@ -282,12 +282,24 @@ public bool IsInNamesMultiWordList(string input, string word)
...
 
-            foreach (var n in name.Contains(' ') ? _namesMultiList : _namesList)
+            // name must contain atleast two chars
+            if (name.Length < 2)
+            {
+                return false;
+            }
+
+            var lookupCollection = name.Contains(' ') ? _namesMultiList : _namesList;
+            if (lookupCollection.Contains(name)) // O(1)
```
ivandrofly: will save of from running O(n) when `name` exists in `name-list`  w/ right casing

PR #3441, opened 2019-03-07
Potential bug-fix

ivandrofly:


libse/SubtitleFormats/Pac.cs
```diff
@@ -1546,18 +1547,7 @@ private static byte[] GetW16Bytes(string text, byte alignment, int encoding)
             return result.ToArray();
         }
 
-        private static bool OnlyAnsi(string line)
-        {
-            string latin = Utilities.AllLettersAndNumbers + " .!?/%:;=()#$'&\"";
```
ivandrofly: There are some character that aren't in **ANSI** range that are present in `Utilities.AllLettersAndNumbers`

libse/CharUtils.cs
```diff
@@ -18,5 +18,10 @@ public static class CharUtils
         /// Checks if character is between A-Z or a-z
         /// </summary>
         public static bool IsEnglishAlphabet(char ch) => ch >= 'A' && ch <= 'z' && (ch <= 'Z' || ch >= 'a');
+
+        /// <summary>
+        /// Check whether the character is within ANSI range (0 t0 0xff)
```
aaaxx: Typo: zero instead of "o" in "to"

PR #3452, opened 2019-03-11
main: remove redundant code

ivandrofly:


niksedk:
Even looks like a bug...

PR #3466, opened 2019-03-15
Don't display paragraph.extra when updating bridge-gaps

ivandrofly:
The SubtitleListView's Extra column is being used to display information of bridge gaps update, mixing it like that can cause confusion


ivandrofly:
image

PR #3467, opened 2019-03-15
OcrFixEngine: minor optimization

ivandrofly:


src/Logic/OCR/OcrFixEngine.cs
```diff
@@ -1350,21 +1350,8 @@ public string FixUnknownWordsViaGuessOrPrompt(out int wordsNotFound, string line
...
-                        foreach (var slashedWord in slashedWords)
-                        {
-                            if (slashedWord.Length < 2)
-                            {
-                                allSlashedCorrect = false;
-                            }
-
-                            if (allSlashedCorrect && !(DoSpell(slashedWord) || IsWordKnownOrNumber(slashedWord, line)))
-                            {
-                                allSlashedCorrect = false;
```
ivandrofly: No `break` after the allSlashedCorrect changes state (the state will never change again inside loop)

PR #3472, opened 2019-03-18
remove redundant methods invoke from SubtitleListView

ivandrofly:


src/Forms/DurationsBridgeGaps.cs
```diff
@@ -126,6 +124,7 @@ private void GeneratePreviewReal()
             FixedCount = Core.Forms.DurationsBridgeGaps.BridgeGaps(_fixedSubtitle, minMsBetweenLines, radioButtonDivideEven.Checked, (double)numericUpDownMaxMs.Value, fixedIndexes, _dic);
 
             SubtitleListview1.Fill(_fixedSubtitle);
```
ivandrofly: `Fill` calles Items.Clear() + BeginUpdate() / EndUpdate()

src/Forms/DurationsBridgeGaps.cs
```diff
@@ -145,7 +144,6 @@ private void GeneratePreviewReal()
             {
                 SubtitleListview1.SetBackgroundColor(index, Color.LightGreen);
             }
-
             SubtitleListview1.EndUpdate();
```
niksedk: You probably should remove "SubtitleListview1.EndUpdate();" also

src/Forms/DurationsBridgeGaps.cs
niksedk: I think this is okay, it clears the list view, and handles a lot of listview stuff (BeginUpdate can be nested I think)

ivandrofly: > I think this is okay, it clears the list view, and handles a lot of listview stuff (BeginUpdate can be nested I think)

Sure, that is why we don't need remove lines above! I don't know if you noticed, I didn't remove BeginUpdate / EndUpdate.

PR #3475, opened 2019-03-19
Main: Context menu opening

ivandrofly:
Fix issue when opening context menu for Nuendo subtitle format it was showing SubStationAlpha features specific

Ps: Nuendo subtitle format is mistakenly being used where **AdvancedSubStationAlpha** / **SubStationAlpha** are used, this is not the only place I found this problem... there is another patch I have been working on the same problem can be visible from there as well.

### DIFF
--- a/src/Forms/Main.cs
+++ b/src/Forms/Main.cs
@@ -7237,12 +7237,18 @@ namespace Nikse.SubtitleEdit.Forms
                 // actor
                 foreach (var p in _subtitle.Paragraphs)
                 {
-                    if (!string.IsNullOrEmpty(p.Actor) && !actors.Contains(p.Actor))
+                    if (string.IsNullOrEmpty(p.Actor))
                     {
-                        actors.Add(p.Actor);
+                        continue;
                     }
-                    actors.Sort();
+                    if (actors.Contains(p.Actor))
+                    {
+                        continue;
+                    }
+                    actors.Add(p.Actor);
                 }
+                actors.Sort();
+
                 setActorForSelectedLinesToolStripMenuItem.DropDownItems.Clear();
                 foreach (var actor in actors)
                 {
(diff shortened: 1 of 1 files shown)


src/Forms/Main.cs
```diff
@@ -7402,18 +7408,43 @@ private void ContextMenuStripListviewOpening(object sender, System.ComponentMode
                     toolStripMenuItemWebVTT.DropDownItems.Add(_language.Menu.ContextMenu.WebVTTRemoveVoices, null, WebVTTRemoveVoices);
                 }
             }
-            else if ((format.Name == "Nuendo" && SubtitleListview1.SelectedItems.Count > 0))
```
ivandrofly: comparing format.Name with `Nuendo` would always return false, there is no subtitle format named `Nuendo` but `Csv Nuendo`

src/Forms/Main.cs
```diff
@@ -7402,18 +7408,43 @@ private void ContextMenuStripListviewOpening(object sender, System.ComponentMode
...
+                    setActorForSelectedLinesToolStripMenuItem.DropDownItems.Add(tss);
+                }
+                setActorForSelectedLinesToolStripMenuItem.DropDownItems.Add(_language.Menu.ContextMenu.NewActor, null, SetNewActor);
+                if (actors.Count > 0)
+                {
+                    setActorForSelectedLinesToolStripMenuItem.DropDownItems.Add(_language.Menu.ContextMenu.RemoveActors, null, RemoveActors);
+                }
+
                 toolStripMenuItemAssStyles.Visible = false;
-                setStylesForSelectedLinesToolStripMenuItem.Text = _language.Menu.ContextMenu.NuendoSetStyle;
```
ivandrofly: for Nuendo subtitle format, we want to update the `Actor` (**Character**) instead of style property

PR #3482, opened 2019-03-22
main-unbreak; fix phantom history item while unbreaking text.

ivandrofly:


src/Forms/Main.cs
```diff
@@ -10964,70 +10964,83 @@ private void Unbreak(bool removeNewLineOnly = false)
             var numberOfNewLines = Utilities.CountTagInText(startText, Environment.NewLine);
             textCaretPos -= numberOfNewLines;
 
+            MakeHistoryForUndo(_language.BeforeRemoveLineBreaksInSelectedLines);
+            // UNDONE: How to make history item for _subtitleAlternate?
```
ivandrofly: ? @niksedk

niksedk: makehistory handles alternate as well

PR #3523, opened 2019-04-06
switch to use zlib.net-multiplatform

NickZ:
This is updated more frequently and is able to be used across .net Framework, .net Core, and .net Standard


niksedk:
Nice :)

Hunspell is probably also a candidate...

NickZ:
@niksedk I actually attempted to switch from NHunspell to WeCantSpell.Hunspell; It did not work as well as NHunspell with the dictionaries in SubtitleEdit. In particular, it seemed to have problems with plurals of common words. However, NHunspell does work fine with .net Core since it doesn't rely on anything but System IO, and all it needs is an updated version pushed to nuGet that targets .net Core. I'm working on doing that.

PR #3524, opened 2019-04-07
[WIP] Working .net Core Port

NickZ:
This is a .net Core port that I am working on.

I have created new projects for both LibSE and SubtitleEdit, such that both the Core versions and the Framework versions will both compile from the same codebase. 

Both versions compile and run on Windows.

I used [.net Core 3 preview 3](https://dotnet.microsoft.com/download/dotnet-core/3.0) and [Visual Studio 2019 Preview](https://visualstudio.microsoft.com/downloads/).

I have had to do the following to get this to work:

- I switched to zlib.net-multiplatform instead of the zlib.net NuGet package.
- I removed the Microsoft Translate API v2, since it was no longer and use and will cease to function at the end of this month, anyway.
- I replaced the SeNetworkService Web Reference with a Service Reference, since the Web Reference classes do not work in .net Core
- Thread.Abort() is not present in .net Core, so it had to be removed from the spell check.

The following things do not work:
- Using a proxy with SeNetworkService does not work yet, because I do not know how to use them with the BasicHTTPBinding.

There are some things preventing this from being multi-platform:
- The biggest is Winforms; while they have been ported to .net Core, they do not work on Linux and Macintosh. Perhaps we could rely on the Mono winforms library?
- The Service Reference relies on Windows-only NuGet packages on .net Core. WCF network references are deprecated; Ideally, the SeNetworkService would be turned into a REST JSON API.


GerardSoleCa:
Hi, thanks for your efforts to migrate SE to .netcore :D

I'm only willing to use the CLI version of SubtitleEdit, is there a chance to be able to run it using netcore in Linux? I really don't need the windows forms, just the subtitle conversion.

Thanks for this great effort, hope to see it merged when finished :)

NickZ:
@GerardSoleCa If you're only using the CLI version, then I don't think there's anything holding up a linux port of that. I don't currently know how to compile this on Linux, though.

GerardSoleCa:
@NickZ I already tried, but the main problem is still the Windows Forms, SubtitleEdit references them, and once they are loaded by dotnet, it crashes. I was trying to extract the CLI part from the SubtitleEdit, so a secondary binary could be compiled and used, but things are too tied there, and without a proper VisualStudio (and time) it's being a mess to split the code.

niksedk:
I think this can be closed as LibSE is dotnot core now...

PR #3539, opened 2019-04-14
paragraph: add helper to detect default/empty paragraph

ivandrofly:


libse/Paragraph.cs
```diff
@@ -51,6 +51,8 @@ public TimeCode Duration
 
         public string Bookmark { get; set; }
 
+        public bool IsDefault => StartTime.TotalMilliseconds == 0 && EndTime.TotalMilliseconds == 0 && string.IsNullOrEmpty(Text);
```
ivandrofly: IsEmpty is also a good candidate...

PR #3544, opened 2019-04-15
stringextensions: improve word count.

ivandrofly:
1. Add `-`
2. Only consider word if contains letter


niksedk:
It's simple on purpose

PR #3562, opened 2019-05-07
Refact CommandLineConvert

xylographe:
1. Refact CommandLineConvert
   - Move command line argument details from Main to CommandLineConvert.
   - Use HashSet<> with StringComparer i/o List<>.
   - Use [FlagsAttribute] i/o List<>.
   - Avoid ToList() where feasible.
2. Try to fix Codacy anomalies
3. Allow multiple track numbers in '/track-number' command line option
4. Add IsRunningOnWindows property to Configuration class
   - Keep IsRunningOn* logic in one class only.
   - Do the Directory.Exists() invocations only once.
5. Add FilePathComparer (Unix, Windows, Native)
6. Use FilePathComparer i/o StringComparer in CommandLineConvert
7. Add GetOffset() to CommandLineConvert
8. Apply GetOffset() in CommandLineConvert

### DIFF
--- a/libse/Configuration.cs
+++ b/libse/Configuration.cs
@@ -40,21 +40,56 @@ namespace Nikse.SubtitleEdit.Core
             _settings = new Lazy<Settings>(Settings.GetSettings);
         }
 
-        public static bool IsRunningOnLinux()
+        private const int _platformWindows = 1;
+        private const int _platformLinux = 2;
+        private const int _platformMac = 3;
+        private static int _platform;
+
+        public static bool IsRunningOnWindows
         {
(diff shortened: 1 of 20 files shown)


xylographe:
I'm afraid, in this case, I do not agree with Codacy. Those 3 Exceptions are all caught within the method they were raised. In fact, this makes error handling in CommandLineConvert more consistent.

xylographe:
The CommandLine-parser is looking good! Might be a good choice for a stand-alone CLI version of SE. The big question is, would there be enough interest in a CLI SE to warrant the effort?

In the current GUI/CLI SE, I believe, it might be better to stick with the MSWindows option style. Current cli users will probably object vehemently to such a radical change.

PR #3566, opened 2019-05-08
only match from [0-9] instead of whole unicode digits

ivandrofly:


libse/SubtitleFormats/UnknownSubtitle90.cs
```diff
@@ -55,7 +55,7 @@ public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string
                     {
                         _errorCount++;
                     }
-                    if (text.Length > 0 && char.IsDigit(text[0]))
+                    if (text.Length > 0 && CharUtils.IsDigit(text[0]))
```
ivandrofly: isn't this the intention?

PR #3589, opened 2019-06-02
Update Tar-File reader

ivandrofly:


libse/TarHeader.cs
```diff
@@ -19,8 +19,8 @@ public TarHeader(Stream stream)
         {
             _stream = stream;
             var buffer = new byte[HeaderSize];
-            stream.Read(buffer, 0, HeaderSize);
```
ivandrofly: we need to store position before we start reading...

src/Forms/Ocr/DownloadTesseract302.cs
```diff
@@ -53,10 +53,10 @@ private void wc_DownloadDataCompleted(object sender, DownloadDataCompletedEventA
             {
                 foreach (var th in tr.Files)
                 {
-                    string fn = Path.Combine(dictionaryFolder, th.FileName.Replace('/', Path.DirectorySeparatorChar));
```
ivandrofly: handled when reading file name from stream...

PR #3594, opened 2019-06-03
recent files: don't keep moved, remove, deleted files from recent list

ivandrofly:
Don't keep moved, removed, deleted files in recent list when updating *recent file list*


xylographe:
In Notepad++ this is optional (Settings > Preferences). Guess what, I've turned it of. This doesn't work well in combination with version control software (svn/git), where files may **seem** deleted, but could easily be ‘resurrected’ on the next check out.

ivandrofly:
> Edit: To clarify, I'm not opposed in principle, but I believe it should be optional.

Agreed, but if you review the code you will see that the previous logic is a bit confusing, because it's written like, if the option is not enabled, all the histories will be deleted... it make more sense this way I think.

PR #3608, opened 2019-06-13
Hearing Impaired: Update

ivandrofly:


libse/Forms/RemoveTextForHI.cs
```diff
@@ -1042,12 +1034,31 @@ private void AddWarning()
...
-            if (text.Trim().Replace("mr. ", string.Empty).Replace("mrs. ", string.Empty).Replace("dr. ", string.Empty).Contains(' '))
+            text = text.Trim(' ', '(', ')', '[', '¿', '¡', ']', '?', '!', '{', '}').ToLowerInvariant();
+
+            // when all uppercase don't add warning
+            if (WarningIndex > 0 && text.ToUpperInvariant().Equals(text, StringComparison.Ordinal) == false
+                && RemoveCommonTitles(text).Contains(' '))
             {
-                AddWarning();
+                Warnings?.Add(WarningIndex);
             }
```
xylographe: I think, the comparison in line 1058 will always be false, because `text` has been converted to lower case in line 1055, unless `text` does not contain characters having upper/lower case variants (decimal digits, punctuation, etc.).

ivandrofly: that is correct, I actually tried to change the variable name to ### textLowerCase to avoid the confusion - but didn't end up committing it.

Update on it's way!

PR #3614, opened 2019-06-16
translate: update + optimization

ivandrofly:


src/Forms/Main.cs
```diff
@@ -22350,16 +22350,23 @@ private void ToolStripMenuItemGoogleMicrosoftTranslateSelLineClick(object sender
                 var p = _subtitle.GetParagraphOrDefault(firstSelectedIndex);
                 if (p != null)
                 {
-                    string defaultFromLanguage = LanguageAutoDetect.AutoDetectGoogleLanguage(_subtitle);
+                    string defaultFromLanguage = string.Empty;
```
ivandrofly: don't run the auto-detect language twice - if we are giving more priority to alternate-subtitle

src/Forms/GoogleTranslate.cs
```diff
@@ -43,20 +43,11 @@ public class ComboBoxItem
...
             public ComboBoxItem(string text, string value)
             {
-                if (text.Length > 1)
-                {
-                    text = char.ToUpper(text[0]) + text.Substring(1).ToLowerInvariant();
-                }
-
-                Text = text;
-
+                Text = text.CapitalizeFirstLetter();
```
ivandrofly: use built-in

src/Forms/GoogleOrMicrosoftTranslate.cs
```diff
@@ -136,15 +139,10 @@ private void Translate()
             }
         }
 
-        private static string FixMsLocale(string from)
-        {
-            if ("ar bg zh-CHS zh-CHT cs da nl en et fi fr de el ht he hu id it ja ko lv lt no pl pt ro ru sk sl es sv th tr uk vi".Contains(from))
```
ivandrofly: inside a loop

PR #3616, opened 2019-06-16
Main: minor fixes

ivandrofly:


src/Forms/Main.cs
```diff
@@ -11203,26 +11203,30 @@ private void ColorToolStripMenuItemClick(object sender, EventArgs e)
...
                 foreach (ListViewItem item in SubtitleListview1.SelectedItems)
                 {
                     var p = _subtitle.GetParagraphOrDefault(item.Index);
                     if (p != null)
                     {
                         SetFontColor(p, color);
                         SubtitleListview1.SetText(item.Index, p.Text);
-                        if (_subtitleAlternate != null && Configuration.Settings.General.AllowEditOfOriginalSubtitle && SubtitleListview1.IsAlternateTextColumnVisible)
-                        {
-                            var original = Utilities.GetOriginalParagraph(item.Index, p, _subtitleAlternate.Paragraphs);
```
ivandrofly: the alternate/original paragraph won't be colored if `p` is null

src/Forms/Main.cs
```diff
@@ -20333,6 +20337,12 @@ private void NormalToolStripMenuItem1Click(object sender, EventArgs e)
...
             var tb = GetFocusedTextBox();
 
+            // minimum length of valid html tag is 3
+            if (tb.SelectionLength < 3)
+            {
+                return;
+            }
+
             if (tb.SelectionLength == 0)
             {
```
xylographe: This condition (`tb.SelectionLength == 0`) can no longer be true?

ivandrofly: > This condition (`tb.SelectionLength == 0`) can no longer be true?

Moved above, thanks!

PR #3374, opened 2019-02-17
BridgeGaps: Update + Unit-test

ivandrofly:

PR #3623, opened 2019-06-18
bridgegaps: fix possible bug (batch-convert)

ivandrofly:
https://github.com/SubtitleEdit/subtitleedit/pull/3374#commitcomment-33978834


libse/Forms/DurationsBridgeGaps.cs
```diff
@@ -19,7 +19,12 @@ public static int BridgeGaps(Subtitle subtitle, int minMsBetweenLines, bool divi
             for (int i = 0; i < count; i++)
             {
                 var cur = subtitle.Paragraphs[i];
-                var next = subtitle.Paragraphs[i + 1];
```
ivandrofly: this is very likely to throw exception when subtitle contains only one paragraph

PR #3624, opened 2019-06-18
splitlonglineshelper: fix possible bugs

ivandrofly:


libse/Forms/SplitLongLinesHelper.cs
```diff
@@ -73,12 +75,12 @@ public static Subtitle SplitLongLinesInSubtitle(Subtitle subtitle, int totalLine
                 oldParagraph.Text = lines[FirstLine];
 
                 // use optimal time to adjust duration
-                oldParagraph.EndTime.TotalMilliseconds = oldParagraph.StartTime.TotalMilliseconds + millisecondsPerChar * oldParagraph.Text.Length - halfMinGaps;
+                oldParagraph.EndTime.TotalMilliseconds = oldParagraph.StartTime.TotalMilliseconds + millisecondsPerChar * HtmlUtil.RemoveHtmlTags(oldParagraph.Text, true).Length - halfMinGaps;
```
ivandrofly: don't include html tags when calculating time/char

libse/Forms/SplitLongLinesHelper.cs
```diff
@@ -139,11 +145,11 @@ private static bool StartsWithDash(string text)
...
         private static bool IsTextClosed(string text)
         {
-            if (string.IsNullOrEmpty(text) || text.Length == 0)
+            string textNoTags = HtmlUtil.RemoveHtmlTags(text);
+            if (string.IsNullOrEmpty(textNoTags))
             {
                 return false;
             }
-            string textNoTags = HtmlUtil.RemoveHtmlTags(text);
             char lastChar = textNoTags[textNoTags.Length - 1];
```
ivandrofly: ths line would throw if text contains only html tags

PR #3637, opened 2019-06-23
[subtitle] optimize adjust time using seconds

ivandrofly:


libse/Subtitle.cs
```diff
@@ -368,16 +373,16 @@ public void AdjustDisplayTimeUsingSeconds(double seconds, List<int> selectedInde
...
                         {
-                            _paragraphs[i].EndTime.TotalMilliseconds = _paragraphs[i].StartTime.TotalMilliseconds + 100;
+                            _paragraphs[i].EndTime.TotalMilliseconds = _paragraphs[i].StartTime.TotalMilliseconds + Configuration.Settings.General.SubtitleMinimumDisplayMilliseconds;
                         }
                         else
                         {
                             _paragraphs[i].EndTime.TotalMilliseconds = newEndMilliseconds;
                         }
                     }
-                    else if (newEndMilliseconds > _paragraphs[i].EndTime.TotalMilliseconds)
```
ivandrofly: this will only be true if adjust seconds variable is 0. So to avoid unnecessary O(n) add the check above

libse/Subtitle.cs
```diff
@@ -368,16 +373,16 @@ public void AdjustDisplayTimeUsingSeconds(double seconds, List<int> selectedInde
 
                     if (seconds < 0)
                     {
-                        if (_paragraphs[i].StartTime.TotalMilliseconds + 100 > newEndMilliseconds)
+                        if (_paragraphs[i].StartTime.TotalMilliseconds + Configuration.Settings.General.SubtitleMinimumDisplayMilliseconds > newEndMilliseconds)
                         {
-                            _paragraphs[i].EndTime.TotalMilliseconds = _paragraphs[i].StartTime.TotalMilliseconds + 100;
+                            _paragraphs[i].EndTime.TotalMilliseconds = _paragraphs[i].StartTime.TotalMilliseconds + Configuration.Settings.General.SubtitleMinimumDisplayMilliseconds;
```
ivandrofly: Use user defined minimums duration instead

PR #1270, opened 2015-09-10
Minor fixes

xylographe:
Enhance Unicode support by capitalizing first grapheme instead of first Char.

PR #3658, opened 2019-07-15
handle all upper word e.g: FOOBAR => Foobar

ivandrofly:


xylographe:
The way to transform `FOOBAR` to `Foobar` is **`s.ToLower().CapitalizeFirstLetter()`** — see [old comment.](https://github.com/SubtitleEdit/subtitleedit/pull/1270#discussion-diff-39163994)

PR #3690, opened 2019-08-11
[profile-rules] optimize getting default profiles

ivandrofly:


xylographe:
Why open a new Form to change the profile name? What am I missing?

PR #3696, opened 2019-08-17
move get subtitle format from name to SubtitleFormat

ivandrofly:


src/Forms/Main.cs
```diff
@@ -336,20 +336,6 @@ private void SetCurrentFormat(SubtitleFormat format)
             }
         }
 
-        private SubtitleFormat FormatNameToFormat(string subtitleFormatName)
```
ivandrofly: Avoid polluting main.cs, also makes more sense to have this method inside SubtitleFormat.cs

PR #3698, opened 2019-08-18
update get dictionaries

ivandrofly:


src/Forms/GetDictionaries.cs
```diff
@@ -269,8 +269,7 @@ private static void ExtractDic(string dictionaryFolder, ZipExtractor zip, List<Z
 
         private void comboBoxDictionaries_SelectedIndexChanged(object sender, EventArgs e)
         {
-            int index = comboBoxDictionaries.SelectedIndex;
-            labelPleaseWait.Text = _descriptions[index];
+            labelPleaseWait.Text = ((OnlineDictionary)comboBoxDictionaries.SelectedItem).Description;
```
ivandrofly: it's better if we display the dictionary description instead of the url, which seem to be more interesting to the user.

src/Forms/GetDictionaries.cs
```diff
@@ -12,15 +12,22 @@ namespace Nikse.SubtitleEdit.Forms
...
-        private List<string> _englishNames = new List<string>();
+
+        private class OnlineDictionary
+        {
+            public string Name { get; set; }
+            public string EnglishName { get; set; }
+            public string Description { get; set; }
+            public Uri Url { get; set; }
+
+            public override string ToString() => EnglishName;
```
ivandrofly: Display the English-Name to the user instead of very long / complicated name

PR #3700, opened 2019-08-18
fix index problem when p.Text.Length is 1

ivandrofly:


ivandrofly:
use the file below to recreate the bug 
[arabic-pacv2.zip](https://github.com/SubtitleEdit/subtitleedit/files/3513476/arabic-pacv2.zip)

PR #3713, opened 2019-08-22
Update Dictionaries

ivandrofly:


xylographe:
Shouldn't the character names (person names) in `hr_names.xml` be included in the global `names.xml`?

Let's assume I would add a Dutch subtitle to Croatian spoken material (film, documentary, or whatever), then it won't help me if the Croatian names of characters/persons are in the names list that is specific to the Croatian language. I will still have to add names that are not in the global names list to the Dutch names list, which doesn't seem very efficient?

xylographe:
One more general observation regarding names lists: I think, names of some famous people should move from global to language specific lists. For example, the name of the famous author **Фёдор Михайлович Достоевский** is written **Fjodor Michajlovitsj Dostojevski** in Dutch, **Fiodor Mikhaïlovitch Dostoïevski** in French, **Fyodor Mihayloviç Dostoyevski** in Turkish, and **Fyodor Mikhailovich Dostoevsky** in English.

niksedk:
The tricky thing about the name list is that some names are also normal words in some languages, which makes it hard for fixes related to casing.
 It's just easier to use local names - especially for users.

xylographe:
I think, that problem should be fixed by the language dependent blacklist, like in the Dutch names list,
```
  <blacklist>
    <name>Anders</name>
    <name>April</name>
    <name>Dan</name>
    <name>Doe</name>
    <name>El</name>
    <name>Gore</name>
    <name>Jeep</name>
    <name>Jen</name>
    <name>Marge</name>
    <name>Mars</name>
    <name>Mei</name>
    <name>Metro</name>
    <name>Tor</name>
    <name>Welk</name>
  </blacklist>
```
Which are all valid Dutch words, that may or may not be capitalised, depending on context.

ivandrofly:
> 
> 
> I think, that problem should be fixed by the language dependent blacklist, like in the Dutch names list,
> 
> ```
>   <blacklist>
>     <name>Anders</name>
>     <name>April</name>
>     <name>Dan</name>
>     <name>Doe</name>
>     <name>El</name>
>     <name>Gore</name>
>     <name>Jeep</name>
>     <name>Jen</name>
>     <name>Marge</name>
>     <name>Mars</name>
>     <name>Mei</name>
>     <name>Metro</name>
>     <name>Tor</name>
>     <name>Welk</name>
>   </blacklist>
> ```
> 
> Which are all valid Dutch words, that may or may not be capitalised, depending on context.

So if we are going to have these names in black list, what is the point of also having them in Dutch-name-list? Subtitle edit will automatically remove those names on loading since they are present in blacklist.

e.g:

` xml
<names>
    <blacklist>
        <name>Foobar</name>
    </blacklist>
    <name>Foobar</name>
</names>
`

xylographe:
> I think the black list makes more sense to have only name that are present in `names.xml`.

Yes, the `<blacklist>` words from my Dutch list are all included in `names.xml`. The actual `<names><name>` elements in my Dutch list contain mostly French names, due to the fact that I mostly process subtitles for French spoken material. :) This is the main reason, I did not add my private `nl_names.xml` to SE, yet. Perhaps, I should add an `nl_names.xml` with only a blacklist.

ivandrofly:
Going in 📥 ?

xylographe:
@ivandrofly Sorry, about the delay. Upgrade of computer hardware turns out to be more complicated than expected. :-(

PR #3724, opened 2019-08-25
Update Change Framerate

ivandrofly:


src/Forms/ChangeFrameRate.cs
```diff
@@ -55,9 +59,9 @@ private string GetFrameRateFromVideoFile(string oldFrameRate)
             if (openFileDialog1.ShowDialog() == DialogResult.OK)
             {
                 var info = UiUtil.GetVideoInfo(openFileDialog1.FileName);
-                if (info != null && info.Success)
+                if (info?.Success == true)
                 {
-                    return info.FramesPerSecond.ToString();
+                    return Math.Round(info.FramesPerSecond, 3).ToString();
```
ivandrofly: Only take three decimals, because if `infor.FramesPerSeconds` is equals to `23.9760228733091` 

``` c#
if (comboBoxFrameRateFrom.Text.Trim() == comboBoxFrameRateTo.Text.Trim())
{
    MessageBox.Show(Configuration.Settings.Language.ChangeFrameRate.FrameRateNotChanged);
     return;
    }

```

will have not effects

PR #3735, opened 2019-08-31
Update Common Errors Fix

ivandrofly:


src/Forms/FixCommonErrors.cs
```diff
@@ -608,16 +608,19 @@ public void FixOcrErrorsViaReplaceList(string threeLetterIsoLanguageName)
...
-                    string text = ocrFixEngine.FixOcrErrors(p.Text, i, lastLine, false, OcrFixEngine.AutoGuessLevel.Cautious);
-                    lastLine = text;
-                    if (AllowFix(p, fixAction) && p.Text != text)
+                    if (AllowFix(p, fixAction))
                     {
-                        string oldText = p.Text;
-                        p.Text = text;
-                        noOfFixes++;
-                        AddFixToListView(p, fixAction, oldText, p.Text);
+                        string text = ocrFixEngine.FixOcrErrors(p.Text, i, lastLine, false, OcrFixEngine.AutoGuessLevel.Cautious);
```
ivandrofly: this method is very expensive and also contains `Application.DoEvents()`, so only call it when the paragraph is **allowed fix**

src/Logic/OCR/OcrFixEngine.cs
```diff
@@ -412,16 +412,7 @@ public string SpellCheckDictionaryName
...
-            while (text.Contains(Environment.NewLine + " "))
-            {
-                text = text.Replace(Environment.NewLine + " ", Environment.NewLine);
-            }
-
-            while (text.Contains(" " + Environment.NewLine))
-            {
-                text = text.Replace(" " + Environment.NewLine, Environment.NewLine);
-            }
+            string text = input.FixExtraSpaces();
```
ivandrofly: More optimized

PR #3737, opened 2019-09-01

diomed:


diomed:
@xylographe could you please come on gitter today or tomorrow, I only have time those two days - so we can talk about custom version.

PR #3746, opened 2019-09-02
Update/Optimization choose encoding

ivandrofly:


src/Forms/ChooseEncoding.cs
```diff
@@ -37,30 +37,29 @@ public void Initialize(string fileName)
...
-                _fileBuffer = new byte[length];
-                file.Read(_fileBuffer, 0, length);
-                file.Close();
             }
             catch
             {
                 _fileBuffer = new byte[0];
             }
 
-            var encoding = LanguageAutoDetect.DetectAnsiEncoding(_fileBuffer);
```
ivandrofly: check the boom first to avoid redundant processing

PR #3785, opened 2019-09-30
[utilities] - minor update regarding to get file name without extension

ivandrofly:


libse/Utilities.cs
```diff
@@ -2779,18 +2779,17 @@ public static bool QualifiesForMerge(Paragraph p, Paragraph next, double maximum
...
         {
-            if (string.IsNullOrEmpty(fileName))
+            if (!Path.IsPathRooted(fileName))
             {
                 return fileName;
             }
 
-            var idx = fileName.LastIndexOf('.');
-            if (idx > 0)
+            if (!Path.HasExtension(fileName))
```
ivandrofly: See => https://docs.microsoft.com/en-us/dotnet/api/system.io.path.hasextension?view=netcore-3.0

PR #3799, opened 2019-10-11
don't invoke the main list view selection handler unnecessary

ivandrofly:


src/Forms/Main.cs
```diff
@@ -5707,7 +5707,9 @@ private void GotoLineNumberToolStripMenuItemClick(object sender, EventArgs e)
                 {
                     if (tabControlSubtitle.SelectedIndex == TabControlListView)
                     {
+                        SubtitleListview1.SelectedIndexChanged -= SubtitleListview1_SelectedIndexChanged;
                         SubtitleListview1.SelectNone();
```
ivandrofly: un-select without generating events

PR #3810, opened 2019-10-21
Remove redundant overrides + Minor refact

xylographe:
The base value of **`IsTimeBased`** is already **`true`**.

### DIFF
--- a/libse/SubtitleFormats/NetflixTimedText.cs
+++ b/libse/SubtitleFormats/NetflixTimedText.cs
@@ -329,13 +329,7 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
             subtitle.Paragraphs.ForEach(p => p.Text = Regex.Replace(p.Text, @"^({\\an[1-7,9]})", string.Empty));
         }
 
-        public override bool HasStyleSupport
-        {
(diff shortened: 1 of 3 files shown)


libse/SubtitleFormats/NetflixTimedText.cs
```diff
@@ -329,13 +329,7 @@ public override void LoadSubtitle(Subtitle subtitle, List<string> lines, string
...
         }
 
-        public override bool HasStyleSupport
-        {
-            get
-            {
-                return false;
-            }
-        }
+        public override bool HasStyleSupport => false;
```
niksedk: I think this one can be removed too as SubtitleFormat `HasStyleSupport` is also default false

xylographe: **NetflixTimedText** inherits from **TimedText10**,
```
public class NetflixTimedText : TimedText10
```
**TimedText10** overrides `HasStyleSupport`,
```
public class TimedText10 : SubtitleFormat

public override bool HasStyleSupport => Configuration.Settings.SubtitleSettings.TimedText10ShowStyleAndLanguage;
```

PR #3827, opened 2019-11-03
Feature/teletext

niksedk:
### Import teletext from transport stream files

TransportStreamParser.cs is extended to handle teletext pes packets which are sent to Teletext.cs which is based on https://github.com/debackerl/telxcc
Colors are kept and top alignment is detected if text is only in lines 1-5 (of 23).

The transport stream track picker now supports both images and teletext.

Batch/cmd convert is saved for later.

#### Notes
Teletext.cs has some static stuff which makes it somewhat hard to use.

https://github.com/CCExtractor/ccextractor/commits/master/src/lib_ccx/telxcc.c might have some commits with fixes that SE should also have.


xylographe:
So far I have used **CCExtractorGUI** to extract ST from teletext before processing the result with SE. Would certainly be convenient to have this functionality included in SE. I will test various transport streams later this week.

niksedk:
> 
> 
> So far I have used **CCExtractorGUI** to extract ST from teletext before processing the result with SE. Would certainly be convenient to have this functionality included in SE. I will test various transport streams later this week.

It has been on my todo list for many years to make a convenient way of getting teletext. I've used **Project X** and that UI is a bit quirky to say the least. 
I recently opened telxcc.c in VS Code and discovered that I could debug it... took some hours but the code was fairly easy to convert to C#.

Well, I look forward to hear how it works for you :)

src/Forms/Main.cs
```diff
@@ -12175,15 +12175,29 @@ private bool ImportSubtitleFromTransportStream(string fileName)
...
             {
                 MessageBox.Show(_language.NoSubtitlesFound);
                 _exitWhenLoaded = _loading;
                 return false;
             }
 
-            int packedId = tsParser.SubtitlePacketIds[0];
-            if (tsParser.SubtitlePacketIds.Count > 1)
+            if (tsParser.SubtitlePacketIds.Count == 0 & tsParser.TeletextSubtitlesLookup.Count == 1 && tsParser.TeletextSubtitlesLookup.First().Value.Count() == 1)
+            {
```
xylographe: @niksedk
typo: **`&`** ==> **`&&`**.
```
tsParser.SubtitlePacketIds.Count == 0 && tsParser.TeletextSubtitlesLookup.Count == 1
```

niksedk: Fixed in https://github.com/SubtitleEdit/subtitleedit/commit/cf36eb9e69328b88818c6249351779075d0a73ad

PR #3860, opened 2019-12-02
Fix AutoBreakLines in VobSubOcr for POSIX environment

xylographe:
In a POSIX environment (e.g. Linux) `Environment.Newline.Length` equals 1, hence,
```
text.Replace(Environment.NewLine, "*").Length + 2 <= text.Length
````
will always be false, because,
```
text.Replace(Environment.NewLine, "*").Length == text.Length
```
is always true.

### DIFF
--- a/src/Forms/Ocr/VobSubOcr.cs
+++ b/src/Forms/Ocr/VobSubOcr.cs
@@ -637,7 +637,7 @@ namespace Nikse.SubtitleEdit.Forms.Ocr
                 text = text.Replace(Environment.NewLine + " ", Environment.NewLine);
 
                 // max allow 2 lines
-                if (checkBoxAutoBreakLines.Checked && text.Replace(Environment.NewLine, "*").Length + 2 <= text.Length)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/Forms/Ocr/VobSubOcr.cs [hunk 1]
```diff
@@ -637,7 +637,7 @@ internal void InitializeBatch(string vobSubFileName, VobSubOcrSettings vobSubOcr
                 text = text.Replace(Environment.NewLine + " ", Environment.NewLine);
 
                 // max allow 2 lines
-                if (checkBoxAutoBreakLines.Checked && text.Replace(Environment.NewLine, "*").Length + 2 <= text.Length)
+                if (checkBoxAutoBreakLines.Checked && Utilities.CountTagInText(text, Environment.NewLine) <= 2)
```
niksedk: Perhaps `GetNumberOfLines` is more "correct" ?

xylographe: Done.

PR #3873, opened 2019-12-05
Check environment variables for directory overrides

NickZ:
This is intended for when the app is run in confinement, when it won't have access to either the `AppData` directory (or `.config` on linux) or its base directory for writing.

### DIFF
--- a/libse/Configuration.cs
+++ b/libse/Configuration.cs
@@ -131,7 +131,12 @@ namespace Nikse.SubtitleEdit.Core
 
             if (IsRunningOnLinux || IsRunningOnMac)
             {
-                if (!Directory.Exists(appDataRoamingPath) && !File.Exists(Path.Combine(BaseDirectory, ".PACKAGE-MANAGER")))
+                var datafolder_override = Environment.GetEnvironmentVariable("DATA_FOLDER_OVERRIDE");
+                if (datafolder_override)
+                {
+                    appDataRoamingPath = Path.Combine(datafolder_override, "");
+                }
+                else if (!Directory.Exists(appDataRoamingPath) && !File.Exists(Path.Combine(BaseDirectory, ".PACKAGE-MANAGER")))
                 {
                     try
                     {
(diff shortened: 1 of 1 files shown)


libse/Configuration.cs
```diff
@@ -180,6 +185,11 @@ private static string GetTesseractDataDirectory()
         {
             if (IsRunningOnLinux || IsRunningOnMac)
             {
+                var tesseract_overide = Environment.GetEnvironmentVariable("TESSERACT4_OVERRIDE");
+                if (tesseract_overide != null)
+                {
+                    return tesseract_overide;
```
xylographe: s/tesseract_overide/tesseract_over**r**ide/

libse/Configuration.cs
```diff
@@ -202,6 +212,11 @@ private static string GetTesseract302DataDirectory()
         {
             if (IsRunningOnLinux || IsRunningOnMac)
             {
+                var tesseract_overide = Environment.GetEnvironmentVariable("TESSERACT302_OVERRIDE");
+                if (tesseract_overide != null)
+                {
+                    return tesseract_overide;
```
xylographe: s/tesseract_overide/tesseract_over**r**ide/

libse/Configuration.cs
```diff
@@ -131,7 +131,12 @@ private static string GetDataDirectory()
 
             if (IsRunningOnLinux || IsRunningOnMac)
             {
-                if (!Directory.Exists(appDataRoamingPath) && !File.Exists(Path.Combine(BaseDirectory, ".PACKAGE-MANAGER")))
+                var datafolder_override = Environment.GetEnvironmentVariable("DATA_FOLDER_OVERRIDE");
+                if (datafolder_override != null)
+                {
```
xylographe: Perhaps, `if (!string.IsNullOrEmpty(datafolder_override) && Directory.Exists(datafolder_override))`?

niksedk:
When/how is SE running in confinement (so I can mention this on the help page) ?
The three overrides (like "DATA_FOLDER_OVERRIDE") should probably be in Settings.xml where the rest of the SE settings are or do you have a special reason for using environment variables?
Should probably work for Windows too.

NickZ:
@niksedk to answer your questions:
SE runs in confinement when in a snap package on Linux. Please see #3952
Unfortunately, this can't be in Settings.xml for snap packaging, I can't get them into there into the user folder when installing. It has to be environment variables. 
I didn't think it was necessary for Windows, because it will only be running in confinement on Linux and MacOS.

NickZ:
Closing in favor of #3971

Issue #3417, opened 2019-02-26
[Request] Add setting for sentence continuation style


Flitskikker:
Similar to the [dialogue style request](https://github.com/SubtitleEdit/subtitleedit/issues/3416), the sentence continuation style is another subtitling element that's different across countries. At the moment, Subtitle Edit only supports one continuation style. I think it would be useful to allow the user to configure the continuation style used.

As far as I can tell, these are the 4 most common continuation styles:

_English / Flemish style (currently the only supported style):_
```
Lorem ipsum dolor sit amet
consectetur adipiscing elit,

donec eget turpis consequat
turpis commodo hendrerit

mauris mollis consectetur nibh,
nec congue est viverra quis.
```

_Dutch TV style:_
```
Lorem ipsum dolor sit amet
consectetur adipiscing elit...

donec eget turpis consequat
turpis commodo hendrerit...

mauris mollis consectetur nibh,
nec congue est viverra quis.
```

_Dutch cinema style:_
```
Lorem ipsum dolor sit amet
consectetur adipiscing elit...

...donec eget turpis consequat
turpis commodo hendrerit...

...mauris mollis consectetur nibh,
nec congue est viverra quis.
```

_Scandinavian style:_
```
Lorem ipsum dolor sit amet
consectetur adipiscing elit -

- donec eget turpis consequat
turpis commodo hendrerit -

- mauris mollis consectetur nibh,
nec congue est viverra quis.
```

As with the dialogue styles, there exist slight variations, for some styles also based on the amount of time between the subtitles, so I think it would be best to also implement generic settings, for example:

![se4](https://user-images.githubusercontent.com/3516155/53379666-c838e800-396a-11e9-96cf-61bc93ce5b29.png)

This way, users can configure it completely.

These settings would then be used across the whole application, for example:
- When splitting a subtitle (suffix and prefix get added)
- When merging two subtitles (suffix and prefix get removed)
- For detecting errors (e.g. missing prefix/suffix)
- For the "Only merge continuation lines" option in Tools > Merge short lines
- If possible, an option to convert between styles

I was working on a plugin myself to automatically convert from the English continuation style to the other continuation styles. Unfortunately, the development got stranded because there was just too much to take into account. Detecting new sentences, detecting lyrics and translated signs, adding suffix/prefix inside tags, adding suffix/prefix inside dialogues... Maybe you could have a try at it, then it would also be integrated into SE directly. I can send you thoughts / things to take into account separately.

[...]

Issue #3416, opened 2019-02-26
[Request] Add setting for dialogue style


Flitskikker:
In the subtitling world, every country has their own guidelines and style. One of the elements that are different across countries, is the dialogue style. At the moment, Subtitle Edit only supports one dialogue style. I think it would be useful to allow the user to configure the dialogue style used.

As far as I can tell, these are the 4 most common dialogue styles:

_English fansubs / Scandinavian subtitles (currently the only supported style):_
```
- Lorem ipsum dolor sit amet.
- Consectetur adipiscing elit.
```

_English retail subtitles:_
```
-Lorem ipsum dolor sit amet.
-Consectetur adipiscing elit.
```

_Dutch professional subtitles:_
```
Lorem ipsum dolor sit amet.
-Consectetur adipiscing elit.
```

_Dutch fansubs / Flemish subtitles:_
```
Lorem ipsum dolor sit amet.
- Consectetur adipiscing elit.
```

However, there are countries that have dialog dashes mid-line, and there are countries that use the en dash (–) instead of a hyphen (-) as a dialogue dash, so I think it would be best to implement generic settings as follows:

![se3](https://user-images.githubusercontent.com/3516155/53378602-7726f500-3966-11e9-8049-35fe77a2c66c.png)

These settings would then be used across the whole application, for example:
- For the "Merge selected lines a dialogue" option
- When splitting a dialogue
- For detecting errors (e.g. ignore "Fix missing space" for retail style)
- If possible, an option to convert between styles

I'm following the professional Dutch style, so at the moment I do everything dialogue-related manually. It would be nice to finally be able to use these features. 😄

Thanks in advance!

Ding-adong:
- denotes different speakers therefore having...
- blah speaker 1
- blah speaker 2
makes perfect sense. Perhaps the Dutch people should evolve to using better subtitling.

Flitskikker:
In Dutch, we call it the "second speaker dash".
A single dash is also used in other languages than Dutch, such as Bulgarian, Croatian, Icelandic, Slovenian, Serbian... probably more.
Therefore, I think many users would benefit from such an option.

darnn:
For what it's worth, in Hebrew it's usually the same as in the above example of Dutch professional subtitles.

OmrSi:
In Arabic, we always use the first one, whether it is for Netflix, Amazon or fansubbing.

marb99:
[...]

PR #3882, opened 2019-12-10
Add additional common profile presets

Flitskikker:
- Netflix for most other languages than English, derived from the Timed Text Style Guides: https://partnerhelp.netflixstudios.com/hc/en-us/sections/203480497-Timed-Text-Style-Guides
- Dutch professional subtitles, derived from the style guide from subtitling company Hoek & Sonépouse (http://www.opensubtitles.org/addons/documents/subtitling_guide_nl.html) and the guidelines composed by the Subtitlers division of the Dutch Authors Union (not yet publicly available).
- Dutch fansubs, derived from the OpenSubtitles fansubbing guidelines (https://forum.opensubtitles.org/viewtopic.php?t=12639) and the Dutch SubtitleEdit manual (https://forum.opensubtitles.org/viewtopic.php?f=38&t=15314).
- Danish professional subtitles, derived from the guidelines compiled by the Danish subtitlers’ association: https://undertekstning.dk/english/
- French subtitles, derived from the SW synchronization norms: http://normes.golgeek.fr/

For the language-specific profiles, I added both a 23.976/24 fps and 25 fps version.
Maybe do this for the Netflix profiles as well? Or maybe use frames for the gap setting, since it's the smallest constituent unit for videos.

Maybe also find a way to update the profiles for users who have version 5.3.11 installed (since AddExtraProfiles() will not be called).

### DIFF
--- a/libse/Settings.cs
+++ b/libse/Settings.cs
@@ -939,7 +939,21 @@ $HorzAlign          =   Center
                 SubtitleMinimumDisplayMilliseconds = 833,
                 SubtitleMaximumWordsPerMinute = 300,
                 CpsIncludesSpace = true,
-                MinimumMillisecondsBetweenLines = 84, // 2 frames for 23.97 videos
+                MinimumMillisecondsBetweenLines = 84, // 2 frames for 23.976 fps videos
+            });
+            profiles.Add(new RulesProfile
+            {
+                Name = "Netflix (Other languages)",
+                SubtitleLineMaximumLength = 42,
+                MaxNumberOfLines = 2,
+                MergeLinesShorterThan = 42,
+                SubtitleMaximumCharactersPerSeconds = 17,
+                SubtitleOptimalCharactersPerSeconds = 12,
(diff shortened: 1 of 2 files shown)


niksedk:
Nice :)

About the settings not being updated, I think a "Reset to default settings" button could solve that - and also other settings that have been adjusted to better values.

I hope to expand the profiles to include "hearing impaired tags" and "dialog dash style" (there's a discussion of dialog styles somewhere in the issues...) in some future version.

Flitskikker:
@niksedk 
Thanks! Yes, a "Reset defaults" button is always useful.

Ah yes, I opened the dialog styles discussion a while ago: #3416
Would you also consider expanding the profiles with sentence continuation style? (The other discussion, in #3417). It also fits with the idea of profiles; different styles for different countries/clients. For example in NL we use dots, and in Denmark/Scandinavia they use dashes.

I'm afraid both are too extensive for me to "fix" in a PR myself.

The HI tags are a good addition, didn't think about that.

PR #3913, opened 2020-01-14
Update hrb_OCRFixReplaceList.xml

diomed:


diomed:
@xylographe I didn't do it good

It should be:

`<RegEx find="spjesn" replaceWith="spješn" />`

xylographe:
Fixed in commit 98e9f77655e5bc0174511283f40e6810826c43b9.

PR #3923, opened 2020-01-17
Add and apply Utilities.GetSubtitleLanguageCultures()

xylographe:
Merge duplicated code.

The current subtitle language exclusion filter suppresses Nynorsk and some other languages that could be supported. The new exclusion filter suppresses only languages that cannot be supported, languages that require an explicit script, e.g. sr-Cyrl.

### DIFF
--- a/libse/Utilities.cs
+++ b/libse/Utilities.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Drawing;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Reflection;
 using System.Security.Authentication;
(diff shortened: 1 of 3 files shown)


libse/Utilities.cs
```diff
@@ -754,6 +754,25 @@ public static List<string> GetDictionaryLanguagesCultureNeutral()
             return list;
         }
 
+        public static IEnumerable<CultureInfo> GetSubtitleLanguageCultures()
+        {
+            var excludes = new HashSet<string>();
+            foreach (var ci in CultureInfo.GetCultures(CultureTypes.NeutralCultures))
+            {
+                if (ci.Name.Length < 4 && ci.Name == ci.IetfLanguageTag)
+                {
```
niksedk: You could add the "ci" to a "temp list here" so we don't have to call "CultureInfo.GetCultures(CultureTypes.NeutralCultures)" twice.

xylographe: Done.

PR #3942, opened 2020-01-23
VobSubOcr: Refact + Add multi-lang items

xylographe:
Because it is hidden (beta), I didn't add _NOCR_ to multi-lang.

Due to the `LogItem` base class for _UnknownWords_, _AutoGuesses_ and _OcrFixes_, string manipulations to extract the line number from the combined text can be avoided.


niksedk:
NOCR was a test with vector lines... worked worse than I expected, but that might by my code ;)
If I had a lot of time for only OCR then Tensorflow or another ANN engine would be interesting to test...

xylographe:
When I started using SE I first tried tesseract (v3). Reasonably fast, but rather poor results, manual post-processing took ages. Then I found that although image compare is rather slow at the start (unrecognised characters), the _overall_ time I needed was less: after the first 60-100 subtitles image compare is fast, and manual post-processing takes only little time. Lately, I've been using tesseract (v4) again, and it's much better than it used to be: I can now use non-interactive tesseract4 (doesn't matter that it's not super-fast), yet the results are within acceptable limits, which, I presume, is also caused by  improved OCR fixing by SE.

I also tried nOCR, but it took too much time: too much training (manual character recognition), and also too much post-processing, mainly due to punctuation (especially dots & commas, IIRC).

Issue #3761, opened 2019-09-15
Would it be possible to publish libse to Nuget


EraYaN:
I'd love to consume libse from another (dotnet core) project. And since the licenses are compatible I should be able to do so. But a release on Nuget would make this a lot easier.

EDIT: Alright I gave it a try running it on dotnet core. And for it to work on netstandard it will need version 2.1 (so the end of this month). And I had to remove some Registry stuff for the configuration. But it DOES work to parse subtitles.

EraYaN:
@niksedk @xylographe So now that #3946 is in, what else do you guys need for this to show up on Nuget? I can add some script stuff that adds a `dotnet pack` to the build, but if you want to split the libse folder out then that would preferably happen first. (example result: https://github.com/EraYaN/libse)

niksedk:
libse is now a a nuget package

PR #3946, opened 2020-01-25
Move LibSE to multiple frameworks (.NET Standard2.1 and .NET 4.0)

EraYaN:
This should help in using this in .NET Core projects. This could be a first step towards #3761 

Building a nuget package from this project now would not be that hard.


EraYaN:
AppVeyor needs some love to select the target framework as .NET 4.0. But I can't see how to do that.

Probably something to do with 
```
# to run your custom scripts instead of automatic MSBuild
build_script:
```
And set it to the build.bat and update that to include the target framework with a switch like `/p:TargetFramework=net40;TargetFrameworkVersion=v4.0`. Or switch to the dotnet CLI completely.

NickZ:
Please see my PR here, it may help you: https://github.com/SubtitleEdit/subtitleedit/pull/3524

In particular, I think you will need to target .net standard 3.0, because there are some missing drawing functions that aren't implemented in .net standard 2.1.

EraYaN:
Well .NET Standard 3.0 does not exist. .NET Core 3.0 (and 3.1) both implement .NET Standard 2.1. It's much preferable to ship the library as a version of .NET standard and move anything that is not supported (and probably not required by any library consumer) to a separate project, that targets some UI or Web focused framework.

niksedk:
@EraYaN: thx for the PR :)
AppVeyor seems somewhat happy now: https://ci.appveyor.com/project/SubtitleEdit/subtitleedit/builds/30509069 (still get some `fatal: Needed a single revision` messages... but hey, it's still green :)
Could you merge master into this branch?

@XhmikosR: Do you see any problems with this?

xylographe:
> Do you see any problems with this?

I think, a few more changes might be necessary, because the LibSE project output folder has changed to `bin\Release\net40\`. Other then that I don't know. I cannot test it. VS2017 doesn't have .NET SDK 3.0,
```
PS> dotnet --list-sdks
1.0.0-preview2-003131 [C:\Program Files\dotnet\sdk]
1.1.14 [C:\Program Files\dotnet\sdk]
2.1.202 [C:\Program Files\dotnet\sdk]
2.1.505 [C:\Program Files\dotnet\sdk]
2.1.507 [C:\Program Files\dotnet\sdk]
2.1.509 [C:\Program Files\dotnet\sdk]
2.1.511 [C:\Program Files\dotnet\sdk]
```
Hence, all I can see when building LibSE target ,NET Standard 2.1 is,
```
The current .NET SDK does not support targeting .NET Standard 2.1.
```
The .NET Framework 4.0 target seems to build ok.

Branch without conflicts is [xg/MultiNet](https://github.com/SubtitleEdit/subtitleedit/tree/xg/MultiNet).

@OmrSi If you would like to try,
```
Build-SubtitleEditBeta.ps1 -Archive https://github.com/SubtitleEdit/subtitleedit/archive/xg/MultiNet.zip
```

xylographe:
@niksedk 
Not tested, but I think, `build.bat` would need a minor adjustment.
From,
```
"%MSBUILD%" SubtitleEdit.sln /t:%BUILDTYPE% /p:Configuration=Release /p:Platform="Any CPU"^
```
to,
```
"%MSBUILD%" SubtitleEdit.sln /r /t:%BUILDTYPE% /p:Configuration=Release /p:Platform="Any CPU"^
```
Add the `/r` (`/restore`) switch to generate `project.assets.json`.

XhmikosR:
@niksedk TBH I have not experience with this. You could just switch to .Net 4 and make it a major version, if there are any real benefits. Otherwise, keeping it simple might be better.

xylographe:
I think, ideally, LibSE would be moved to its own repository (SubtitleEditLibrary, perhaps), and published as a NuGet package, which could be used in other applications as well. This NuGet package would be added to SE, and in the long run SE would move to .NET 5.0 (the successor of .NET Framework 4.8).

EraYaN:
I can split it out into it separate repo (including all history) I’ve done that many times, and then you can push it to this accounts’ repositories.

SubtitleEdit itself can stay on 4.0 (or go to 4.8 for speed improvements) on the main framework until 5.0 hits.

EraYaN:
I rebased and cherrypicked @xylographe's commit onto this branch. This branch should be open to pushes from contributors.

OmrSi:
@xylographe I don't have much experience with this, so this is probably an error because of something missing from my end, but the normal version is building without issues:
image

xylographe:
@OmrSi Thanks.
From the first line: _Asset file not found. Run a NuGet package restore._

It means that to build the libSE project with MSBuild directly (no Visual Studio) the `/restore` switch is required. I will update Build-SubtitleEditBeta.ps1 ASAP.

EraYaN:
What also works is using the dotnet CLI. Then any other implementation details will be taken care of, even if they might change. Although building a solution is not recommended, building (and publishing) a csproj is the way to go.

niksedk:
I think this is a good start :)
And I think we should keep it like this for a few months + do some tests on Linux/Mac.

I guess `System.Drawing` is the main issue - a bad lib to be dependent on, so other libs (ImageSharp?) should probably be used instead (or remove bitmap stuff from LibSE)

EraYaN:
In Jellyfin we use https://github.com/mono/SkiaSharp for all image processing. It's a quite good image processing library from google. It's worked for us so far. But not sure about your actual drawing (graphics) requirements.

OmrSi:
@xylographe It builds, but I got these.
image

xylographe:
@OmrSi 
According to [Microsoft Docs](https://docs.microsoft.com/en-us/dotnet/api/system.net.securityprotocoltype?view=netstandard-2.1#fields) `SecurityProtocolType.Ssl3` is stll supported (_“provided for backward compatibility only”_) in .NET Standard 2.1.

NickZ:
Is there a reason that it's still enabled? SSL 3.0 is deprecated and incredibly unsafe to use because of [POODLE](https://en.wikipedia.org/wiki/POODLE) since 2014.

niksedk:
I guess there's no need for `SecurityProtocolType.Ssl3` (also, many sites now only supports TLS 1.2+).

milahu:
> Add the `/r` (`/restore`) switch to generate `project.assets.json`.

this breaks in offline builds:

> "/build/source/SubtitleEdit.sln" (Restore target) (1) ->
> (Restore target) -> 
>   /nix/store/pm6il5m5i0jqscz1x4hz1jqxvpw0qrhn-msbuild-16.10.1+xamarinxplat.2021.05.26.14.00/lib/mono/msbuild/Current/bin/NuGet.targets(131,5): error : Unable to load the service index for source https://api.nuget.org/v3/index.json. [/build/source/SubtitleEdit.sln]

why not add `src/libse/obj/project.assets.json` to the repo?

PR #3948, opened 2020-01-26
Do not display Dictionaries link on Linux/MacOS

NickZ:
The "Open Dictionaries folder" link on the "Download additional dictionaries" dialog will crash the program when clicked on Linux and MacOS. This PR disables them on Linux and MacOS.


niksedk:
Hm, the open folder link works fine for on Ubuntu / Arch (Manjaro) ...

NickZ:
@niksedk you're right, it doesnt happen on Ubuntu 18.04 but it does happen on my Ubuntu 19.10. I'm closing this and instead just adding exception handlers for those errors here: #3951

PR #3954, opened 2020-01-28
Use distro defaults for linux default font

NickZ:
Instead of using DejaVu Serif, use the default fonts for the respective distros.

These are:

- Ubuntu: Ubuntu
- Manjaro: Noto Sans
- Fedora/GNOME: Cantarell
- others: Dejavu Sans

Also adjust UI font size based on font chosen.

### DIFF
--- a/libse/Configuration.cs
+++ b/libse/Configuration.cs
@@ -3,6 +3,8 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Drawing;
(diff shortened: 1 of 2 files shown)


niksedk:
SE is build for Unicode... and works with e.g. music symbols a lot.
`Dejavu Sans` is a ttf Unicode font that is hopefully available on all distros.
(I've was somewhat chocked too see how poor Unicode support is on Linux compared to Win10)

PR #3965, opened 2020-01-31
Some minor updates

ivandrofly:


niksedk:
Hi @ivandrofly - thx for the PR :)

1) Could you please make smaller pull requests.
2) Please do not mix different things in the same PR.
3) Please for each PR add a description of "what" and especially "why".
4) Please test any optimizations in real world scenarios. I just tested the `ProgramMapTableParser` on a 2 GB .ts file + put both new optimized method and old method in a 100 count loop - and the old method took 1 ms and the new optimized method took 2 ms...
4) Your commit messages are already good, but you could improve them a tiny bit by prefixing them with a verb, such as "Add", "Remove", "Optimize", "Fix" (IMO)

ivandrofly:
> 4\. Please test any optimizations in real world scenarios. I just tested the `ProgramMapTableParser` on a 2 GB .ts file + put both new optimized method and old method in a 100 count loop - and the old method took 1 ms and the new optimized method took 2 ms...

The speed can be noticed when there is key in HashMap, 'cause List contains is always O(n) while hashmap is O(1)

PR #3971, opened 2020-02-07
Environment variable overrides for data directories and linux font

NickZ:
Adds environment variable overrides for the Data Directory and the Tesseract directories. Also adds an override to the Linux font.

The reason for the data directories override being Environment variables instead of configuration file changes is because in certain environments (like my snap package) the data directory will not be known until run time, and we will not be able to change the configuration file because it will be installed to a read-only file system. 

The reason for the Linux font override is, while DejaVu Sans certainly has better compatibility with unicode than most, it's not perfect and does not support certain Asian scripts. In the case of people using certain Asian scripts, I want this to be configurable so that the user can switch to a font that works with the script that they use.

These environment variables are able to be used on Linux, MacOS, and Windows.


niksedk:
Why change `DejaVu Serif` to `DejaVu Sans` ?

(Off topic rant: Unicode support on Linux is like the "good" old WinXP days...)

NickZ:
Because:
1) Deja Serif is pretty ugly as a UI font, honestly
2) It doesn't handle Arabic script properly, and Deja Sans does.

niksedk:
"Serif" gives people a chance to see the difference between uppercase "i" and lowercase "L"... used by Latin languages (many people), so I still prefer that.

NickZ:
OK, in that case, could we split this into default Linux UI font and default Linux subtitle font?

NickZ:
i.e. the UI font would be DejaVu Sans, and the subtitle font would be DejaVu Serif

AnterCreeper:
Can Adds environment variable overrides for the UI fonts? Some of the languages displays only squares.

AnterCreeper:
Pretty Thanks!

PR #3973, opened 2020-02-08
[installer] Conditionally update "Dictionaries" XML files

xylographe:
The `SetupSupport.dll` assembly makes it easy to upgrade user dictionary and no-break-after files by merging the distributed file with the existing file. The update behaviour can be set for each word list separately via the "Update" attribute value: "Merge" (default), "Never" or "Replace" ("Overwrite").

For consistency the other file types (OcrFix and Names) also take an "Update" attribute, whose value can be "Replace"/"Overwrite" (default) or "Never".

To make the "Update" attribute persistent, a few minor changes to the SE code are necessary: `XmlNode.RemoveAll()`, which removes _child nodes and attributes_, has been replaced with `XmlNode.CloneNode(false)`, which removes child nodes while _preserving attributes_.

_Note 1:_ SetupSupport and SubtitleEdit projects must specify the same `<TargetFrameworkVersion>`.
_Note 2:_ SetupSupport `<PlatformTarget>` must be `x86`, because Inno Setup builds 32-bit installers.

### DIFF
--- a/Dictionaries/da_DK_user.xml
+++ b/Dictionaries/da_DK_user.xml
@@ -1,4 +1,6 @@
-﻿<words>
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Set "Update" attribute to "Merge" (default), "Never" or "Replace" -->
+<words Update="Merge">
   <word>1/2</word>
   <word>aflæsningszone</word>
   <word>afterparty</word>
@@ -51,7 +53,7 @@
   <word>grape</word>
   <word>grev</word>
   <word>heroinmisbruger</word>
-  <word>Hey</word>
+  <word>hey</word>
   <word>hm</word>
   <word>hmm</word>
   <word>hos</word>
--- a/Dictionaries/da_NoBreakAfterList.xml
+++ b/Dictionaries/da_NoBreakAfterList.xml
@@ -1,4 +1,6 @@
-﻿<NoBreakAfterList>
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Set "Update" attribute to "Merge" (default), "Never" or "Replace" -->
+<NoBreakAfterList Update="Merge">
   <Item>Dr</Item>
   <Item>Dr.</Item>
   <Item>Frk.</Item>
--- a/Dictionaries/da_names.xml
+++ b/Dictionaries/da_names.xml
@@ -1,6 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <!-- This list contains names with specific casing - and specific to Danish only -->
-<names>
+<!-- Set "Update" attribute to "Replace" (default) or "Never" -->
+<names Update="Replace">
   <blacklist>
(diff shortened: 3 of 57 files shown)


xylographe:
Apparently, the AppVeyor invocation of `vstest.console` has changed, causing unit test _FixCommonOcrErrorsSlashIsL_ to fail.

Before (OK),
```
vstest.console /logger:Appveyor /TestCaseFilter:"TestCategory!=Local" "C:\projects\subtitleedit\src\Test\bin\Debug\Test.dll"
```
After,
```
vstest.console /logger:Appveyor "C:\projects\subtitleedit\src\Test\bin\Debug\Test.dll"
```
Fails with,
```
  X FixCommonOcrErrorsSlashIsL [86ms]
  Error Message:
   Assert.AreEqual failed. Expected:<The font is ita/ic!>. Actual:<The font is italic!>.
```

niksedk:
Sorry about the appveyor fail - I'm trying to msbuild work on appveyor without warnings/errors.
There's a problem with a `git rev-parse --verify refs/heads/master` I think + a problem with the test project not able to overwrite binary files - see https://ci.appveyor.com/project/SubtitleEdit/subtitleedit/builds/30659356
Any ideas?

xylographe:
AppVeyor clones only one branch,
```
git clone -q --branch=feature/test-run-settings https://github.com/SubtitleEdit/subtitleedit.git C:\projects\subtitleedit
```
Therefore, **`git rev-parse --verify refs/heads/master`** must fail. However, this doesn't matter for AppVeyor builds. The **`latestRepositoryVersion`** is only used to determine if the AssemblyInfo template file needs to be updated. This can only happen when you create the next Release in your local repository.

In an AppVeyor build **`latestRepositoryVersion.RevisionGuid.Length > 0`** will always be false (because there is no master branch), and **`currentRepositoryVersion.RevisionGuid.Length == 0`** will always be false too (because the git repository exists), Hence, **`newVersion = currentRepositoryVersion`** is always executed, where **`currentRepositoryVersion`** contains the information retrieved from the repository HEAD.

xylographe:
About the overwrite errors,
```
Test Run deployment issue: Failed to copy file 
'C:\projects\subtitleedit\src\Test\bin\Debug\zlib.net.dll' to
'C:\projects\subtitleedit\src\Test\bin\Debug\zlib.net.dll'

Test Run deployment issue: Failed to copy file 
'C:\projects\subtitleedit\src\Test\bin\Debug\libse.dll' to
'C:\projects\subtitleedit\src\Test\bin\Debug\libse.dll'

etc.
```
Why are these assembly files copied to their own location? This can never succeed.

niksedk:
I don't know why the dll files are copied to their own folder yet...

xylographe:
I think, to avoid copying those files, you would add,
```
  <MSTest>
    <DeploymentEnabled>False</DeploymentEnabled>
  </MSTest>
```
to the `Local.runsettings` file.

PR #3978, opened 2020-02-09
Fix opening URLs, files, and folders on Linux with recent versions of Mono

NickZ:
Recent versions of Mono have broken "system.diagnostics.process.start" for files, urls, and folders on Linux (regular executables work just fine). I have replaced that method with "xdg-open", which will work across all Linux distributions and should work on every mono version.

### DIFF
--- a/libse/Utilities.cs
+++ b/libse/Utilities.cs
@@ -854,7 +854,18 @@ namespace Nikse.SubtitleEdit.Core
             }
             try
             {
-                System.Diagnostics.Process.Start(helpFile + parameter);
+                if (Configuration.IsRunningOnWindows || Configuration.IsRunningOnMac)
+                {
+                    System.Diagnostics.Process.Start(helpFile + parameter);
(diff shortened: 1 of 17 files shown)


libse/Utilities.cs
```diff
@@ -852,8 +852,25 @@ public static void ShowHelp(string parameter)
             {
                 helpFile = "https://www.nikse.dk/SubtitleEdit/Help";
             }
-
-            System.Diagnostics.Process.Start(helpFile + parameter);
+            try
```
niksedk: Could this be just `UiUtil.OpenURL(helpFile + parameter)` ?

Looks good :)
Nice to encapsulate "process start" as it also seem broken in dot net core: https://github.com/dotnet/runtime/issues/17938

NickZ: This file is in libse, and I didn't want to create a dependency on SubtitleEdit for this.

Issue #3972, opened 2020-02-08
UTF-8 or UTF-8-BOM?


uckthis:
Hello,
Recently one of my client sent back one of the file I sent saying it was UTF-8-BOM when they want a UTF-8. Frankly, I didn't and don't the difference between the two. Encoding was set to UTF-8 in SE, but when I checked the encoding in Notepad++ it was UTF-8-BOM and I converted it with Notepad++. 
Is it possible to rename the current encoding as UTF-8-BOM and add another encoding which is UTF-8?
Thank you.

xylographe:
Which subtitle format?

uckthis:
@xylographe SRT. Any file. Open with Notepad+= and under encoding it says UTF-8-BOM.

niksedk:
@uckthis: To change the encoding "UTF-8" in SE to save without BOM is a bad idea as SE has always written UTF-8 with BOM (except for a few formats that requires the absence of BOM). Also, some programs require presence of an UTF-8 BOM to function correctly.

In a perfect world UTF-8 without BOM would suffice, but there are still a lot of people using other encoding types, e.g. ANSI encoding (which naturally sucks) . The BOM is an excellent way to tell computers how a file is encoded and is by default added to UTF-8 files when using .NET framework.
Some programs (especially on Linux) cannot handle BOM and file appending can also give problems regarding BOM.

In SE, perhaps "UTF-8" could be split into "UTF-8 with BOM" and "UTF-8 without BOM"...

uckthis:
@niksedk You know best. Splitting is how they do it in Notepad++. 
Default should be UTF-8-BOM and if someone needs without BOM, he can just switch it.

Thank you.

niksedk:
@uckthis: How is latest beta: https://github.com/SubtitleEdit/subtitleedit/releases/download/3.5.13/SubtitleEditBeta.zip ?

uckthis:
@niksedk Perfect!

niksedk:
Closed via commit 0f28d058ccb2c5942d95cdbf182c526b02641a63

PR #3979, opened 2020-02-09
Add new encoding "UTF-8 without BOM"

niksedk:
Notepad and Notepad++ (now) have two different encodings for UTF-8 - with and without BOM, and there seems to be more and more requests about this - like #3972.

This PR will rename the old "UTF-8" to "UTF-8 with BOM" and add a new encoding to the encodings combo-box in the UI:
image
Also, the "secret" setting `WriteUtf8Bom` has been removed.

ATM, BOM is not really part of the `Encoding` object in C#, so it requires a bit more care than just an normal `Encoding`.


src/Logic/CommandLineConvert/CommandLineConverter.cs
```diff
@@ -219,13 +219,31 @@ private static int Help(string[] arguments)
...
+                    {
+                        targetEncoding = new TextEncoding(Encoding.UTF8, TextEncoding.Utf8WithBom);
+                    }
+                    else if (encodingName.Equals("utf-8-no-bom", StringComparison.OrdinalIgnoreCase) ||
+                             encodingName.Equals(TextEncoding.Utf8WithoutBom.Replace(" ", string.Empty), StringComparison.OrdinalIgnoreCase) ||
+                             encodingName.Equals(TextEncoding.Utf8WithoutBom.Replace(" ", "-"), StringComparison.OrdinalIgnoreCase))
+                    {
+                        targetEncoding = new TextEncoding(Encoding.UTF8, TextEncoding.Utf8WithoutBom);
+                    }
                     if (encodingName.Length > 0)
```
xylographe:
```
else if (encodingName.Length > 0)
```

OmrSi:
@niksedk Is the default without BOM?

niksedk:
Yes, the default is with BOM as it has always been (you can change the default in Option - Settings though)

OmrSi:
Cool. It must have been a problem with my settings, it defaulted to "without BOM" for some reason.
Thank you.

niksedk:
Oh, if you open a file already in "UTF-8 wihtout BOM" that encoding will be kept.

PR #3993, opened 2020-02-16
Add setting for dialog style

niksedk:
Different types of dialog style has been requested many times (mostly by Dutch people).
We have a plugin (DialogueOneHyphen) but a deeper integration is needed for this to work properly inside SE.

So, therefore this new setting for dialog style (in settings profile) with these options:
 - DashBothLinesWithSpace
 - DashBothLinesWithoutSpace
 - DashSecondLineWithSpace
 - DashSecondLineWithoutSpace

This should work in `Merge as dialog` and `Split line` (EDIT: + `Toggle dialog dashes` and `Remove text for HI`)
`Fix common errors` should be able to convert between dialog styles.
OCR fixes should not touch anything regarding dialogs.

Fix common errors shows which dialog style is applied:
image

This PR is also explained nicely in #3416

image

Beta updated: https://github.com/SubtitleEdit/subtitleedit/releases/download/3.5.13/SubtitleEditBeta.zip


OmrSi:
Well, here is one for ya.
"Toggle dialog dashes" shortcut still uses "Dash both lines with space" even if you select otherwise.

niksedk:
Thx @OmrSi :)
Beta updated with "Toogle dashes" fix: https://github.com/SubtitleEdit/subtitleedit/releases/download/3.5.13/SubtitleEditBeta.zip

OmrSi:
In general, it works.
In one case, if you have a line with no punctuation at the end of the first one, it just gives the original one with spaces.
Like on this:
```
You're good
Thanks
```
Will always, no matter the style, give:
```
- You're good
- Thanks
```

OmrSi:
Also, since you're working on this, any chance you can make "Toggle dialog dashes" work on the original?
Like, if you have a line you want to toggle the dashes in, it only affects the Normal and doesn't work on the original, you have to add it manually there.

Flitskikker:
Wow nice, great work! I will do some testing. :)

OmrSi:
@niksedk Thank you.
Now there are 2 issues,
The first one is, now when you toggle on dashes for a line that only has an original text, you can't toggle them off.
image

The second one is, you made "Toggle" only work if the text could be dialog.
Here you have a problem, I will only speak about my language, not all people use punctuations, so when we have a line like:
```
لقد رحل
إلى أين؟
```
Toggle dashes wouldn't work.

OmrSi:
Well, this is a doozy. ^^"
The original is toggling well now, thank you.
But the other one, if you have a line with no punctuation, it always uses the first kind.

OmrSi:
@niksedk Okay, imagine this with me.
You are translating a file with translator mode turned on.
You arrive at a line that has dialog dashes in the original text box.
You translate the line in the normal text box and toggle dashes.
In this case, you will have to use the toggle dashes shortcut twice, once to remove the dashes in the original subtitle and once to toggle them on in both.
This shouldn't happen, just like toggling italics or music notes from inside the text box, it should only affect the current text box.
Using it on the whole line should affect both.

OmrSi:
The original thing is all good now. Thank you. :D

niksedk:
@OmrSi: thx for testing :)
Hopefully close now!

@Flitskikker: Do you know if the profiles are correct regarding "dialog style": https://github.com/SubtitleEdit/subtitleedit/blob/master/libse/Settings.cs#L1071 ?

OmrSi:
Thank you for all this hard work, everything is looking good on my end. :D

Flitskikker:
> @Flitskikker: Do you know if the profiles are correct regarding "dialog style": https://github.com/SubtitleEdit/subtitleedit/blob/feature/add-setting-for-dialog-style/libse/Settings.cs#L1159 ?
> (Don't know how French dialog style is supposed to be...)

#3994 :)

OmrSi:
https://github.com/SubtitleEdit/subtitleedit/issues/4002

OmrSi:
You really put a lot of work into this. Thank you.
Are you waiting for more testing to merge?

niksedk:
It's almost done... I need to test OCR with dialogs and see what happens.

Ding-adong:
There was some changes in how fix (add dash) and fix (remove dash), fix common errors, works last year. In the last few months I had noticed a single speaker has a dash per line, making the assumption there are two speakers. I had to correct it while watching the programme. It happens when there are two sentences, broken into two lines like below. Today I just found out why.

 ```
- So what is that?
What does that mean?
```

Fix (add dash) change it into
 ```
- So what is that?
- What does that mean?
```
when it is suppose to be
 ```
So what is that?
What does that mean?
```

This is wrong. The purpose of fix (add dash) is to fix 'second speaker dash' such as dutch subs into dash for each speaker in eng sub. See top of thread for examples.

 ```
So what is that?
- What does that mean?
```

becomes
 ```
- So what is that?
- What does that mean?
```
is correct.

**Comment by Ding-adong** [comment]
This is an example why 'break long lines' in fix common errors need to done after 'fix (remove dash)'.

```
you would've been okay?
- Then I would still have it, yeah. (35 characters which is more than max chars per lines of 34)
```
becomes
```
you would've been okay? - Then
I would still have it, yeah.
```
then 'fix dialogs on one line' becomes
```
you would've been okay?
- Then I would still have it, yeah.
```
back to the same as the first. A horrible loop that refused to use 'fix (add dash)' until I uncheck 'break long lines.'

It should work like this.
```
you would've been okay?
- Then I would still have it, yeah. (35 characters which is more than max chars per lines of 34)
```
becomes (fix (add dash))
```
- you would've been okay?
- Then I would still have it, yeah. (35 characters which is more than max chars per lines of 34)
```
becomes (break long lines)
```
1; start time; end time; duration; You would've been okay?
2; start time; end time; duration; Then I would still have it, yeah.
```

else (fix (add dash))
```
- you would've been okay?
- Then I would still have it. (less than max chars per lines of 34)
```
No break long lines required.

Ding-adong:
Great, that worked for me. Cheers.

PR #4004, opened 2020-02-19
timer autobackup: do not force initialization for optional component!

ivandrofly:


niksedk:
:)

A little text in the PR description is always nice - could have been `Do not load auto-backup timer into memory if not used`.

Would crash on line 4530 with `_timerAutoBackup.Stop();` if not started with auto backup active.

PR #4007, opened 2020-02-23
Fix encoding detection on non-Windows platforms

NickZ:
This pull request fixes ANSI encoding detection for MacOS and Linux by changing the dependency necessary for detection from MLang (mlang.dll, i.e. MultiLang) to UTF.Unknown, which is compatible across all platforms and is compatible with .net Standard 1.0 and up.

I only use UTF.Unknown for detecting input encoding. I left the MLang dependency in because LibSE uses it for detecting output encoding, which UTF.Unknown doesn't do, but SubtitleEdit doesn't use it.


niksedk:
Looks good to me - I tested with a few ANSI files and actually got better results :)

PR #4012, opened 2020-02-26
Improve auto-break lines

ivandrofly:
Improves dialog matching for Arabic language!

### DIFF
--- a/libse/Utilities.cs
+++ b/libse/Utilities.cs
@@ -469,17 +469,19 @@ namespace Nikse.SubtitleEdit.Core
                 if (noTagLines.Count == 2)
                 {
                     const string LineCloseChars = ".?!)]♪";
-                    var arr0 = noTagLines[0].Trim().TrimEnd('"', '\'').TrimEnd();
+                    char[] quoteChars = { '"', '\'' };
+
+                    var arr0 = noTagLines[0].Trim().Trim(quoteChars).Trim();
                     if (language == "ar")
                     {
-                        if (arr0.EndsWith('-') && noTagLines[1].TrimStart().EndsWith('-') && arr0.Length > 1 && (LineCloseChars.Contains(arr0[0]) || arr0.StartsWith("--", StringComparison.Ordinal) || arr0.StartsWith('–')))
+                        if (arr0.EndsWith('-') && noTagLines[1].TrimEnd(quoteChars).EndsWith('-') && arr0.Length > 1 && (LineCloseChars.Contains(arr0[0]) || arr0.StartsWith("--", StringComparison.Ordinal) || arr0.StartsWith('–')))
                         {
                             return text;
                         }
                     }
                     else
                     {
-                        if (arr0.StartsWith('-') && noTagLines[1].TrimStart().StartsWith('-') && arr0.Length > 1 && (LineCloseChars.Contains(arr0[arr0.Length - 1]) || arr0.EndsWith("--", StringComparison.Ordinal) || arr0.EndsWith('–')))
+                        if (arr0.StartsWith('-') && noTagLines[1].TrimStart(quoteChars).StartsWith('-') && arr0.Length > 1 && (LineCloseChars.Contains(arr0[arr0.Length - 1]) || arr0.EndsWith("--", StringComparison.Ordinal) || arr0.EndsWith('–')))
                         {
                             return text;
                         }
--- a/libse/Utilities.cs
+++ b/libse/Utilities.cs
@@ -457,7 +457,7 @@ namespace Nikse.SubtitleEdit.Core
 
         public static string AutoBreakLinePrivate(string text, int maximumLength, int mergeLinesShorterThan, string language, bool autoBreakLineEndingEarly)
         {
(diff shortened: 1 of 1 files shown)


libse/Utilities.cs
```diff
@@ -468,17 +468,20 @@ public static string AutoBreakLinePrivate(string text, int maximumLength, int me
...
                 {
-                    var arr0 = noTagLines[0].Trim().TrimEnd('"', '\'').TrimEnd();
+                    const string LineCloseChars = ".?!)]♪";
+                    char[] quoteChars = { '"', '\'' };
+
+                    var arr0 = noTagLines[0].Trim().Trim(quoteChars).Trim();
                     if (language == "ar")
                     {
-                        if (arr0.EndsWith('-') && noTagLines[1].TrimStart().EndsWith('-') && arr0.Length > 1 && (".?!)]♪".Contains(arr0[0]) || arr0.StartsWith("--", StringComparison.Ordinal) || arr0.StartsWith('–')))
+                        if (arr0.EndsWith('-') && noTagLines[1].TrimEnd(quoteChars).EndsWith('-') && arr0.Length > 1 && (LineCloseChars.Contains(arr0[0]) || arr0.StartsWith("--", StringComparison.Ordinal) || arr0.StartsWith('–')))
```
ivandrofly: In arabic we want to trim from the end not start, since the hyphen will be placed from end of the line!

libse/Utilities.cs
```diff
@@ -468,17 +468,20 @@ public static string AutoBreakLinePrivate(string text, int maximumLength, int me
                 var noTagLines = HtmlUtil.RemoveHtmlTags(text, true).SplitToLines();
                 if (noTagLines.Count == 2)
                 {
-                    var arr0 = noTagLines[0].Trim().TrimEnd('"', '\'').TrimEnd();
+                    const string LineCloseChars = ".?!)]♪";
+                    char[] quoteChars = { '"', '\'' };
+
+                    var arr0 = noTagLines[0].Trim().Trim(quoteChars).Trim();
```
ivandrofly: Trim quotes chars from both side... the language is not checked yet so we don't know if it is going to be Arabic or not!

OmrSi:
@ivandrofly The problem here is that we don't always reverse it, so we don't always have to trim from the end.
If you're working on a file for a company, it's required that you don't reverse the punctuation marks.

ivandrofly:
> @ivandrofly The problem here is that we don't always reverse it, so we don't always have to trim from the end.
> If you're working on a file for a company, it's required that you don't reverse the punctuation marks.
> The same thing applies when you use a Unicode control character for reversing.

Could you be more specific? are you talked about Arabic dialogue-paragraph or dialogue-paragraph general?

OmrSi:
Just Arabic.
Here are the 3 cases: [Test.zip](https://github.com/SubtitleEdit/subtitleedit/files/4256399/Test.zip)
The first one is not reversed, which is used when making a file for a company, for work.
The second one is reversed via normal reverse, where you but the punctuation at the end, which is used in fansubs.
The third one is reversed using Unicode control characters, as we do for files taken from Netflix.
What I mean is, it's not always added to the end, we use "toggle dialog" then reverse if it's not for work.
I don't know if this is related to your request, I'm just stating in case it is.

ivandrofly:
> Just Arabic.
> Here are the 3 cases: [Test.zip](https://github.com/SubtitleEdit/subtitleedit/files/4256399/Test.zip)
> The first one is not reversed, which is used when making a file for a company, for work.
> The second one is reversed via normal reverse, where you but the punctuation at the end, which is used in fansubs.
> The third one is reversed using Unicode control characters, as we do for files taken from Netflix.
> What I mean is, it's not always added to the end, we use "toggle dialog" then reverse if it's not for work.
> I don't know if this is related to your pull request, I'm just stating in case it is.

Okay, I get you point. Is not mandatory for the arabic dialog end with hyphen in order to indicate a dialog? 

I will update the PR to considerer this option. Thank you!

OmrSi:
No.
I'm saying you have to add a "-" to indicate dialog, and it has to show up at the start of the line.
In order to make it show up at the start, you have to reverse, then it will be written at the end of the line because it's RTL.
But if you were doing a file for a company, they ask you not to reverse, so then it's just left at the start.

OmrSi:
There are 2 ways of reversing, using Unicode control chars leaves the dashes at the start of the line.
Using Reverse RTL moves them to the end.
When fansubbing, we just the "Reverse RTL" method, when reversing a file taken from Netflix, we use the control chars.

OmrSi:
This is a fansub: https://subscene.com/subtitles/jumanji-the-next-level/arabic/2144188
This is a file from Netflix: https://subscene.com/subtitles/locke-and-key-first-season/arabic/2140427
And a work file is just a file that is not reversed.

ivandrofly:
> I'm saying you have to add a "-" to indicate dialog

Sure, I get that! what I mean is that is not mandatory for arabic sub to have "-" at the end - which you also mentioned.

There is no updated required though!

OmrSi:
Great. I just wanted to point out that fact in case it affects the pull request. :D

PR #4016, opened 2020-03-02
Update ru-RU.xml

Elheym:


src/Languages/ru-RU.xml
```diff
@@ -82,6 +82,7 @@
     <DeleteCurrentLine>Удалить текущую строку</DeleteCurrentLine>
     <Width>Ширина</Width>
     <Height>Высота</Height>
+    <Collapse>Collapse</Collapse>
```
niksedk: Is this missing? (context menu in shortcuts)
image

Elheym: Yeah, you're right.

PR #4020, opened 2020-03-03
Improve line break functionality

ivandrofly:
- Improves title check
- Improves dialog check
- allow single line dialog auto-break
    * AutoBreak button can now auto-break single line dialog e.g: 

> \- What? - Bak Man, if you see...

### DIFF
--- a/libse/Utilities.cs
+++ b/libse/Utilities.cs
@@ -192,43 +192,59 @@ namespace Nikse.SubtitleEdit.Core
 
         internal static bool CanBreak(string s, int index, string language)
         {
-            char nextChar;
+            char charAtPos;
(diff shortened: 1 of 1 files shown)


libse/Utilities.cs
```diff
@@ -499,7 +515,10 @@ public static string AutoBreakLinePrivate(string text, int maximumLength, int me
             }
 
             string s = RemoveLineBreaks(text);
-            if (HtmlUtil.RemoveHtmlTags(s, true).Length < mergeLinesShorterThan)
```
ivandrofly: By doing this, if you pass a single line dialog which its length is < mergedLinesShorterThan, it will be skipped

PR #4027, opened 2020-03-08
Subtitle: Improve AdjustDisplayTimeUsingSeconds

ivandrofly:
- do not loop O(n) when only one paragraph is going to be re-adjusted its duration
- extract codes to a method to avoid duplications


niksedk:
Good catch that it was not using "MinimumMillisecondsBetweenLines" :)

Issue #4059, opened 2020-03-22
[Feature request] Add (or copy) support for maximum single line length based on pixels (safe area)


Flitskikker:
In theory, the maximum line length of subtitles is defined as maximum characters per line. However, because not all letters have the same width, lines with the same number of characters can have varying widths.

Spot, the software most professional Dutch subtitlers use, therefore detects "too long lines" not based on the amount of characters, but based on exceeding the safe area.

Take for example this subtitle in SE, which is marked too long because it exceeds 42 CPL:
![image](https://user-images.githubusercontent.com/3516155/77255510-71363d00-6c68-11ea-82a5-7511d992f46e.png)

However, as Spot shows, it does fit in the safe area, because the second line contains a lot of narrow characters:
![image](https://user-images.githubusercontent.com/3516155/77255534-96c34680-6c68-11ea-853e-d100a8e38d1d.png)

Fortunately, SE offers the "Check line width" plugin, which I use extensively to check my subtitles' "actual" width. I did some comparing and I found that these settings are a close enough match to Spot's safe area:
![image](https://user-images.githubusercontent.com/3516155/77255624-1d782380-6c69-11ea-8740-7d36b511570c.png)
(I chose Arial 16, which is Spot's editor font, but it's changed to 15,75 for some reason.)

My request is to copy this functionality to SE itself, because it's quite a hassle to open the plug-in everytime, scroll to the line I want to check, make changes etc.

My idea would be to just add a label next to the CPL counter, and populating it with the values as seen in the plug-in. For example:

![image](https://user-images.githubusercontent.com/3516155/77255697-b0b15900-6c69-11ea-99a4-c72963bc2831.png)
_(Exceeds CPL, but fits within safe area)_

![image](https://user-images.githubusercontent.com/3516155/77255742-fbcb6c00-6c69-11ea-84b8-d0ec816caf43.png)
_(Does not exceed CPL, but exceeds safe area)_

The pixel limit and used font would be added to the settings.
The text in the ListView will be highlighted red when it exceeds either the CPL or pixel value.
The "Auto br" could change into "Auto br using pixel width" (as in the plug-in) on Shift+Click/Alt+Click or depending on what error.

I looked into adding this feature myself, but the Main.cs form has a lot going on. It's also quite a change to the "basics' UI", so maybe it's best to run it by you first.

Thanks! :)

OmrSi:
There is a plugin called "Check line width". It might have something to do with this.

[...]

PR #4065, opened 2020-03-29
Add line width (in pixels) syntax validation, labels and statistics

Flitskikker:
This is for #4059.

It's now possible to toggle the actual widths in pixels next to the character counter.
As not all characters have the same width, the "maximum amount of characters per line" is not always reliable for checking if the subtitles don't exceed the safe area. Now you can safely check this using this feature.

image

image

(also works in translator mode)

Syntax coloring can be toggled from the settings:

image

Click the button to modify the actual calculation settings, and to toggle the new UI options.

image

By default, all options are turned off. The default calculation settings are based on Spot's safe area settings (used by professional Dutch subtitlers/broadcasters). ([Reference](https://user-images.githubusercontent.com/3516155/77255534-96c34680-6c68-11ea-853e-d100a8e38d1d.png): Arial 24pt, 720 px - 10% left and right safe area)

Also generates statistics when enabled:

image

Notes:
- I didn't add it to profiles, because I think it's quite niche and not really widely used across all languages. Can always be added later.
- Text widths are calculated using `TextRenderer.MeasureText`. It seems like `Graphics.DrawString`, as used by the plugin, is less accurate. `TextRenderer.MeasureText` is also not 100% accurate (it leaves some padding even when padding is disabled). I tried drawing the text to a Bitmap and then calculating the exact width (not using GetPixel but reading directly from memory using BitmapData), which gave more accurate results, but it was too slow. Now using `TextRenderer.MeasureText` in combination with a correction value, based on testing and comparing with Spot. Not ideal, but it works in most cases and is at least more accurate than the CheckLineWidth plugin.
- All settings and i18n added.

Please test it and let me know what you think!


niksedk:
Looks good to me :)

Two/three minor issues:
1) TextWidth should dispose the `measureFont`.
2) No need for a setting regarding statistics - just always show it.
3) If the main win width is very small - perhaps the pixels lengths should be hidden?

OmrSi:
Why always show it? If a person isn't interested in this, he wouldn't activate it and it wouldn't show up in statistics.
I'm saying, if the option to `color text if too wide` is activated, show it in statistics, if not, don't.
This is just my opinion. :D

OmrSi:
I also think it should always show the width under the text box if `color text if too wide` is activated, or else you would get a colored line without an apparent reason:
image

Why is it red here? It's not clear.
@niksedk What do you think?

Flitskikker:
@OmrSi @niksedk 
Thanks for your comments! I'll remove the two checkboxes and toggle those options based on whether "Color text if too wide" is enabled or not.

Flitskikker:
@niksedk 
Issues fixed. Thanks for your feedback.

PR #4069, opened 2020-03-30
audiovisulizer: refact / update

ivandrofly:
- fix method naming
- prefer struct when the type member isn't being changed frequency

### DIFF
--- a/src/Controls/AudioVisualizer.cs
+++ b/src/Controls/AudioVisualizer.cs
@@ -21,7 +21,7 @@ namespace Nikse.SubtitleEdit.Controls
             End
         }
 
-        public class MinMax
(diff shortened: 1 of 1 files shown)


src/Controls/AudioVisualizer.cs
```diff
@@ -2137,7 +2137,7 @@ private bool EndParagraphDueToLowVolume(Subtitle subtitle, int blockSizeMillisec
             return subtitleOn;
         }
 
-        private MinMax GetMinAndMax(int startIndex, int endIndex)
+        private WaveStats GetWaveStats(int startIndex, int endIndex)
```
ivandrofly: Use more descriptive method name

Issue #4026, opened 2020-03-08
[Feature request] Split doesn't change end time


OmrSi:
Right now, when you split a line, it adds the gap to the end time.
Can we have an option to keep the end time fixed when splitting?
Most times, you already set the end time to audio, so you don't want it to change when you split.
I remember someone already mentioned this here before.

niksedk:
OK, now gap time will be subtracted from the two new paragraphs duration (half each).
Does that work for you?

OmrSi:
It's usually subtracted from the second one.
As you are splitting at the place you want the first line to end with, so you're already ending the first line where you want it to end.
By "usually " I mean on Netflix's subtitling platform.

niksedk:
OK, if user supplies split time, then gap will be subtracted from second sub - if no split time from user, then gap is taken from both subs evenly... sounds/works okay?

OmrSi:
Sounds good to me, and it's working well.
Thank you, as always.

PR #4075, opened 2020-03-30
Add setting for split behavior

Flitskikker:
Related to commit 0c657ae245c947b20f06cd9237f6e70d468e88b4 and #4026:
Make split behavior configurable (from which subtitle the gap will be substracted). For me it's more logical to set the gap to the left of the split point (Spot also does this).

image

Setting & i18n added.


OmrSi:
@Flitskikker Don't you think it should go to the second line when it's: "add gap to the left"?

Flitskikker:
@OmrSi I'm not quite sure what you mean. Could you please explain?

The idea is, when it's set to "Add gap to the left of split point":

I set the split point at the position where I want the second line to start (which is a few frames before the start of speech):

image

It then sets the start time of line 2 to the split point, and adds the gap at the left side of the split point, deducting time from line 1:

image

OmrSi:
I mean, since you split it where the second starts, should the selection now be on the second?
I working on Netflix's platform and this is what happens, when you split it leaves the gap to the left and moves to the second line. (The selection moves to the second line)
In your picture, line 100 would be selected.

Also, what do you get from the Spectrograms?

Flitskikker:
@OmrSi 
Ah, I understand. Yes, good idea actually! I'll see if I can change it tomorrow.
Do you think it should selected the left subtitle when it's set at "Add gap to the right of the split point"?

I find the spectograms really useful to distinguish speech, especially when there's a lot of background noise:

image

OmrSi:
Yes, it already does that.
Because when it's to the right, you'd be setting the end of the first line.
While when it's on the left, you'd be setting the beginning of the second line.

Nice, I didn't know it can be used for that.

OmrSi:
@Flitskikker You just need to add go to next to the condition of `SplitBehavior == 0` right?

Flitskikker:
@OmrSi 
Yes. Had some GitHub issues yesterday.
Just pushed the fix: #4082

PR #4076, opened 2020-03-30
Waveform improvements (align gridlines, snap to shotcuts, single-click select)

### LINKED ISSUE #3413: [Improvement] Several waveform improvements
Here are some improvement suggestions I've collected:

- [ ] **Add an option for single-clicking to select subtitles**
At the moment, one can only double-click to select a subtitle from the waveform. It could be useful to also add an option to use a single-click for this.

- [ ] **Properly align gridlines**
At the moment, the gridlines are aligned differently on every scroll tick. I'm not sure how these are being drawn, but I think it would be better to align them to the seconds markers,
se2

- [ ] **Snap to shot cuts**
I'm always syncing my subtitles to shot cuts. I'm using a seperate application that I wrote to snap all my cues to shot cuts if they're not exactly on them. It would be nice to have SE automatically snap subtitles to shot cuts when moving and resizing subtitles in the waveform. Preferably (if possible) with an accuracy setting on the Waveform tab and the ability to ignore snapping while holding the Alt key (same key as in e.g. PowerPoint).

Flitskikker:
This fixes #3413.

- Added an option for single-clicking to select subtitles in waveform
- Properly align gridlines with the time ticks on every scroll tick and zoom level
- Added an option to snap to shot cuts when resizing or moving subtitles, and when making selections

Added settings in UI:

image


OmrSi:
Good stuff.
What about adding a shortcut for "snap start/end to scene change"?

Flitskikker:
@OmrSi 
Thanks. I added the following shortcuts:

image

Would those work for you?

While I was at it, I also added two shortcuts that I've been missed since switching from SubtitleWorkshop to SubtitleEdit (a long time ago), extend to previous/next subtitle.

There was already an "Extend until next subtitle", but it was capped to the subtitle maximum length and did only work with 1 selected subtitle.

Also added them to the context menus (listview and waveform):

image

OmrSi:
This is some awesome work. <3

Flitskikker:
And also added the "Go to next/previous subtitle" feature - that was only accessible using a shortcut - to the waveform contextmenu, so easily go to the previous or next subtitle instead of endlessly scrolling. 😁

image

OmrSi:
@Flitskikker The issues I mentioned are related to this.
Extending to previous doesn't affect the original when the previous is +1 seconds away from the current which causes the text box to be grayed out.

PR #4079, opened 2020-04-02
Audio visualizer: Update / refact

ivandrofly:
remove unneeded method and variable

### DIFF
--- a/src/Controls/AudioVisualizer.cs
+++ b/src/Controls/AudioVisualizer.cs
@@ -1133,7 +1133,7 @@ namespace Nikse.SubtitleEdit.Controls
                 {
                     double seconds = RelativeXPositionToSeconds(e.X);
                     var milliseconds = (int)(seconds * TimeCode.BaseUnit);
-                                        
+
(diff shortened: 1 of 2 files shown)


src/Controls/AudioVisualizer.cs
```diff
@@ -2108,15 +2108,19 @@ private double GetAverageVolumeForNextMilliseconds(int sampleIndex, int millisec
...
+            {
+                return 0;
+            }
+
             double v = 0;
-            int count = 0;
             for (int i = from; i < max; i++)
             {
                 v += _wavePeaks.Peaks[i].Abs;
-                count++;
```
ivandrofly, line 2116: there is no need be counting here!

src/Forms/Main.cs
```diff
@@ -17738,11 +17738,6 @@ private void VideoEnded(object sender, EventArgs e)
             mediaPlayer.Pause();
         }
 
-        private static VideoInfo ShowVideoInfo(string fileName)
```
ivandrofly, line 17741: There is no need to have this method!

PR #4115, opened 2020-04-13
Add total pixel width in main window

Flitskikker:
To complete the pixel width feature, this PR will add the total pixel width to the totals label under the textbox:

image

(Only if it's enabled in syntax options)


Flitskikker:
@niksedk 
Forgot to explicitly test this yesterday with the translator mode, but just tested it and it works fine for that as well!

PR #4121, opened 2020-04-16
Fix extend option/shortcuts not committing changes

Flitskikker:
After using the extend to previous/next shortcuts, the changes would be undone when selecting another subtitle.


OmrSi:
@niksedk Can you merge this?

PR #4126, opened 2020-04-18
Main: ensure empty lines are removed after removing formatting

ivandrofly:
Empty lines will remaining if contains only formatting text, after running "Remove all formatting" they will still be kept

Original subtilte:
image

After: 
image

TEST SUBTITLE: https://subscene.com/subtitles/fantasy-island/english/2191181

### DIFF
--- a/src/Forms/Main.cs
+++ b/src/Forms/Main.cs
@@ -26935,20 +26935,21 @@ namespace Nikse.SubtitleEdit.Forms
             ToggleBookmarks(true);
         }
 
-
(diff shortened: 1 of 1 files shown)


niksedk:
Nice catch and nice description with images and sample file :)

niksedk:
Hm, was somewhat slow... apparently indexing `SelectedIndices` a quite slow!
The commit 432d05ec01977ff211bd3a048f00e898b5bf2269 make "Remove all formatting" go from 4000 ms to 17 ms on the above sub :)

PR #4137, opened 2020-04-21
FIxCommas : Improvements

ivandrofly:
- remove unnecessary group capturing
- cache regex for inverted commas
- optimize comma removing before closing characters e.g: ?!
- include "]" & ")" in closing characters
- introduce method which as more meaningful names, 
- add unit-test for "]" and ")"


niksedk:
Nice :)
Very good performance in the "RemovePrefixCommaSymbol" method!

(also don't know why I had the string.Replace in a while statement... dough)

PR #4144, opened 2020-04-23
Some continuation style improvements

Flitskikker:
Some fixes:
- Improved detection of actual quotes versus 'explanation quotes' around one phrase (e.g. `This is a 'test'` > `This is a 'test'...`, instead of `This is a 'test...'`)
- Improved detection of dialogs when merging (sometimes a dialog dash got removed because it was incorrectly seen as a continuation dash)

Added tests accordingly.

### DIFF
--- a/libse/ContinuationUtilities.cs
+++ b/libse/ContinuationUtilities.cs
@@ -121,7 +121,7 @@ namespace Nikse.SubtitleEdit.Core
             // Remove single-char quotes from the ending
             if (checkString.Length > 0 && Quotes.Contains(checkString[checkString.Length - 1]))
             {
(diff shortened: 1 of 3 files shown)


niksedk:
Nice, some of the unit tests are commented out... should they be deleted or commented in?

Flitskikker:
@niksedk 
Thanks! The tests that are commented out can be deleted for now.
They were for some changes that might be desirable for Dutch subtitles, but not by default.

For example:
`We are doing "this,"` will now be changed into `We are doing "this..."` when using dots. Personally, I'd like `We are doing "this"...` better, but it would be wrong to force that.

For now, it will just follow the source position. So `We are doing "this,"` will be `We are doing "this..."`, and `We are doing "this",` will be `We are doing "this"...`.

PR #4151, opened 2020-04-25
namelist: prevent adding names when all word are already in name list

ivandrofly:
*_Prevent adding name list **Bill Clinton** when both **Bill** and **Clinton** are already in single word name list._

Ps: There is another PR that is on it's way, which will remove all names with all words already in name list

### DIFF
--- a/libse/Dictionaries/NameList.cs
+++ b/libse/Dictionaries/NameList.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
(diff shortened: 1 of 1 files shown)


xylographe:
IIRC **Bill** is blacklisted in `en_names.xml`. Hence, it would make sense to have **Bill Clinton** in the name list.

ivandrofly:
> IIRC **Bill** is blacklisted in `en_names.xml`. Hence, it would make sense to have **Bill Clinton** in the
> global name list.

Good catch :)...

PR #4192, opened 2020-05-20
spelling correction

vinod-svj:
It looked like a spelling mistake in the code where "&Cancel" was typed as "C&ancel".  Please review and commit if it ok.


niksedk:
Hi vinod-svj,

Thx for the PR :)

Yes, it does look a bit funny, but that's on purpose. The letter after `&` will be a shortcut, reachable via Alt+letter.

PR #4216, opened 2020-06-04
Cloud Speech-To-Text API

MehdiZouaoui:
Hello everyone,

Sorry, I'm really new to Github and I didn't know where to submit my request but I wonder if it is possible that we have the ability to add Cloud Speech-To-Text API key to Subtitle Edit. There is only Google Translate which is paid service. Speech to Text will save the time of a lot of people because once they have the script in text they can directly translate it using normal Google Translate and just polish the translation. I downloaded a Json file from my Google Cloud console and but I have no idea what to do with it.

Thank you again and sorry if I'm writing that in the wrong place.


xylographe:
Please, open a [new issue](https://github.com/SubtitleEdit/subtitleedit/issues/new) to post your request.

PR #4286, opened 2020-07-28
add script file to clear test-results

ivandrofly:
This test results folder is never cleared if not manually, this script will help clearing the junk left by unit-test
it can also be integrated with visual studio "Clean"


niksedk:
OK, I did not know that... I had 500+ mb.
But I think this should just be manual though...

Issue #4225, opened 2020-06-07
[Bug] Merging text changes comma


OmrSi:
Hey, I have just noticed this.
I had two lines:
```
1
00:00:01,000 --> 00:00:02,053
يا فتيات، لديكن 5 دقائق

2
00:00:02,136 --> 00:00:03,000
لإنهاء كل التفاصيل.
```
And when I use and type of Merge, the Arabic comma is changed to an English one:
```
يا فتيات, لديكن 5 دقائق
لإنهاء كل التفاصيل.
```

niksedk:
Hm, it does not do that here... how do you merge?

OmrSi:
Well, I've been testing with the settings file for 40 minutes.
I got the cause, it happens when you have: `<ContinuationStyle>NoneTrailingDots</ContinuationStyle>`

OmrSi:
It seems like it's not doing `ConvertBackForArabic` but only `ConvertToForArabic`.

Flitskikker:
Thanks @niksedk for the swift fix!

OmrSi:
@niksedk This is also causing original English text to be changed to Arabic punctations.
Before split:
![image](https://user-images.githubusercontent.com/20923700/89210002-0149a300-d5c8-11ea-8c38-b6a8bfc47303.png)

After spilt:
![image](https://user-images.githubusercontent.com/20923700/89209998-ff7fdf80-d5c7-11ea-8c6d-63b9e7774832.png)

The same happens with commas.

PR #4293, opened 2020-08-03
Fix bug in split with continuation style / Arabic

OmrSi:
Fix https://github.com/SubtitleEdit/subtitleedit/issues/4225#issuecomment-668146463


niksedk:
Good catch :)

Issue #4323, opened 2020-08-24
New shortcut needed (a modification of an existing one) "Extend selected lines to next scene change "


m0ck69:
Hello.
I found a shortcut which is very helpful when making subtitles.
"Extend selected lines to next scene change (or subtitle)".

However, according to some general translating rules, a subtitle cannot end at a scene change, it must end before it.
Usually it is 2 frames gap.

Can you add a new shortcut **"Extend selected lines to next scene change (or subtitle) with a gap"**, please?

It would help a lot and speed up the subtitling process significantly!

OmrSi:
Or maybe add an option somewhere that lets you select which behavior you want...

OmrSi:
@niksedk Can you check the pull request? Is it enough? Do you like where the option is?

m0ck69:
I'll just add that the gap doesn't always have to be 2 frames.
The program should read the gap value from here:

![gap](https://user-images.githubusercontent.com/46628545/91142936-feffd380-e6b1-11ea-9c34-357bcfa81025.jpg)

niksedk:
Beta updated with fix from OmrSi :)
https://github.com/SubtitleEdit/subtitleedit/releases/download/3.5.16/SubtitleEditBeta.zip

PR #4325, opened 2020-08-24
Add Extend to next scene change behavior

OmrSi:
Netflix rules say that the out-cue should be 2 frames before the "Shot change".
The existing shortcut only extends the out-cue to the Shot change.
In this pull request, I've added an option to change this behavior, making it according to Netflix's timing rules.

I've noticed that the Combobox is a couple of pixels to the right, I didn't know how to fix that.

Closes https://github.com/SubtitleEdit/subtitleedit/issues/4323


niksedk:
I like the separate shortcut option better... and it should use the "min gap between subtitles ms" value...

OmrSi:
I'll do another pull request with a shortcut then.

PR #4326, opened 2020-08-24
Fonts

ivandrofly:
Remove duplication and more readable

### DIFF
--- a/src/Logic/UiUtil.cs
+++ b/src/Logic/UiUtil.cs
@@ -463,31 +463,35 @@ namespace Nikse.SubtitleEdit.Logic
                     font = new Font(Configuration.DefaultLinuxFontName, 8F);
                 }
 
-                try
+                if (IsFontPresent(font.Name))
                 {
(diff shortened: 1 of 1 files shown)


src/Logic/UiUtil.cs
```diff
@@ -463,31 +463,35 @@ private static Font DefaultSystemFont
...
+                if (IsFontPresent(font.Name))
                 {
-                    // Bold + italic + regular must be present
-                    font = new Font(font.Name, 9, FontStyle.Bold);
-                    font = new Font(font.Name, 9, FontStyle.Italic);
-                    font = new Font(font.Name, 9, FontStyle.Regular);
+                    _defaultSystemFont = font;
                 }
-                catch
+                else if (IsFontPresent(SystemFonts.DefaultFont.Name)) // system default
```
niksedk, line 470: I don't think we need to check if SystemFonts.DefaultFont exists on the current system...

ivandrofly: yes!

PR #4332, opened 2020-08-26
replace with stringbuilder to avoid creating new string instance

ivandrofly:


niksedk:
Slower in my tests :(

PR #4336, opened 2020-08-29
Work on batch filters

OmrSi:
Added a new batch filter for MKV language code:
image

And made "Filename contains..." apply when importing files manually.
This is in case there are a lot of files that the user wants to select a range of them and apply the "filename" filter to then.

### DIFF
--- a/src/Forms/BatchConvert.cs
+++ b/src/Forms/BatchConvert.cs
@@ -269,6 +269,7 @@ namespace Nikse.SubtitleEdit.Forms
             comboBoxFilter.Items[2] = l.FilterMoreThanTwoLines;
             comboBoxFilter.Items[3] = l.FilterContains;
             comboBoxFilter.Items[4] = l.FilterFileNameContains;
+            comboBoxFilter.Items[5] = l.MkvLanguageCodeContains;
             comboBoxFilter.SelectedIndex = 0;
             comboBoxFilter.Left = labelFilter.Left + labelFilter.Width + 4;
             textBoxFilter.Left = comboBoxFilter.Left + comboBoxFilter.Width + 4;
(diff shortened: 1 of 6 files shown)


src/Forms/BatchConvert.cs
```diff
@@ -519,202 +520,223 @@ private void buttonInputBrowse_Click(object sender, EventArgs e)
 
         private void AddInputFile(string fileName)
         {
-            try
+            if (comboBoxFilter.SelectedIndex == 4 && textBoxFilter.Text.Length > 0 && !fileName.Contains(textBoxFilter.Text, StringComparison.OrdinalIgnoreCase))
```
OmrSi, line 523: This is the first change.

src/Forms/BatchConvert.cs
```diff
@@ -519,202 +520,223 @@ private void buttonInputBrowse_Click(object sender, EventArgs e)
...
+                        {
+                            item = new ListViewItem(fileName);
+                            item.SubItems.Add(Utilities.FormatBytesToDisplayFileSize(fi.Length));
+                            listViewInputFiles.Items.Add(item);
+                            item.SubItems.Add("Matroska/VobSub - " + lang);
+                            item.SubItems.Add("-");
+                        }
+                        foreach (var lang in mkvSrt)
+                        {
+                            if (comboBoxFilter.SelectedIndex == 5 && textBoxFilter.Text.Length > 0)
```
OmrSi, line 679: And this is the second change.
The rest is the same.

OmrSi:
@niksedk Is the code good enough or should I close this and find a way to improve the logic?

OmrSi:
@niksedk What about this one? Close or merge?

src/Forms/BatchConvert.cs
```diff
@@ -519,202 +520,223 @@ private void buttonInputBrowse_Click(object sender, EventArgs e)
...
+                            {
+                                if (lang.Contains(textBoxFilter.Text, StringComparison.OrdinalIgnoreCase))
```
OmrSi, line 681: I don't know if there is a better way to formulate this without having to write the same code twice.

OmrSi:
Closed for a better logic and a crash fix.

PR #4350, opened 2020-09-04
Fix a crash in batch convert + add new filter

OmrSi:
Fixed a crash in batch convert, It used to crash when you press "Delete" when the list view is empty.

Added a new batch filter for MKV language code:
image

And made "Filename contains..." apply when importing files manually.
This is in case there are a lot of files that the user wants to select a range of them and apply the "filename" filter to then.


niksedk:
Cool :)

(sorry, I don't have too much time atm, so things will be a bit slower than usual)

Issue #3565, opened 2019-05-08
[Request] "Allow overlap" setting in Adjust durations dialogue


Moerges:
The "Adjust durations" dialogue notes that "Display time will not overlap start time of next text", and in most cases, this is exactly the desired behaviour. However, for some operations, it would be useful to have the option to allow overlap.

Hence my question: would it be possible to add an "Allow overlap" checkbox to the "Adjust durations" window?

Thank you for your consideration—and for the wonderful tool that Subtitle Edit is in general.

OmrSi:
I think this issue should be closed.
The PR did exactly what the author said, but Nik said he doesn't think it's really needed.
Writing a plugin for it is not worth it, it would be exactly the same with just one extra option.

Issue #3276, opened 2019-01-04
[Bug?] Adjust Durations...


OmrSi:
It seems like from version 3.5.5, adjusting duration doesn't work on already overlapped lines.
It's good that it doesn't cause overlap when changing the duration, but shouldn't it work on already overlapped lines because the overlap is probably intended?

OmrSi:
Closed because fixing https://github.com/SubtitleEdit/subtitleedit/issues/3565 would fix this.

Issue #4125, opened 2020-04-18
Bug in "Adjust durations" with lines shorter than minimum duration


Flitskikker:
I found a bug in the "Adjust durations" feature:

When using the Adjust via seconds option:
Lines that are shorter than the minimum duration (1400 ms in my case) don't get adjusted properly.

For example, take this subtitle:

![image](https://user-images.githubusercontent.com/3516155/79638692-26381880-8187-11ea-9e0c-a5c3035af282.png)

When I add 40 ms, the line that is too short gets reset to the minimum duration, causing overlap:

![image](https://user-images.githubusercontent.com/3516155/79638702-3f40c980-8187-11ea-835a-0e77628cdbd9.png)

When I substract 40 ms, the line that is too short will remain unchanged:

![image](https://user-images.githubusercontent.com/3516155/79638756-76af7600-8187-11ea-9f47-755b7ec2d5cb.png)

OmrSi:
An issue related to "Adjust duration" https://github.com/SubtitleEdit/subtitleedit/issues/3276

OmrSi:
I think this is closed with https://github.com/SubtitleEdit/subtitleedit/commit/349c562a8f4a66438cb5f717c03dc6088e9b7574

PR #4353, opened 2020-09-06
Add "Allow overlap" option in "Adjust duration"

OmrSi:
image
 I don't think it should work with "Recalculate" because you would want just one line on screen with optimal reading speed.

Also added to "Batch convert" now:
image

Closes https://github.com/SubtitleEdit/subtitleedit/issues/3565, Closes https://github.com/SubtitleEdit/subtitleedit/issues/3276
https://github.com/SubtitleEdit/subtitleedit/issues/4125 is related to "Adjust durations" and also closed.


OmrSi:
The overloading mechanism is used 5 times throughout the file, I don't know if I should change anything...

OmrSi:
The optional parameters method is used 5 times throughout the file, I don't know if I should change anything...

OmrSi:
I don't think I need to change anything...

OmrSi:
@niksedk I don't think it should work with "Recalculate" because you would want just one line with optimal reading speed.

Do you want me to use overloading to solve the Codacy issue? Optional parameters are used in 4 other places in the file, I think it should all stay the same.

Please let me know if something needs editing when you have time.

OmrSi:
Added to "Batch":
image

OmrSi:
@niksedk Please let me know if you want me to change anything.

OmrSi:
@niksedk I don't think it should work with "Recalculate" because you would want just one line on screen with optimal reading speed.

Do you want me to use overloading to solve the Codacy issue? Optional parameters are used in 4 other places in the file, I think it should all stay the same.
But I can change it in the whole file if you want.

Also added to "Batch convert" now:
image
Please let me know if you want me to change anything.

OmrSi:
Closing this as @niksedk thinks it would be better as a plugin.
I don't think adding a plugin for the same function with the same code just for an extra option is good, but that's not my call.

PR #4358, opened 2020-09-08
ensure encoding form dispose + store encoding in tag

ivandrofly:


src/Forms/BatchConvert.cs
```diff
@@ -2681,20 +2681,22 @@ private void buttonBrowseEncoding_Click(object sender, EventArgs e)
             openFileDialog1.Filter = UiUtil.SubtitleExtensionFilter.Value;
             if (openFileDialog1.ShowDialog(this) == DialogResult.OK)
             {
-                var chooseEncoding = new ChooseEncoding();
-                chooseEncoding.Initialize(openFileDialog1.FileName);
-                if (chooseEncoding.ShowDialog(this) == DialogResult.OK)
+                using (var chooseEncoding = new ChooseEncoding())
```
ivandrofly, line 2684: add `using`

src/Forms/ChooseEncoding.cs
```diff
@@ -63,11 +63,12 @@ public void Initialize(string fileName)
             foreach (var enc in Configuration.AvailableEncodings)
             {
                 var item = new ListViewItem(new[] { enc.CodePage.ToString(), enc.WebName, enc.EncodingName });
-                listView1.Items.Add(item);
+                item.Tag = enc;
```
ivandrofly, line 66: store encoding in item tag for faster retrieve later on...

src/Forms/Main.cs
```diff
@@ -13102,13 +13102,15 @@ private void ToolStripMenuItemManualAnsiClick(object sender, EventArgs e)
             openFileDialog1.Filter = UiUtil.SubtitleExtensionFilter.Value;
             if (openFileDialog1.ShowDialog(this) == DialogResult.OK)
             {
-                var chooseEncoding = new ChooseEncoding();
-                chooseEncoding.Initialize(openFileDialog1.FileName);
-                if (chooseEncoding.ShowDialog(this) == DialogResult.OK)
+                using (var chooseEncoding = new ChooseEncoding())
```
ivandrofly, line 13105: add `using`

src/Forms/ChooseEncoding.cs
niksedk: Is actually slower in my tests...

ivandrofly: That is strange... Encoding.GetEncoding is actually calling `EncodingProvider.GetEncodingFromProvider(name)` which loops through a array of encoding... maybe there is some cache of something I will investigate more!

image

image

Issue #4359, opened 2020-09-09
extend sel. lines to previous scene with min. gap


m0ck69:
Hello.

Lately @OmrSi fulfilled my request and @niksedk added a very helpful shortcut "Extend selected lines to next scene change with min. gap". (https://github.com/SubtitleEdit/subtitleedit/issues/4323)
Thanks once again!

I think there should be also the opposite version "Extend selected lines to previous scene change with min. gap".

When you have time, please add this shortcut as well.

OmrSi:
@niksedk Before I start. Do you want this in SE?

m0ck69:
Yes, the same like here (#4323)

OmrSi:
Done.
Like the last time, he will merge it if he wants it.

m0ck69:
That was fast. Thx so much!

OmrSi:
Anytime.
Well, I have a college exam tomorrow, so... anything to avoid studying. xD
Also, this is more fun.

PR #4360, opened 2020-09-09
Add "Extend to previous scene change with gap" shortcut

OmrSi:
Closes https://github.com/SubtitleEdit/subtitleedit/issues/4359

I don't see a use for it but it's was a request and now it's your call.


m0ck69:
There is a use because some clients (Disney for instance) requires that subtitle should start not at a scene change but some frames after it so this shortcut can be very helpful in that case while working on synchronization.

PR #4365, opened 2020-09-10
batch convert: cache subtitle formats

ivandrofly:


src/Forms/BatchConvert.cs
```diff
@@ -2468,7 +2471,7 @@ private void ScanFiles(IEnumerable<string> fileNames)
                                 var format = sub.LoadSubtitle(fileName, out _, enc, true, null, false);
                                 if (format == null)
                                 {
-                                    foreach (var f in SubtitleFormat.GetBinaryFormats(true))
```
ivandrofly, line 2471: for every file dropped a list of new formats will be created...  (don't do that!)

ivandrofly:
@niksedk should we consider this PR, if not feel free to close!

PR #4380, opened 2020-09-20
Add "Ignore Arabic Diacritics" + Fix bugs

OmrSi:
I didn't expect to run into these things when I started working on this.
This has a lot of changes, I hope you can review it before merging.

### 1. Bug in FCE characters count:
It used to show the newline as a character:
image

### 2. Bug in "Split line!"
It used to show up sometimes -when the words are long- even if the line can be broken into 2:
image

### 3. Bug in Remove line breaks in short text with only one sentence
It didn't work with Arabic because `s != s.ToUpperInvariant()` can't be True for Arabic and other languages that don't use casing.
Those other languages should be added as well.

### 4. Added "Text can fit on one line" to Netlix's Quality checker
In Netflix's rules, if the text can fit on one line, it should be written in one line.
I added this to "Number of lines".

### 5. Made SE not count the control character used by it `\u202B` as a character in characters count and CPS.
Other control character can be added.

### 7. Added a hidden option to ignore Arabic diacritics from characters count and CPS.

### DIFF
--- a/libse/Forms/FixCommonErrors/EmptyFixCallback.cs
+++ b/libse/Forms/FixCommonErrors/EmptyFixCallback.cs
@@ -53,26 +53,14 @@ namespace Nikse.SubtitleEdit.Core.Forms.FixCommonErrors
             get { return new SubRip(); }
         }
 
-        private string _language = "en";
-
-        public string Language
-        {
-            get { return _language; }
-            set { _language = value; }
-        }
+        public string Language { get; set; } = "en";
 
         public bool IsName(string candidate)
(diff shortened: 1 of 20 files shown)


niksedk:
Looks good to me :) 
I'll do some testing.

(only related changes should in a PR... e.g. netflix + statistics should have been separate PRs - next time...)

OmrSi:
I said that to myself after I made the pull request. 😅
It's like the issues, each issue opened should be related to one issue.
Won't happen again.

Issue #303, opened 2014-10-08
Progress status in title bar


niksedk:
Translate progress: percent with subtitle orginal and new subtitle. (percent per line translated)
New translation progress: percent with only new subtitle (blank) and translated per movie time.
movie 60min: Translated to 30 minutes is 50%

(some sort of auto-detection of mode or ?)

Global90:
Please add percent in SubtitleEdit for both methods.
1. new subtitle for translate with movie file(get Current time video and Calculation percent).

2. news subtitle for translate with open orginal subtitle (translate mode) . get Current line translated and Calculation percent.

![untitled](https://cloud.githubusercontent.com/assets/8215910/6041209/74df1966-ac8f-11e4-8fef-2c92eae943a8.gif)

niksedk:
@Global90: Could you test this version: http://www.nikse.dk/SubtitleEdit.zip (no installer, beta) ?

Global90:
i tested this version. but i dont see Progress(percent) in the SE.

i write in lines. and open orginal subtitle(translation mode) but any Progress.

Thankyou

niksedk:
it's in the lower left corner - statusbar...
![untitled]http://www.nikse.dk/Test.png)

niksedk:
But if you use something like GT to get a start it won't work...

Global90:
MR. Nikolaj i dont have GT

My PC:
Win7 x86
.Net 4.5

Step1: open SRT file (1700 Lines)
Step2: open orginal Subtitle (this File)
Step3: remove Lines to 10 Line
Step4: ReWrite Line Orginal Text
Step5: All Row (Column Text) to empty
Step6: Write in Column Text

![untitled](https://cloud.githubusercontent.com/assets/8215910/9294195/1992c166-445b-11e5-9dee-96a9f5592e9e.jpg)

![1](https://cloud.githubusercontent.com/assets/8215910/9294194/197e5f28-445b-11e5-9a38-0ed76e7b013a.jpg)

niksedk:
Did you use the beta I linked to?
Included the Settings.xml? (ShowProgress must be true)

Global90:
i download http://www.nikse.dk/SubtitleEdit.zip (3.4.8 build 107)

and add <ShowProgress>True</ShowProgress> in Settings.xml in tag <General> Because ShowProgress not found in Settings.xml.

again dont show Progress

niksedk:
This should be really easy:
[...]

PR #4387, opened 2020-09-24
Don't show progress for imported plain text

OmrSi:
Related to https://github.com/SubtitleEdit/subtitleedit/issues/303#issuecomment-462065206

It used to show a big percentage:
image


OmrSi:
@niksedk This is a minor one. :)

PR #4399, opened 2020-10-01
Update ru-RU.xml

Elheym:


niksedk:
Could you re-add the UTF-8 BOM header... just to be sure.

PR #4401, opened 2020-10-01
Make "Move word up-down in current" work with original

OmrSi:
If the listview is focused, it works on normal.
If the normal textbox is focused, it works on normal.
If the original textbox is focused, it works on the original.

Please take a look to make sure I didn't miss anything. :D

### DIFF
--- a/src/Forms/Main.cs
+++ b/src/Forms/Main.cs
@@ -14675,14 +14675,28 @@ namespace Nikse.SubtitleEdit.Forms
                 MoveFirstWordInNextUp();
                 e.SuppressKeyPress = true;
             }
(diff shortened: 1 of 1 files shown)


darnn:
Oh, great, this will be really handy!

src/Forms/Main.cs
```diff
@@ -15799,6 +15821,52 @@ private void MoveWordUpDownInCurrent(bool down)
             }
         }
 
+        private void MoveWordUpDownInCurrent(bool down, TextBox tb)
```
niksedk, line 15824: What about removing the old `private void MoveWordUpDownInCurrent(bool down)` ?

OmrSi:
@niksedk Is it good this way?

PR #4402, opened 2020-10-01
Make Move words between subs work with original

OmrSi:
If the listview is focused, it works on normal.
If the normal textbox is focused, it works on normal.
If the original textbox is focused, it works on the original.

### DIFF
--- a/src/Forms/Main.cs
+++ b/src/Forms/Main.cs
@@ -9480,13 +9480,20 @@ namespace Nikse.SubtitleEdit.Forms
             UpdatePositionAndTotalLength(labelTextLineTotal, textBoxListViewText);
         }
 
-        private void MoveFirstWordInNextUp()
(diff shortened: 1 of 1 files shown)


src/Forms/Main.cs
```diff
@@ -9496,26 +9503,41 @@ private void MoveFirstWordInNextUp()
...
         }
 
-        private void MoveLastWordDown()
+        private void MoveLastWordDown(TextBox tb)
         {
             int firstIndex = FirstSelectedIndex;
             if (firstIndex >= 0)
             {
-                var p = _subtitle.GetParagraphOrDefault(firstIndex);
+                var p = _subtitle.GetParagraphOrDefault(firstIndex); ;
```
niksedk, line 9532: empty statement ;)

OmrSi: Ugh. Sorry. I wonder why Codacy didn't catch it.

PR #4403, opened 2020-10-01
Add "Control + Shift + N" new group shortcut for "Multiple replace groups"

OmrSi:


niksedk:
Some weird extra change of shortcut key order + extra non related commit...

OmrSi:
@niksedk I just used "else if" instead of "if" in some places because only one shortcut can be used, no need to check for others.

And I changed the old `e.Control && !e.Shift && !e.Alt` with what you've been using `(Control | ...)`

And the weird changes are rearranging conditions parameters, I don't think it has an effect, it's just easier to read as you usually press a modifier then a key, so I thought it made more sense.

niksedk:
Too many different things... refactor + `Add shortcuts for moving groups to top and bottom` 
Normally if statements should have the fastest (normally the most specific condition) first (due to short circuiting).

OmrSi:
Cool, I'll separate.

PR #4405, opened 2020-10-01
Allow for "Control + Alt" shortcuts when a textbox is focused

OmrSi:
Shortcuts using "Control + Alt" didn't use to work when a Textbox is focused.


OmrSi:
@niksedk Was this combination emitted on purpose?

niksedk:
>Was this combination omitted on purpose?

No, I don't think so 👍

PR #4407, opened 2020-10-02
Redefine "CountCharacters" to ignore all Unicode Control Chars

OmrSi:
This now ignores all control characters.

I don't know if this makes it better or worse.
I just figured why ignore the tags here while you can remove them?
Please let me know what you think.

### DIFF
--- a/libse/Forms/FixCommonErrors/FixLongLines.cs
+++ b/libse/Forms/FixCommonErrors/FixLongLines.cs
@@ -16,7 +16,7 @@ namespace Nikse.SubtitleEdit.Core.Forms.FixCommonErrors
                 bool tooLong = false;
                 foreach (string line in lines)
                 {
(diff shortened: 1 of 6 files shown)


niksedk:
This method is optimized as it's used a lot where performance matters.
Added a fix for control chars here: https://github.com/SubtitleEdit/subtitleedit/commit/41d94783e97714d3fd1c3070ec5351badffd2ca2

We should change `HtmlUtil.RemoveHtmlTags(line, true).CountCharacters` to just `CountCharacters`, right?

OmrSi:
> We should change `HtmlUtil.RemoveHtmlTags(line, true).CountCharacters` to just `CountCharacters`, right?

Yes, I do think so.

> Added a fix for control chars here: [41d9478](https://github.com/SubtitleEdit/subtitleedit/commit/41d94783e97714d3fd1c3070ec5351badffd2ca2)

What about "Unicode" Control characters? 
Right now it just ignores "\u202B".

OmrSi:
They way I see it, Unicode control chars are like html tags, they shouldn't be counted as they aren't really text.

PR #4408, opened 2020-10-02
Work on multiple replace shortcuts

OmrSi:
Group context menu:
image

Rules context menu:
image

Also fixed some issues where the items aren't visible nor focused after moving them.


OmrSi:
@niksedk What about adding all those shortcuts to the context menu? Wouldn't that be better and clearer?

OmrSi:
@niksedk Working on it.

PR #4414, opened 2020-10-04
Fix: commandline export plaintext format bugs,  save/load some settin…

raidsan:
Fix: commandline export plaintext,   TimeCodeSeparator is miss,
and I found some save/load some settings value process wrong.


niksedk:
Nice fixes - thx :)

Issue #3063, opened 2018-08-26
[Enhancement after the next version] For measurement converter.


OmrSi:
Maybe Make it official instead of beta, and like "Find" when you open it, it takes the highlighted number.
Also, it would be useful to remember the last units you converted from and to, cause if you're working on a file with a lot of numbers, calculating the weight of something for example, it's hard to set the units every time and remember the window's position.

OmrSi:
Like This:
![screenshot 392](https://user-images.githubusercontent.com/20923700/45591514-95230f00-b95d-11e8-944e-5700cb103d51.png)

OmrSi:
@ivandrofly Maybe you could make it as a plugin?

ivandrofly:
@niksedk can you please assign me to this just so I don't forgot, would also make it easier to find in issues later on :)

darnn:
Does this exist currently? How can I get to it?

OmrSi:
@darnn  There Exists a version of it...
![screenshot 696](https://user-images.githubusercontent.com/20923700/48672813-53703980-eb43-11e8-9d9c-b7a27a9ed2d6.png)

You have to activate beta stuff for the settings file:
`<ShowBetaStuff>True</ShowBetaStuff>`

Also, Unrelated, I found it useful to activate: `<ShowProgress>True</ShowProgress>` if you haven't already.

darnn:
Oh, cool! What does ShowProgress do?

darnn:
Ah, I see. Neat!

OmrSi:
@ivandrofly Can you work on this one? :D

ivandrofly:
> @ivandrofly Can you work on this one? :D

I'll take look at this maybe in two weeks or so...

OmrSi:
@ivandrofly What about now? hehe

ivandrofly:
> ShowProgress Shows you how much of the file is done.
> If you have an original file opened, it shows the percentage in relation to the original.
> Else, It shows the percentage in relation to the video's time.

What result are you expecting? 😕

OmrSi:
@ivandrofly I'm not expecting any other result. ^^"
He was just asking what it does.

PR #4418, opened 2020-10-07
Enhance Measurement Converter

OmrSi:
Closes: https://github.com/SubtitleEdit/subtitleedit/issues/3063

image

1. Added a shortcut for "Measurement Converter".
2. Added categories and made it save the selected category/from/to upon closing
3. Added an "insert" button which inserts the converted value to the text box or replaces the selected text.


src/Forms/Main.cs
```diff
@@ -26677,8 +26678,46 @@ private void toolStripMenuItemDost_Click(object sender, EventArgs e)
...
+            var selectedText = string.Empty;
+            if (tabControlSubtitle.SelectedIndex == TabControlSourceView)
+            {
+                selectedText = textBoxSource.SelectedText;
+            }
+            else
+            {
+                if (textBoxListViewText.Focused)
+                {
+                    selectedText = textBoxListViewTextAlternate.SelectedText;
```
niksedk, line 26690: Should be `textBoxListViewText.SelectedText;`  ?

src/Forms/Main.cs
```diff
@@ -26677,8 +26678,46 @@ private void toolStripMenuItemDost_Click(object sender, EventArgs e)
...
+                }
+                else
+                {
+                    selectedText = textBoxListViewText.SelectedText;
```
niksedk, line 26694: Sould be `selectedText = textBoxListViewTextAlternate.SelectedText;`  ?

src/Forms/Main.cs
```diff
@@ -26677,8 +26678,46 @@ private void toolStripMenuItemDost_Click(object sender, EventArgs e)
...
+            _measurementConverter.Show(this);
+
+            _measurementConverter.ReplaceClicked += measurementConverter_ReplaceClicked;
+        }
+
+        private void measurementConverter_ReplaceClicked(object sender, MeasurementConverter.ReplaceEventArgs e)
+        {
+            if (IsSubtitleLoaded)
+            {
+                textBoxListViewText.SelectedText = e.Result;
```
niksedk, line 26719: Should also know which textbox to use.

OmrSi: I was thinking it should always use the main one, you don't usually change the units in the original file. 

What do you think? 
Also, if you think it should be the textbox that was focused, how would I get that? Since it loses focus when switching to the converter.

src/Forms/Main.cs
OmrSi: Ah, the confusion happened when I switched the conditions to include the case when text is selected but the focus is in the list view.

src/Forms/Main.cs
niksedk: I think it should work for both boxes. You can save flag in the "measurement converter" regarding which box was active.
Also check the  `GetFocusedTextBox()` method.

niksedk: Could you also merge `master` into your branch? (there's a conflict in Settings.cs)

OmrSi: I solved the conflict. :D
It was the "Remove unneeded initialization".

OmrSi:
@niksedk I actually did it, the issue was that if you the Converter is working and then you click outside the Converter and change the textbox then get back to the Converter using the mouse, it doesn't detect the change in focus.

I was trying to figure out how to solve that but couldn't find a way except using an event, and I don't think that's a good option, or making a global property which saves the focused textbox, and then the "Activated" event in the Converter would change the "FocsuedTextBox" for the event in the converter.

image

image

What do you think?

niksedk:
Made a possible fix: https://github.com/SubtitleEdit/subtitleedit/commit/a2b14f5fae7afab68be36bea752f2387771b142f

Also refactored a little: https://github.com/SubtitleEdit/subtitleedit/commit/a2b14f5fae7afab68be36bea752f2387771b142f

OmrSi:
That was a great idea. :)

PR #4425, opened 2020-10-10
Add a separate comment for Overlap in NQC

OmrSi:
Right now it shows the "2 frames gap" comment, I think it's better to show a separate comment for Overlap.


OmrSi:
@niksedk Can you take a look at this? 😃

niksedk:
Better :)

PR #4427, opened 2020-10-11
Unify shortcut name "ToggleTranslationOriginalInPreviews"

OmrSi:
It was named "MainToolsToggleTranslationOriginalInPreviews" in the settings file and "MainEditToggleTranslationOriginalInPreviews" in the shortcuts so it wasn't being imported correctly.


OmrSi:
@niksedk This is still needed after, "Export" saves it as `MainEditToggleTranslationOriginalInPreviews` while import tries to find `MainEditToggleTranslationOriginalInPreviews`.

PR #4428, opened 2020-10-11
Alpha style: update

ivandrofly:
- Ensure objects are disposed when not needed
- Reuse reusable objects.
- Increment shadow value by 1 since the decimal will be trimmed

image


niksedk:
thx 👍 
Many disposes missing...

(Please try not to mix refact and functionality change!)

PR #4437, opened 2020-10-16
textbox: don't create n instance of string  while mouse-keydown on textbox

ivandrofly:


src/Controls/SETextBox.cs
```diff
@@ -244,12 +244,17 @@ protected override void WndProc(ref Message m)
             if (m.Msg == WM_LBUTTONDOWN)
             {
                 long milliseconds = (DateTime.UtcNow.Ticks - _gotFocusTicks) / 10000;
-                if (milliseconds > 10)
+                if (milliseconds > 10 && SelectionLength - SelectionStart > 0)
                 {
                     _dragText = SelectedText;
```
ivandrofly, line 249: as long as the user keep the left-mouse-key down there will be this string copy

Issue #3267, opened 2019-01-01
[Feature request] Show video chapters in video player


OmrSi:
As you know, some videos have chapters, which I think helps in editing time for new versions, it's not a very important one, just thought I'd mention it as MPC and MPV already support it.
So they would show up on the video's progress bar and have a way to move between them.
![screenshot 47](https://user-images.githubusercontent.com/20923700/50576610-f93ee580-0e1d-11e9-88a2-69f8f4509a73.png)

As you said before,  `my time is limited and requests seem not to be`

Sorry about all these requests. 😅
So Take your time with this and thanks a lot for everything you do!

PR #4439, opened 2020-10-16
Add chapters support

OmrSi:
Closes: https://github.com/SubtitleEdit/subtitleedit/issues/3267

Added a menu item to show chapters for Matroska files on video and waveform.
Added shortcuts to move between chapters.

Some changes are just switching to auto property, I'm sorry if I should have done it in a separate commit.

Please let me know if you'd like me to make any changes.


OmrSi:
@niksedk Do you think it would be better to have the option as an item in the Video menu instead of a hidden option?

OmrSi:
I'm thinking of making this an item in the video menu instead of a hidden option.

OmrSi:
Okay, now it's shown in the video menu if the file is MKV, and shows a message if there are no chapters in the file.
I think it's better this way.

niksedk:
Ups... wrong branch, sorry ;)

OmrSi:
And to think all of this can be done with just `mediaPlayer.chapters = Task.Run(() => matroska.GetChapters());` in .NET 4.5. ^^"

niksedk:
Just give it a quick run... could you add the chapter text too?
 (we can also add that later)

niksedk:
Yes, after 4.5+ some code could be improved :)

OmrSi:
Yes, I can parse the chapters text, put where should it show up?

niksedk:
The text could show up in the waveform.

OmrSi:
Do you think it would be better as tuples or class?

Also, where on the waveform? I imagine it would cause an issue if it's overlapping with the subtitle text.
Wo you think it should show up on the right side in the middle of the chapter line?

niksedk:
I would prefer classes... very easy to make :)

OmrSi:
This was my first experience with tooltips, so please do check out the code. :D

OmrSi:
@niksedk I've noticed that the tooltip is flickering, I couldn't find a way to fix it.

niksedk:
Not really sure what you mean about tooltip? Screenshot?

Looks good :)
Perhaps you can use a background like bookmarks, as chapter text is a bit hard to read:
image
And perhaps another color than white for text/vertical line? Yellow or ?

OmrSi:
A tooltip shows up on video with the chapter's name :D
image

OmrSi:
I'm not very good with colors, so please do change it if it's not good.

niksedk:
The color looks fine to me ;)
Could you put the color in settings? 
And perhaps use it in the video player progress bar too?

OmrSi:
Should the user be able able to choose it from the UI or just in the settings file?

niksedk:
Settings is fine :)
Yeah, you're right. Don't look clear enough.

OmrSi:
Do you think red looks better for an initial color instead of purple?
image

And on a related note, should the color on the video take the color from the settings as well?

Issue #4384, opened 2020-09-22
CPU usage even when no file loaded


dw817:
I'm concerned that even when no SRT or other subtitle file is loaded the executable can use anywhere from 5-10% CPU doing ... absolutely nothing. If I leave it for hours, it jumps up to 15%.

What could possibly be happening in the executable when there is no file even loaded in it ?

OmrSi:
This is not happening with me:
![image](https://user-images.githubusercontent.com/20923700/93915468-c256e480-fd10-11ea-98e3-6c9f2ee10ddd.png)

niksedk:
Also zero cpu% here:
![image](https://user-images.githubusercontent.com/1261331/93915759-c121b980-fd08-11ea-9700-2ec3f567cb98.png)

dw817:
Glad it's not happening with you, my question still remains. Or let's put it this way. What could CAUSE that to happen as it is here ?
I just brought it back up NOW with no file and it's using 5.29% CPU.
I'd like to hear from the people that DO have this problem as it may be a rare or isolated incident and what they think might be causing it.
Simply stating "Not me" doesn't answer this question.

dw817:
![subtitleedit cpu](https://user-images.githubusercontent.com/4763989/93943384-08ec1500-fcf8-11ea-879a-9484d5c798ea.png)

dw817:
![09-22-20 System Info](https://user-images.githubusercontent.com/4763989/93944452-6ed99c00-fcfa-11ea-821a-2c0241315dfc.jpg)

dw817:
![09-22-20 System Info](https://user-images.githubusercontent.com/4763989/93944574-b95b1880-fcfa-11ea-8939-88fc4cb16084.jpg)

OmrSi:
Simply stating "Not me" is a first step in the course of identifying the cause of the issue.

dw817:
Reminds of a bunch of kids, really. Very well, OmrSi, do you have anything more concrete to to contribute towards the correct answer to my initial question ?

I've posted my system status. Not sure what other data in my computer would be deemed helpful to solving this question.

vinodthebond:
@dw817 , please be respectful to the hard work these guys are putting in.  This piece of software is by far the best free subtitling tool available out there.   There could be some memory leakage happening.  It would certainly take some time to look into this and find out where the leakage is happening.

[...]

PR #4462, opened 2020-10-29
Remove asterisk if the last line in an unsaved file is deleted

OmrSi:
The asterisk used to not go away even if the last line in a file was deleted and the file wasn't saved.
image

(The force pushes are just renaming the commit)

### DIFF
--- a/src/Forms/Main.cs
+++ b/src/Forms/Main.cs
@@ -8315,6 +8315,11 @@ namespace Nikse.SubtitleEdit.Forms
                 labelAlternateCharactersPerSecond.Text = string.Empty;
                 labelTextAlternateLineLengths.Text = string.Empty;
                 labelTextAlternateLineTotal.Text = string.Empty;
+
+                if (Text.Contains('*'))
+                {
+                    Text = Text.RemoveChar('*').TrimEnd();
+                }
             }
         }
 


OmrSi:
@niksedk What's the issue? You want it to be done in the same method where the asterisk is added?
That method returns when there are no lines so it's not removing it.

niksedk:
If your clear all lines, the asterisk should still show...

OmrSi:
That's if the file is saved, I think.
But here you just deleted everything, so nothing actually changed.
Or do you mean it's to show changes has been made and therefore undo and redo can be used?

niksedk:
This PR is a little buggy: When I open a subtitle and delete all lines, it shows no changes.

I've tried to fix the issue here: https://github.com/SubtitleEdit/subtitleedit/commit/35d5faff7309a1b6d8dfd33c62711e49229260f4
( was also close to just deleting the if statement from this https://github.com/SubtitleEdit/subtitleedit/issues/4384 )

OmrSi:
@niksedk Yes. This is what I meant here:
> That method returns when there are no lines so it's not removing it.

OmrSi:
Also, note that it was showing no change before this PR.

OmrSi:
Removing all lines in SE doesn't do anything. If you delete all lines and save the file, nothing is changed because you can't save an empty file.

Issue #2901, opened 2018-04-18
Video player suggestions


mkver:
Hello,

I have a few suggestions for the video players:
1. MPC-HC can also be run in a portable mode, but SE can only work with an installed version of MPC-HC and some special paths. It would be easier to let the user choose the path directly.
2. The same goes for mpv-1.dll. I know it is probably more uncommon in this scenario, but if one uses ffmpeg anyway it is advantageuos to use shared dlls.
3. It seems that mpv doesn't use hardware acceleration at all even when hardware acceleration is available for a codec. Maybe one could use the --hwdec=auto (or auto-copy if displaying the subtitles requires it) option by default (and because there are always problems for some, this default should be overrideable)?
4. Is there actually a way to specify which audio track to use if a video file contains more than one?

Greetings
mkver

ldgbc:
When you first load the video there is a popup asking which Audio track you want to use. As far as I know you can change it mid-way. You have to close and reopen it.

mkver:
The popup you mention only appears when one wants to add a waveform; and even then it only selects the language that you want to have the waveform for, but it does not affect the language that is played at all. (This is actually what led me to question whether it is impossible to change the language.) But now I found out that it is indeed possible: Just go to video and use "choose audio track". It works without closing and reopening. So sorry for 4.

niksedk:
1) MPC-HC does not have an api, so it's mostly amazing it does anything...

3) I've not looked to much into it, but I don't really see any difference... you can try different values in Options -> Video player -> download mpv (here you can also set the "video output")
https://mpv.io/manual/master/#video-output-drivers
Perhaps I should test `--hwdec=auto` on an older computer.

4) It's available in the video menu (at least for mpv and vlc).

mkver:
1. Before I opened this issue I have already checked whether there might be a hidden option for the location of MPC-HC; I found one in the settings "MpcHcLocation", looked a bit at the source and tested it, but it didn't work for me and so I thought that this is a stub from an earlier unfinished attempt to allow selection of a portable MPC-HC version. Now I took a closer look and found out why it didn't work. See the description of my pull request.
2. I have found a workaround for my 2. regarding alternative mpv-1.dll: I can just use hardlinks or symbolic links to use one ffmpeg/mpv version on my system and yet be able to use it in Subtitle Edit. Given that this is a very niche thing I retract this request.
[...]

PR #4465, opened 2020-10-31
Add waveform multitrack support

OmrSi:
@niksedk Well, I don't think this is done yet, but it's working and I wanted to get your notes and let you test it out.
So, please do and let me know what you think and what's missing/if something isn't working as it should.

Closes https://github.com/SubtitleEdit/subtitleedit/issues/2901, maybe.

### DIFF
--- a/libse/WaveToVisualizer.cs
+++ b/libse/WaveToVisualizer.cs
@@ -345,7 +345,7 @@ namespace Nikse.SubtitleEdit.Core
     {
         #region Movie Hasher -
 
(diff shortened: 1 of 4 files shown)


niksedk:
Wow, that works well 👍 
I've tested it a bit it and it just works! 
I don't see anything missing - do you?

OmrSi:
I don't see anything missing, but I didn't know if you had more ideas about this that you wanted to implement. :D

PR #4470, opened 2020-11-02
NQC - Update Netflix Shot Change timing rules + some fixes

OmrSi:
Netflix updated some timing rules with shot changes today, so I updated them in the tool.
Fixed some bugs with shot changes timing as well.

I've also made a "suggested fix" column available.
I might work more on the tools in a future PR.


matmaggi:
Hello @OmrSi 

I appreciate this, but I've noticed that when the out-cue is 1 frame before the shot change OR 3-11 frames before the shot change, the checker doens't warn you that something is wrong.

image

PR #4472, opened 2020-11-03
Add a hidden option to set Continuation Pause

OmrSi:
Netflix said today, `A pause has been defined in all TTSGs as being 2 seconds or longer`.

Right now, Fix continuation style suggests removing the ellipses when the gap between 2 lines is 300 or less, meaning it considers the pause as 300 ms, if the gap is more than 300 ms and the first line has ellipses, it doesn't suggest removing the ellipses.
Netflix said that a pause is 2 seconds, which means if the gap between two lines is 2 seconds or more, the ellipses should be removed, SE doesn't handle it this way.
In this PR I made the pause customizable, with the default of 2 seconds.

### DIFF
--- a/libse/ContinuationUtilities.cs
+++ b/libse/ContinuationUtilities.cs
@@ -1261,7 +1261,7 @@ namespace Nikse.SubtitleEdit.Core
 
         public static int GetMinimumGapMs()
         {
-            return Math.Max(Configuration.Settings.General.MinimumMillisecondsBetweenLines + 5, 300);
+            return Math.Max(Configuration.Settings.General.MinimumMillisecondsBetweenLines + 5, Configuration.Settings.General.ContinuationPause);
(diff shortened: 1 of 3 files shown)


OmrSi:
@niksedk Better?

niksedk:
We also have a setting `Configuration.Settings.Tools.BridgeGapMilliseconds`... perhaps that setting should be used. 
It guess it could have a default value of 2000.

OmrSi:
Hmm, BridgeGapMilliseconds is not the same.
A pause is where you would use ellipses or not.
Like, Netflix says gaps that are less than 12 frames, 500 ms, should be bridged, but a pause is 2000 ms, they are not related.
You can have 2 lines with more `gap > 500 && gap < 2000` which shouldn't be considered a pause.

PR #4474, opened 2020-11-04
Don't change default frame rate from source view

OmrSi:
If you select a frame rate then do some changes in source view, the default frame rate used to change.

### DIFF
--- a/src/Forms/Main.cs
+++ b/src/Forms/Main.cs
@@ -24451,13 +24451,11 @@ namespace Nikse.SubtitleEdit.Forms
                     return;
                 }
 
-                var frameRate = CurrentFrameRate;
                 var format = new Subtitle().ReloadLoadSubtitle(textBoxSource.Lines.ToList(), null, currentFormat);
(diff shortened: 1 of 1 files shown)


OmrSi:
How to replicate: 
1. Go to the source view. 
2. Add sone lines. 
3. Change the frame rate. 
4. Go back to list view. 
5. Create a new file. 

You will notice that the default frame rate changed.

niksedk:
Hm, I think either the line should be removed, and we should update the UI combobox... perhaps just remove it?

OmrSi:
Doesn't changing it in source view affect the subtitle? 
If it does, then shouldn't the "current frame rate" be changed?

niksedk:
Yes, just deleting the line is probably the most simple solution - normally the frame rate should be ignored, but I guess when starting from blank, the frame rate could be useful... ?

OmrSi:
Yes, I'm pasting a frame-based file and the frame rate should matter, I think.

PR #4499, opened 2020-11-12
Change frame color + Don't set font size for 3.5

OmrSi:
You were setting the font size to 14, that's too big.
Also, we're still in 3.5, which would cause the font size to change each time we open 3.5.18.

Change frame color, the old one makes the text box seem inactive.


OmrSi:
@niksedk This is still valid.

niksedk:
Closed via https://github.com/SubtitleEdit/subtitleedit/pull/4502/commits/5476733383042ccb7a18a3d8935f887d7499ea1d

PR #4502, opened 2020-11-14
Feature/use rich text box

niksedk:
Use RichTextBox control for subtitle text box which enables syntax coloring like this:

image

Should also make live spell check possible in the future.

Syntax coloring can be turned off (which will make SE use a normal TextBox control):
image


OmrSi:
Opening SE with the normal TextBox causes it to not use the selected font, but default to another font:
image

My font:
image

OmrSi:
Now the font isn't right when using syntax coloring:
image

niksedk:
Thx :)
Hopefully font is working now!

OmrSi:
Changing the text box type makes the text box active even without any lines:
image

PR #4511, opened 2020-11-15
Set Frame color in one more place

OmrSi:
When opening a file…


niksedk:
What looks wrong?

niksedk:
I love text and images in PRs :)

OmrSi:
I agree.
I just thought you'd remember this from the previous PR.

So, when you open a file, the frame color is:
image

While it should be:
image

Which changes when you enter the textbox for the first time.

OmrSi:
It looks inactive without that Window Frame color.

PR #4532, opened 2020-11-22
Fix center in RTL mode

OmrSi:
@niksedk What do you think?
Center wasn't working in RTL mode.


niksedk:
Cool, did not know that center and RTL could be combined - that works fine?

OmrSi:
Yes, it seems to work fine.
RTL moves it to the right if it was on the left even. :D

PR #4535, opened 2020-11-23
Fix ref to SubtitleEdit.resources.dll

OmrSi:
image


niksedk:
I don't experience the problem, but this should be fine.

PR #4545, opened 2020-11-30
Fix another typo for RichTextBox in Paste

OmrSi:
@niksedk **Also, double-clicking a word or using Control + Shift + Arrrow in the normal text box is wrong:**

image

I noticed you removed:
```
            if (m.Msg == WM_DBLCLICK || m.Msg == WM_LBUTTONDBLCLK)
            {
                UiUtil.SelectWordAtCaret(this);
                return;
            }
```

From WndProc, but it doesn't work even if I add it back.

### DIFF
--- a/src/ui/Controls/SETextBox.cs
+++ b/src/ui/Controls/SETextBox.cs
@@ -738,7 +738,7 @@ namespace Nikse.SubtitleEdit.Controls
             {
                 _textBox.Paste();
             }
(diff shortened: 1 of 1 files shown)


niksedk:
@OmrSi: Does double-clicking a word or using Control + Shift + Arrrow in the normal text box work better here:
https://github.com/SubtitleEdit/subtitleedit/pull/4547

PR #4547, opened 2020-12-01
Fix selecting a word in text box

niksedk:
Fix selecting a word so it stops at quotes etc.

Related to comment on https://github.com/SubtitleEdit/subtitleedit/pull/4545


OmrSi:
Hey, Nik.
Double-clicking is back to normal.

But I guess Control Shift Arrow has always been like this?
image

It doesn't select the quote in RichTextBox, though:
image

Issue #4560, opened 2020-12-06
SE 3.5.18 Beta 169: Shortcut for setting focus on Waveform not working


GrampaWildWilly:
I have a shortcut defined for setting the focus (in list view undocked) on the waveform but it isn't working.

This is a kind of minor issue I discovered as I was preparing this bug report.  In order to verify that I indeed had this shortcut defined, I did a search on the Shortcuts tab of the Settings dialog using "Ctrl" in the search string.  There was exactly one shortcut that appeared & it wasn't the one for waveform focus.  (I'll leave it as an exercise for you to find it.)  I thought that was strange so I looked more closely.  It turns out all but this one shortcut use the term "Control" in their definitions.  This is a rather disconcerting inconsistency.  I recommend using "Ctrl" across the board since it's short & I believe it is what's printed on the key on most keyboards.  On the other hand, it would be an easier fix to change the one "Ctrl" shortcut that exists today.  Easier, but I say less user-friendly.  I vote to change them all to read "Ctrl."

GrampaWildWilly:
Oops.  I need to amend this.  The shortcut works if focus is in the caption list.  The shortcut does not work if focus is in the text box.

PR #4564, opened 2020-12-08
Work on text box

OmrSi:
So, this is a bunch of stuff, I'll explain the best I can.

**Firstly**, The frame color is now changed when initializing the text box.
The frame color used to not change when you add the first line in a file or import a file from a video and such.
image

**Secondly**, there is no need to color the `H` in ass color tags, so I removed its color:
image

**Third**, fix an issue where the background color used to not change for light colors:
image

**Fourth**, I worked a bit on center alignment, I made it so it doesn't change each time the text is changed for the simple textbox, but rather when the text box is initialized, and for the rich text box, not it sets the alignment for when it enters the rich text box if it's empty instead of checking each time the text is changed, or when the Right to Left mode changes, it seems to work well with my tests.

**Fifth**, Closes: https://github.com/SubtitleEdit/subtitleedit/issues/4560


OmrSi:
@niksedk Sorry if this is too much for on PR, but they are all related to text box.

Issue #4530, opened 2020-11-22
[Bug report] bilingual split issue in 3.5.18


LeonCheung:
Basically, the end char of the selection in the text box (in the 2nd line of course) will be split into the next new subtitle, which is different and unexpected from what's in 3.5.17. Thanks.

LeonCheung:
In the latest beta, try using the bilingual split tool on a small example below, you'll see what I'm talking about:

```
Try split this line.
一二三四五六
```

OmrSi:
How do I test this exactly?
If I just use the shortcut, I get this:
![image](https://user-images.githubusercontent.com/20923700/101661528-0eccd900-3a51-11eb-9c1c-27d51ad30810.png)

LeonCheung:
@OmrSi Please see #2995 on how to use this magic. :)

LeonCheung:
![1111](https://user-images.githubusercontent.com/494788/101720872-aaa62580-3ae1-11eb-9fe3-34705305899c.gif)
Basically, there are two issues here:
1. cursor position for the non-English line is unexpected. There seems to be an offset.
2. Even splitting with expected selection, the result would also offset.

Both issues occurred for non-English lines.

OmrSi:
Does it also happen when "syntax coloring" is turned off?

LeonCheung:
Ah! You are right, hmm

LeonCheung:
Ah, so that's it, everything works fine by turning it off...

OmrSi:
@LeonCheung Should be fixed now.

LeonCheung:
Thank you so much for the fix,  OmrSi! Closing. All the best.

PR #4577, opened 2020-12-11
Fix for bilingual split in RichTextBox

OmrSi:
Fixes https://github.com/SubtitleEdit/subtitleedit/issues/4530

### DIFF
--- a/src/ui/Forms/Main.cs
+++ b/src/ui/Forms/Main.cs
@@ -9486,8 +9486,8 @@ namespace Nikse.SubtitleEdit.Forms
             {
                 if (_subtitle.Paragraphs.Count > 0 && SubtitleListview1.SelectedItems.Count >= 1 && SubtitleListview1.SelectedItems.Count < 10)
                 {
-                    e.SuppressKeyPress = true;
                     SplitSelectedLineBilingual();
+                    e.SuppressKeyPress = true;
                 }
             }
             else if (e.KeyData == _shortcuts.MainListViewAlignmentN1)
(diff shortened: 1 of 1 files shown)


src/ui/Forms/Main.cs
```diff
@@ -10551,6 +10551,7 @@ private void SplitSelectedLineBilingual()
             newParagraph.Text = text2;
             SubtitleListview1.Fill(_subtitle, _subtitleAlternate);
             SubtitleListview1.SelectIndexAndEnsureVisible(idx, true);
+            RefreshSelectedParagraph();
```
OmrSi, line 10554: @niksedk That makes more sense.
But I think this line is still needed.

OmrSi: @niksedk Look:
image

src/ui/Forms/Main.cs
```diff
@@ -9486,8 +9486,8 @@ private void TextBoxListViewTextKeyDown(object sender, KeyEventArgs e)
             {
                 if (_subtitle.Paragraphs.Count > 0 && SubtitleListview1.SelectedItems.Count >= 1 && SubtitleListview1.SelectedItems.Count < 10)
                 {
-                    e.SuppressKeyPress = true;
                     SplitSelectedLineBilingual();
+                    e.SuppressKeyPress = true;
```
OmrSi, line 9490: Don't you think this looks better switched? Like all the others, `e.SuppressKeyPress = true;` should be at the end.

Issue #4248, opened 2020-06-13
Jump to top when change made in waveform (beta)


Monovox:
In 3.5.15/Beta, when making a change in the waveform while main window in source view, both source and list jump to start of file.

OmrSi:
Don't close an issue and open it again, it's really pointless.
The developer sees everything and fixes what he can when he can.
If you had doubts about an issue you closed, you can reopen it. https://github.com/SubtitleEdit/subtitleedit/issues/4223

Monovox:
@OmrSi Sorry, I'm new to this platform.
When should I close an issue?

xylographe:
In most cases one would close an issue when the problem has been solved in **SubtitleEditBeta**.

PR #4586, opened 2020-12-15
Some work on list/source view

OmrSi:
Only update Source view when visible.
Save position in Source view on time change.
Focus shortcuts support for Source view.

Add "Go to source view" to List view's context menu.
Make a context menu for Source view and add "Go to list view" in it.

Closes https://github.com/SubtitleEdit/subtitleedit/issues/4248

### DIFF
--- a/LanguageMaster.xml
+++ b/LanguageMaster.xml
@@ -1258,7 +1258,6 @@ Go to "Options -&gt; Settings -&gt; Tools" to enter your Google translate API ke
     <Controls>
       <SubtitleFormat>Format</SubtitleFormat>
       <FileEncoding>Encoding</FileEncoding>
-      <ListView>List view</ListView>
       <UndoChangesInEditPanel>Undo changes in edit panel</UndoChangesInEditPanel>
       <Previous>&lt; Prev</Previous>
       <Next>Next &gt;</Next>
--- a/LanguageMaster.xml
+++ b/LanguageMaster.xml
@@ -1259,7 +1259,6 @@ Go to "Options -&gt; Settings -&gt; Tools" to enter your Google translate API ke
       <SubtitleFormat>Format</SubtitleFormat>
       <FileEncoding>Encoding</FileEncoding>
       <ListView>List view</ListView>
-      <SourceView>Source view</SourceView>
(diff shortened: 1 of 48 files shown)


niksedk:
Commit cddda749c5a6cb209380ba952ae956b5255ce793 should be moved to a separate PR.
Commit 9fa0ae4b9e639c7836367befc2a4f1aa48e00648 should have the refact removed.

It would be nice to remove the tab control that contains list-/source-view... but that can be later too.

OmrSi:
Yeah, I'm working on removing the tab control, it's why this is still a draft. :D

I've already done it, I think.
But I still need to figure out the events linked to the old tab control.
image

niksedk:
Looks good, probably about 2 extra rows displayed now :)

To re-map events, just do it in the UI - you need to remap these:
image

OmrSi:
@niksedk It's actually the Tab-control events that need reamping.
`TabControlSubtitleSelectedIndexChanged` and `TabControlSubtitleSelecting`.

OmrSi:
I changed `TabControlSubtitleSelectedIndexChanged ` to `VisibleChanged` of the `splitContainerListViewAndText`,
so the one I'm thinking about is `TabControlSubtitleSelecting`.

OmrSi:
Just to give you some context, since the tab-control is removed, now I'm working with `textBoxSource` and `splitContainerListViewAndText`.

OmrSi:
Here goes nothing… :)

OmrSi:
That was crazy!

OmrSi:
@niksedk Go a chance to test it out?

niksedk:
Yes, I've tested a while it works nicely :)

I think the list/source toggle shortcut should have a default value... perhaps F2?

OmrSi:
> Yes, I've tested a while it works nicely :)
Great!

> I think the list/source toggle shortcut should have a default value... perhaps F2?
Yeah, seems like a good idea.

Monovox:
Hi 
Getting rid of the tabs, will it be possible to add an optional icon in the tool bar for this list/source toggle?
Maybe it makes sense.

Issue #4598, opened 2020-12-18
New F2 Feature Issue


LeonCheung:
In latest beta, when trying to press F2 while working with formats like EBU_STL, it shows "Not supported!" message, but it cannot be switched back to list view by simply pressing F2 again.

PR #4599, opened 2020-12-18
Fix not going back to list when source isn't supported

OmrSi:
Fixes https://github.com/SubtitleEdit/subtitleedit/issues/4598


OmrSi:
@niksedk Is there a better way to break out of the main if statement here in order to not repeat the activation code?
Because it just needs to continue from `if (inListView)`.
`goto` is an option, but I don't know if it's better.

Issue #2361, opened 2017-04-30
Hotkey to switch between Translate and Create Tabs


JanuszChmiel:
Dear C# specialists, 
I AM visually impaired user of Subtitle Edit. Yours app is 100 % accessible but to have better results while synchronizing with no possibility to use wave form or Visual Sinc, I would like to have hod key to switch between Translate and Create Tabs. Subtitle Edit have already hod key to switch between listview and Edit box. Algorithm work reliably and do not cause app or system crashes. Would somebody of us try to add hodkey to switch between Create and Translate TAB? Please do not include algorithm to switch to Duration TAB if you will consider to do that, it would be perfect to have separate hodkey for this task.
Thank you very much for yours time and programmers effort.
I have setup IDE to be able to compile from source code. I have successfully compiled stable source code of Subtitle edit 3.5.2. But Subtitle edit is very complex app and I AM helpless to find which C source code would had to modify.

niksedk:
Closed by PR #4609

PR #4609, opened 2020-12-19
Remove modes tab control and use buttons

OmrSi:


OmrSi:
@niksedk I forgot, Closes #2361

PR #4611, opened 2020-12-19
Some work on dark theme

OmrSi:
1. The tool strip had a line under it that was not colored.
image
image

2.  ComboBox/ToolStripComboBox/NumericUpDown styles should be like this so they don't take a full white color.
3. No need for `_isConfigUpdated` since it's now set from the settings.
4. I do believe the list view's gridlines look better, the list view looks too empty without them, so I added a hidden option to show them if someone said they want them.
5. The header should be a bit darker than the rest.
6. Only change waveform color if the user hasn't set it to another color before.
7. Don't change listview's color when drawing the header.
8. Add check icon for ToolStripMenuItems:
image


OmrSi:
@niksedk Check it out. :)

niksedk:
Looks nice - except 4 + 6...

OmrSi:
4 is a hidden option with a default of "false".
It shouldn't be an issue.

As for 6, I'll see what I can do.

OmrSi:
@niksedk And now?

niksedk:
> 4 is a hidden option with a default of "false".
> It shouldn't be an issue.

OK, it just turned on last time I tried the branch.

OmrSi:
This is why they say don't force push. xD

OmrSi:
@niksedk Good now?

niksedk:
Hm, I still get the grid + the old settings... ?

OmrSi:
Use a clean settings file?

GridLines was probably set to true before, but it should default to false on new settings files.

>the old settings... ?

Apparently, I have to use `ToArgb()` as `==` compares more than just the values.
Also, you weren't setting the ListView's forecolor, which means it needed a restart to take effect.
Should be fixed in the latest commit.

niksedk:
OK, crashed here:
image

Needs something like `Math.Max(BackColor.R - 9, 0)` ...

OmrSi:
Makes sense. My bad, didn't take that into account.

niksedk:
> 
> 
> Makes sense. My bad, didn't take that into account.

np, have made the exact same error more thatn once... :)

OmrSi:
I've been trying to figure out how the header should be drawn in RTL, but I couldn't.

niksedk:
Looks good :)

OmrSi:
The header still needs a lot of work, as I've mentioned in my email.

PR #4615, opened 2020-12-20
Refactoring & Cleanup Translation features &  SentenceMerging-Algorithm

Brummochse:
### Refactoring & Cleanup Translation features

- decoupling calling context from TranslationService-specific code
- split up the current approach with the ITranslator-interface into 2 different encapsulated strategy-pattern:
1. **TranslationService** (with concrete implementation of MicrosoftTranslationService, GoogleTranslationService and possibility to extend like DeepLTranslationService ...)
2. **TranslationStrategy**: handle different translation stratgeies for a specific TranslationService (like GoogleTranslationVersion1Strategy, GoogleTranslationVersion2Strategy, GoogleTranslationVersion3Strategy)
- introduce new concept **TranslationProcessor** which handles complex translation tasks and cares about:
1. input text chunking (complying translation service specific constraints)
2. translationn status report & cancellation handling
3. how to pass the paragraphs into the translator (paragraph line hadnling)
- refactor the calling context of "GoogleTranslate.cs" to a abstract "GenericTranslate.cs": remove all google or microsoft specific code and only couple it with AbstractTranslationService and ITranslationProcessor

### Translation Improvement with new SentenceMerging-Algorithm

- Basic idea: current machine translation services works best with whole sentences (better than passing the subtitle paragraphs individualy)
- implement a **SentenceMergingTranslationProcessor** which merges paragrpahs together and cut it into sentences. This sentences get send to a TranslationService and results get split back to paragraphs again


Brummochse:
> nikse: There's a problem with hearing impaired texts, see:
grafik

I minimized this problem with 2 additional strategies:
1. open and close brackets will interpreted as sentence end delimiters (this means this will not merged together with other paragraphs).
2. I use a more sophisticated way to assign the new translated target texts to the source chunks to optimize the proportionality between source and target chunks (**AbstractStringSplitsChunkAssigner**). I am not a pro in theoretical informatics, but I assume it is a NP-completeness problem to find the best suitable solution. Beside the deterministic **GreedyStringSplitsChunkAssigner** a new random-based approach (**RandomizedStringSplitsChunkAssigner**) was invented , which shows way much better result. (Maybe an approach on basis of genetic algorithms could improve this even more).

For string splitting a new class **StringSplitEngine** was improved, which can handle rules (**IUnsplittableMarker**) to exclude specific string patterns from splitting. At the moment **XmlTaggedUnsplittableMarker** and **UrlUnsplittableMarker** are supported to prevent xml-tags and website-addresses from breaking.

I changed the paragraph-selection interpretation in the translation window. In the past all paragraphs were translated following the first selected paragraph in the list. I changed it: Now only selected parapgraphs get translated. With this change it is possible to use different line-handling for specific paragraphs (in the case, the sentence merge algorithm don't produce satisfying results)

PR #4619, opened 2020-12-20
Work on list view's column header

OmrSi:
Is there a reason for reversing `Words/min` and `Chars/sec`?
They were causing an issue for the dark theme.
image
The issue left now is the column header in RTL mode with the dark theme.

I also tested and found there is no reason to set `TextAlign` based on `RightToLeftLayout`.

Let me know what you think.


OmrSi:
Okay, so the second commit is to hide the empty column, which's header takes the wrong color in dark theme.

How it was:
image

How it is now:
image

PR #4636, opened 2020-12-25
Hide HScroll when opening a file

OmrSi:
Okay, this kinda seems dumb, so feel free to close it.
But it's really annoying and I haven't found another way to fix it, even after searching for 2 days!

When you open a file, a Horizontal scrollbar shows up for no reason, even though the last column is sized to exactly the list view's width:
image

If you scroll down or click on the scrollbar, it just disappears, like it updates itself or something.
In this commit, I send an empty scroll message to the list view's scrollbar so it would disappear right when the last column's width is set. :)

### DIFF
--- a/src/ui/Controls/SubtitleListView.cs
+++ b/src/ui/Controls/SubtitleListView.cs
@@ -1140,6 +1140,7 @@ namespace Nikse.SubtitleEdit.Controls
             if (Columns.Count > 0)
             {
                 Columns[Columns.Count - 1].Width = ClientSize.Width - width;
+                NativeMethods.SendMessage(Handle, 0x1014, 0, 0);
             }
         }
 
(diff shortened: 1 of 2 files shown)


niksedk:
Kind of weird that the horizontal scroll bar remains visible... the `set last column width` is called like 10 times which also causes some flicker - I've tried a fix in commit 6a662e8d0ba9703c466812ed6d6c0f8c7abff727 - call only once (with a delay if 50ms).
Does that also work for you?

OmrSi:
@niksedk It works, but how? Doesn't the timer start each time the method is called?

OmrSi:
@niksedk Does it work because 50ms doesn't pass when calling it in a row?

niksedk:
Not really sure why it did not work before ;)
But your edit line sounds likely.

Issue #4347, opened 2020-09-04
spell check live?


m0ck69:
Hello.
Is it possible (or will it be) to spell check "live" while typing in text box.

Something very similar to MS Word.
It could look like this.

![spell](https://user-images.githubusercontent.com/46628545/92211387-accc6880-ee90-11ea-9532-0efdf9be3364.jpg)

PR #4637, opened 2020-12-27
Work on live spell check

OmrSi:
How it looks:
image

TODO:
1. ~~Add an option to add the word to the dictionary.~~
2. ~~Add an option to activate in the UI.~~
3. ~~More work on isolating the words.~~
4. ~~Make loading dictionaries async.~~

@niksedk Please check it out and let me know what you think.

Closes https://github.com/SubtitleEdit/subtitleedit/issues/4347 when it's done. :)


OmrSi:
@niksedk Do refactor the code :)

OmrSi:
Loading dictionaries asynchronously stops the timer for a bit while they are being loaded, I don't know if this is good or if a new timer should be added.

niksedk:
Very close :)

1) The live spell check should not spell check the word being typed (like word).

2) I opened a Danish subtitle, and the live spell check looked very much like English.

OmrSi:
About 1: How can you tell the word is done? You detect a space or a new line, but what if the user changed a work that already exists?

About 2: it uses `LanguageAutoDetect.AutoDetectLanguageName("", subtitle)` to detect that language, is google's better?

OmrSi:
@niksedk Okay, number 2 was a bigger challenge than I thought.
It's not enough to detect the language, you have to check if the dictionary files for it are available.
I think I got it working.
It should let the user download the dictionary if it's not downloaded.

I'm still not clear on how number 1 can be achieved.

OmrSi:
@niksedk Okay, I think this is the best I can do.
New words that are added to the end of the line will only be checked if certain keys were pressed, but old words will be checked while being written.

To be honest, I don't think I have anything more to add to this.

niksedk:
@OmrSi: Happy new year :)
Will give it a test run later today.

PR #4639, opened 2020-12-27
Try to fix some text box issues (selection length + text when switching syntax coloring on)

OmrSi:
The first issue is that when using "Find", the word isn't selected, therefore using "Find next" doesn't work:
image

The second issue is that if you had "Syntax coloring" off, then you turn it on, the selected line is lost:

I don't know if this is the best solution for either.

### DIFF
--- a/src/ui/Controls/SETextBox.cs
+++ b/src/ui/Controls/SETextBox.cs
@@ -33,15 +33,18 @@ namespace Nikse.SubtitleEdit.Controls
         {
             ContextMenuStrip oldContextMenuStrip = null;
             var oldEnabled = true;
+            var oldText = string.Empty;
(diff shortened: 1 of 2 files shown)


src/ui/Controls/SETextBox.cs
```diff
@@ -344,7 +348,7 @@ public int SelectionLength
                 }
                 else if (_uiTextBox != null)
                 {
-                    var target = _uiTextBox.SelectionLength;
+                    var target = value;
```
OmrSi, line 351: This is for the "Selection Lenght" issue, I don't know it it was intended the way it was.

niksedk: That's a plain bug :)

Issue #4608, opened 2020-12-19
"Reload" button for Compare tool


LeonCheung:
Usually, we compare in order to both check and/or fix things, it's nice to keep eyes on things up to date easily with one single click, to make sure no issues left before quitting.

PR #4650, opened 2020-12-28
Add "Reload" buttons to Compare

OmrSi:
Closes: https://github.com/SubtitleEdit/subtitleedit/issues/4608

image


OmrSi:
Ah, language tags...

LeonCheung:
Hi, I just found the "Reload" buttons are disabled when dragging files into the panel to open them. Would you please have a look?

Issue #3226, opened 2018-12-11
[Feature Request] Temporarily Hide Preview Texts on Video Player


LeonCheung:
Is there a way to hide Preview texts? Sometimes it just easily blocked the important information on the video that I want to get. Thx!

![se](https://user-images.githubusercontent.com/494788/49785698-5930f780-fd5c-11e8-9251-4b8158477bbd.png)

darnn:
Yeah, toggle preview on/off would be nice. Currently I just shift all the subtitles two hours forward using "adjust all times".

OmrSi:
Or you can just hit F12, setting the end of the line hides it, and this is a bug in my opinion, then Undo.

OmrSi:
You can also use the new alignment shortcut, for me it's Ctrl + Num8 to make the subtitle go up, then use it again to remove the tag. Using it once adds {\an8} and twice removes it.
This is better the setting the end, I think.

OmrSi:
Okay, I think this should be closed now, we waited and saw.
As you can read in the PR, Nik doesn't think this is needed, or at least he doesn't think hiding the preview completely is a good idea.
He suggested a toggle for "MPV handles text preview", but that would just put the text under the video, which resembles what I suggested by using the alignment shortcut and putting the subtitle at the top of the screen, which is why I don't think it's a good solution.

niksedk:
Added shortcut General - "Toggle preview on video":
![image](https://user-images.githubusercontent.com/1261331/103889268-6c9f2000-50e6-11eb-8e82-2f0bbf9bf4f5.png)

Should solve the issue :)

OmrSi:
@niksedk Can the size difference between MPV handles text preview and the other options be fixed somehow?
Like, here is my subtitle with MPV handles text preview:
![image](https://user-images.githubusercontent.com/20923700/103889697-7da05f00-50ef-11eb-88a6-7cc702520fbc.png)

Here it is without:
![image](https://user-images.githubusercontent.com/20923700/103889715-885af400-50ef-11eb-9901-c97bac88da90.png)


The font size is too large and the second line is out of the screen.

I noticed that the subtitle is shown via a rich text box, can't that be used to set the font and the font size to work as the user chose them?

PR #4667, opened 2021-01-02
Add "Toggle subtitle visibility" shortcut

OmrSi:
Closes https://github.com/SubtitleEdit/subtitleedit/issues/3226


OmrSi:
@niksedk Would you like me to change the name to "Toggle subtitle preview" or something?

niksedk:
Hm, what about "Toggle on-video-preview"... (mpv handles preview text)

OmrSi:
I'm with you on that.
I actually wanted it to work just with MPV, but I thought you'd want it to include all.

OmrSi:
How is it now?

OmrSi:
I guess the name was too vague...

niksedk:
I was thinking about a toggle for this:
image

(vlc might potentially get the same in the future - if it ever catches up to mpv api :)

OmrSi:
Hmm, I think it makes more sense this way.
I mean, you're not changing the preview type altogether, you're just hiding the subtitle for a little while just to see what's behind it or for whatever reason, and you're not losing anything since the textbox is already in front of you.
It makes a lot of sense when you're put in this situation.
Also, you never know why a user would want to turn the preview on or off.

It can also be applied to VLC when it gets implemented.

Issue #4444, opened 2020-10-18
Original subtitles in video player v2


m0ck69:
With reference to https://github.com/SubtitleEdit/subtitleedit/issues/4400
The topic is closed and I don't know If I such topics are monitored so I created a new one.

Thank you so much @niksedk for adding this option.

However, I feel like a little of customization is needed. 
Basically two things are missing: changing the font of original subtitles **(especially the size)** and the **placement** (it would be perfect solution If we could drag the subtitles and put them whenever we want on the video).

Now original subs and translation are too close to one another and it doesn't look good.

I'd like to have the original subs on the top but If sb else doesn't like it, they should be allowed to put them whenever they want.

Please look the the photos below:

![Clipboard01](https://user-images.githubusercontent.com/46628545/96365772-2bc8e800-1143-11eb-94bc-e33dab129b52.jpg)

![94806662-9e6d6000-03ee-11eb-98a5-1e097011b332](https://user-images.githubusercontent.com/46628545/96365789-4d29d400-1143-11eb-99e1-1ca323481ad5.jpg)

OmrSi:
I think this is not possible right now because he's not using MPV to handle the text.
Since MPV isn't handling the text, you can't change the font or the size or the place.
It might be a little tricky to make MPV handle two subtitles, or he would have done it in the first place.

Just letting you know what I think. :D

m0ck69:
What handles the size and placement then? @niksedk original subs are smaller than mine and look slightly better.


![Screenshot](https://user-images.githubusercontent.com/46628545/96366088-9bd86d80-1145-11eb-87c4-c76a5663dcc3.png)

niksedk:
The subtitle at the bottom is displayed via a text label inside SE by SE.
The top one is displayed on the video by the video mpv/vlc player... if you put `{\an8}` on a line it will actually display on top.
(the above functionality has been requested by another user... when I first added mpv it auto-loaded subtitles - like directshow will also do if you have directvobsub installed. Some people likes this even though it's not on purpose).

OmrSi:
@niksedk Can't something be done with MPV's "--secondary-sid=<ID|auto|no>" option for example?
I haven't seen how SE works with MPV, that's why I'm asking.

PR #4687, opened 2021-01-07
Improve opening a second subtitle

OmrSi:
Okay, this improves opening a second subtitle with MPV in the following way:
1. It now works with MPV handles preview, the second subtitle is shown on top.
2. Added a `close second subtitle` option if MPV is used.

I tried to make `close second subtitle` work with VLC, but its API doesn't have a clear method for input slaves in `media_player`.

`LanguageDeserializer` and `LanguageMaster` need updating.

Closes https://github.com/SubtitleEdit/subtitleedit/issues/4444

### DIFF
--- a/src/ui/Logic/VideoPlayers/LibVlcDynamic.cs
+++ b/src/ui/Logic/VideoPlayers/LibVlcDynamic.cs
@@ -1,5 +1,4 @@
-﻿using Nikse.SubtitleEdit.Core;
-using Nikse.SubtitleEdit.Forms;
+﻿using Nikse.SubtitleEdit.Forms;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -159,7 +158,7 @@ namespace Nikse.SubtitleEdit.Logic.VideoPlayers
         private libvlc_media_player_next_frame _libvlc_media_player_next_frame;
 
         [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
-        private delegate int libvlc_media_player_add_slave(IntPtr media, int type, byte[] filePath, bool select);
+        private delegate int libvlc_media_player_add_slave(IntPtr mediaPlayer, int type, byte[] filePath, bool select);
         private libvlc_media_player_add_slave _libvlc_media_player_add_slave;
 
         [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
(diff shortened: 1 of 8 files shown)


OmrSi:
@niksedk Why not, though? Is there a problem with making it open in MPV?

OmrSi:
The "close" option for MPV was also a good addition, I think.

niksedk:
The current "Open second subtitle..." is not too good, I know. But it works a little...
It should probably prompt for location (top) and color (e.g. yellow) - a bit like the plugin for merge subs. This does not close the issue.
Also, this code just does not feel right...

OmrSi:
Hmm.
Well, I don't know if you would read the code explanation, but...

MPV provides `secondary-sid` option you can use to set the secondary subtitle which shows up at the top by default.
This works if you have the subtitle loaded and you use it to show that subtitle as the secondary subtitle with its ID.

SE keeps removing and adding and refreshing the subtitle, which causes the secondary subtitle to disappear sometimes, which is why there is a need to make sure it's visible each time `sub-load` of `sub-remove` is used.

In addition to this, using `sub-load` with a file that already has a bunch of mixed subtitles makes the loaded subtitle the last track in MPV, which is why I made a list of all subtitle tracks and set the secondary to the last one.

src/ui/Logic/VideoPlayers/LibVlcDynamic.cs
```diff
@@ -159,7 +158,7 @@ public class LibVlcDynamic : VideoPlayer, IDisposable
         private libvlc_media_player_next_frame _libvlc_media_player_next_frame;
 
         [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
-        private delegate int libvlc_media_player_add_slave(IntPtr media, int type, byte[] filePath, bool select);
+        private delegate int libvlc_media_player_add_slave(IntPtr mediaPlayer, int type, byte[] filePath, bool select);
```
OmrSi, line 161: And by the way, this is "mediaPlayer", the one in "media" has different parameters.
There is no `Clear` method for input slave in "media player", but there is in "media".
That's why I couldn't get the "Close second subtitle" to work on this.

OmrSi:
I would work on adding the option to set its font size and color, but I guess there is no point if the coding isn't good enough.
I'll be waiting to see how you'll implement it.

OmrSi:
For future refrence:
```
--secondary-sid=<ID|auto|no>
Select a secondary subtitle stream. This is similar to --sid. If a secondary subtitle is selected, it will be rendered as toptitle (i.e. on the top of the screen) alongside the normal subtitle, and provides a way to render two subtitles at once.
```

But I guess you'd probably use the normal `sub-load` and set its position.

PR #4689, opened 2021-01-08
Fix typo in the word "toggle"

OmrSi:
It was written "toogle" in some places.


niksedk:
He, probably because I've typed `Google` to many times ;)

PR #4695, opened 2021-01-09
Apply Win10 dark theme for titlebar, ListView, ListBox, and Treeview

OmrSi:
I don't think you'll like this, but I had to offer it since I'm using it and liking it.

In this, API calls are used in windows 10 to use its dark theme to color the titlebar and the scrollbar of ListView and ListBox.
It's also used on the TreeView.

The dark theme looks better with it, and I'm not having any issues with using it.

I had to add a manifest so it would detect the correct version of windows to work this.

Note that It will only try to work on the supported Windows 10 versions.

### DIFF
--- a/src/ui/Logic/DarkTheme.cs
+++ b/src/ui/Logic/DarkTheme.cs
@@ -11,6 +11,41 @@ namespace Nikse.SubtitleEdit.Logic
 {
     public static class DarkTheme
     {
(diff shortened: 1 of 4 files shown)


niksedk:
This is cool :)

I don't like calling too many DLL files, but this is okay (and does not break Linux).

Issue #4705, opened 2021-01-11
shortcut doesn't respect gap


m0ck69:
Hi.
A quick question.
Does the shortcut "move start 1 frame back (keep gap to previous If close" work properly?
My gap is set to 2 frames but this shortcut seems to leave a bigger gap.
Look:

https://user-images.githubusercontent.com/46628545/104168922-12ac9c00-53ff-11eb-8735-520c09dc1750.mp4

@OmrSi

OmrSi:
Yeah, That's how @niksedk meant for it to work.
`keep gap to previous If close`, he hardcoded `close` to 199 ms or smaller.

m0ck69:
Well... I don't think it was a good idea to hardcode a gap. And why 199 ms and not 198 or 200 ms?
The gap is a variable value set in the settings and its length should be taken from there.
But thanks for your explanation.

niksedk:
A minor bug...

(The `Min gap` has not existed forever... I've fixed it tons of places, but a few might be still left)

OmrSi:
@niksedk This shortcut is pretty new. https://github.com/SubtitleEdit/subtitleedit/issues/4078#issuecomment-624071718
I'm sure that `Min gap` existed then.
That's why I thought it is intentional.

niksedk:
It's not intentional.

OmrSi:
Want me to change it?

m0ck69:
@OmrSi 
Sorry for the offtop, but what's the name of the defaulf subtitle preview font?
It's not Adobe Devanagari.
![f2](https://user-images.githubusercontent.com/46628545/104299226-3fc38200-54c5-11eb-9bf7-1cd9ec2cc6da.jpg)

I mean this font.

![f1](https://user-images.githubusercontent.com/46628545/104299231-418d4580-54c5-11eb-935c-d7fd433d5511.jpg)

OmrSi:
@m0ck69 Good catch.
It's actually Tahoma, which is the default.

PR #4708, opened 2021-01-12
Fix for default font name in VideoPreviewFont

OmrSi:
Fix for the issue mentioned here https://github.com/SubtitleEdit/subtitleedit/issues/4705#issuecomment-758544771

### DIFF
--- a/src/libse/Common/Settings.cs
+++ b/src/libse/Common/Settings.cs
@@ -1205,15 +1205,6 @@ $HorzAlign          =   Center
             DefaultFrameRate = 23.976;
             CurrentFrameRate = DefaultFrameRate;
             SubtitleFontName = "Tahoma";
-            if (Configuration.IsRunningOnLinux)
-            {
-                SubtitleFontName = Configuration.DefaultLinuxFontName;
-            }
-            else if (Environment.OSVersion.Version.Major < 6) // 6 == Vista/Win2008Server/Win7
-            {
-                SubtitleFontName = "Times New Roman";
-            }
-
             SubtitleTextBoxFontSize = 12;
             SubtitleListViewFontSize = 10;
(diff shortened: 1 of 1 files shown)


src/libse/Common/Settings.cs
```diff
@@ -1323,6 +1314,17 @@ public GeneralSettings()
             DialogStyle = DialogType.DashBothLinesWithSpace;
             ContinuationStyle = ContinuationStyle.None;
 
+            if (Configuration.IsRunningOnLinux)
+            {
+                SubtitleFontName = Configuration.DefaultLinuxFontName;
+                VideoPlayerPreviewFontName = Configuration.DefaultLinuxFontName;
+            }
+            else if (Environment.OSVersion.Version.Major < 6 || Environment.OSVersion.Version.Major == 6 && Environment.OSVersion.Version.Minor < 2) // Vista/Win2008Server/Win7
```
OmrSi, line 1322: As you can see here https://stackoverflow.com/a/2819962/14349868
Windows 7 is 6, but the minor is 1, so it should be like this, I think.

niksedk: I think you can remove "else stuff"" as  WinXP is no longer supported.
"Times New Roman" was one of the few fonts on WinXP that had some Unicode support.

OmrSi: So Windows 7 has "Tahoma"? The comment mentions Windows 7 as well.

Side note: I moved from XP to 10, never worked with 7. ^^"

OmrSi: Okay, given that the code was using Tahoma on Win7 and no one complained, I guess it's safe to say that it works on Win7.

niksedk: Yes, Unicode support in Win7 is fine :)

PR #4715, opened 2021-01-13
Use a dark tab control + some work on colors

OmrSi:
Welp, it's exactly what it says:
image

image

I found this in `gitextensions`'s dark theme and edited it for SE, the license is `GNU General Public License v3.0`, and I mentioned where I got it from.
Is this okay?

### DIFF
--- a/src/ui/Forms/FixCommonErrors.cs
+++ b/src/ui/Forms/FixCommonErrors.cs
@@ -1433,7 +1433,7 @@ namespace Nikse.SubtitleEdit.Forms
             UiUtil.GetLineLengths(labelSingleLine, text);
 
             var s = text.Replace(Environment.NewLine, " ");
-            labelTextLineTotal.ForeColor = Color.Black;
+            labelTextLineTotal.ForeColor = UiUtil.ForeColor;
             buttonSplitLine.Visible = false;
             var abl = Utilities.AutoBreakLine(s, _autoDetectGoogleLanguage).SplitToLines();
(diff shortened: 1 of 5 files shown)


niksedk:
That's okay :)

Did you test with RTL + Linux ?

OmrSi:
Hmm, I don't think ant tab control in SE uses RTL. Do you know where it does?

I don't have Linux installed, I do have Ubuntu on VMWare, installing it there now to check.

OmrSi:
Okay, forced it to use RTL, seems to be working well:
image

Checking on Linux is left.

OmrSi:
@niksedk This is my first time with mono, do I have to do anything special?
SE opens for a sec then closes

OmrSi:
Linux:
image


Good enough?

OmrSi:
@niksedk About this last commit, I have photos from windows 10 and 7 showing that the back color of groupBoxEdit used to be white, it has the color of a control on Linux for some reason, but I believe it used to be white on windows:
image

I don't know why removing the tab control caused it to turn to the control color.

niksedk:
> 
> 
> @niksedk About this previous commit, I have photos from windows 10 and 7 showing that the back color of groupBoxEdit used to be white, it has the color of a control on Linux for some reason, but I believe it used to be white on windows:
> 
> image
> 
> The tab control must have been giving it the white color.

I probably set to white at some point... but the default value should probably be used.

OmrSi:
I don't know, it seems like it came out straight from an old OS with the control color.

I don't think you sat it to white. On Linux, it's not white in 3.5.18. I do believe the tab control was giving it the white color.

PR #4719, opened 2021-01-14
Keep dark theme for ListView header in RTL

OmrSi:
I've been at it all day, but something is very wrong with the listview header in RTL, you can't use `e. Graphics.DrawString` or `e. Graphics.FillRectangle`, they just don't work. The only thing that works is `TextRenderer.DrawText` for some reason.

This is how it is:
image

It's the best result I could get.

### DIFF
--- a/src/ui/Logic/DarkTheme.cs
+++ b/src/ui/Logic/DarkTheme.cs
@@ -423,7 +423,7 @@ namespace Nikse.SubtitleEdit.Logic
             {
                 int posY = Math.Abs(e.Bounds.Height - e.Font.Height) / 2;
                 e.Graphics.DrawString(e.Header.Text, e.Font, fc, e.Bounds.X + 3, posY, strFormat);
-                if (e.ColumnIndex != 0)
+                if (Configuration.Settings.General.DarkThemeShowListViewGridLines && e.ColumnIndex != 0)
                 {
                     e.Graphics.DrawLine(new Pen(ForeColor), e.Bounds.X, e.Bounds.Y, e.Bounds.X, e.Bounds.Height);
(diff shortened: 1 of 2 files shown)


src/ui/Logic/DarkTheme.cs
```diff
@@ -395,16 +395,17 @@ private static void ListView_DrawSubItem(object sender, DrawListViewSubItemEvent
...
 
-            e.DrawDefault = false;
-            using (var b = new SolidBrush(Color.FromArgb(Math.Max(BackColor.R - 9, 0), Math.Max(BackColor.G - 9, 0), Math.Max(BackColor.B - 9, 0))))
+            using (var slightlyDarkerBrush = new SolidBrush(Color.FromArgb(Math.Max(BackColor.R - 9, 0), Math.Max(BackColor.G - 9, 0), Math.Max(BackColor.B - 9, 0))))
             {
-                e.Graphics.FillRectangle(b, e.Bounds);
+                e.Graphics.FillRectangle(slightlyDarkerBrush, e.Bounds);
             }
 
             var strFormat = new StringFormat();
```
niksedk, line 411: Is StringFormat disposable?

OmrSi: Yep, it implements IDisposable.

OmrSi:
@niksedk Is it true that using TextRenderer which works with GDI is better than Graphics.DrawString which works with GDI+?

niksedk:
> 
> 
> @niksedk Is it true that using TextRenderer which works with GDI is better than Graphics.DrawString which works with GDI+?

Sorry, I don't know...

PR #4725, opened 2021-01-17
Some work on dropdown menus

OmrSi:
1. Fix dynamic dropdown menus colors in dark theme like this one:
image

2. Remove WinXP font check as it is no longer supported.

3.  Set last column width in Styles Manager and Timed Text Styles.

4. Add a checkmark to some dropdown menus like selected audio track and selected style and actor:
image

### DIFF
--- a/src/libse/Common/FastBitmap.cs
+++ b/src/libse/Common/FastBitmap.cs
@@ -41,21 +41,6 @@ namespace Nikse.SubtitleEdit.Core.Common
         {
             _workingBitmap = inputBitmap;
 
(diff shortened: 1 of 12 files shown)


OmrSi:
@niksedk I think there is an issue with the format `CsvNuendo`.
It used sometimes as `formatType == typeof(CsvNuendo)` and sometimes as `format.Name == "Nuendo"`

niksedk:
> 
> 
> @niksedk I think there is an issue with the format `CsvNuendo`.
> It used sometimes as `formatType == typeof(CsvNuendo)` and sometimes as `format.Name == "Nuendo"` while its name is `Csv Nuendo`

One is a plugin...

PR #4728, opened 2021-01-17
Update TextBoxSyntaxColor state on launch

OmrSi:
Before:

image

After:

image


niksedk:
Hm, I would like to keep the button positions, to make it more clear, that they are dependent.

OmrSi:
Makes sense, it just didn't look so when I saw it.

Do you think this is better?

image

niksedk:
I think it's better like this:
image

OmrSi:
Hmm, they are too far, I didn't know you meant for them to be connected when I first saw them. But, okay.

niksedk:
> 
> 
> Hmm, they are too far, I didn't know you meant for them to be connected when I first saw them. But, okay.

Yes, they are probably a bit too far right - you can move them a little to the left...

OmrSi:
image

PR #4736, opened 2021-01-19
Yet another try to fix Arabic comma in RichTextBox

OmrSi:
It was adding a Unicode char for each comma in the line.

### DIFF
--- a/src/ui/Controls/SETextBox.cs
+++ b/src/ui/Controls/SETextBox.cs
@@ -8,6 +8,7 @@ using System;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 
@@ -103,7 +104,6 @@ namespace Nikse.SubtitleEdit.Controls
             _richTextBoxTemp = null;
             if (useSyntaxColoring)
             {
-
                 _richTextBoxTemp = new RichTextBox();
                 _uiTextBox = new RichTextBox { BorderStyle = BorderStyle.None, Multiline = true };
                 InitializeBackingControl(_uiTextBox);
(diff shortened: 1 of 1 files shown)


OmrSi:
@niksedk Something strange is happening and I don't know why.
Sometimes the input language changes on its own when I type a space in the richtextbox:
Untitled Project

niksedk:
> 
> 
> @niksedk Something strange is happening and I don't know why.
> Sometimes the input language changes on its own when I type a space or a dot or a new line in the richtextbox:
> Untitled Project
> 
> Notice how the language gets changed each time I enter a space.
> it happens when I:
> 
>     1. Open SE with the language set to English.
> 
>     2. Add a line and write an English letter.
> 
>     3. Delete that letter and switch the language.
> 
>     4. Write a word and press space.
> 
> 
> I think it has something to do with the temp richtextbox's rtf.

I cannot re-create this with Danish... you can try to remove the temp and just use the existing rtf...

OmrSi:
@niksedk It works after removing the temp, but the issue it the updates to the text are visible to the user, like when it selects the whole text to set the color, the user can see the highlight.

niksedk:
> 
> 
> @niksedk It works after removing the temp, but the issue it the updates to the text are visible to the user, like when it selects the whole text to set the color, the user can see the highlight.

does SuspendLayout/ResumeLayout help?

OmrSi:
No, it doesn't seem to help:
Untitled Project

I tried to make it only highlight again when the tags are changed, but that causes an issue at the beginning and end of lines.

niksedk:
Perhaps with this: https://stackoverflow.com/questions/3282384/richtextbox-syntax-highlighting-in-real-time-disabling-the-repaint

OmrSi:
Yes, it works very well.

PR #4737, opened 2021-01-20
Some work on RichTextBox highlighting

OmrSi:
Try to fix an issue with the Arabic comma.
Removed the temp RichTextBox.
Make the highlighting only happen when a spell check was requested or there is a tag in the line.

Do check the code and test if you can. :D

Side note: Added a method called `DoFormattingActionOnRtb` which takes an action and does it on the text of the _uiTextBox while saving the cursor position.


OmrSi:
I think the only thing left is to try changing its color while disabled.

PR #4744, opened 2021-01-22
Translation update

PMitsakis:


src/ui/Languages/el-GR.xml
```diff
@@ -6,7 +6,7 @@
     <TranslatedBy>Μετάφραση: Παναγιώτης Μητσάκης</TranslatedBy>
     <CultureName>el-GR</CultureName>
     <HelpFile />
-    <Ok>&amp;Εντάξει</Ok>
```
niksedk, line 9: This will remove the "alt+E" shortcut - is that on purpose?

PMitsakis: No. I will fix it.

PR #4750, opened 2021-01-23
Work on richtextbox selection length and highlighting

OmrSi:
Well, this is getting tedious. xD

### DIFF
--- a/src/ui/Controls/AdvancedTextBox.cs
+++ b/src/ui/Controls/AdvancedTextBox.cs
@@ -34,7 +34,7 @@ namespace Nikse.SubtitleEdit.Controls
         private string _currentDictionary;
         private string _uiTextBoxOldText;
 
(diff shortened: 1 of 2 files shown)


niksedk:
> 
> 
> Well, this is getting tedious. xD

Then we should add a few more tests:
https://github.com/SubtitleEdit/subtitleedit/blob/master/src/Test/SETextBoxControl/SeTextBoxTests.cs

It's really annoying that the TextBox and RichTextBox behaves so differently.
Both should really just use Environment.NewLine!

OmrSi:
>Both should really just use Environment.NewLine!

I hadn't noticed this until you used `base.Text` instead of `Text`.
It's annoying that we have to consider each case.
I think the only way to actually find the issues is to work on a subtitle and see what comes up.
I haven't done that for a while, COVID slowed work down.

OmrSi:
Anyway, this PR should be one on the correct path.
SelectionLength was wrong across lines, and the spell check wasn't being done each time the text changed.

PR #4763, opened 2021-01-27
Add Categories for ASSA

OmrSi:
1. Remove Ass page from the settings.
2. Add Categories to Assa styles manager.
3. Add icons for Assa styles manager and Assa properties.

image


OmrSi:
@niksedk This is ready.
You just need to test and add the icons to the icons file, whenever I try to set the icons the `.resx` file and the designer file gets all messed up because of my DPI.

The AssaStylesManager icon should be named `toolStripButtonAssaStylesManager`.
The AssaProperties icon should be named `toolStripButtonAssaProperties`.

PR #4767, opened 2021-01-30
Add icons for Assa styles manager

OmrSi:
@niksedk You just need to test and add the icons to the icons file, whenever I try to set the icons the .resx file and the designer file gets all messed up because of my DPI.

The AssaStylesManager icon should be named `toolStripButtonAssaStylesManager`.
The AssaProperties icon should be named `toolStripButtonAssaProperties`.


niksedk:
OK :)
Added the icon here: 78b21d09f28497735e2cdea3c1ebb054601cfc20

Issue #4702, opened 2021-01-10
Default profile for new .ass files


ChocOranger:
Can we have a default profile set for new .ass files please? Very useful for default resolution and the editing author / language. No point in having to fill them every sub. Users that don't need it can leave it as it is.

ie the details for File > AAS properties can have pre-filled defaults (including the resolution).
A loaded movie should override the defaults eg title and resolution. Although you could add a option not to do this.

niksedk:
Default profile can now be made in the ASSA style manager:
![image](https://user-images.githubusercontent.com/1261331/108590612-be34fe80-7364-11eb-978c-32eb11f3ce0f.png)

Beta link: https://github.com/SubtitleEdit/subtitleedit/releases/download/3.6.0/SubtitleEditBeta.zip

ChocOranger:
Currently , the beta Style Storage doesn't keep any of the properties in the Advanced Sub Station Alpha Properties windows.

I was asking for defaults here.
![Ass_properties](https://user-images.githubusercontent.com/77179521/108606524-67eeac80-73b2-11eb-8167-32dd805c08e1.png)

OmrSi:
My bad, thought from the title it's for the styles.

@niksedk I guess this should be reopened, although I don't know if you're interested in doing defaults for the properties.

ChocOranger:
I'd be happy enough for a default for the Resolution.
I sub for a single resolution and use black borders for any video that requires them.

Issue #4483, opened 2020-11-07
auto load ass styles


z3us:
Hi
Is it possible that the program load automatically my ass styles file at the start?
Thanks

niksedk:
I'm not sure what you mean... could you explain a bit more?
SE loads styles when opening an ass file.

z3us:
When I want to assign styles in a ass file, I only have the default style. Then I have to go to the import button, search my styles file and load it. I want my styles yo be automatically imported in the program

niksedk:
OK, how many "saved styles" would you normally need? 
GUI suggestions are welcome :)

z3us:
It would be enough if the program just remembered the previously imported styles

niksedk:
Style storage implemented in latest beta with @LeonCheung design:
https://github.com/SubtitleEdit/subtitleedit/releases/download/3.5.18/SubtitleEditBeta.zip

z3us:
> Style storage implemented in latest beta with @LeonCheung design:
> https://github.com/SubtitleEdit/subtitleedit/releases/download/3.5.18/SubtitleEditBeta.zip

And how does this feature solve my issue? I still have to reload my styles each time I open a subtitle

niksedk:
So a solution could be an option to load the styles from the "style storage" automatically to new .ass subtitles?

z3us:
Yes, as soon as u open an ass subtitle file, have the styles available

ChocOranger:
It would be even better to have the "style storage" available when working with .ass files rather than just on load.
eg when switching format from .srt to .ass
Basically load "style storage" on program load - it shouldn't impact non-style users.

Additionally, and you might consider this a separate issue, but since you are looking at .ass styles, the style GUI does not contain nor save the following values which are in the .ass style header:
Spacing (this is so useful)
Scale X% / Y%
Angle

As it is, loading "style storage" will lose these values.

LeonCheung:
@z3us To be honest I don't quite agree with that idea. Styles should be project-based. Most people don't want to auto-load extra styles they don't expect. Say, someone has 100 styles in the storage, is it really cool to auto-load all of them into the current project???

Your idea makes sense only when working on projects that are quite similar to each other. Imho, the only workable solution would be to implement kinda **Autoload Style List** separately (or, to define styles as auto-load ones by toggling them from the Storage table. For that, we may need a separate column). Not sure if it is worth doing though...

niksedk:
[...]

PR #4768, opened 2021-01-30
Remove style page from the settings and add categories to Assa

OmrSi:
@niksedk There have to be together, adding categories and removing the style page makes the default Ass styles be taken from the storage's default category, which means the `SubtitleSettings` Ass options are no longer needed because they can't be set from any place.

Closes https://github.com/SubtitleEdit/subtitleedit/issues/4702, Closes https://github.com/SubtitleEdit/subtitleedit/issues/4483

### DIFF
--- a/LanguageMaster.xml
+++ b/LanguageMaster.xml
@@ -2588,6 +2588,10 @@ Continue?</RestoreDefaultSettingsMsg>
     <AddToStorage>Add to storage</AddToStorage>
     <StyleStorage>Style storage</StyleStorage>
     <OverwriteX>Overwrite {0}?</OverwriteX>
+    <Category>Category</Category>
+    <NewCategory>New category</NewCategory>
+    <CategoryName>Category name</CategoryName>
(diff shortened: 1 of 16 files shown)


OmrSi:
@niksedk Can you check this before adding more ASSA changes? It's causing conflicts each time.

niksedk:
This is for after 3.6.0, right?

OmrSi:
It was supposed to be, but I see you're still adding some things for ASSA.

It's up to you, but it's ready.
If you leave it to 3.6.0, tell me when you want to review it so I can resolve any conflicts that will show up in the future.

niksedk:
I guess it for after 3.6.0 due to language tags... alternately you could "borrow" if other windows have the same texts...

I'm not adding things for ASSA - at least not visible things. I've been fixing a few bugs though.

OmrSi:
I guess so.
I couldn't find "Categories" in other places, it's called "Groups" for Multiple Replace.
Also, I think the note is a big dealbreaker.
I'm sure `Note: "Default" styles will be applied to new ASSA files` isn't available anywhere else.

PR #4769, opened 2021-01-30
Add hidden settings for MPV preview

OmrSi:
And a UI setting for opaque box.

About the Ssa style change, it's only using an object from SsaStyle to set MPV's font style and only changes it when the style needs updating, it's better than generating the style each time MPV updates the subtitle.


src/ui/Controls/VideoPlayerContainer.cs
```diff
@@ -399,6 +400,30 @@ public void SetSubtitleText(string text, Paragraph p, Subtitle subtitle)
             }
         }
 
+        private SsaStyle _mpvStyle;
+        public void UpdateMpvStyle()
+        {
+            if (_mpvStyle != null)
```
niksedk, line 406: Is this null check + null assignment needed?

src/libse/Common/SsaStyle.cs
```diff
@@ -28,6 +28,8 @@ public class SsaStyle
         public string RawLine { get; set; }
         public bool LoadedFromHeader { get; set; }
 
+        public const string DefaultAssStyleFormat = "Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding";
```
niksedk, line 31: Should/could this be used more places?

src/ui/Controls/VideoPlayerContainer.cs
```diff
@@ -399,6 +400,30 @@ public void SetSubtitleText(string text, Paragraph p, Subtitle subtitle)
             }
         }
 
+        private SsaStyle _mpvStyle;
```
niksedk, line 403: It would be a little better if it was an "_mpvStyleHeader", so generating the header will happen less often.

src/libse/Common/Settings.cs
```diff
@@ -1292,7 +1292,7 @@ public GeneralSettings()
             MpvVideoOutputLinux = string.Empty; // could also be e.g. "x11";
             MpvHandlesPreviewText = true;
             MpvPreviewTextPrimaryColor = Color.White;
-            MpvPreviewTextOutlineWidth = 2;
+            MpvPreviewTextOutlineWidth = 1;
```
niksedk, line 1295: 👍

src/libse/Common/SsaStyle.cs
OmrSi: Yes, I used it in my categories PR as well. :D

src/ui/Controls/VideoPlayerContainer.cs
OmrSi: I guess not, when I wrote it I thought the function would be used in more places and might require resetting the object.

OmrSi:
@niksedk Better?
Do you think there might be a better idea for the `_mpvPreviewStyleHeader`?

PR #4778, opened 2021-02-03
Try to fix the blur of Windows DPI scaling.

nkh0472:
Above changes have not been tested…

### DIFF
--- a/src/ui/app.config
+++ b/src/ui/app.config
@@ -5,4 +5,5 @@
   <startup>
     <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.7.2"/>
   </startup>
+  <add key="EnableWindowsFormsHighDpiAutoResizing" value="true" />
 </configuration>
--- a/src/ui/app.manifest
+++ b/src/ui/app.manifest
@@ -23,4 +23,15 @@
       <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}" />
     </application>
   </compatibility>
+  
+    <!-- Indicates that the application is DPI-aware and will not be automatically scaled by Windows at higher
+       DPIs. Windows Presentation Foundation (WPF) applications are automatically DPI-aware and do not need 
+       to opt in. Windows Forms applications targeting .NET Framework 4.6 that opt into this setting, should 
+       also set the 'EnableWindowsFormsHighDpiAutoResizing' setting to 'true' in their app.config. -->
+
+  <application xmlns="urn:schemas-microsoft-com:asm.v3">
+    <windowsSettings>
+      <dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true</dpiAware>
+    </windowsSettings>
+  </application>
 </assembly>


nkh0472:
According to http://www.kynosarges.org/WindowsDpi.html, the current state of this software seems to be "Applications with no DPI awareness at all", which causes "new DPI virtualization" then to be fuzziness.

> Applications with no DPI awareness at all — These are either very old or very poorly written, and today are fortunately becoming a rarity. Here lazy developers use system fonts in their GUI, but rather than querying Windows for the actual font sizes they simply hardcode the 96 DPI sizes into their window layout, naturally screwing up the GUI when higher DPI settings increase those sizes.
> 
> Such applications require the new DPI virtualization, fuzziness be damned; otherwise you’ll encounter problems ranging from cut-off text to overlapping controls, sometimes making the GUI entirely unusable. Over the years I’ve collected some sample screenshots of broken applications.

In my computer, which using Windows 10 21H2 with 1920x1080 screen resolution with 125% display scaling, the default state of SubtitleEdit looks like some kind of blurry, but at least the display of buttons and so on is not misplaced.
image

------------
However, if I mark the dpiAware flag as true, it seems to be "Applications without DPI awareness that set the DPI-aware flag".

> Applications without DPI awareness that set the DPI-aware flag — These are even worse than totally DPI-unaware applications. Examples include GUIs that scale decently to 120 DPI but not above, or Java Swing applications without scaling options. There’s nothing you can do here, as Windows won’t force such programs to use DPI virtualization. Once the DPI-aware flag is set, the application must scale itself, however poorly. You can only nag the developers to fix their product – or use something else. (Or upgrade to Windows 10 “Creators Update” which can finally enforce DPI virtualization.)

At that time, the situation might be similar to: When I enable “Override High DPI scaling behavior” and set the Scaling Behavior to “Application”, "Windows won’t force such programs to use DPI virtualization", the font of the program is clear. However, the text in some places will be truncated because it is beyond the display range, such as "UTF-8 with BO……" at the top of the window.
image

Although this commit can fix the problem that the interface of the software is blurred under the high-resolution screen, I am worried that the truncation of text will be more serious under the higher resolution(or bigger font?), so that it can not be used normally.

-------------
All of the above are my guesses. I don't have visual studio installed and I don't have a larger monitor for my testing. So I hope these can provide some reference for developers.

Also, there is another link that might be useful for developers.
[High DPI Desktop Application Development on Windows](https://docs.microsoft.com/zh-cn/windows/win32/hidpi/high-dpi-desktop-application-development-on-windows)

niksedk:
I also don't have a high res monitor, so I cannot test nor merge this.

nkh0472:
No need to change your monitor to test this.
Just right-click the desktop, open Display settings, and change "Scale and layout" from 100% to another value. 
Then maybe you need a sign-out or a restart to let it apply.

niksedk:
I've tested this a little... and I think it's a huge task!
SE has about 174 forms which all needs to be checked and actually most need to be fixed - some will probably take several hours to fix.
Still a good idea though, and this PR contains some of the starting code :)

Issue #4754, opened 2021-01-24
Release 3.6.0


niksedk:
Time for a new release :)

This release contains a lot of code/project restructuring to ensure SE will also work on .NET in the future (like .NET 5).

SE now uses .NET framework 4.7.2 (has been updated from 4.0). There are many reasons for this: makes much stuff much easier to code, works on recent Linux versions, RichTextBox, security fixes, performance. The downside is that WinXP/Vista will no run SE 3.6+, but SE 3.5.18 is a very good and stable release.

The "source view" is now hidden by default which gives a little more space for the list view. 
It's possible to toggle between "source view" and "list view" by "F2" (or context menu).

Some of the new stuff:
* Edit of bdsup/bdnxml/ts-w-img files
* Dark theme (finally out of beta - OmrSi also made many nice fixes here)
* Textbox syntax coloring + live spell check in text box (much of which was coded by OmrSi)
* Internal store for ASSA styles - thx Leon/z3us
* Copy-paste translate (pretty fast once you get the hang of it - probably works with all translation sites)

Change log: https://raw.githubusercontent.com/SubtitleEdit/subtitleedit/master/Changelog.txt

Beta link: https://github.com/SubtitleEdit/subtitleedit/releases/download/3.5.18/SubtitleEditBeta.zip

Do let us know if you known/find any bugs :)

TODO:
- [x] Test text boxes
- [x] Test ASS style manager
- [x] Test dark theme
- [x] Test networking
- [x] Test Batch convert
- [x] Test image export
- [x] Test with latest libmpv
- [x] Test OCR
- [x] Test translate
- [x] General testing  (+ fixing bugs)

m0ck69:
I cannot type Polish "ć" letter (alt+c) in text box. Capital "Ć" works tough.

moob158:
@niksedk 

pt-PT language update
[pt-PT.zip](https://github.com/SubtitleEdit/subtitleedit/files/5867670/pt-PT.zip)

OmrSi:
@m0ck69 Check your shortcuts, it might be colliding with one of them.

m0ck69:
@OmrSi nope, that's not it. Everything was OK before but after I updated SE to the newest beta the problem occurred.

OmrSi:
https://github.com/SubtitleEdit/subtitleedit/issues/4500#issuecomment-731059255 I guess this is back.

OmrSi:
[...]

PR #4788, opened 2021-02-08
Add a separate timer for live spell check

OmrSi:
Related to https://github.com/SubtitleEdit/subtitleedit/issues/4754#issuecomment-774634035

@niksedk Do you the language check and the initialization should be separated?
They work well as is.
If we separate them, the initialization will have to be done in multiple places and all functions that include it will have to be `async`.

### DIFF
--- a/src/ui/Forms/Main.cs
+++ b/src/ui/Forms/Main.cs
@@ -94,6 +94,8 @@ namespace Nikse.SubtitleEdit.Forms
         private bool _sourceViewChange;
         private int _changeSubtitleHash = -1;
         private int _changeOriginalSubtitleHash = -1;
+        private int _changeSubtitleTextHash = -1;
+        private Timer _liveSpellCheckTimer;
         private int _subtitleListViewIndex = -1;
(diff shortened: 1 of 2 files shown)


src/ui/Forms/Main.cs
```diff
@@ -4469,6 +4469,7 @@ private void ResetSubtitle(bool forceVideoReload = false)
             _subtitle = new Subtitle(_subtitle.HistoryItems);
             _changeOriginalSubtitleHash = -1;
             _changeSubtitleHash = _subtitle.GetFastHashCode(GetCurrentEncoding().BodyName);
```
OmrSi, line 4471: @niksedk Shouldn't this be `_changeSubtitleHash  = -1` since we're resetting it?

niksedk: Yes, I think so.

niksedk: Fixed in 801a8a9e7bf32dea73a59bc48bcfb054dd7e897a

PR #4791, opened 2021-02-09
Fix some issues with live spell check

OmrSi:
@niksedk Does this solve it?


src/ui/Controls/AdvancedTextBox.cs
```diff
@@ -879,14 +880,14 @@ private void UiTextBox_KeyPress(object sender, KeyPressEventArgs e)
         {
             if (IsLiveSpellCheckEnabled)
             {
```
OmrSi, line 882: Is this what's call short-circuiting?

PR #4797, opened 2021-02-13
Fix some issues in Live Spell Check

OmrSi:
Live spell check was trying to run in the original text box each time, I think it shouldn't.

Live spell check timer was being disposed of each time the settings is closed even if the live spell check option didn't change.

Right-clicking on the original text box was showing suggestions for the wrong word even if the mouse cursor wasn't on it.

The bounds for showing suggestions on right-mouse-click were wrong.

If you open SE with live spell check disabled, then enable it, it didn't work until you restart SE. (Wasn't subscribing to events if Live Spell Check is disabled on launch.)

### DIFF
--- a/src/ui/Controls/AdvancedTextBox.cs
+++ b/src/ui/Controls/AdvancedTextBox.cs
@@ -23,7 +23,7 @@ namespace Nikse.SubtitleEdit.Controls
         private bool _checkRtfChange = true;
         private int _mouseMoveSelectionLength;
 
-        private bool IsLiveSpellCheckEnabled => Configuration.Settings.Tools.LiveSpellCheck;
+        private bool IsLiveSpellCheckEnabled => Configuration.Settings.Tools.LiveSpellCheck && Parent?.Name == "textBoxListViewText";
         private Hunspell _hunspell;
         private SpellCheckWordLists _spellCheckWordLists;
         private List<SpellCheckWord> _words;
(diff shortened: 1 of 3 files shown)


OmrSi:
@niksedk Do you think it would be better to re-initialize the textbox instead of checking `IsLiveSpellCheckEnabled` each time the event fires?

src/ui/Controls/AdvancedTextBox.cs
```diff
@@ -23,7 +23,7 @@ public sealed class AdvancedTextBox : RichTextBox, IDoSpell
         private bool _checkRtfChange = true;
         private int _mouseMoveSelectionLength;
 
-        private bool IsLiveSpellCheckEnabled => Configuration.Settings.Tools.LiveSpellCheck;
+        private bool IsLiveSpellCheckEnabled => Configuration.Settings.Tools.LiveSpellCheck && Parent?.Name == "textBoxListViewText";
```
niksedk, line 26: Perhaps

`Parent?.Name == nameof(textBoxListViewText)`
or
`Parent == textBoxListViewText`

?

OmrSi: It would be better, but isn't `textBoxListViewText` an instance in Main? How would this class see it? Am I missing something?

niksedk: Nope... would have been nice.

OmrSi: Well, I can use a property in Main that returns the nameof it, what do you think?

niksedk: Could be a little better, so we avoid the string value... but I don't a see a super easy solution (I only see a public static property being set when creating the main form)

OmrSi: This seems a bit sloppy https://github.com/OmrSi/subtitleedit/commit/d86ed449cd8ec79ba9aa17677cf75e06ab87afa5

PR #4807, opened 2021-02-16
Fix fpc support

ngaullier:
I had vietnamese .fpc files that could not be loaded.
It was possible to fix it in PacUnicode.cs but I think it is much better to merge the implementations with Pac.cs which has been certainly much more tested during years.


niksedk:
Sorry, I prefer to have two different formats

ngaullier:
I understand you prefer to keep 2 distinct classes/source files, each with its dedicated file extension etc.
But do you mind if we make these two implementations strictly converge ? It is easier to maintain two source files with exact same structure and same PAC parsing algorithms. Only text decoding differs between PAC and PAC unicode, so it should be the same for the implementations : only the text decoding should differ.
Would you take into account a new pull request where I would not modify PAC but where PAC unicode would be completely rewritten to be a strict clone of PAC with only text decoding being different ?

niksedk:
OK, merging two formats in one file is fine - SE already does this for some formats like https://github.com/SubtitleEdit/subtitleedit/blob/master/src/libse/SubtitleFormats/FinalCutProXml19.cs

This PR contains some unrelated changes... it should not.

Could you also attach a sample file (vietnamese .fpc) or email it to me?

ngaullier:
OK, I will revise my implementation to have something like fcpxml.

Concerning the two other unrelated commits, would you like to have 1 PR for each ? Does they sound right to you ?

niksedk:
Just let the other stuff be... this PR was perhaps pretty close without them!

PR #4808, opened 2021-02-17
Use null-coalescing operator in SETextBox

OmrSi:


OmrSi:
@niksedk Just out of curiosity, was something wrong with this, or is it just more readable the way it is?

niksedk:
Yes, a bit more readable...

PR #4814, opened 2021-02-19
Allow more MPV options in "MpvExtraOption"

OmrSi:
I was trying to add `sub-justify=left` as an option to MPV, but it didn't work.
Turns out you have to do `sub-ass-justify=yes` as well for it to work.

Right now that's not possible, as we are only allowed one option, so I thought it would be better if more options separated by a space are allowed.

Made it split at spaces and trim any `-` at the start of each option.
The way I see it, this can now be used as you would use it in CLI, like arguments.
For example, `--sub-ass-justify=yes --sub-justify=left`
Couldn't find any [options](https://github.com/mpv-player/mpv/blob/master/DOCS/man/options.rst) that contain a space in them, so I think this is safe.

### DIFF
--- a/src/libse/Common/Settings.cs
+++ b/src/libse/Common/Settings.cs
@@ -1122,7 +1122,7 @@ $HorzAlign          =   Center
         public string VlcLocationRelative { get; set; }
         public string MpvVideoOutputWindows { get; set; }
         public string MpvVideoOutputLinux { get; set; }
-        public string MpvExtraOption { get; set; }
+        public string MpvExtraOptions { get; set; }
         public bool MpvLogging { get; set; }
         public bool MpvHandlesPreviewText { get; set; }
         public Color MpvPreviewTextPrimaryColor { get; set; }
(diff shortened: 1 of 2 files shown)


OmrSi:
@niksedk Okay, made it split at spaces and trim any `-` at the start of each part.
The way I see it, this can now be used as you would use it in CLI, like arguments.
Couldn't find any [options](https://github.com/mpv-player/mpv/blob/master/DOCS/man/options.rst) that contain a space in them, so I think this is safe.

OmrSi:
@niksedk Is there a problem with this? 
A user contacted me asking for a way to justify the text, I looked and found that you need the two options I mentioned and I'm waiting for it to be merged so I can send him a beta. 

I don't want to send of it if it's not going to get merged.

niksedk:
> 
> OmrSi wrote:
> @niksedk Is there a problem with this?

No, looks very nice <3
Nice code and nice PR description :)


I just wanted to test it, before merging.

Issue #4628, opened 2020-12-22
EBUSTL Command Line export issue on Linux (using Windows build with Mono)


batatop:
I'm using Subtitle Edit's Windows build on Linux. I use `Mono` to use the Windows package on Linux. I used it for many types such as `Spruce STL`, `SRT`, `VTT`, `SCC`, and `Cheetah`. I never had any issues but with `EBU STL`, I'm getting this error:
```
1: test.srt -> .../Desktop/subtitleEdit/test.stl...
ERROR: No data is available for encoding 20269. For information on defining a custom encoding, see the documentation for the Encoding.RegisterProvider method.
  at System.Text.Encoding.GetEncoding (System.Int32 codepage) [0x001d1] in <a17fa1457c5d44f2885ac746c1764ea5>:0 
  at Nikse.SubtitleEdit.Core.SubtitleFormats.Ebu+EbuTextTimingInformation.GetBytes (Nikse.SubtitleEdit.Core.SubtitleFormats.Ebu+EbuGeneralSubtitleInformation header) [0x001b8] in <c91ab6774be146d987d9212242ee7d60>:0 
  at Nikse.SubtitleEdit.Core.SubtitleFormats.Ebu.Save (System.String fileName, System.IO.Stream stream, Nikse.SubtitleEdit.Core.Subtitle subtitle, System.Boolean batchMode, Nikse.SubtitleEdit.Core.SubtitleFormats.Ebu+EbuGeneralSubtitleInformation header) [0x005b2] in <c91ab6774be146d987d9212242ee7d60>:0 
  at Nikse.SubtitleEdit.Core.SubtitleFormats.Ebu.Save (System.String fileName, Nikse.SubtitleEdit.Core.Subtitle subtitle, System.Boolean batchMode, Nikse.SubtitleEdit.Core.SubtitleFormats.Ebu+EbuGeneralSubtitleInformation header) [0x00006] in <c91ab6774be146d987d9212242ee7d60>:0 
[...]

Issue #2173, opened 2017-01-07
EBU STL: On attempt to read/write "No data is available for encoding 20269" error @ Ubuntu 16.10


p-kraszewski:
Environment:
* Version: _Subtitle Edit 3.5.1_, original SE351.zip and SE351PL.zip 
* OS: _Ubuntu 16.10 x86_64_
* Runtime: _Mono 4.2.1.102_
* Locale: _pl_PL.utf8_, happens also in _en_US_ and _POSIX_ (both 8bit ASCII-compatible)

*Any* operation on **EBU STL** format - read or write - results in `No data is available for encoding 20269` error. While there actually IS some wicked 20269 codepage, it has nothing to do with **any** of my OS settings. It doesn't matter if I try to open some valid, existing .stl file or try to save a new one with one line of text. Read results in a stacktrace, write just gives `No data is available for encoding 20269` infobox with OK button.

```
---------------------------
Windows Forms Thread Exception
---------------------------
An application error occurred in Subtitle Edit.
Please report at https://github.com/SubtitleEdit/subtitleedit/issues with the following information:

Error Message:
No data is available for encoding 20269.

Stack Trace:
  at System.Text.Encoding.GetEncoding (Int32 codepage) <0x7f1d22be8d80 + 0x00360> in <filename unknown>:0 
  at Nikse.SubtitleEdit.Core.SubtitleFormats.Ebu.GetCharacter (System.Boolean& skipNext, Nikse.SubtitleEdit.Core.SubtitleFormats.EbuGeneralSubtitleInformation header, System.Byte[] buffer, Int32 index) <0x40f93000 + 0x000c7> in <filename unknown>:0 
  at Nikse.SubtitleEdit.Core.SubtitleFormats.Ebu.ReadTextAndTiming (System.Byte[] buffer, Nikse.SubtitleEdit.Core.SubtitleFormats.EbuGeneralSubtitleInformation header) <0x40f717a0 + 0x0092f> in <filename unknown>:0 
  at Nikse.SubtitleEdit.Core.SubtitleFormats.Ebu.LoadSubtitle (Nikse.SubtitleEdit.Core.Subtitle subtitle, System.Collections.Generic.List`1 lines, System.String fileName) <0x40f71090 + 0x002a3> in <filename unknown>:0 
[...]

PR #4819, opened 2021-02-20
Ebu

jason-rayles-nbcuni:
I have added support for non-ASCII characters when encoding 20269 is unavailable.

This is a fix for linux users who use mono with Subtitle Edit.

I have observed these characters in Polish subtitles. I have tested converting from Ebu to SubRip, and this change fixes the display of non-ascii characters that previously did not display correctly after converting.

Relates to #4628 and #2173


src/libse/SubtitleFormats/Ebu.cs
```diff
@@ -23,6 +23,48 @@ public class Ebu : SubtitleFormat, IBinaryPersistableSubtitle
 
         private const string LanguageCodeChinese = "75";
 
+        private static readonly Dictionary<int, SpecialCharacter> SpecialASCIICodes = new Dictionary<int, SpecialCharacter>
```
niksedk, line 26: Could this just be a `Dictionary<int, string>` ?

Issue #4818, opened 2021-02-20
An idea on new ASSA category feature


LeonCheung:
Maybe it's nice to have an "As Default" button for users to pick which one would be autoloaded, so that they can more easily switch among projects without re-defining styles in the Default category time and time again for different projects, maybe like this:

![proto](https://user-images.githubusercontent.com/494788/108596748-fc144000-73c1-11eb-9d06-73c699ea988a.jpg)

niksedk:
Ah yes, much better solution IMO :)

OmrSi:
@niksedk Want me to work on this?

niksedk:
> 
> @niksedk Want me to work on this? Button or checkbox?

I think a button is fine... there should be some indication in the list view (e.g. "*") and remove/delete should not be possible :)
Give it a go :)

OmrSi:
Wouldn't disabling the "Set as default" button be enough? It would be disabled so the user would know he is in the default category.

niksedk:
> 
> 
> Wouldn't disabling the "Set as default" button be enough? It would be disabled so the user would know he is in the default category.

Could be... I will test, perhaps others will too :)

OmrSi:
@LeonCheung Do test it out :D
[SubtitleEditBeta.zip](https://github.com/SubtitleEdit/subtitleedit/files/6016388/SubtitleEditBeta.zip)

PR #4820, opened 2021-02-20
Add "Set as default" for ASSA categories

OmrSi:
Closes https://github.com/SubtitleEdit/subtitleedit/issues/4818

There are bound to be a couple of bugs, but that's what testing is for. :)

### DIFF
--- a/LanguageMaster.xml
+++ b/LanguageMaster.xml
@@ -2607,6 +2607,7 @@ Continue?</RestoreDefaultSettingsMsg>
     <Category>Category</Category>
     <NewCategory>New category</NewCategory>
     <CategoryName>Category name</CategoryName>
+    <CategorySetDefault>Set as default</CategorySetDefault>
     <CategoryNote>Note: "Default" styles will be applied to new ASSA files</CategoryNote>
   </SubStationAlphaStyles>
   <PointSync>
--- a/src/ui/Forms/Styles/SubStationAlphaStyles.Designer.cs
+++ b/src/ui/Forms/Styles/SubStationAlphaStyles.Designer.cs
@@ -1,6 +1,4 @@
-﻿using System.Windows.Forms;
-
-namespace Nikse.SubtitleEdit.Forms.Styles
+﻿namespace Nikse.SubtitleEdit.Forms.Styles
 {
     sealed partial class SubStationAlphaStyles
     {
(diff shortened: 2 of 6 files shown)


OmrSi:
Well, that was quick. xD

niksedk:
Well, looks good to me :)

LeonCheung:
Ah, so nice! Okay, here is the discussion:

1. Not sure why, I can only see the word "Set as":
screenshot1
Perhaps it's too long for UI display... Hmm, maybe some short word like "**Pin**" would also work?

2. This "**Remove**" button also needs a confirmation popup. 

3. Not sure if a "**Rename**" button is necessary, anyway. :)

LeonCheung:
And: 
screenshot2

1. IMO the **Remove All** button should also turn gray in this case, right?
2. When trying to export an empty category, nothing happens, so it may confuse users a bit, maybe we either turn it gray, or make a message popup like "Nothing to export"?

LeonCheung:
But another thing is, how to export multiple categories anyway? 😅

OmrSi:
Related to this commit:
1. I think it's the font size that's causing this, looks a bit big The button might need to be bigger.

Related to categories.
2. Shouldn't be hard to add.
3. That would be up to @niksedk, what do you think is the best way to support renaming?

From the initial implementation:
4.5. Are from even before I added categories.

And import/export don't support categories yet.

LeonCheung:
@OmrSi Cool, thanks. I think The word "Pin" would be better yet shorter as it exactly functions. Do you agree?

LeonCheung:
@niksedk @OmrSi I know this is a closed merge, but I think it's better to discuss it further right here. :)

Just have an idea, that could be quite workable and of high expandability for something we mentioned above:

proto3

We can put things like Set as default, rename, or even re-order list... anything in the panel. 
For a default category, we can put a pin to indicate that. Oh, just found a "pin" character: 🖈   Unicode?

And, IMO we can even use "+"for "New" and "-" for "Remove".  😄  Not a big deal, just brainstorming.

OmrSi:
@LeonCheung Yes, Nik talked about something like this and I will probably on it in the near future.
It will be a `...` button that contains the rest of the functionality.

LeonCheung:
Cool, just made another mockup, FYI:
proto4

PR #4829, opened 2021-02-21
Fixes related to ASSA Style Manager

OmrSi:
1. Fix  "Move Up/down/top/bottom" disappearing when the current category contains less than two styles.
2. Prompt for category removal.
3. Don't log name changes for storage. This was causing an issue where, when renaming a style in storage, the style changes for a line in the subtitle. I don't think it's even needed because only styles in the current file should be saved.
4. Increase the size of the "Set as default" button.
5. Manage ASSA style manager buttons state based on selected items, remove all and export and other buttons shouldn't be enabled when there is no style in the list or if no style is selected.

The next step is to make the import/export work on categories instead of styles, and maybe adding "Move to category" to the context menu.

Also, what do you think we should do about renaming categories? Should I make it so the user can write in the combobox and by that the name is changed?

We might also need to allow multiple selection.

### DIFF
--- a/LanguageMaster.xml
+++ b/LanguageMaster.xml
@@ -2609,6 +2609,7 @@ Continue?</RestoreDefaultSettingsMsg>
     <CategoryName>Category name</CategoryName>
     <CategorySetDefault>Set as default</CategorySetDefault>
     <CategoryNote>Note: "Default" styles will be applied to new ASSA files</CategoryNote>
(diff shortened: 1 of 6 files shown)


OmrSi:
@niksedk What of these did you solve in your commit so I can remove them?

niksedk:
> 
> 
> @niksedk What of these did you solve in your commit so I can remove them?

Hi OmrSi, probably only the last one ("Manage ASSA style manager buttons state based on selected items")

PR #4835, opened 2021-02-22
Fix a crash in RTL when activating syntax coloring

OmrSi:
I do not know if this is the best fix for the issue.

Here is what happens.
1. Run SE in RTL mode with syntax coloring disabled.
2. Go to the settings and activate syntax coloring.
3. A crash happens.

Somehow, OnRightToLeftChanged is causing a change in the Text property with a null value,
so SE crashes on `base.Text = string.Join("\n", s.SplitToLines());`


niksedk:
Looks good to me :)

PR #4837, opened 2021-02-23
Fix issue with RichTextBox in RTL with arrow keys

OmrSi:
This was causing an issue in RTL, where if you were at the end of the line and pressed `Right`, it should go right into the line, but it doesn't and vice versa.

Also in left-to-right mode, if a word at the beginning of the line is selected and you press left, it should remove the selection but it used to not do anything.


niksedk:
Good catch :)

Issue #4833, opened 2021-02-22
ASSA: export issue


LeonCheung:
When trying to export IN-FILE styles:
1. Newly added styles cannot be exported.
2. Newly removed styles can still be exported.

Unless click Apply or OK.  (latest beta) While for Storage, everything is fine.

PR #4838, opened 2021-02-23
Take another approach in exporting Assa styles

OmrSi:
In this commit, I switched exporting Assa styles from a header to a list of SsaStyle objects.
I think the coding is pretty straightforward.

It also closes https://github.com/SubtitleEdit/subtitleedit/issues/4833


niksedk:
Nice :)

Issue #4827, opened 2021-02-21
sync via other sub bottom scroll bar bug


AliNet97:
https://user-images.githubusercontent.com/75434680/108629567-ca5cb100-7475-11eb-987f-f71ff7cbea48.mp4

have a problem on sync via other sub with second (bottom) scroll bar when I want to see text only text tab not change

PR #4849, opened 2021-02-26
Allow the last column to be wider than the listview's width

OmrSi:
Sadly, I don't see another way to do this.
Feel free to close it if there is.

Closes https://github.com/SubtitleEdit/subtitleedit/issues/4827


OmrSi:
About the third commit, I think it would be easier than having to write the whole thing each time.

OmrSi:
@niksedk Since you closed the related issue, I'm guessing this is not a good option, so I will close it as well.
Do tell me if you want me to open it again.

PR #4860, opened 2021-03-03
Fix auto detect Bulgarian language

josdion:
Because of the poorly selected words, Bulgarian is always detected as Russian. This should fix the issue.


niksedk:
Nice, `"[Нн]е" should probably be removed as two letter words are best to avoid if possible - but do add a few more words (you can also use File -> Statistics to help select words).

josdion:
"[Нн]е" is one of the problems with detection of Bulgarian, it is because this is common word for both Bulgarian and Russian. The same is the case with - "вас", "за", "нас". These word should be removed or counted for both languages. Currently counted common words are: "[Нн]е", "Да", "как", "Но", "вас", "за", "нас"

niksedk:
I guess `[Нн]е` could be kept as it's used in the Russian word list... 
Could you possibly add a few more, longer (half) common words?

PR #4865, opened 2021-03-03
Update spa_OCRFixReplaceList.xml

paconaranjo:


niksedk:
The OCR fix replace list is only meant to contain fixes related to OCR errors, like "Hola" instead of "HQla"... 
(I don't read or speak Spanish although it would be nice)
Does this PR do more or is it only OCR related?

paconaranjo:
Sorry, my bad. Am afraid it’s not. Thanks for all your hard work.

PR #4869, opened 2021-03-06
Add a manager for ASSA Categories

OmrSi:
A manager for Assa Categories:

image

@niksedk @LeonCheung Please do test and tell me what bugs you find :)
Also your thought about how to make it look better
[SubtitleEditBeta.zip](https://github.com/SubtitleEdit/subtitleedit/files/6094487/SubtitleEditBeta.zip)


LeonCheung:
Great work OmSi. This is pretty cool! Okay, something I found so far:

1. When clicking "Copy" under Style Storage area one or twice, SE will crash.
2. Should we use some marker to indicate which one is currently set as default?
3. When trying to import categories, sometimes I get this:
error1

OmrSi:
@LeonCheung Thanks for testing :D

The default category is now colored green in the manager and the styles form.
Also added the ability to move styles between categories and made some shortcuts like Select All and Inverse selection possible in both the storage and the current styles list views.

The issue with import should be fixed.

As for the crash when using "Copy", I could not reproduce it:
Untitled Project

Here is the latest: [SubtitleEditBeta.zip](https://github.com/SubtitleEdit/subtitleedit/files/6096301/SubtitleEditBeta.zip)

LeonCheung:
@OmrSi Cool! The new feature feels so handy. Two cents:

1. About the crash issue, I mean ANOTHER Copy button. :)
2. When moving one more more styles to another category, what if there are already style(s) with the same name(s) in the target category?

OmrSi:
@LeonCheung You mean the copy button in the storage section or the "Move to storage" one?

Right now the name is just kept the same.
When changing a style's name, the text box of the names becomes red if it has the same name as another style, but if you ignore it, it just keeps them both with the same name.
I'm not sure if @niksedk intended it this way or not.
So, I thought why would we change its name when moved to another category if the user can have two styles with the same name in that category?

Do you think I should not allow 2 styles with the same name in the same category?

LeonCheung:
1. Yes, I mean the copy button in the storage section. It still crashes with the lastest test build above.
2. For the style name, maybe a warning message is welcome.
3. Plus: The tip here seems not dynamic, maybe it can be improved some? Or simply say "The styles in green will be applied to new ASSA files"
tip

OmrSi:
@LeonCheung Everything should be fixed. [SubtitleEditBeta.zip](https://github.com/SubtitleEdit/subtitleedit/files/6097671/SubtitleEditBeta.zip)
Give it a test. :D

OmrSi:
@LeonCheung Everything should be fixed. [SubtitleEditBeta.zip](https://github.com/SubtitleEdit/subtitleedit/files/6097671/SubtitleEditBeta.zip)
Give it a test. :D

I think the note is too long now, I might have to remove `(colored in green)` from it.

niksedk:
Looks good :)
Import and export works very well.

I think it would be nice with a "number of styles" column + an "is default" column.
The form is missing a min size or was it mean to be a fixed size dialog?

After resize, something is behind the import button.
image

OmrSi:
Adding a `Number of styles` column is a great idea.
Currently, the default category is colored green, would it be better to add a `Default` column as well?

Yes, I should do something about the size as well as the size of the buttons.

OmrSi:
@niksedk Should be a bit better now. :D

image

Although, I could not reproduce the thing behind the `Import` button after resizing.

OmrSi:
@niksedk Should be a bit better now. :D

image

Although, I could not reproduce the thing behind the `Import` button after resizing.

OmrSi:
@LeonCheung Here is the latest: [SubtitleEditBeta.zip](https://github.com/SubtitleEdit/subtitleedit/files/6102084/SubtitleEditBeta.zip)

LeonCheung:
@OmrSi Thanks for the update. Something I found so far:

1. When nothing is selected in the Categaties Manager window, I think the Remove button should not be clickable.
2. Currenly, empty name is allowed to create. Maybe we should not allow that?
3. When exporting categories, I just got one error message but I cannot reproduce it so far:
error2

OmrSi:
@LeonCheung Thank you for testing. :D

The crash was happening when you close the manager with no category selected, should be fixed now.
Empty names are no longer allowed and the remove button is disabled when no category is selected.

Here is the latest: [SubtitleEditBeta.zip](https://github.com/SubtitleEdit/subtitleedit/files/6103356/SubtitleEditBeta.zip)

LeonCheung:
One more thing to fix, and it would be perfect! It seems we can't remove multiple categories in one click:

error-1

OmrSi:
@niksedk Should be ready now.

PR #4878, opened 2021-03-09
Fix for modify selection last column size

OmrSi:
It used to leave a white part.

### DIFF
--- a/src/ui/Forms/ModifySelection.cs
+++ b/src/ui/Forms/ModifySelection.cs
@@ -615,15 +615,18 @@ namespace Nikse.SubtitleEdit.Forms
                                                  - listViewFixes.Columns[1].Width
                                                  - listViewFixes.Columns[3].Width
                                                  - 30;
-                return;
+            }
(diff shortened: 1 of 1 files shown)


niksedk:
Did you test this with and without styles? (e.g. srt and ass)

OmrSi:
@niksedk Yes. It seems to be working well here.

With srt:
image

With ass:
image

OmrSi:
@niksedk Also, can I make a method in `UiUtil` to this operation `listViewFixes.Columns[listViewFixes.Columns.Count - 1].Width = -2;` instead of haveing to write it all each time?

Something like: 
```
internal void SetListViewLastColumnSize(this ListView listview)
{
    listview.Columns[listview.Columns.Count - 1].Width = -2;
}
```

niksedk:
Had to revert as the horizontal scroll bar kept appearing.

OmrSi:
@niksedk I guess I will have to take another approach.

PR #4882, opened 2021-03-10
Modify selection last column patch

OmrSi:
Okay, this should be working now. Fingers crossed.

Also added a util method for setting the last column's size instead of having to write the whole thing each time.


niksedk:
Nice :)

PR #4891, opened 2021-03-14
Work on setting color and removing formatting in text boxes

OmrSi:
1. `Remove all formatting` shortcut used to not work for the focused text box.
2. Setting the alignment for the original text box when focused was not possible.
3. It wasn't possible to use the custom colors on a section of the text, it used to affect the whole text box, and it wasn't possible to use it inside the focused text box only. I moved the logic of setting the color from the text box's context menu to a method and used that method to set the color of the selected text if a text box is focused and for setting the color from the context menu..


niksedk:
Cool, setting the color on selected text was on my todo :)

I think that the setting of color on selected text should select the new html too - so we can toggle without added new color tags.

OmrSi:
@niksedk Good point. Reverted.
I forced-pushed because it's not worth a commit, I'm just going back to the original behavior.

OmrSi:
> Cool, setting the color on selected text was on my todo :)

The logic was already there, I just had to move it to a separate method. :D

PR #4905, opened 2021-03-19
Add "Subtitle formats" section to the settings

OmrSi:
Needs testing.

The only issue I couldn't solve is that the formats in the Combobox require SE to be restarted in order to apply the reordering and show the favorites first.

Since you have:
```
if (_allSubtitleFormats != null)
{
    return _allSubtitleFormats;
}
``` 

`_allSubtitleFormats` is always returned the same.
And regenerating `_allSubtitleFormats` takes a lot of time, maybe because it uses reflection to get the plugins formats.

As for the second commit, it seems that specifying the type of the array is enough for Resharper, no need to use the Cast. I did it in the same PR because I used the same logic in another file


niksedk:
Looks very  nice :)

The formats list view should probably have a search box as SE has *so* many formats :)
image

You could move your `Move the favorite formats to the top` code to a public method, that also could be called after `ShowSettings` in ´Main.cs´.

Will be very useful - also for me!

niksedk:
The default save format could be `(Auto)` or ` - Auto - ` or similar...

OmrSi:
@niksedk Fixed the updating issue and added the search box. :D

OmrSi:
@niksedk Fixed the updating issue and added the search box and made it `- Auto -` instead of black.

PR #4924, opened 2021-03-24
Ignore Alt alone in normal text box

OmrSi:
When pressing the `Alt` key alone while a text box is focused, SE used to ignore that keypress because it's annoying.

After adding the RichTextBox, we found that the code is causing an issue for the Alt codes, so we removed this feature.
But why remove it completely?
It used to word for the normal text box, so why not leave it for the normal text box?

In this PR, I just returned the feature for the normal text box, as it doesn't cause any issues in it.

### DIFF
--- a/src/ui/Forms/Main.cs
+++ b/src/ui/Forms/Main.cs
@@ -14396,6 +14396,13 @@ namespace Nikse.SubtitleEdit.Forms
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


OmrSi:
@niksedk This works with the normal text box with no issues, it's been working for a long time.
Or is the problem that you want the same behavior for both types?

PR #4932, opened 2021-03-26
Attempt to fix some issues with auto-translate selected lines

OmrSi:
Renumbering the selected lines without renumbering the translation result used to cause a crash.

Also passing the same object of paragraphs then renumbering is causing the selected lines to have the new numbers, which is why I made it pass a new paragraph object as a deep copy.

After the translation of the selected lines is done, only the text needs to be copied, no need to copy anything else.

I don't know if this is the best fix.


niksedk:
It might not be the best fix - like fixing the translator engine dependency of line numbers - but it's a good fix for now as the other fix will be a major change.

OmrSi:
@niksedk Yes, the whole thing should depend on the ID, but that does need some work.

PR #4949, opened 2021-04-02
Batch dialog: change cancel button from "Ok" to "Cancel"

cvzi:
I changed the label of the cancel button from "Ok" to "Cancel" in the "Batch convert" dialog.

If believe the label "Ok" is confusing for this button because if you press the button, the dialog closes and you lose all the settings in the dialog and you have to start from scratch.


OmrSi:
Well, no. If you change some settings and click `Ok`, the changed settings are saved.

Like, if you activate some options and then click `Ok` then open Batch Convert again, you'll see that the options you changed are saved.
I think `Cancel` suggests that all changes will be lost.

This is just my opinion.

cvzi:
When I reopen Batch Convert again, it's lost for me, the form is empty again

OmrSi:
The form is empty, but the options are saved.

PR #4958, opened 2021-04-05
Make extend to scene change with gap work for multiple scene changes

OmrSi:
It's used to only work for the scene change directly after the end of the line, this should make it work for all scene changes after the end.


OmrSi:
@niksedk Can you revert this please? I was using the shortcuts and found it more annoying.
If the line ends less the MinGap before the scene change and you use it, it moves it to the next scene change instead of applying the MinGap.

PR #4988, opened 2021-04-15
BUGFIX: FPC Paragraph Decode Fix

tomglover24:
While working on a library of subtitle validation tools using the LibSE I noticed some files would fail to validate, due to a missing tag or missing character.

Looking closely at the files in a Hex Editor I noticed that the last character was missing from single line paragraphs, and the second line of multi-line paragraphs. 

The change on **Line 128**  fixes the missing character, 11 is required as time codes use 8 bytes then length uses the next 2, so it needed to skip 1 more from the `timeStartIndex` to get the actual `maxIndex`.

The changes to **line 150/151** rectify an issue when reading the first paragraph where some valid text would be missed, see Story / Lang.

The two changes can been seen in the screenshot below 3.5.11 was the version i had installed locally and found the issue in, 3.6.0 was a build after making the two changes below.

This change has been tested against 100+ FPC subtitles, all are decoding correctly. 

Apologies in advance but I cannot provide the FPC subtitles as they are my employers content.

image

### DIFF
--- a/src/libse/SubtitleFormats/PacUnicode.cs
+++ b/src/libse/SubtitleFormats/PacUnicode.cs
@@ -125,7 +125,7 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
                 return null; // probably not correct index
             }
 
-            int maxIndex = timeStartIndex + 10 + textLength;
+            int maxIndex = timeStartIndex + 11 + textLength;
 
             var sb = new StringBuilder();
             index = feIndex + 3;
@@ -142,13 +142,29 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
                         {
                             if (buffer[j] == 0xff)
                             {
+                                // I never hit this on a breakpoint...
                                 buffer[j] = 0x2e; // replace end of line marker
                             }
                         }
-
+                        
                         sb.AppendLine(Encoding.UTF8.GetString(buffer, textBegin, textIndex - textBegin));
-                        textBegin = textIndex + 7;
-                        textIndex += 6;
+                        for (var v = 0; v < 5; v++)
+                        {
+                            if (buffer[textIndex + 1 + v] == 0x1F &&
+                                buffer[textIndex + 2 + v] == 0xEF &&
+                                buffer[textIndex + 3 + v] == 0xBB &&
+                                buffer[textIndex + 4 + v] == 0xBF)
+                            {
+                                // skip utf8 BOM
+                                textBegin = textIndex + 7;
+                                textIndex += 6;
+                            }
+                            else
+                            {
+                                // ASCI or Header Paragraph no BOM 
+                                textBegin = textIndex + 1;
+                            }
+                        }
                     }
                 }
                 else if (buffer[textIndex] == 0xFF)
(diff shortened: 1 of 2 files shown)


src/libse/SubtitleFormats/PacUnicode.cs
```diff
@@ -124,8 +124,8 @@ private static Paragraph GetPacParagraph(ref int index, byte[] buffer)
             {
                 return null; // probably not correct index
             }
-
-            int maxIndex = timeStartIndex + 10 + textLength;
+            
+            int maxIndex = timeStartIndex + 11 + textLength;
```
niksedk, line 128: Just made a quick test - this seems to give a lot of lines ending with `.` - even some already ending with `.` or `!`...

tomglover24: Thanks will investigate further as it is 100% chopping off content in certain paragraphs.

tomglover24: So running a further test on an FPC (Item2) and PAC(Item1) side by side in English and Burmese we are 1 character short if I reset Line 128 back to 10, however as you can see there are extra dots on the FPC Paragraph 1 & 3 (Item 2), these are 0xFF in the files, which i'm continuing to dig into, its also adding extra space on the second line of Paragraph 1.

The test files were generated in WinCaps which is the Broadstream / Screen Subtitling current commercial Subtitle editor.

I'll continue digging.

image

src/libse/SubtitleFormats/PacUnicode.cs
```diff
@@ -142,13 +142,29 @@ private static Paragraph GetPacParagraph(ref int index, byte[] buffer)
...
+                                // skip utf8 BOM
+                                textBegin = textIndex + 7;
+                                textIndex += 6;
+                            }
+                            else
+                            {
+                                // ASCI or Header Paragraph no BOM 
+                                textBegin = textIndex + 1;
+                            }
+                        }
```
tomglover24, line 167: This checks and resolves an issue with the "Header" Paragraphs using ASCI which was skipping content, and no BOM, while keeping the BOM skip in, stopping double line breaks later in the decode / display. 

Header Packets written from Subtitle Edit won't have this issues, as every paragraph is treated the same and the BOM is applied to every line.

src/libse/SubtitleFormats/PacUnicode.cs
tomglover24: Kept it extend to 11 as the full stops are present in the file at the correct byte, next to a control character which signifies a line break or paragraph change.

I'm not sure how we can check or change this.

If you count forward from TimeStartIindex 10 you end up with 03 (ETX) control character, instead of the first character of the text, 11 corrects that, then adds the text length to it. 

Please correct me if I'm wrong, I may have gotten completely the complete wrong end of the stick.

Thanks,
Tom

ngaullier:
Hi,
I can confirm PACUnicode.cs est buggy. PAC and PACUnicode share the exact same file structure, text paragraph etc. Only encoding differs. So I have sent yesterday a pull request to simply drop PACUnicode.cs code and extend PAC.cs to support unicode. Have tested it with screen tools and seems ok.
https://github.com/SubtitleEdit/subtitleedit/pull/5109
Nicolas

tomglover24:
#5109 is the way to go, as such i'll be closing this and testing that fix shortly.

Apologies for the delay work has been fairly hectic the last couple of weeks.

PR #4993, opened 2021-04-16
Try to fix an issue with format change and filename

OmrSi:
I emailed you about the issue I first noticed, I thought it had something to do with Undo:

https://user-images.githubusercontent.com/20923700/115031288-a72c0c00-9ed0-11eb-80ca-9ef22fba8669.mp4

It turned out it's because the `ComboBoxSubtitleFormatsSelectedIndexChanged` code is getting called when the format stays the same, even when you unsubscribe from it first.

I passed a `subscribed` boolean to the function, so it doesn't call the code when it's not subscribed.

I hope this is the proper fix.

### DIFF
--- a/src/ui/Forms/Main.cs
+++ b/src/ui/Forms/Main.cs
@@ -197,12 +197,12 @@ namespace Nikse.SubtitleEdit.Forms
             }
         }
 
-        private void SetCurrentFormat(string formatName)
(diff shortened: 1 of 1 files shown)


niksedk:
thx - already fixed it though

OmrSi:
@niksedk Sorry, should've worked on it in silence or told you I'm working on it.

PR #5000, opened 2021-04-20
Try to fix an issue with Undo and format

OmrSi:
The `ComboBoxSubtitleFormatsSelectedIndexChanged` event was sometimes being called even after you unsubscribe from it.
This was causing SE to record a history item for a format change to the same format.
Using `_makeHistoryPaused = true;` alone wasn't enough, because the history is recorded in `_subtile.MakeHistoryForUndo`.

The issue that was being caused is that this was showing up even when I don't change the format:

image

Which cases issue when using `Redo`.

Steps to recreate: Just open any file, make a change and use `Undo`, then check the `History`.

### DIFF
--- a/src/ui/Forms/Main.cs
+++ b/src/ui/Forms/Main.cs
@@ -197,12 +197,12 @@ namespace Nikse.SubtitleEdit.Forms
             }
         }
 
-        private void SetCurrentFormat(string formatName)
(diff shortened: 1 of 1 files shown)


OmrSi:
PR number 5000, Nice!

PR #5063, opened 2021-05-24
Use ngen64 on 64-bit

XhmikosR:
Notes:

1. Not sure what the current Inno Setup version is, but it looks like it's 6.x. The check should be adapted later

https://github.com/SubtitleEdit/subtitleedit/blob/8f3e5ae193b6f1fd8a74f665d7b43f1da8fdb020/installer/Subtitle_Edit_installer.iss#L25-L27

2. IIRC Inno Setup 6.x is only Unicode thus this check is probably moot:

https://github.com/SubtitleEdit/subtitleedit/blob/8f3e5ae193b6f1fd8a74f665d7b43f1da8fdb020/installer/Subtitle_Edit_installer.iss#L29-L31

**Untested!** Should work but someone should test it before landing it (and take care of the above TODO later)

### DIFF
--- a/installer/Subtitle_Edit_installer.iss
+++ b/installer/Subtitle_Edit_installer.iss
@@ -15,10 +15,10 @@
 ;* GNU General Public License for more details.
 ;*
 ;* You should have received a copy of the GNU General Public License
-;* along with Subtitle Edit.  If not, see <http://www.gnu.org/licenses/>.
+;* along with Subtitle Edit.  If not, see <https://www.gnu.org/licenses/>.
 
 ; Requirements:
-; Inno Setup Unicode: http://www.jrsoftware.org/isdl.php
+; Inno Setup Unicode: https://jrsoftware.org/isdl.php
 
 
 ; preprocessor checks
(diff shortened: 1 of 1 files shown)


installer/Subtitle_Edit_installer.iss
```diff
@@ -897,15 +900,16 @@ function InitializeSetup(): Boolean;
...
-    if not WizardSilent() then 
+    if not WizardSilent() then
     begin
       if SuppressibleMsgBox(CustomMessage('msg_AskToDownNET'), mbCriticalError, MB_YESNO or MB_DEFBUTTON1, IDNO) = IDYES then
         ShellExec('open','https://go.microsoft.com/fwlink/?LinkId=2085155','','',SW_SHOWNORMAL,ewNoWait,ErrorCode);
         Result := False;
       end;
     end;
   end;
-end.
```
niksedk, line 911: Hi @XhmikosR,

thx for the PR, works fine, except for the change in line 911 which give a compile error now.

XhmikosR: @niksedk can you try now? I'm having trouble building the project since I'm using Inno Setup 6.x.

niksedk: Works very well now :)

niksedk:
thx :)

PR #5109, opened 2021-06-07
Merge PAC (using codepages) and FPC (using unicode)

ngaullier:
Unicode is auto-detected and thus supported even with incorrect .pac extension.
Fixes FPC which was buggy for some encodings (some vietnamese characters at least).


niksedk:
Thx for the PR :)

I cannot open any of my .fpc files now...

Do you have a sample that works with your code, and not the current code base? Email is fine.

PR #5138, opened 2021-06-18
fix zero length string replace

ivandrofly:
image


niksedk:
Nice, how can I re-create the issue?

ivandrofly:
Settings
image

Then run Fix common errors with (maybe you won't need to check all options but I ran it these options): 
image

PR #5154, opened 2021-06-25
Export PNG Xml: Don't waste process

ivandrofly:
Simplifies loop / condition inside loop


src/ui/Forms/ExportPngXml.cs
```diff
@@ -751,12 +750,10 @@ private void ButtonExportClick(object sender, EventArgs e)
                 if (errors.Count > 0)
                 {
                     var errorSb = new StringBuilder();
-                    for (int i = 0; i < 20; i++)
+                    int min = Math.Min(20, errors.Count);
```
niksedk, line 753: Could you call the variable `maxErrorsToDisplay` or similar?

PR #5189, opened 2021-07-09
Fix "Text can fit in one line" Netflix Quality Check

### LINKED ISSUE #5188: Netflix quality check "Text can fit in one line" is broken
The netflix quality check "Text can fit in one line" does not always work correctly. Sometimes a subtitle can fit in one line but it is not flagged.

See attached subtitle where both lines should be flagged, however only the second one is with no apparent reason.
[split_lines.zip](https://github.com/SubtitleEdit/subtitleedit/files/6793493/split_lines.zip)

image

image

Qbort:
Fixes #5188

Need to merge lines shorter than `SingleLineMaxLength + 1` for AutoBreakLine to behave as expected


niksedk:
thx :)

PR #5345, opened 2021-09-17
Add `Clear` key to list of shortcut keys

OmrSi:
I don't know if it was intentionally omitted, but I wanted to use it and couldn't find it so I thought I'd add it.
I hope that's okay.


niksedk:
If it works, then it's fine :)

PR #5350, opened 2021-09-20
Update zh-Hans.xml

nkh0472:
For commits 58fa783 78387fd c703bca 8114795.

---
Line 2478 
`<BDOpensIn>BD sup/bdn-xml opens in</BDOpensIn>`
Does it ok to translate it to "Open BD sup/bdn-xml file..."?
If this is the title of a window, it might be ok. But I'm not sure where this string is used.


niksedk:
It will be used in Options - Setttings - Tools:
image

nkh0472:
Thanks, now it's ok for merge.

PR #5389, opened 2021-10-14
Bug fix for saving SMPTE

OmrSi:
I did a rename that renamed too much before making the previous pull request...


OmrSi:
@niksedk In the second commit, I noticed that the waveform doesn't set Smpte when opening SE with an already opened file, I tried to fix it the best I know how, do you think it's a proper solution or could it cause other issues?

OmrSi:
@niksedk In the third commit, I noticed that `Video Container` already has a property for `Smpte`, so I unified both properties.
No need for each one to be taken from a different place.

OmrSi:
@niksedk Just tested it, it works great as far as I can tell. :D

PR #5536, opened 2021-11-26
Some work on spell check design

OmrSi:
I never liked that the using a bold font was forced, maybe it looked good for English but It didn't look good for other languages, Arabic at least.
Also, the buttons that are not filling the groupbox didn't look good, and there wasn't enough proper space for the Undo button.
The way the Suggestions section was in a proper section of its own was better as well in my opinion.

So I did some tweaks, before:
image

After:
image

I added the `Bold` option as a hidden setting, one might activate it if the font is too small for them.

@niksedk What's your take on this? Do you think there should be visible options to select the font and its size and boldness in the UI?

### DIFF
--- a/src/libse/Common/Settings.cs
+++ b/src/libse/Common/Settings.cs
@@ -1186,7 +1186,6 @@ $HorzAlign          =   Center
         public bool FixContinuationStyleHideContinuationCandidatesWithoutName { get; set; }
         public bool FixContinuationStyleIgnoreLyrics { get; set; }
         public string SpellCheckLanguage { get; set; }
-        public bool SpellCheckFontBold { get; set; }
         public string VideoPlayer { get; set; }
         public int VideoPlayerDefaultVolume { get; set; }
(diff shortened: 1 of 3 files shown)


OmrSi:
Here is why it looks bad in Arabic:

image

niksedk:
Looks nice - probably the suggestion list should be higher.
Perhaps the text box should take the font for the text box? Else we could add a UI setting in a future version...

OmrSi:
> Perhaps the text box should take the font for the text box?

Yes, I was actually thinking about that.
Should it just take everything like the SEtextbox and then we wouldn't need the hidden setting? I think it's a good idea.

niksedk:
> > Perhaps the text box should take the font for the text box?
> 
> Yes, I was actually thinking about that. Should it just take everything like the SEtextbox and then we wouldn't need the hidden setting? I think it's a good idea.

Please do :)

OmrSi:
@niksedk How does this look?

image

PR #5567, opened 2021-12-06
Add some new strings

bovirus:
@niksedk 

I hope that now is ok.


bovirus:
@niksedk 

Could you please check and apply if is it ok?

PR #5619, opened 2021-12-18
libse: add .net6 target and improve Matroska zlib code

Bond-009:
Uses System.IO.Compression instead of a 3rd party library for zlib when targeting
.net6

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,6 @@
 version: "{build}"
 
-image: Visual Studio 2019
+image: Visual Studio 2022
 
 platform:
   - Any CPU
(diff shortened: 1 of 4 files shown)


niksedk:
Hi @Bond-009, thx for the PR :)
We already have a dot net 6 branch here: https://github.com/SubtitleEdit/subtitleedit/tree/feature/dotnet6-test

I'm not really sure what to do about the dotnet framework version. 
The dotnet framework 4.x SE versions work on Linux with Mono, but I'm not sure if dotnet 6 Winforms can be made working too? I was unable to make it work.
I thought perhaps a light version could be made in MAUI - but that's also not working on Linux atm.

Bond-009:
@niksedk My main interest is having libse target .net 6, we use it in https://github.com/jellyfin/jellyfin, should I open a new PR with the zlib changes against the .net 6 branch?

niksedk:
Yes, that would be fine :)

Issue #5634, opened 2021-12-23
Add group and description (if any) to "Multiple replace" results.


Y4kuzi:
This way, we can easily see what group and what specific rule is responsible for the proposed replace.
Useful if we want to improve/change/temporarily disable rules without going through the hassle of finding the correct rule.

PR #5643, opened 2021-12-27
Add affecting rule info to multiple replace

OmrSi:
Closes https://github.com/SubtitleEdit/subtitleedit/issues/5634

@niksedk How is this? I think only the log is missing now. I'll probably do it in a separate PR if you don't mind.


niksedk:
Looks good - no need for a log IMO :)

PR #5649, opened 2021-12-29
Unbreak line when using merge if needed

OmrSi:
When merging two lines, I think it would be better if the two lines are put on one line if they can fit in it?

Like if you have:
```
1
Hey.

2
How are you?
```

Isn't it better for it to become `Hey. How are you?` instead of:
```
Hey.
How are you?
```

when the combined line is shorter than the max length?

This might be a big change to "Merge" behavior, so I would understand if it's not okay.

### DIFF
--- a/src/ui/Forms/Main.cs
+++ b/src/ui/Forms/Main.cs
@@ -11468,9 +11468,17 @@ namespace Nikse.SubtitleEdit.Forms
                         .Replace(" " + Environment.NewLine, string.Empty)
                         .Replace(Environment.NewLine, string.Empty);
                 }
-                else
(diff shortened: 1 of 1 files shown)


niksedk:
We already have multiple shortcuts for this, merge and unbreak, merge and auto-break.

OmrSi:
@niksedk This is not the same.

`Merge and unbreak` will remove the line break in any case, and will do auto-break if the line is too long, which might change how the liens are split.
`Merge and auto-break` will do auto-break in any case.

This edit is not this nor that, it will only make the normal merge function unbreak if the text fits on one line.
Like, how would the user know that the text can fit on one line when merging and know that he should use `Merge and unbreak`?

PR #5674, opened 2022-01-07
Update zh-Hans.xml

nkh0472:
For https://github.com/SubtitleEdit/subtitleedit/commit/24ce68dece0f35066c75909d3505fd7bc9ed035f

---
I`m not sure why Line 8 was modified in commit 569bcc2f6d44f966cfbf53dd42d7efb18454def4.
No matter this item is useless or something, it is still shown in XML Content Translator. And after translating, it turns back to before.


niksedk:
About line 8, is not a tag to be translated but rather an url pointing to a translated help page, like https://github.com/SubtitleEdit/subtitleedit/blob/master/src/ui/Languages/es-ES.xml#L8

nkh0472:
Since there doesn`t have a Chinese version of the helpfile, I replaced it with the English version.

niksedk:
> Since there doesn`t have a Chinese version of the helpfile, I replaced it with the English version.

That will work fine, could you just leave it empty? When empty the default English link will also be used.

nkh0472:
> > Since there doesn`t have a Chinese version of the helpfile, I replaced it with the English version.
> 
> That will work fine, could you just leave it empty? When empty the default English link will also be used.

That`s OK... But the Translator will always show it as untranslated, looks quite annoying.

LeonCheung:
@nkh0472 Hi, could you help to check if the #5675 issue occurs to you? Thanks!

PR #5677, opened 2022-01-07
Rename reading speed in recalc shortcuts

OmrSi:


OmrSi:
@niksedk Just say the work if you don't think CPS is clear enough. :)

PR #5683, opened 2022-01-09
Use comma not semicolon in CSV export

OmrSi:
Was the semicolon intended?
When I opened my exported CSV file is was just one column.

The Encoding was also not set properly.


niksedk:
Looks good to me :)

(I don't have Excel/Word anymore)

Issue #3822, opened 2019-10-31
Slow reading / opening of files (Performance)


mgutt:
I'm using the windows client. If I open a mkv file that is located at a 10G NAS it opens super slow and takes several minutes although I can usually access it with ~400 MB/s:
![2019-10-31 08_30_01](https://user-images.githubusercontent.com/10757176/67927446-e8374000-fbb8-11e9-88ab-c12f61ca33ca.jpg)

If the file is located on my PC its much better, but I think only "ok" as Subtitle Edit reads files only with under 100 MB/s and the CPU is not really used, too:
![2019-10-31 08_28_32](https://user-images.githubusercontent.com/10757176/67927391-c63dbd80-fbb8-11e9-953b-9563e5973be9.jpg)

Copying a file is usualy possible with >1 GB/s:
![2019-10-31 08_27_06](https://user-images.githubusercontent.com/10757176/67927204-4dd6fc80-fbb8-11e9-9ab6-03414c50a7a7.jpg)

Why is Subtitle Edit so slow / not using the available performance?

PR #5754, opened 2022-02-02
 Use 16 MB buffer for parsing MKV

gvisokinskas:
Subtitle stream opening from MKV files looks to be rather slow when compared to other tools. Here's a benchmark for opening a single PGS subtitle stream from ~28GB MKV file:

| Program       | Time (seconds) |
|---------------|----------------|
| ffmpeg        | 27.583         |
| mkvtoolnixgui | 25             |
| subtitleedit  | 40.58          |

I've benchmarked this on my machine with a Gigabyte AORUS NVMe Gen4 SSD. Basically SubtitleEdit reads the file at ~400MB/s, whereas both FFmpeg and MKVToolNix read the file at ~1.2GB/s. This translates to the SubtitleEdit being around 1.6 times slower when opening MKV files.

I've started digging deeper, and it seems that the currently used `MemoryMappedFile` does not provide any noticeable performance benefit when compared to a plain `FileStream`. As a proof of concept, I've decided to try parsing the file with a 16 MB buffer. However, due to [performance issues](https://www.jacksondunstan.com/articles/3568) with the `FileStream`, I had to wrap it (inspired by https://stackoverflow.com/a/9643111) in a `ByteBuffer`. This led to the SubtitleEdit having the fastest reading time of all of the benchmarked applications, and (in one of the best scenarios) managed to read the file in ~19 seconds. 

Roughly speaking, by processing the file with a buffer of 16 MB, we are able to match (and even exceed) the speed of other MKV parsers, with a rough reading speed of ~1.5 GB/s.

I'm opening this PR for consideration, as I'm not sure if the same approach could be used in machines with HDDs or when reading from network drives (https://github.com/SubtitleEdit/subtitleedit/issues/3822)


niksedk:
Hm, I get much faster results with the `MemoryMappedFile`... I've no idea why I get different results than you.
I also use SSD disc (on a core i7 with 32GB RAM).

According to your linked stack overflow article, `BufferedStream` is only needed in dot net 3.5 or older.

The article from [Jackson Dunstan](https://www.jacksondunstan.com/) does sound interesting and could be worth trying out...

Also, how does this perform for you?
https://github.com/SubtitleEdit/subtitleedit/tree/feature/dotnet6-test
Binary: https://github.com/SubtitleEdit/subtitleedit/releases/download/3.6.4/SEBetaDotNet6.zip

PR #5768, opened 2022-02-06
Saving and restoring of the number of active audio track

chshkhr:
Hi,

I like your application very much. Great job!

I need to open several times the same video file on the same audio track which is not the first in the list. 
Any time I had to switch it back manually...
I do not have any experience in C# but managed to solve the task by following your code.
I think it could be a useful new feature.

By the way, I use the application for studying foreign languages combining it with QTranslate. It's rather convenient but could be even better with something like keystroke macro.

### DIFF
--- a/src/libse/Common/Settings.cs
+++ b/src/libse/Common/Settings.cs
@@ -2832,7 +2832,7 @@ $HorzAlign          =   Center
                     videoFileName = listNode.Attributes["VideoFileName"].Value;
                 }
 
-                string audioTrack = "1";
+                string audioTrack = "-1";
                 if (listNode.Attributes["AudioTrack"] != null)
                 {
                     audioTrack = listNode.Attributes["AudioTrack"].Value;
--- a/src/libse/Common/Settings.cs
+++ b/src/libse/Common/Settings.cs
@@ -20,6 +20,7 @@ namespace Nikse.SubtitleEdit.Core.Common
         public string FileName { get; set; }
         public string OriginalFileName { get; set; }
         public string VideoFileName { get; set; }
+        public int AudioTrack { get; set; }
         public int FirstVisibleIndex { get; set; }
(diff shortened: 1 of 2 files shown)


niksedk:
Cool, it works :)

src/libse/Common/Settings.cs
```diff
@@ -2830,6 +2832,12 @@ private static Settings CustomDeserialize(string fileName)
                     videoFileName = listNode.Attributes["VideoFileName"].Value;
                 }
 
+                string audioTrack = "1";
```
niksedk, line 2835: Could you used "-1" ?

src/ui/Forms/Main.cs
```diff
@@ -22536,6 +22544,18 @@ private void GotoSubPositionAndPause(double adjustSeconds)
...
+                   if (mediaPlayer.VideoPlayer is LibVlcDynamic libVlc)
+                    {
+                        libVlc.AudioTrackNumber = VideoAudioTrackNumber;
+                    }
+                    else if (mediaPlayer.VideoPlayer is LibMpvDynamic libMpv)
+                    {
+                        libMpv.AudioTrackNumber = VideoAudioTrackNumber;
+                    }
+                }
+
```
niksedk, line 22558: Could you move the above code to "VideoLoaded" ?

src/ui/Forms/Main.cs
```diff
@@ -2820,7 +2820,7 @@ public double CurrentFrameRate
 
         private void OpenSubtitle(string fileName, Encoding encoding)
         {
-            OpenSubtitle(fileName, encoding, null, null);
+            OpenSubtitle(fileName, encoding, null, 1, null);
```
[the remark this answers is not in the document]
chshkhr: Done

PR #5809, opened 2022-03-02
Adding Brazilian Portuguese to GoogleTranslationService

thefalked:
It was missing my language on the list.

I build the app and worked. If there's anything else to do let me know. Thanks.

### DIFF
--- a/src/libse/Translate/Service/GoogleTranslationService.cs
+++ b/src/libse/Translate/Service/GoogleTranslationService.cs
@@ -133,6 +133,7 @@ namespace Nikse.SubtitleEdit.Core.Translate.Service
                 new TranslationPair("PERSIAN", "fa"),
                 new TranslationPair("POLISH", "pl"),
                 new TranslationPair("PORTUGUESE", "pt"),
+                new TranslationPair("PORTUGUESE (BRAZILIAN)", "pt_br"),
                 new TranslationPair("PUNJABI", "pa"),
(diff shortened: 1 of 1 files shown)


hugok79:
@niksedk By default, the language code for Portuguese is **pt** **(Portuguese spoken in Portugal and only official language)**. However, since there are some variants such as Brazilian Portuguese, there is **pt_BR**. My question here is, if we add this pull to the list, should we also change the Portuguese from **pt** to **pt_PT** Portuguese (Portugal) so there is no confusion since Subtitle Edit uses the **pt-PT** and the **pt-BR** in its interface?
https://en.wikipedia.org/wiki/Portuguese_language

thefalked:
> @niksedk By default, the language code for Portuguese is **pt** **(Portuguese spoken in Portugal and only official language)**. However, since there are some variants such as Brazilian Portuguese, there is **pt_BR**. My question here is, if we add this pull to the list, should we also change the Portuguese from **pt** to **pt_PT** Portuguese (Portugal) so there is no confusion since Subtitle Edit uses the **pt-PT** and the **pt-BR** in its interface? https://en.wikipedia.org/wiki/Portuguese_language

Thinking of what you said, i just did some digging and the Google Translator API does not differentiate the `pt_PT` of `pt_BR`. It works but it doesn't differentiate. I also test some subtitles but they came the same with `pt`, `pt_PT` and `pt_BR`. Unfortunally i guess it's better to just close the pull request.

PR #5814, opened 2022-03-05
Update zh-Hans.xml

nkh0472:
For 327d53ebe37dbf58b5ebcb7eed6208afbe88830d

---
Question: "CPS only" means only displaying cps number or it has another meaning?


niksedk:
> Question: "CPS only" means only displaying cps number or it has another meaning?

Yes, "CPS only" means it's only for CPS (and not for calculating line length)

PR #5835, opened 2022-03-16
Remove names already present in "names.xml"

ivandrofly:
Remove names from local names list that are already present in "names.xml"


niksedk:
nice :)

PR #5886, opened 2022-04-05
Fix annoying bug when opening a file with video offset

OmrSi:
I've spent 2 hours trying to fix it and this is what I got. I'm not sure this is the way.

How to recreate:
1. Open any file.
2. Apply a video offset.
3. Focus the listview on a line that is not the first line, for example, double click on the fourth line.
4. Close the file.
5. Reopen the file.
The first line will get a double offset and the timecode with be corrupted.

When the file first opens, `timeUpDownStartTime` doesn't have the timecode with the offset, so this code works:
```
var startTime = timeUpDownStartTime.TimeCode;
if (startTime != null && last != null && _subtitle != null &&
	Math.Abs(last.StartTime.TotalMilliseconds - startTime.TotalMilliseconds) > 0.5)
{
	var dur = last.Duration.TotalMilliseconds;
	last.StartTime.TotalMilliseconds = startTime.TotalMilliseconds;
	last.EndTime.TotalMilliseconds = startTime.TotalMilliseconds + dur;
	SubtitleListview1.SetStartTimeAndDuration(_subtitleListViewIndex, last, _subtitle.GetParagraphOrDefault(_subtitleListViewIndex + 1), _subtitle.GetParagraphOrDefault(_subtitleListViewIndex - 1));
	showSource = true;
}
```

I tried setting `timeUpDownStartTime` when a file with an offset first opens.

### DIFF
--- a/src/ui/Forms/Main.cs
+++ b/src/ui/Forms/Main.cs
@@ -4394,6 +4394,7 @@ namespace Nikse.SubtitleEdit.Forms
             if (rfe.VideoOffsetInMs != 0)
             {
                 _subtitle.AddTimeToAllParagraphs(TimeSpan.FromMilliseconds(-Configuration.Settings.General.CurrentVideoOffsetInMs));
+                InitializeListViewEditBoxTimeOnly(_subtitle.GetParagraphOrDefault(rfe.FirstSelectedIndex));
                 _changeSubtitleHash = GetFastSubtitleHash();
                 if (_subtitleOriginal != null && _subtitleOriginal.Paragraphs.Count > 0)
                 {


niksedk:
@OmrSi: Hm, I cannot re-create this in latest beta... any hints?

OmrSi:
@niksedk Here are the settings:
 
image

Here is the file: [Episode 6.zip](https://github.com/SubtitleEdit/subtitleedit/files/8444833/Episode.6.zip)
Open it with any video.

Notice how the first line moves to 2 hours:
image

OmrSi:
https://user-images.githubusercontent.com/20923700/162252891-17e86d29-ca8d-486d-9f45-85c3f61f7ab4.mp4

OmrSi:
Note that this doesn't happen when the first line is selected when closing the file.

OmrSi:
@niksedk Here is what I think is happening:
In `ReopenSubtitleToolStripMenuItemClick` we have `OpenRecentFile(rfe);` and then `SetRecentIndices(rfe);`

We have in `OpenRecentFile`:
```
            OpenSubtitle(rfe.FileName, null, rfe.VideoFileName, rfe.AudioTrack, rfe.OriginalFileName, false);
```

In it is:
```
                _subtitleListViewIndex = -1;
                if (SubtitleListview1.Items.Count > 0)
                {
                    SubtitleListview1.SelectIndexAndEnsureVisible(0);
                }
```

The selected index changes from -1 to 0, but the offset then is still 0 so `timeUpDownStartTime` takes the exact start time of the first line, `01:00:15,792` without adding any offset.

Then `SetRecentIndices` works and changes the selected index with `SubtitleListview1.SelectIndexAndEnsureVisible(rfe.FirstSelectedIndex, true);`

So we are now moving from line 0 to line 3 for example, and we have: `var last = _subtitle?.GetParagraphOrDefault(_subtitleListViewIndex);
`
`_subtitleListViewIndex` is 0, but it takes the new timing of the first line: `00:00:25,792`
Then we have:
```
                    var startTime = timeUpDownStartTime.TimeCode;
                    if (startTime != null && last != null && _subtitle != null &&
                        Math.Abs(last.StartTime.TotalMilliseconds - startTime.TotalMilliseconds) > 0.5)
                    {
                        var dur = last.Duration.TotalMilliseconds;
                        last.StartTime.TotalMilliseconds = startTime.TotalMilliseconds;
                        last.EndTime.TotalMilliseconds = startTime.TotalMilliseconds + dur;
                        SubtitleListview1.SetStartTimeAndDuration(_subtitleListViewIndex, last, _subtitle.GetParagraphOrDefault(_subtitleListViewIndex + 1), _subtitle.GetParagraphOrDefault(_subtitleListViewIndex - 1));
                        showSource = true;
                    }
```
Here, `timeUpDownStartTime` still has the old time code which is `01:00:15,792`, but `last` is at `00:00:25,792` so the condition is met and the timing of `last` which is the first line, is changed.

 So I made SE set the timing of `timeUpDownStartTime` when opening a file and there is an offset, which will update it before moving on to the selected index in the file.

PR #5898, opened 2022-04-10
Fix bug when opening a file with video offset

OmrSi:
NOTE:
I closed the previous PR because I was setting the timing to the wrong index.
Also, I do not know if this is the best fix, but I've explained the problem as best as I can and spent as much time as I can trying to figure it out.

How to recreate:
1. Open any file.
2. Apply a video offset.
3. Focus the listview on a line that is not the first line, for example, double click on the fourth line.
4. Close the file.
5. Reopen the file.
The first line will get a double offset and the timecode with be corrupted.

Here is the file: [Episode 6.zip](https://github.com/SubtitleEdit/subtitleedit/files/8444833/Episode.6.zip)
Open it with any video.
The file needs to have an offset before you offset just the video.

https://user-images.githubusercontent.com/20923700/162252891-17e86d29-ca8d-486d-9f45-85c3f61f7ab4.mp4

### DIFF
--- a/src/ui/Forms/Main.cs
+++ b/src/ui/Forms/Main.cs
@@ -4394,6 +4394,7 @@ namespace Nikse.SubtitleEdit.Forms
             if (rfe.VideoOffsetInMs != 0)
             {
                 _subtitle.AddTimeToAllParagraphs(TimeSpan.FromMilliseconds(-Configuration.Settings.General.CurrentVideoOffsetInMs));
+                InitializeListViewEditBoxTimeOnly(_subtitle.GetParagraphOrDefault(_subtitleListViewIndex));
                 _changeSubtitleHash = GetFastSubtitleHash();
                 if (_subtitleOriginal != null && _subtitleOriginal.Paragraphs.Count > 0)
                 {


OmrSi:
@niksedk Here is what I think is happening:
In `ReopenSubtitleToolStripMenuItemClick` we have `OpenRecentFile(rfe);` and then `SetRecentIndices(rfe);`

We have in `OpenRecentFile`:
```
    OpenSubtitle(rfe.FileName, null, rfe.VideoFileName, rfe.AudioTrack, rfe.OriginalFileName, false);
```

In it is:
```
    _subtitleListViewIndex = -1;
    if (SubtitleListview1.Items.Count > 0)
    {
        SubtitleListview1.SelectIndexAndEnsureVisible(0);
    }
```

The selected index changes from -1 to 0, but the offset then is still 0 so `timeUpDownStartTime` takes the exact start time of the first line, `01:00:15,792` without adding any offset.

Then in `OpenRecentFile` we have `Configuration.Settings.General.CurrentVideoOffsetInMs = rfe.VideoOffsetInMs;` so the offset is set.

Then `SetRecentIndices` in `ReopenSubtitleToolStripMenuItemClick` works and changes the selected index with `SubtitleListview1.SelectIndexAndEnsureVisible(rfe.FirstSelectedIndex, true);`

So we are now moving from line 0 to line 3 for example, and we have in `SubtitleListView1SelectedIndexChange()`:
`var last = _subtitle?.GetParagraphOrDefault(_subtitleListViewIndex);`
`_subtitleListViewIndex` is 0, but it takes the new timing of the first line with the offset: `00:00:25,792`

Then we have:
```
    var startTime = timeUpDownStartTime.TimeCode;
    if (startTime != null && last != null && _subtitle != null &&
        Math.Abs(last.StartTime.TotalMilliseconds - startTime.TotalMilliseconds) > 0.5)
    {
        var dur = last.Duration.TotalMilliseconds;
        last.StartTime.TotalMilliseconds = startTime.TotalMilliseconds;
        last.EndTime.TotalMilliseconds = startTime.TotalMilliseconds + dur;
        SubtitleListview1.SetStartTimeAndDuration(_subtitleListViewIndex, last, _subtitle.GetParagraphOrDefault(_subtitleListViewIndex + 1), _subtitle.GetParagraphOrDefault(_subtitleListViewIndex - 1));
        showSource = true;
    }
```
Here, `timeUpDownStartTime` still has the old time code without the offset which is `01:00:15,792`, but `last` is at `00:00:25,792` so the condition is met and the timing of `last` which is the first line, is changed.

So I made SE set the timing of `timeUpDownStartTime` when opening a file and there is an offset, which will update it before moving on to the selected index in the file.

niksedk:
Thx for this - surely difficult... seems like the design of this could be better.

PR #5971, opened 2022-05-21
fixes zero bytes files detection

ivandrofly:
image


src/ui/Forms/Main.cs
```diff
@@ -3656,7 +3656,7 @@ private void OpenSubtitle(string fileName, Encoding encoding, string videoFileNa
             }
 
             // check for all binary zeroes (I've heard about this a few times... perhaps related to crashes?)
-            if (format == null && file.Length > 50 && FileUtil.IsSubtitleFileAllBinaryZeroes(fileName))
```
ivandrofly, line 3659: Let "IsSubtitleFileAllBinaryZeroes" handle the size

PR #5972, opened 2022-05-21
store read infos into model + display "..." when name is too long

ivandrofly:
- read dictionary info from zip to model
- when name is too long to be displayed on combobox-dictionary suffix it with "..." 

image

### DIFF
--- a/src/ui/Forms/GetDictionaries.cs
+++ b/src/ui/Forms/GetDictionaries.cs
@@ -109,7 +109,7 @@ namespace Nikse.SubtitleEdit.Forms
                 // ensure ellipses suffix on text overlaps
                 using (var graphics = Graphics.FromHwnd(IntPtr.Zero))
                 {
-                    double comboboxWidth = comboBoxDictionaries.Width;
+                    double comboboxWidth = comboBoxDictionaries.DropDownWidth;
                     // format display value
                     foreach (var dictionaryItem in dictionaryItems)
                     {
--- a/src/ui/Forms/GetDictionaries.cs
+++ b/src/ui/Forms/GetDictionaries.cs
@@ -2,6 +2,8 @@
 using Nikse.SubtitleEdit.Logic;
 using System;
 using System.Collections.Generic;
+using System.Drawing;
+using System.Drawing.Drawing2D;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


src/ui/Forms/GetDictionaries.cs
```diff
@@ -12,9 +13,33 @@ namespace Nikse.SubtitleEdit.Forms
...
+            /// Native Name
+            /// </summary>
+            public string NativeName { get; set; }
+            public string Description { get; set; }
+            public string DownloadLink { get; set; }
+
+            public override string ToString()
+            {
+                var displayName = $"{Name}{(string.IsNullOrEmpty(NativeName) ? "" : $" - {NativeName}")}";
+                if (displayName.Length > 55)
```
niksedk, line 32: Could you use `comboBoxDictionaries.DropDownWidth` instead of 55?

ivandrofly: `comboBoxDictionaries.Width or comboBoxDictionaries.DropDownWidth` is not working properly.

I also tried to use `TextWiddth.CalcPixelWidth` but there are way too different in value when compared with `comboBoxDictionaries.DropDownWidth`

src/ui/Forms/GetDictionaries.cs
```diff
@@ -108,7 +105,30 @@ private void LoadDictionaryList(string xmlResourceName)
...
+                        if (width > comboboxWidth)
+                        {
+                            double pixelChar = width / text.Length;
+                            var charCount = (int)Math.Floor(comboboxWidth / pixelChar);
+                            displayText = text.Substring(0, charCount - 3) + "...";
+                        }
+
+                        dictionaryItem.DisplayText = displayText;
+                    }
+                }
```
ivandrofly, line 128: @niksedk updated

src/ui/Forms/GetDictionaries.cs
```diff
@@ -77,18 +96,40 @@ private void LoadDictionaryList(string xmlResourceName)
...
+                            Description = description,
+                            DownloadLink = downloadLink
+                        });
+                    }
+                }
+
+                // ensure ellipses suffix on text overlaps
+                using (var graphics = Graphics.FromHwnd(IntPtr.Zero))
+                {
+                    double comboboxWidth = comboBoxDictionaries.Width;
```
niksedk, line 112: Should probably be DropDownWidth, right?

ivandrofly: > Should probably be DropDownWidth, right?

Changed :) but they return the same value!

PR #5991, opened 2022-05-28
fixes always true

ivandrofly:


src/libse/Forms/FixCommonErrors/FixMusicNotation.cs
```diff
@@ -117,18 +117,18 @@ private static string HandleQuestionMarks(string input)
             }
 
             // be stub when looking for question marks
-            string noTagsText = HtmlUtil.RemoveHtmlTags(text, true);
+            var noTagsText = HtmlUtil.RemoveHtmlTags(text, true);
 
             // check if text starts and ends with?
-            if (noTagsText.StartsWith('?') && noTagsText.EndsWith('?') && text.Length > 1)
+            if (noTagsText.StartsWith('?') && noTagsText.EndsWith('?') && noTagsText.Length > 2)
```
ivandrofly, line 123: This is the fixes, the remaining changes are the **explicit** type to **var**

PR #6118, opened 2022-08-04
Final fix for go to with SMPTE

OmrSi:
Turns out I've been going about it the wrong way.
The problem was that you have two ways to set `CurrentPosition`:
1. `mediaPlayer.VideoPlayer.CurrentPosition` which doesn't make any edit to timing when setting or getting for SMPTE
2. `mediaPlayer.CurrentPosition` which does the needed fix.

It used to be that sometimes `mediaPlayer.VideoPlayer.CurrentPosition` is used instead of `mediaPlayer.CurrentPosition` so the time isn't adjusted properly,
I've moved to use `mediaPlayer.CurrentPosition` in every position, hoping this would fix everything as it would get the timing without SMPTE and set it with SMPTE.

### DIFF
--- a/src/ui/Forms/Main.cs
+++ b/src/ui/Forms/Main.cs
@@ -15948,7 +15948,7 @@ namespace Nikse.SubtitleEdit.Forms
             {
                 if (mediaPlayer.IsPaused)
                 {
-                    var pos = mediaPlayer.VideoPlayer.CurrentPosition;
+                    var pos = mediaPlayer.CurrentPosition;
                     var paragraph = _subtitle.GetFirstParagraphOrDefaultByTime(pos * TimeCode.BaseUnit);
                     if (paragraph != null &&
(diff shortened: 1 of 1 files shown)


src/ui/Forms/Main.cs
```diff
@@ -29560,7 +29554,7 @@ public void GotoPrevSubPosFromvideoPos()
         {
             if (mediaPlayer.VideoPlayer != null && _subtitle != null)
             {
-                double ms = mediaPlayer.VideoPlayer.CurrentPosition * TimeCode.BaseUnit;
+                double ms = Math.Round(mediaPlayer.CurrentPosition * TimeCode.BaseUnit, MidpointRounding.AwayFromZero);
```
OmrSi, line 29557: The round here is because I've had cases where the video position is at something like 39.999999 and the start time is 40, and since it's ms, I don't see a need for the numbers after the decimal point.

PR #6132, opened 2022-08-11
Allow empty splitting for split at video/cursor

OmrSi:
I know, I know, just… hear me out on this. "^^

So, let's say I have this line:
image

The text written in the text box is for the section that is before the video position, so now I want to split the line at video position and keep the whole text in the first line, while the second line is empty and takes the rest of the duration:
image

Right now, there is no easy way to do this, you'd have to type an extra word and split before it, or set end time, add a new line, and set its end time as well.
But what's the harm in allowing the user to split at the start or end of the line when using split at video/cursor?
He would know where the cursor is, thus knowing that he's making a new empty line for the rest of the duration.
Right now, using split at video/cursor does nothing when it's at the start or end of the line, so I would see this as an extra feature. :D

Are there any side effects for removing the index condition in the split function? I could not find any.

### DIFF
--- a/src/ui/Forms/Main.cs
+++ b/src/ui/Forms/Main.cs
@@ -11193,7 +11193,7 @@ namespace Nikse.SubtitleEdit.Forms
 
                 string oldText = currentParagraph.Text;
                 var lines = currentParagraph.Text.SplitToLines();
-                if (textIndex != null && textIndex.Value > 1 && textIndex.Value < oldText.Length - 1)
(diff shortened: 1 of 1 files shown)


niksedk:
> @niksedk I can tell you're having doubts, tell me what you're thinking. :D

He, I'm somewhat AFK atm, but I would like to test that other stuff that calls this split method, will still work as before... will do so soonish.

Issue #6137, opened 2022-08-11
[Shortcut suggestion] Repeat selected line


cyzs233:
Most language learning tools benefit from the ability to A-B cycle given dialogues, implementing this functionality will make subtitleedit more desirable to language learners who seek precise replay control on personal favored linguistic Corpus in the form of digital media, ie, movies.
Other usecase would be to actually transcribe each line in subtitleedit.

Keyword-Stuffing：A-B loop, Auto-repeat, replay, cycle, fractal, quantum physics.

OmrSi:
I think A-B loop is a bit hard, but a good idea in my opinion might be an infinite loop shortcut for a line, the start and end of the line would be the A and B.

niksedk:
I think SE already has this - via translate mode:

![image](https://user-images.githubusercontent.com/1261331/185037457-fd32f90e-2827-438f-b6c6-02f41646ac19.png)


Or alternatively use F5:

![image](https://user-images.githubusercontent.com/1261331/185037367-be653a07-d6f2-4d97-8410-5a7f8f97914b.png)

niksedk:
> transcribe each line in subtitleedit

Do you mean like, right-click in list view?
![image](https://user-images.githubusercontent.com/1261331/185037806-63d31353-991c-47bb-8764-9df906eb7560.png)

OmrSi:
@niksedk True, but it's somewhat limited since it's only in `translate mode`.
As for play selected lines, it plays them once while loop would keep playing it.

I do think maybe a better logic is needed?
Like a pause event for the video where it would reset the variables that should be reset when pausing a video.

cyzs233:
Yes, besides the inconvenience of it being a `translate mode` only utility as covered by OmrSi, this shortcut was initially requested more towards the language learning direction on top of the already usable `AUTO Repeat`. 
When you're just started getting your hands on a language, your listening skills are limited and most of the time you end up looping a given dialogue over and over until enlightenment finally descends upon you. 
The current 'AUTO Repeat' is designed for that specific widget, the fact that it has text box always focused, only has redefined repeat count option, having to **use your mouse** to click that not-so-obvious 'play' trigger button, only viable in 'translate mode', I suggest we make it a standalone shortcut.

niksedk:
[...]

PR #6156, opened 2022-08-16
Add `Loop selected line` shortcut

OmrSi:
Somewhat related to https://github.com/SubtitleEdit/subtitleedit/issues/6137

I would understand if you think it doesn't look too good.


niksedk:
@OmrSi : Could you make it work with "selected lines" (like F5 Play selection)?

OmrSi:
@niksedk I really tried to make it work like that but I faced one problem that I could not solve.
Should I push a commit and you could take a look?

niksedk:
Sure :)
What is the issue?

OmrSi:
@niksedk This line:
image

I causing an issue, when looping it keeps skipping to the end of the first line.
So if you have 3 lines, it skips the first one and loops the other two.
I can't figure out what it's for or what condition I need to add here.

niksedk:
Yes, there were some problems with `Play selection` when `Auto-continue` was on.
Also, going back to a previous video position is pretty slow which gives some problems - should probably have been async or have a callback for when setting video position is done.... perhaps at some later point...

So, I've updated Main.cs - which mean you might need to re-implement your main changes (and now you can use `_playSelectionIndexLoopStart` to set the paragraph start index for the first selected sub in the selection loop - I hope).

OmrSi:
@niksedk This seems to be working as far as I can see. I removed the debugging lines left by your last commit.

The loop and selection play weren't being reset when pausing via a click on the video, I tried solving that by making a new event and resetting the fields, let me know if that behavior was intended so I can remove this change.

It's probably better if you can also do a couple of tests of your own.

niksedk:
Only found a minor issue with `play selected lines` after `loop selected lines`. Those two command might not work on slow computers, but it works for me (except when using mpc-hc).

PR #6195, opened 2022-08-31
Add support for OCR via Google's Cloud Vision API

Flitskikker:
Hello again,

I made some progress with Cloud Vision for OCR. :)
It works surprisingly well for distorted and unclear images, where Tesseract has a hard time and where Binary image compare stands no chance.

image

It does need an API key (can be the same as the Google Translation v2 API key), and it's not entirely free. Still, it's not super expensive (see [pricing](https://cloud.google.com/vision/pricing)) and you get amazing results.

I tried to copy the strategy pattern used for Translation APIs for easier expansion to other services.

Other OCR improvements (mainly focused on Binary image compare which I often use):
- On the OCR Character window: 
  - When you click "Abort" and the subtitle was not empty, the existing text will now not be overwritten anymore.
  - When you click "Skip", the inputted character will be used (but not saved to the DB), instead of always '\*'. When no character was entered, a '\*' is used as usual.
- For [VSF](https://sourceforge.net/projects/videosubfinder/) users:
  - Fix timecodes being all zeroes when importing images with double file extensions (e.g. 1_43_03_440__1_43_07_079_0054600317006630072001280.jpeg.png)
  - If you're OCR'ing processed images (e.g. from the RGBResults folder), but the original images (in the RGBImages folder) are still available, the option "Send original images" will appear, to send those to Cloud Vision instead.

One drawback is that it does one API request per subtitle. In theory, the API supports up to 16 images per request.
That would probably be faster, but it appears this would be a fundamental change to the logic in VobSubOcr.cs to update more subtitles at once.

Please let me know what you think. :)

### DIFF
--- a/src/libse/Common/Settings.cs
+++ b/src/libse/Common/Settings.cs
@@ -2065,6 +2065,7 @@ $HorzAlign          =   Center
         public int UnfocusedAttentionPlaySoundCount { get; set; }
         public string CloudVisionAPIKey { get; set; }
         public string CloudVisionLanguage { get; set; }
+        public bool CloudVisionSendOriginalImages { get; set; }
 
         public VobSubOcrSettings()
         {
@@ -2095,6 +2096,7 @@ $HorzAlign          =   Center
             UnfocusedAttentionPlaySoundCount = 1;
             CloudVisionAPIKey = string.Empty;
             CloudVisionLanguage = "en";
+            CloudVisionSendOriginalImages = false;
         }
     }
 
@@ -7395,6 +7397,12 @@ $HorzAlign          =   Center
                 settings.VobSubOcr.CloudVisionLanguage = subNode.InnerText;
             }
 
+            subNode = node.SelectSingleNode("CloudVisionSendOriginalImages");
+            if (subNode != null)
+            {
+                settings.VobSubOcr.CloudVisionSendOriginalImages = Convert.ToBoolean(subNode.InnerText, CultureInfo.InvariantCulture);
+            }
+
             foreach (XmlNode groupNode in doc.DocumentElement.SelectNodes("MultipleSearchAndReplaceGroups/Group"))
             {
                 var group = new MultipleSearchAndReplaceGroup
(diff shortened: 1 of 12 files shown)


niksedk:
Hi @Flitskikker,

Super nice work :)

> In theory, the API supports up to 16 images per request.
OK, if you pay per request then this should probably be fixed at some point - otherwise it's not a concern I think. Nice that you keep the same httpclient for the session, so performance should still be nice.

"Language hint" should probably just be language, and the combobox should probably contain full language names... but that's details.

niksedk:
Included in latest beta: https://github.com/SubtitleEdit/subtitleedit/releases/download/3.6.7/SubtitleEditBeta.zip

Flitskikker:
> Hi @Flitskikker,
> 
> Super nice work :)

Thank you!


> > In theory, the API supports up to 16 images per request.
> 
> OK, if you pay per request then this should probably be fixed at some point - otherwise it's not a concern I think. Nice that you keep the same httpclient for the session, so performance should still be good.

Sorry, my mistake.
The request body of the `images:annotate` method asks for a `requests` array, which supports up to 16 requests (with 1 image per request).
So if you send 16 images, you still pay for 16 requests I think. No extra money lost. 😁


> "Language hint" should probably just be language, and the combobox should probably contain full language names... but that's details.

The list of supported languages seemed to be dynamic:
https://cloud.google.com/vision/docs/languages

I added the supported languages to the list, but users can still enter one of the other codes for the experimental languages of course.

nshmyrev:
You can integrate something open instead like

https://github.com/FangShancheng/ABINet
https://huggingface.co/spaces/tomofi/ABINet-OCR

or even easyocr

https://github.com/JaidedAI/EasyOCR

Flitskikker:
> $1.50 for 1000 images is quite expensive. On average, DVD subtitles have 2-3k lines.

Whoops, I thought it was $1.50 for 1001 - 5,000,000 images. 🫥
Personally, I use Binary image compare as my main driver and now Cloud Vision as a back-up for unclear images, so that reduces costs already.
Also, Google offers $300 in free credits for the first 12 months for new users - at least when I signed up - so maybe that's also something to consider.
 
> Wouldn't it be possible to concatenate a bunch of lines to pass them through? I'm pretty sure there could be a way to distinguish where each subtitle starts and ends, so that it's separated nicely. I don't know how deep changes in the SE core would be required to be done, though.

I'm afraid that goes beyond my capabilities, so that would be something for @niksedk.
But it surely is a good idea for future cost/image optimalization.

> You can integrate something open instead like
> 
> https://github.com/FangShancheng/ABINet https://huggingface.co/spaces/tomofi/ABINet-OCR
> 
> or even easyocr
> 
> https://github.com/JaidedAI/EasyOCR

I have tried to make this feature extendable, so you could add support for those services by using `IOCRService` and `IOCRStrategy`, providing that they have an online API.

nissansz:
Any outsourcing work for subtitle extraction? I accept such work. steve8000818@gmail.com

AommyPatty:
It might be a dumb question but what is API key?
I got JSON credential file with a lot of information, but I don't know which text I need to copy into the program.

PR #6199, opened 2022-09-01
Some Cloud Vision improvements

Flitskikker:
Some improvements after testing:

- Cloud Vision doesn't seem to like transparent images, so the sub images will now not be made transparent for Cloud Vision
- Since the languages box is now selection-only, I added the [experimental languages](https://cloud.google.com/vision/docs/languages) too. They will get supported eventually.
- Fix language code not being saved (it saved the name)
- Fix language list not sorted and back to English after starting OCR
- Add missing translation for group box title
- Some comments / minor refactoring


cyzs233:
VSF generate tons of extra images(usually double-triple the actually lines). A grid window that can filter images would be a huge money saver.
Blueprints:
A pop-up window that displays all the images, preferably displays them in grid view for maximum processing(manually deselect) speed. Ideally a simple click on images to select and deselect an image. And a must have is a "Duration limit" to filter out all the images with a duration under a certain threshold (0.5ms for example), which are usually empty images produced by VSF.

Besides VSF supports, I also have some other thoughts that might be useful.

- I think we also needs a specific OCR fix list for **OCR through APIs**. 
Those services do their own spelling check and the result are most of the time exceedingly good. For Altaic language family, it usually means specing detection between character. For other languages, it probably just some punctuation errors. The current **auto correction** is recommended disable for speed concerns. 

- Maybe a redesign of this OCR window?
A standalone popup window for OCR results proofreading would be most ideal, we don't have to refactor the current interface in this way, just adding more popup windows.😉
Here is a screenshot of a tool with a nice interface. There is usually little to none errors for you to fix, so I find be able to scriming through as fast as possiable should be the top priority.
Subtitleedit OCR proofreading window

PR #6213, opened 2022-09-03
Show HEX and ASSA color with RGB and copy on click

OmrSi:
Thought they would make nice additions from my personal experience. :)

### DIFF
--- a/LanguageMaster.xml
+++ b/LanguageMaster.xml
@@ -236,6 +236,9 @@ Note: Do check free disk space.</WaveFileMalformed>
   </AssaResulationChanger>
   <ImageColorPicker>
     <Title>Image color picker</Title>
+    <CopyColorHex>Copy to clipboard as HEX color {0}</CopyColorHex>
+    <CopyColorAssa>Copy to clipboard as ASSA color {0}</CopyColorAssa>
+    <CopyColorRgb>Copy to clipboard as RGB color {0}</CopyColorRgb>
   </ImageColorPicker>
   <AssaSetBackgroundBox>
     <Title>Generate background box</Title>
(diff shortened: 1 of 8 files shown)


niksedk:
Perhaps it's better with a context menu like?
````
  Copy to clipboard as RGB HEX "FFBB55"
  Copy to clipboard as ASSA HEX "BBFF55"
  Copy to clipboard as RGB "RGB(255,66,44)"
````

OmrSi:
Hmm. Wouldn't the color change when you try to click on the menu item?
I don't see a point in copying RGB for subtitling purposes, and copying when clicking makes more sense, I think.
You know, since it saves the color when you click.

OmrSi:
I agree that a right-click to copy ASSA color isn't the brightest idea, but it seems a good trick for a user to discover. xD

OmrSi:
@niksedk You'll just have to update the `.resx` file yourself. I write the designer code manually because the sizes get all messed up when I changed something, so the `.resx` file isn't updated.

niksedk:
Nice, thx :)

A bit related, I've some problems with colors in ASSA after burn-in. ASSA colors are changed a little. which is not good... I've compensated by using ` var colorMatrix =  "bt601:bt709";` in the image color picker, but I guess I will have to move this to the burn-in and change the color matrix for the burn-in assa file in SE. 
Any thoughts on this?

OmrSi:
No, sorry, I haven't come across this before. :\

The only problem that comes to mind is that usually when I use external color pickers, I notice that they don't exactly match the color I take from the video for some reason, so I take the color and use a lighter version so it visually matches the video.
Now I'm thinking, would that color then match after I burn the subtitle in? Strange.

PR #6218, opened 2022-09-04
Add "Convert colors to dialog" tool

Flitskikker:
I integrated another tool I wrote a few years ago into SE.
This adds a feature to convert subtitles with different colors (in Teletext subtitles, for example) to dialogs with dashes.

Available via the Tools menu:

image

(Only visible when the subtitle contains color tags)

image

It follows the current dialog style from the settings.

Teletext subtitles regularly have multiple speakers per line, so possibly you don't want to add new lines by default. Then you can uncheck the second option.

Also available in Batch convert:

image

Example:

```
<font color="yellow">Are you sure you want to do this?</font>
<font color="white">Yeah, you?</font> <font color="yellow">Yes.</font> <font color="white">Okay, bye.</font>
```

Result:

```
Are you sure you want to do this?
-Yeah, you? -Yes. -Okay, bye.
```
(dialog style = only dash second line without space)

It also works when the default color has no color tags, like this:

```
<font color="yellow">Are you sure you want to do this?</font>
Yeah, you? <font color="yellow">Yes.</font> Okay, bye.
```


niksedk:
SE also has this as a plugin... I'll do some testing...

niksedk:
These do not work:

````
1
00:00:01,000 --> 00:00:03,000
<font color="#ffffff">How ar you?</font>
I'm fine.

2
00:00:08,124 --> 00:00:10,124
How ar you?
<font color="#ffffff">I'm fine.</font>

````

niksedk:
Some unit tests would be nice - see e.g. https://github.com/SubtitleEdit/plugins/blob/master/UnitTests/ColorToDialogUnitTest.cs

PR #6224, opened 2022-09-07
"Convert colors to dialog" improvements

Flitskikker:
I implemented your suggestions. :)

- Add some unit tests (I used yours and added some more)
- Don't add dashes if there are already dashes (based on your unit tests)
- Use SafeSubstring everywhere just to be sure...


niksedk:
Nice :)

PR #6231, opened 2022-09-07
Work on generate blank video

OmrSi:
I don't know if it's intentional, but I think it's better for the empty video to take the name of the subtitle file if a subtitle file is loaded.


OmrSi:
`Merged into main`, gonna have to get used to that. xD

niksedk:
He, yes. `main` is the new default :)
Seems to work...

PR #6235, opened 2022-09-09
Fix switched shortcuts

OmrSi:
Also, are you sure this is how you want the toggle shortcut to work?
Right now, if you toggle when the video is paused, it will reset when you play and pause.


niksedk:
thx :)

PR #6256, opened 2022-09-16
Add configurable insert option for "Skip" action during OCR

Flitskikker:
When working with non-clean images, you may want to input characters without saving them to the OCR database.
Before, you had to skip them and replace the asterisks afterwards.

This adds an option to either add an asterisk or add the current input when clicking Skip:

image

This replaces my previous commit (https://github.com/SubtitleEdit/subtitleedit/commit/5c7c2d84b1e5eabed514c2911ef1ece183393310), because you couldn't enter "nothing" for spots/streaks in the image.


niksedk:
I like the button :)
The two skip options should probably be `Insert *` and `Insert nothing`... 
A few fixes for dark theme might be needed.

Flitskikker:
> I like the button :) The two skip options should probably be `Insert *` and `Insert nothing`... A few fixes for dark theme might be needed.
@niksedk 
Thank you! Credits to MS. ;)

If the second option is selected, it will insert the current text that's in the TextBox.
So for example, if a letter with OCR artefacts around it is selected, you can enter the letter in the box and press "Skip".
Then it will still insert the letter, but the "dirty" image will not be saved to the OCR database.

I will have a look at the dark theme.

Flitskikker:
Replaced by #6268

PR #6266, opened 2022-09-19
Update pt-PT.xml

hugok79:
For commits:
 
 * Testing naudio record ([480152a](https://github.com/SubtitleEdit/subtitleedit/commit/480152a2dfe1dd0f482406499425e1554b4f722d))
 * Add missing "renumber" after mp4 vobsub import ([96baa0b](https://github.com/SubtitleEdit/subtitleedit/commit/96baa0b5bb4ab41ce07f1c6f186aebbb505b0222))


hugok79:
@niksedk
The **LanguageBaseEnglish.xml** file is in **Linux (LF)** and the **pt-PT.xml** was in **Windows (CR LF)** so I saved it as **Linux (LF)** also to be the same file type as the source file.
I don't know if this makes much difference because I'm not very knowledgeable in this matter. I use Linux and Windows but generally I just use SubtitleEdit in Windows.
Normally I use **Notepad++** and **WinMerge** (wich I also translate to Portuguese language) and I always end up editing the **pt-PT.xml** file with these two programs.
Whenever I opened **Winmerge** to compare the two files, it always warned me that the source file and the translation were in a different file type....

niksedk:
It should not matter but as SE is developed on Windows, I try to keep files with CR LF... to me it looks like `LanguageBaseEnglish.xml` is saved with CR LF...

hugok79:
I openned both `LanguageBaseEnglish.xml` and `LanguageMaster.xml` with Notepad++ to compare and the first one is in Unix LF and the second in Windows CR LF ....
Nevertheless in the next patch I will save it in Windows CR LF

PR #6268, opened 2022-09-20
"Use once" button for OCR

Flitskikker:
@niksedk 

Hello, I made a new proof of concept. Is this more what you mean? :)

image

Only "Edit last" during Binary compare doesn't work for used-once images, but that's not that of a big deal for me.
Maybe in the future, we can reset the index of the loop to the last added used-once image.
I tried to do that now, but there are many flags and counters going on.


niksedk:
Yes, looks good to me :)

PR #6277, opened 2022-09-22
Fix crash in setting AdavncedTextBox text

OmrSi:


OmrSi:
It's a mistake from when I removed the Arabic comma logic, it used to crash when activating `Syntax coloring`.

PR #6403, opened 2022-11-10
Try to fix bookmark showing on first spell check line

OmrSi:
I'm not sure this is the proper fix but I couldn't another way to do this.

It used to be that when the is a wrong word in the first line, SE shows a bookmark for it even if it isn't bookmarked:

image


niksedk:
Looks good to me :)

PR #6464, opened 2022-11-26
Add hidden option for MPV font size scaling

OmrSi:
Okay, hear me out on this. :)

Say I'm using font size 24 for my MPV preview, which looks like this:
image

Just a proper font size.
And then I load a second sub:
image

The difference in font size between MPV and normal preview just doesn't work, I've been struggling with this for a while now, so why not at least have a custom hidden option for MPV font size factor?
That way I can set my font size to something like 16 and the factor to 1.5 which give me 24 for MPV preview:

I tried doing this for MPV using something in the `MpvExtraOptions` setting but sadly couldn't find any option that would only change the default font size for the preview.

### DIFF
--- a/src/libse/Common/Settings.cs
+++ b/src/libse/Common/Settings.cs
@@ -1374,6 +1374,7 @@ $HorzAlign          =   Center
         public bool MpvPreviewTextOpaqueBox { get; set; }
         public string MpvPreviewTextAlignment { get; set; }
         public int MpvPreviewTextMarginVertical { get; set; }
+        public decimal MpvPreviewTextFontSizeFactor { get; set; }
         public string MpcHcLocation { get; set; }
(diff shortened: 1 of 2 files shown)


OmrSi:
Or will this cause some confusion when the font size in SE is different from the actual font size outside SE?

I don't know, but something needs to be done about this. ^^"
It's really annoying for both to not have the same size, and editing the font size each time from the settings is troublesome, any ideas?

OmrSi:
Doing something for the normal preview is probably better than messing with MPV's.

OmrSi:
@niksedk I reached a good solution for me: https://github.com/OmrSi/subtitleedit/commit/96af946c5dc99ba0c592e97f1bdfa75474166ef9

In case you're interested, let me know so I can open a pull request.

PR #6532, opened 2022-12-16
optimize merge line with same text

ivandrofly:
We want to out the nested for-loop as soon as ((p+1).StartTime - (p).EndTime)) > `max_allowed_gaps`
otherwise we will run useless `n-1` operation as it's certain that non of the paragraph following the (p+1) will have lesser gaps


ivandrofly:
Another suggest if to consolidate the two changed classes as they are basically doing the same thing!

PR #6557, opened 2022-12-27
Don't round up for NQC timing

OmrSi:
Well, according to Netflix: `We will not be measuring for a one-frame discrepancy.`
So I guess when the frame rate is 25, a gap of 12 frames is acceptable, meaning there is no need to round it up to 13.
I'm still convinced that it should be rounded up because it makes more sense, this is Netflix quality check so my opinion doesn't matter. :3


niksedk:
OK, no need. But it is still okay round I think... ?

OmrSi:
Yes, rounding is okay but no need to round upwards.

niksedk:
@OmrSi: I'm closing this as rounding up should be okay (not necessary though).
Let me know if I'm wrong :)

PR #6663, opened 2023-02-10
Update pt-PT.xml

hugok79:
* For commit: 8871c7e9f3d5afd3304938c925371272fb247d26
* Other small correction

### DIFF
--- a/src/ui/Languages/pt-PT.xml
+++ b/src/ui/Languages/pt-PT.xml
@@ -139,7 +139,7 @@ Quer transferir e usar o FFmpeg</FfmpegNotFound>
     <ExtractingSeconds>A extrair o áudio: {0:0.0} segundos</ExtractingSeconds>
     <ExtractingMinutes>A extrair o áudio: {0}:{1:00} minutos</ExtractingMinutes>
     <WaveFileNotFound>Não foi possível encontrar o ficheiro de ondas extraído!
-É necessário o VLC media player 1.1.x ({0}-bit) ou superior para executar esta operação.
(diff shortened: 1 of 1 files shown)


src/ui/Languages/pt-PT.xml
```diff
@@ -139,7 +139,7 @@ Quer transferir e usar o FFmpeg</FfmpegNotFound>
     <ExtractingSeconds>A extrair o áudio: {0:0.0} segundos</ExtractingSeconds>
     <ExtractingMinutes>A extrair o áudio: {0}:{1:00} minutos</ExtractingMinutes>
     <WaveFileNotFound>Não foi possível encontrar o ficheiro de ondas extraído!
-É necessário o VLC media player 1.1.x ({0}-bit) ou superior para executar esta operação.
```
niksedk, line 142: Are you it's not use "bit" ? Normally (in English) I think you say "VLC 64-bit".

hugok79: Portuguese is a more complete and complex language than English.
32 and 64 are plural numbers
Here in Portugal we say 32 bits ir 64 bits
For example: https://support.microsoft.com/pt-pt/windows/windows-de-32-bits-e-64-bits-perguntas-mais-frequentes-c6ca9541-8dce-4d48-0415-94a3faa2e13d

hugok79: @ivandrofly Do you agree? Estás de acordo?(pt)

ivandrofly: > @ivandrofly Do you agree? Estás de acordo?(pt)

Yup

PR #6965, opened 2023-05-25
Fix "Remove all line breaks in short text (all except dialogs)" to work with RTL characters

MichaelYochpaz:
The functionality of the "Remove all line breaks in short text (all except dialogs)" common error fix checks for dialogs (to not include) by looking at the first character on the second line, and checking whether it's a hyphen (`-`).

The issue is, that in RTL languages (in my case, Hebrew), the first character is in many times an invisible Unicode control character indicating that it's an RTL string, causing in the dialog check to fail and return dialog strings as well.

In this PR, I've updated the sanitization of the string (which up to now only removed HTML tags) that was used in conditions to also remove Unicode control characters.

### Comparison screenshots:
**Before (current version):**

25 05 2023-SubtitleEdit-417

**After (PR version):**
25 05 2023-SubtitleEdit-418

Here's the file (converted to `txt` since apparently GitHub doesn't allow uploading `srt` formats) I used for the testing, in case you want to test it yourself:
[sample.txt](https://github.com/SubtitleEdit/subtitleedit/files/11568969/sample.txt)

### DIFF
--- a/src/libse/Common/Utilities.cs
+++ b/src/libse/Common/Utilities.cs
@@ -480,13 +480,13 @@ namespace Nikse.SubtitleEdit.Core.Common
             // do not auto break dialogs or music symbol
             if (text.Contains(Environment.NewLine) && (text.IndexOf('-') >= 0 || text.IndexOf('♪') >= 0))
             {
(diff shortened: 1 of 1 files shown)


niksedk:
Thx for the PR :)

PR #6989, opened 2023-06-05
Add better visual markings during syncing of 2 subtitles.

Vasolik:
I added visual markings for syncing 2 subtitles. When 1 line in subtitle is selected, other subtitle shows lines closed to this line grayed based of how distant those lines are. Also it makes those close lines visible on the screen.

View list is renamed for consistancy. And Color utils methods are added for blanding colors.

Btw I have made myself translitaration translator for converting between sr-latn to sr-cyrl and vice verse. But I did not prapare that PR as I how to localize the name of translator and also if it is ok to rename sr-cyrl to sr language. Or at least if I can make that naming of translated file ends with .sr-cyrl instead of just .sr

### DIFF
--- a/src/libse/Common/ColorUtils.cs
+++ b/src/libse/Common/ColorUtils.cs
@@ -1,0 +1,29 @@
+﻿using System;
+using System.Drawing;
+
+namespace Nikse.SubtitleEdit.Core.Common
+{
+    public class ColorUtils
+    {
+        public static Color Blend(Color baseColor, Color targetColor, double percentage = 0.5)
+        {
+            percentage = Math.Abs(percentage);
+            
+            if (percentage == 0)
+                return baseColor;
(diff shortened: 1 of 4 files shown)


niksedk:
Hi @Vasolik,

thx for the PR :)

Could you make it work with "Dark Theme" too?
image

(It's a bit bad practice to do rename and functionality in the same PR as it makes it harder to review changes)

niksedk:
He, the dark theme can be enabled here:

image

Vasolik:
I made something that somewhat fix the problem with dark theme. One problem remain when you hover on marked text, it bacome black on black. And I am not sure what would be a good choice of color in that case. I am better in programming then visual design :)

Vasolik:
image


I thought that this looked ok on dark theme.

I update my branch from upstream as well and applied those changes.

PR #6997, opened 2023-06-06
Remove duration cap when ripping from VobSub or TransportStream

Flitskikker:
Every time I ripped a DVD, I noticed the durations were capped at 7 seconds.
However, Dutch subtitles often exceed 7 seconds, so I had to manually check all subtitles that were 7,000 seconds to extend them if needed.
I figured it was a limitation of VobSub or something.

Recently, when fiddling with my settings in SE, I noticed that the duration cap changed.
I checked the source code and found that SE hard caps the durations at the configured maximum display time, both for DVDs and TS files.

I think it's better to just rip the subtitles with their original timings, after which users can use "Adjust duration limits" if desired.
So, this PR removes these checks for VobSub and TransportStreams.

----

The maximum display time check is also done in several subtitle formats.
Perhaps, the same change should be made there (allow the timings to exceed the maximum duration to keep the original timings -- only recalculate when the duration is negative), but I'm not familiar with these formats.

image

### DIFF
--- a/src/libse/ContainerFormats/TransportStream/TransportStreamParser.cs
+++ b/src/libse/ContainerFormats/TransportStream/TransportStreamParser.cs
@@ -260,7 +260,7 @@ namespace Nikse.SubtitleEdit.Core.ContainerFormats.TransportStream
                         {
                             sub.EndMilliseconds = list[i + 1].PresentationTimestampToMilliseconds() - 25;
                         }
-                        if (sub.EndMilliseconds < sub.StartMilliseconds || sub.EndMilliseconds - sub.StartMilliseconds > (ulong)Configuration.Settings.General.SubtitleMaximumDisplayMilliseconds)
+                        if (sub.EndMilliseconds < sub.StartMilliseconds)
                         {
                             sub.EndMilliseconds = sub.StartMilliseconds + (ulong)Configuration.Settings.General.SubtitleMaximumDisplayMilliseconds;
                         }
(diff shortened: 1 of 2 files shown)


niksedk:
It's kinda on purpose, so that the subtitles follow the current settings - also, some formats do not have an end time.

Flitskikker:
> It's kinda on purpose, so that the subtitles follow the current settings - also, some formats do not have an end time.

@niksedk 
Alright, for the other formats, it's okay to leave them as-is.
But for VobSub and TransportStreams, I still think removing the cap would be better. 

Some thoughts:
- The long durations are part of the original file (and thus the intent of the subtitler). If I open a WebVTT with durations above my current set maximum, they would not get capped either. After opening/importing, users can see which subtitles are too long with their current settings and make corrections if needed.
- While too long durations are capped, too short durations are untouched.
- A common practical example: people on subtitle sites rip DVD subtitles regularly. Some users have a maximum duration of only 6 seconds set in SE for their own translations. Professional Dutch subtitles can go up to 8 seconds. All the ripped subtitles they upload have their durations capped at 6 seconds, which they are probably not aware of. When you watch the movie with their rips, subtitles might disappear too early (before the actors have finished speaking), because the durations on the original DVD were 7,x seconds.
- An option for VobSub could be added to `DvdSubRip.cs` for example, but I would advocate to set it to "Don't cap durations" by default, for the reasons above.
- I'm not sure where to add the option for TS files, because there is not really a UI (except when there is more than one subtitle in the file).

What do you think?

PR #6999, opened 2023-06-06
Adds ability to set layer for the current subtitle.

JonSchram:
When manually positioning a lot of subtitles, which subtitle is rendered on top can often be incorrect, so a layer input would be useful.

The layer property in Paragraph is a 32-bit integer so the numeric up/down control accepts all 32-bit integers to be able to display any possible value (If the value were instead restricted to anything smaller, such as -100 to 100, I'm worried it would modify a valid layer value).


JonSchram:
After looking at this a little more, I think this should register an entry in the undo list. I see some other code for adding undo entries so if this otherwise looks good, I can add the ability to undo layer changes.

niksedk:
Thx for the PR - could it merge "main" in the PR? (it's currently has conflicting changes).
Also, the "Layer" should only be visible when using format ASSA.

JonSchram:
Thanks for the comments! I modified it to make sure it only shows the control for ASSA files. I also saw that you added a layer dialog so I generated the history entry with the language structure similar to how it was done for the dialog.

PR #7105, opened 2023-07-16
Add "Beautify time codes" feature

Flitskikker:
Okay, I think it's time for the big one. Development has been on and off for the past few months, even going as far back as two years ago, but I think it's finally ready for a first public release. Also seeing that I will be going on holiday next month; the merge conflicts when syncing with main are already a constant battle. 😅

---

This adds the feature that enables users to snap all time codes to shot changes, fully configurable using extensive rules.
Besides that, several improvements and new features that leverages this functionality.

I have been testing it using a personal beta version for the last two months, and I'm very happy with the results. Finally, no need to use SubTimingsBeautifier separately anymore. It's even a lot faster, because it was written from the ground up.

New and updated features:
- "Beautify time codes" added to Tools and as icon to the toolbar
- Of course, the "Beautify time codes" feature itself
  - Align timecodes to frame timecodes
  - Snap timecodes to shot changes
  - Very configurable, including visual examples
  - Three presets as a basis for the most widely used styles
  - "Simple mode" for easy configuring
  - The ability to extract exact frame timecodes using ffprobe
- Snapping in the waveform now uses your rules profile
  - Snapping distance is now dynamic, based on your profile and the zoom level
- "Adjust durations" now has an option to don't extend past shot changes
  - I also added a toggle for enforcing minimum and maximum durations (it always did this, which you might not want)
- "Apply duration limits" now also has an option to don't extend past shot changes
- The shortcuts to extend to previous/next shot change now use your rules profile
   - The separate shortcuts (extend to shot change & extend to shot change minus gap) have been replaced with single dynamic shortcuts that follow your rules profile
- New shortcuts to extend to previous/next subtitle with chaining gap
   - This will extend to the previous/next subtitle minus the chaining gap, e.g. 500 ms for NF
- Some small control alignment fixes on existing forms

### Screenshots

image            image

_You can go to Tools > Beautify time codes... or use the icon on the toolbar (old or new style)_

---
image
_Then you can choose to extract the exact timings using ffprobe (useful for 23.976 FPS content due to rounding -- not really needed for 25 FPS), and you can toggle shot change snapping._

---
image
_The rules profile is extensively configurable. Netflix' rules (not my personal preference) are shown in the example. There are also presets for e.g. SDI and the "default" preset from good ol' SubTimingsBeautifier._

---
image
_A "simple mode" helper wizard thingy if you find the previous dialog to be too overwhelming._

---
image
_The wave form now also uses your rules profile._

---
image

image
_"Adjust durations" and "Apply duration limits" are now also shot change-aware._

---

### Notes

This is a first MVP-like version. It is already very usable, but of course, additional improvements and features can be added.
Things that could be added in future revisions:
- Beautify only selected lines
- Add Beautify time codes to the batch converter
- Make "Adjust durations" and "Apply duration limits" in the batch converter also shot change-aware
- Generating time codes in bulk in "Batch generate waveforms"

I wrote some tests for the public utility functions, but I'm still struggling with tests for the internal functions and logic itself. Everything is very "visual" and has many variables, such multiple subtitles, shot changes and rules that all affect eachother.
Nevertheless, my real-world testing in the last month went smoothly, and I have added extensive comments in the main logic class.

I really hope you will like this new feature. Please let me know what you think.


niksedk:
Impressive work Martijn :)
Will look at this tonight.
It has been on my TODO list for several years. Also, much requested (but not for all).

So ffprobe is better with drop frame? Sounds like a new in-program download...

>the merge conflicts when syncing with main are already a constant battle

He, already there.... sorry.

niksedk:
Considering bumping SE version number to 4.0 - new gfx and now this :)

niksedk:
If "FFmpegLocation" is empty it crashes:
image

Also, I think just using the FFmpegLocation is fine.

niksedk:
In master now :)

I'll add the ffprobe auto dl...

Flitskikker:
> Impressive work Martijn :)
> Will look at this tonight.
> It has been on my TODO list for several years. Also, much requested (but not for all).

Thank you! 😄

> So ffprobe is better with drop frame? Sounds like a new in-program download...

Not exactly sure about the drop frame to be honest, but I have had some occasions in 23,976 fps files where the actual extracted timecode would differ from a calculated one.
Stepping through the video in SE + mpv with the "One frame forward/back" shortcuts will show the correct timecodes in the video player though. Maybe mpv can be leveraged in some way?

I have been using ffprobe instead of ffmpeg to extract the timecodes because it was much faster in my testing.
Of course, it could be because of my specific command line. If there is a faster way with ffmpeg itself, then users won't have to download ffprobe separately as well. Maybe it could even be combined with the ffmpeg command line to detect shot changes.

Here are some command lines:

`ffmpeg -i "video.mkv" -vf "select='gte(scene,0)', showinfo" -threads 0 -an -f null -`

`ffprobe -select_streams v -show_frames -show_entries frame=pkt_dts_time -of csv -threads 0 "video.mkv"`

If you can find this file, you'll have a good example of extracted times vs. calculated times:
<details> 
  <summary>Click to show base64'd name</summary>
  <p>Qmx1ZS5MaWdodHMuUzAxRTA0LkhEVFYueDI2NC1YRU4wTg==</p>
</details>

This one is 25 FPS. Instead of ,000 / ,040 / 0,80, it goes ,005 / ,045 / ,085.

>> the merge conflicts when syncing with main are already a constant battle
>
> He, already there.... sorry.

Not sure how my original message came across... How I meant it:
It happened a lot that I wasn't able to work on the project for some weeks or even months. Of course, SE was actively being worked on in the meantime, so whenever I picked up the project and synced with master to continue working on it, and due to my updates in frequently changed files such as Main.cs and Settings.cs (and due to VS reordering controls in .Designer.cs files for some reason), there would be merge conflicts almost every time. I'm happy that it is finally ready now!
Apologies for the confusion.


> If "FFmpegLocation" is empty it crashes:
> Also, I think just using the FFmpegLocation is fine.

Thank you for catching this, and for fixing it.
I also fixed some other bugs (every time _after_ publishing...), PR coming up.


> In master now :)
> 
> I'll add the ffprobe auto dl...

Thank you so much!

TarantinoMariachi:
Using 4.0.8 and found some problems, I had a subtitle pretty ready, only thing needed to fix was the 2 gap seconds between subs, I could apply the gap feature only but that could add problems with the change shots rules. So I dediced to give a go with this feature. But after applied it causes an issue, some subtitles on scene change appears 1 frame off. It starts 1 frame before the scene change when it should be on the exact new frame after the scene change. Sometimes it starts one frame after the scene change.

I have experience with aegisub, and when I create with xvid the .pass file with the keyframes I am able to fix this issue with that software, but would be great if thisf eature worked properly. Dunno if the problem is on the scene change detector. Would be great if it accepted the xvid .pass file. That way I could at least test and check where the problem comes from.

Flitskikker:
@TarantinoMariachi 
Could you please share:
- The subtitle file you try to fix
- A .txt file with the exported shot changes (Video > Remove/export shot changes > Export)
- If possible, (a link to) the video file
- A screenshot of your Beautify time codes settings/profile

Then I can have a look and debug.

Also, if you could share an example of a .pass file, I could look into adding support for it.

TarantinoMariachi:
I cannot provide those in specific, but I'll pick another subs and video, to recreate the same test scenario.

TarantinoMariachi:
Here are the 3 most common keyframe generated files out there: https://www.swisstransfer.com/d/171fa3ef-9084-4b97-9b23-673ea3a5238c

One of them using VirtualDub2 with XviD, doing 1 pass.

Another one using scxvid.exe with this (XviD2pass):
`ffmpeg -i %video% -f yuv4mpegpipe -vf scale=640:360 -pix_fmt yuv420p -vsync drop -loglevel quiet - | SCXvid.exe %video2%_keyframes.txt`

(both can be found here: https://doki.co/support/doki-timing-guide/)

And the last one using vapoursynth-wwxd: https://raw.githubusercontent.com/LightArrowsEXE/fansub-utils/master/generate_keyframes.py

TarantinoMariachi:
The video for testing can be found here: https://www.swisstransfer.com/d/24d01dae-7510-4aa4-b4be-b74c90951b3b
And the test subtitle: https://www.swisstransfer.com/d/fcff4473-a800-4f8a-8052-4a5a17ca79ea

I open Beautify, and extracted the timecodes:
image

I honestly miss a button there so we can remove that specific timecode we created. Because if we want to generate again for some reason, we need to go to timecodes folder to delete, but the names are all with numbers so we dunno what is what. Having a way at least to delete or regenerate would be great. And would be great if we could export (save as) the timecode to another folder, with  name given by us. That way I could even use for other softwares if I wanted. The current hack is to have the timecodes folder clean and run it and grab it from there.

Then I generated the shot changes:
image

And I loaded Netflix preset. But saw something awkard. If I click Simple Mode, then Netflix Rules and click ok, this value goes to 120 ms:
image

With netflix preset it goes to 180. Shouldn't be the same?!

Another issue I saw was that 2 frames are counted as 83ms, and I understand why, after all, 2 / 23.976 * 1000 = 83.416~
But those 83 doesnt cope with the netflix profile rule that is at 84:
image

In this case probably the profile shoud be changed to 83? (At least it was what I did on my side)


Now back to the shot changes... when I hit ok to do the tpp (time post-processed), it do all the changes, with my values:
image

But ofcourse it will only respect the shot changes it detects, if it detects the wrong frame, it will be off.
And from this test scenario we can see at least line 46 that the line doesnt start on the scene change but 2 frames after.
And I did an extra test, I changed one of my keyframe files to be compatible with this feature import, and using xvid2pass keyframe it did a good work (at least on line 46). It seems to do a proper work at least. better then the own file it generates.

Now I decided to test again my original files to see if I use a good keyframe file, if it do a good job. But it doesnt, Aegisub does a proper job with their TPP. I beleive it can be related with the problem between ms and frames. Because 1 frame can be like 4ms, and i would say that aegisub seems to always choose the 3st ms of the frame, and this feature here doesnt do the same, and therefor it can be the reason why it shows 1 frame off. Because the time SE put was: 0:01:29.84 (1 frame off on the players, as MPC) to 0:01:32.01 and when I do the fix on Aegisub it goes to 0:01:29.82 to 0:01:32.01, so it adjust by 2 ms, half a frame. The 1st frame after the scene change is at 80 and for aegisub when I choose that frame and set the time it put 82. On SE player it seems that even if the time is set as 84 it still shows when the player is at 80ms, but with most of the players that doesnt happen (MPC, MPV).

Hope this information can help.

Flitskikker:
@TarantinoMariachi 
Thank you for the files and your extensive reply.

> Here are the 3 most common keyframe generated files out there

Thanks for the files. I will work on import support.

> I honestly miss a button there so we can remove that specific timecode we created. (...)

That is a good point. I will add this to the list. In the meantime, I suppose you could also sort the files in the TimeCodes folder by creation date. That way, you don't have to clear it first. For shot changes, you can regenerate/remove them through the Video menu.

> With netflix preset it goes to 180. Shouldn't be the same?!

Yes, you are right. The Simple Mode dialog calculates the value based on the gap ms * 1.5, but for the NF profile, it still has 180 ms hardcoded. I will update the profile accordingly.

> In this case probably the profile shoud be changed to 83?

These should be at 83 already, actually. They were updated a long time ago, because of the reasons you mention.
https://github.com/SubtitleEdit/subtitleedit/blob/main/src/libse/Settings/GeneralSettings.cs#L464

I believe SE doesn't update existing profiles, so if you have been using the current profiles for a long time (years), then they would still be at 84. I'm afraid it only recreates them if you clear them all, unfortunately.

> But ofcourse it will only respect the shot changes it detects, if it detects the wrong frame, it will be off.

This is true, this feature relies on the shot change data being correct.

---

> And from this test scenario we can see at least line 46 that the line doesnt start on the scene change but 2 frames after.

Now it gets interesting. I loaded the subtitle file and video (House Rules Ep.01), and generated shot changes (ffmpeg) and time codes for it.
Line 46 does start exactly on the shot change for me; when I step through the frames and navigate to the white line, the shot changes exactly on the white line, which is on 00:02:41,360 (00:02:41:09), the same as the in-cue for line 46.

image

The BITC is not in sync, though. It appears the video actually starts at 00:00:00,080 instead of 00:00:00,000, so there is a 2-frame offset. I can visually correct this using Video > Set video offset.

image

(or -00:00:00:02 in frames mode)

Now everything lines up:

image

And now I can run Beautify time codes to snap everything into place.

---

When I import your keyframes.txt, however, I do indeed see that the shot changes are shifted by 2 frames.
The same is the case when I generate shot changes with the external tool of my choice (EZTitles).

So, it seems that somewhere, something (in the video? libmpv? a codec? SE?) causing a 2 frame / 80 ms offset.

Inspecting the generated timecodes file, the very first line reads "0.08", which corroborates what I noticed when stepping through the frames in SE backwards: it doesn't go beyond 00:00:00,080.

The same is seen when I open the video in TMPGEnc MPEG Smart Renderer 6, which has frame-level editing.
The first image with BITC 00:00:00:00 is shown 3 times, then it goes to 00:00:00:01. So a 2-frame offset for some reason.

I think the keyframe and pass files assume the video always starts at 00:00:00:00, but for some reason, that is not the case with this video. Because the video is shifted 2 frames, the shot changes are too.
So, the keyframes.txt says 38 is a shot change, but with the offset, it should be 40.

SE has an option to account for video offset that will shift the waveform, but not the shot changes.
So, I think a solution would be the ability to shift the imported shot changes?
That has been on the list for a while, actually, because I have had the need for it a couple of times already.

I'm not sure what Aegisub does differently here. But it appears that SE (and ffmpeg, and ffprobe) are merely reading the video file like it is, as having a 2-frame offset if you compare it to the BITC (which is admittedly highly ignoring, being a professional subtitler myself).

Perhaps @niksedk could chime in and help analyzing the House Rules Ep.01.mp4 video file you provided.
All signs say the offset is in the video itself, but perhaps it is something with libmpv, a codec, or SE after all.

--

> Now I decided to test again my original files to see if I use a good keyframe file, if it do a good job. But it doesnt, it seems to be better but still not perfect. 

I think it's the same issue as above; some offset in the video that is causing everything to be shifted.

> Because 1 frame can be like 4ms (...) so it adjust by 2 ms, half a frame

What is the framerate or your original file? If 1 frame is 4 ms, the framerate would be 250 fps. 🤯

niksedk:
I think this video is "fishy"... if you place the video cursor at 30 seconds (both mpv and directhsow), the video says: "TIME: 00:00:29:04"... that can not be correct.

If I generate a 25 fps video via ffmpeg, it looks like this at 30 seconds:
image

TarantinoMariachi:
> These should be at 83 already, actually. They were updated a long time ago, because of the reasons you mention.
> https://github.com/SubtitleEdit/subtitleedit/blob/main/src/libse/Settings/GeneralSettings.cs#L464
> 
> I believe SE doesn't update existing profiles, so if you have been using the current profiles for a long time (years), then they would still be at 84. I'm afraid it only recreates them if you clear them all, unfortunately.

Ahhh, that explains it. Ok, removed, open SE again and voila! Thanks!



> What is the framerate of your original file?

23.976, I'll see if I can test more files so I can send a proper video, because that one is indeed a mess to not say other word xD



> I think this video is "fishy"... if you place the video cursor at 30 seconds (both mpv and directshow), the video says: "TIME: 00:00:29:24"... that can not be correct.

well, the last 2 numbers are not ms but frames, that video is 25fps, so 29:24 is 1 frame off to be 30:00, I tested with MPC-HC and MPC-BE (external) and it is frame accurate. mpv, that I never was a fan, it is 2 frame off (external), but i am using it on SE too, I took a look at SE, and with mpc and mpv is 2 frame off for me, but not for all frames. LOL
This is correct:
image

This is not:
image

Same for my original video, that is not a messed up video:
image
image
610 is 14 and 611 appear as 15, but 15 on 23.976 video should be at 625 so its off for 14ms
image

This happens with all episodes

If I create keyframes with the 3 mthods I talked about, the vapoursynth doesnt detect all changes, would say 95% of them, and the other two detects 100% of them, and the vdub one detects more false positives, but hey better more then less as vapour solution. And using any of the 2 keyframes that works better on scene detection it works perfectly on aegisub.

So, if I could use those on SE, I could do more tests to see if that is enough to solve all this offset problems.

TarantinoMariachi:
I am able from one of the keyframes to do some cleaning on the txt file to make it work on SE, and after some tests, found out that even having the correct keyframe, and having the 2 gap frame rule, the subtitle before the scene is 2 gap away, correct, but the next subtitle that should appear on the exact 1st frame of the new scene it appears on the 2nd frame. 

My current way to fix this, is going to aegis and aplly their TPP with 42ms on the keyframes so it will snap to the keyframe any subtitle that is 1 frame away. But some cases it seems is 2 frames away (need to do more testing, if its really a thing then I am screw since I cannot automatically fix that, since aegis doesnt have the feature for gaps and apply a 0 gap chaining them together :'(.)

TarantinoMariachi:
This is how it looks when the timing is frame accurate, 2 frames gap and the next sub is on the exact spot.
image

niksedk:
I will try to look more at mpv and see if seeking can be made more precise...

niksedk:
@TarantinoMariachi: How is this? https://github.com/SubtitleEdit/subtitleedit/releases/download/4.0.8/SubtitleEditBeta.zip

TarantinoMariachi:
Thanks! But same Problem :( 1 frame offset

niksedk:
SE now always matches time in video player with time/frame in video. Is that not working for you?

TarantinoMariachi:
> SE should now always match time in video player with time/frame in video. Is that not working for you?

No, it is not. :(

PR #7107, opened 2023-07-16
Translation update

PMitsakis:
What layer refers to in the ASSA menu?


niksedk:
Layer is the "image layer number" for assa subtitles (like Photoshop and Paint.NET has layers).

When doing more stuff at the same time, setting a layer will make sure that other subtitles at the same time does not interfere with each other.
It also determines what goes on-top or behind each other.

PMitsakis:
Thank you for the help, Nikolaj.

PR #7110, opened 2023-07-16
Some "Beautify time codes" fixes

Flitskikker:
Some bugs squashed.
- The selected options on the "Beautify time codes" dialog are now being saved correctly
- The overall progress is now calculated correctly
- Extracting exact timecodes had some bugs with seconds vs. milliseconds mismatches


OmrSi:
Extend and snap shortcuts seem to not work well anymore?

OmrSi:
@Flitskikker `GetNextShotChangeInMs` doesn't seem to be actually return Ms?

Flitskikker:
> @Flitskikker `GetNextShotChangeInMs` doesn't seem to be actually return Ms? Maybe the same for `GetPreviousShotChangeInMs`?

Thanks! #7113

PR #7113, opened 2023-07-17
Another quick "Beautify time codes" bugfix

Flitskikker:
Thanks @OmrSi!

Sorry for the back and forth...

### DIFF
--- a/src/libse/Common/ShotChangeHelper.cs
+++ b/src/libse/Common/ShotChangeHelper.cs
@@ -91,7 +91,7 @@ namespace Nikse.SubtitleEdit.Core.Common
         {
             try
             {
-                return shotChanges.Last(x => SubtitleFormat.MillisecondsToFrames(x * 1000) <= SubtitleFormat.MillisecondsToFrames(currentTime.TotalMilliseconds));
+                return shotChanges.Last(x => SubtitleFormat.MillisecondsToFrames(x * 1000) <= SubtitleFormat.MillisecondsToFrames(currentTime.TotalMilliseconds)) * 1000;
             }
             catch (InvalidOperationException)
             {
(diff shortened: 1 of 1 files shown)


OmrSi:
@Flitskikker Thank you!
One more thing, why not make `Extend selected lines to next subtitle with chaining gap` work with shot changes as well?

Flitskikker:
> One more thing, why not make `Extend selected lines to next subtitle with chaining gap` work with shot changes as well?

Could you explain this some more? It already uses your current "Chaining" rules when the next subtitle's incue (or the previous subtitle's outcue) is on a shot change.

OmrSi:
@Flitskikker If there isn't a line after the shot change and the audio ends 500ms before the shot change, you would want to set the out-cue to the Chaining gap before the shot change.

It's not the Chaining gap per se, but it's 500ms in Netflix's case no matter what the frame rate is, which matches the Chaining gap.

Flitskikker:
@OmrSi 
Ah, I see. "Free" out cues on shot changes are handled differently from chained out cues on shot changes.

image

So I think it would be new separate shortkeys. Something like:
- Enforce green zone of nearest shot change for in cue of selected lines
- Enforce green zone of nearest shot change for out cue of selected lines

And depending on which side of the shot change the cue is, it will either put it 12 frames (NF) before or after it.

OmrSi:
@Flitskikker Yes, this is it. But won't that make for 4 new shortcuts?
For in-cue before or after the shot and the same for out?

Flitskikker:
@OmrSi 
I added the shortcuts in today.

Initially, I wanted to do only 2 shortcuts and move the cue to the left or right depending on on which side of the shotchange the cue is. But if an outcue is "on" a shot change, for NF it is actually 2 frames before the shotchange, so it would always move to the left, to 12 frames before the shotchange. If you wanted to overshoot the shotchange, you would first have to extend it a bit and then press the shortcut. I could imagine that would be tedious, so I split them up anyway.

Although it's indeed 4 shortcuts now, I think it works really well. 😄

image

Very analogical to my other shortcuts:

image

And Alt + < and Alt + > to jump 500 ms back and forth (also in Originator)

PR #7122, opened 2023-07-19
Add "Beautify time codes" to batch converter & much more

Flitskikker:
Some more new goodies.
- "Beautify time codes" added to batch converter and command line
   - It will load the available timecodes and shotchanges (if any) if a video can be found (e.g. movie.srt > movie.mkv)
- "Extract time codes" checkbox added to "Batch generate waveforms"
   - Also added more status messages for each step
- "Respect shot changes" checkbox added to "Adjust durations" in batch converter
   - It will load the available timecodes and shotchanges (if any) if a video can be found (e.g. movie.srt > movie.mkv)
- "Respect shot changes" checkbox added to "Apply duration limits" in batch converter
   - Ditto
   - Also added the same UI as the standalone version to "Apply duration limits" itself
- New "Set in/out cue to minimum distance before/after closest shot change" shortcuts added
   - Thanks @OmrSi for the idea, really useful!
- Fixed a bug where cues close to 00:00:00,000 would be set to -00:00:00,042 because the fallback value (-1) would be the closest shot change...
- Some other small improvements such as null/empty checks

I think that was everything from my list for now. :D

---

image

---

image

---

image

---

image

---

image


src/ui/Forms/Main.cs
```diff
@@ -19089,6 +19109,161 @@ private void ExtendSelectedLinesToPreviousShotChange(bool withGap = false)
             }
         }
 
+        private void SetCueToClosestShotChangeGreenZone(bool isInCue, bool isLeft)
```
ivandrofly, line 19112: Wouldn't it be better to have this method somewhere else as the `Main.cs` is already very big with +30k lines which makes the IDE struggle  /cc @niksedk  @Flitskikker

Flitskikker: @ivandrofly 
Yeah, I noticed the IDE struggling. Although I don't think moving just these ~150 lines will make much difference. I would suggest instead to move all shortcut-related functions to another file.

Issue #7000, opened 2023-06-07
Not all subtitles are shown in waveform view when there are too many.


JonSchram:
If I have a lot of subtitles in a file, the waveform view stops displaying them partway through.

I was looking [AudioVisualizer.cs](https://github.com/SubtitleEdit/subtitleedit/blob/92b04ff61f7151a944213d45be6c6641e19f0acb/src/ui/Controls/AudioVisualizer.cs#L442) and it looks like there's a hard cap of 100 paragraphs at a time / filtered down to 30 later on. This also looks like it was made to address big files better.

For me, the 15 additional seconds of rendered paragraphs can force nothing to be rendered when there are > 100 subtitles outside of the visible area, or causes the rendering to stop halfway through the waveform.

I wonder if it would be worth processing paragraphs a little differently. Maybe some combination of:
- Retain paragraphs in the visible area first when there are > 100
- Prune excessive paragraphs during the initial 100-subtitle load loop so that the 30-paragaph limit is put to better use.
- Update visible paragraphs instead of clearing and re-loading them. (might get complicated if some were pruned from the above optimizations but then they no longer need to be pruned, maybe check if that paragraph number is loaded?)

Do these optimizations sound reasonable? I think this is something I could address but I'm curious to know if there's anything I should watch out for here.

niksedk:
I think the main issue is performance - I've seen 30-100MB ASSA files where things will be extremely slow if number of visible subs are not pruned.

JonSchram:
Thanks for the advice, sorry for the delay. I had a subtitle file that was 3 MB in size, never realized that they could get as large as 100 MB.

Just for testing, I tried removing all pruning and the performance was pretty bad. I tried a couple approaches but I think this one works the best.

The previous algorithm discarded paragraphs that started within 90 ms of the previous subtitle, but this doesn't work well with LINQ. Instead, I group them into 'buckets' 90ms wide. Then, I have it select paragraphs from buckets evenly so that it can render multiple subtitles with the same start time in case they fit under the max displayable paragraph limit. This helps when the paragraphs don't also have the same end time. Subtitles outside the visible are are pruned with a slightly simpler algorithm. A 100-paragraph limit is applied at the very end.

This new algorithm took around 20ms max on my computer, and it tended to get faster the more I scrolled (runtime optimization maybe?).

PR #7139, opened 2023-07-22
Changes pruning algorithm for audio visualizer.

JonSchram:
- Selects paragraphs in the visible portion of the visualizer first to ensure you can see the current timeline.
- Select paragraphs across the entire timeline before selecting multiple paragraphs that start at the same time.
- After visible paragraphs are pruned, add some invisible ones to assist with scrolling (and prune again to limit rendered paragraphs).

After these changes, the timeline renders paragraphs even when there are many just outside the visible range.

#7000


src/ui/Controls/AudioVisualizer.cs
```diff
@@ -431,40 +431,52 @@ private void LoadParagraphs(Subtitle subtitle, int primarySelectedIndex, ListVie
...
+
+            const int maxDisplayableParagraphs = 100;
+            IEnumerable<Paragraph> displayableParagraphs = visibleParagraphs;
+            if (visibleParagraphs.Count > maxDisplayableParagraphs)
+            {
+                /*
+                 * Group & select is done so that it draws paragraphs across the entire timeline before drawing overlapping paragraphs, up to the display limit.
+                 * Materialize to list so that it can be counted below without pruning twice.
+                 */
+                displayableParagraphs = visibleParagraphs.Where(p => p.DurationTotalMilliseconds >= 0.01)
```
ivandrofly, line 453: Too many LINQ. This can impact the performance!

JonSchram: Thanks for taking a look! I didn't think the performance impact was too large with LINQ but I don't mind rewriting it to use traditional loops.

After creating this PR I noticed there's a SecondsToXPosition method that looks better than grouping by the same number of milliseconds no matter the zoom level. So I'll get both of these changes in this PR.

ivandrofly: @niksedk is this something of your interest?

JonSchram:
Hi, sorry for such a delay here! I had a lot of things to do and couldn't devote any time to this.

Recently I've been able to work on this again. I started out with a simple rewrite of my first attempt but wasn't happy with the results. Since then, I've gone through a few revisions but this is an algorithm I'm really happy with.

I've done a lot of profiling and optimizing to make sure it is fast even with a huge number of paragraphs, and I added some test cases to ensure it is choosing the 'best' paragraphs for display.

PR #7144, opened 2023-07-23
Replace magic numbers with named constants

ivandrofly:
This commit replaces magic numbers in the retry mechanism with named constants. In "UpdateLanguageFiles", the hardcoded delay of 10ms between retries has been replaced by the variable 'delayBetweenRetries'. Similarly, in "UpdateAssemblyInfo", the number of retries set to 10 has been replaced with 'maxRetries' and 'delayBetweenRetries' is used for the time gap between each attempt. This makes the code more readable and maintainable.

### DIFF
--- a/src/UpdateLanguageFiles/Program.cs
+++ b/src/UpdateLanguageFiles/Program.cs
@@ -109,7 +109,7 @@ namespace UpdateLanguageFiles
                         throw;
                     }
 
-                    System.Threading.Thread.Sleep(10);
+                    System.Threading.Thread.Sleep(delayBetweenRetries);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/UpdateLanguageFiles/Program.cs
```diff
@@ -109,7 +109,7 @@ private static void SaveWithRetry(Action saveAction)
                         throw;
                     }
 
-                    System.Threading.Thread.Sleep(10);
+                    System.Threading.Thread.Sleep(delayBetweenRetries);
```
ivandrofly, line 112: Fixes from pre PR where the variable was left unused!

PR #7145, opened 2023-07-23
Add more null and count checks

Flitskikker:
Like in 8752323fb3cb3c0b081d03761550b632de277312

### DIFF
--- a/src/libse/Common/ShotChangeHelper.cs
+++ b/src/libse/Common/ShotChangeHelper.cs
@@ -11,7 +11,6 @@ namespace Nikse.SubtitleEdit.Core.Common
 {
     public static class ShotChangeHelper
     {
-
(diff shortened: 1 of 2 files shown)


src/libse/Common/ShotChangeHelper.cs
```diff
@@ -89,6 +88,11 @@ public static void DeleteShotChanges(string videoFileName)
 
         public static double? GetPreviousShotChange(List<double> shotChanges, TimeCode currentTime)
         {
+            if (shotChanges == null || shotChanges.Count == 0)
```
ivandrofly, line 91: To avoid duplication `shotChanges == null || shotChanges.Count == 0` could've been moved to a method 
called `Isvalid` or something similar :)

Issue #7069, opened 2023-07-06
Show current time of playhead in the selected waveform segment


rRobis:
It would be very useful to see at what time the playhead is inside the current subtitle waveform timeline. 

This would help to time ASS effects more efficiently (currently i have to experiment with timing based on the total time of the subtitle segment).

So for example if i have a segment that is 2 seconds long, i want to see at what time inside that segment my playhead is at.

![image](https://github.com/SubtitleEdit/subtitleedit/assets/19517183/eb2a0b0c-b45a-4606-b339-79eb988b8585)

niksedk:
It's in the video player - lower right corner.

rRobis:
@niksedk Hi!
Was my last comment clear enough to understand, what i mean?
 
If it's required, i could make some visual examples, of what i mean, if it's required. I would really appreciate, if this could be implemented.

rRobis:
The calculation for relative current time inside selected segment would be:

Current playhead absolute time (from the beginning of the video) - Currently selected segment start time (in ms) = relative time inside selected segment

PR #7149, opened 2023-07-23
#7069 Improve AudioVisualizer - show relativeTime inside SelectedParagraph

rRobis:
Working solution to #7069 issue. Adding relative Playhead time to Selected Paragraph

(If necessary have to refactor, because this is the first time I code in C#)

**Potential lacks:**
1. May look buggy if there are overlapping paragraphs.
2. Performance?

What could be improved more:
✅ Move var calculation after checking if it needs to be added or not (and not calculate it on each paragraph. 
❌ Update the relativeTime variable only on pause and mouse click on the Waveform
❌ Move the relativeTimeInMilliseconds variable into a function to return the value and use that instead. (Still not sure where to do it)
❌ Make it an optional setting

### DIFF
--- a/src/ui/Controls/AudioVisualizer.cs
+++ b/src/ui/Controls/AudioVisualizer.cs
@@ -1067,7 +1067,19 @@ namespace Nikse.SubtitleEdit.Controls
                 // paragraph number
                 if (n > 15)
                 {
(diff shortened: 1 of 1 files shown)


rRobis (commented):
Improve the previous commit a little bit

rRobis:
image

PR #7165, opened 2023-07-26
Remove git cached files

ivandrofly:
Removes git cached ignored files

For example these files 
src/ui/DLLs/Voikkox64.dll
src/ui/DLLs/Voikkox86.dll
src/ui/Hunspellx64.dll

Are now references as NuGet packages but since they were added before the .gitignore they are still being 'tracked'

https://stackoverflow.com/a/1139797/2766753


ivandrofly:
Appveyor is complaining about `QuartzTypeLib`... checking

src/ui/SubtitleEdit.csproj
```diff
@@ -2288,6 +2288,9 @@
     <None Include="Resources\color-picker-small2.png" />
     <None Include="Resources\rec32.png" />
     <None Include="Resources\Help32.png" />
+    <Content Include="DLLs\Interop.QuartzTypeLib.dll">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
ivandrofly, line 2292: Just one doubts does SE requires this file at the runtime?

Because it was previously in csproj as  `<None Include="DLLs\Interop.QuartzTypeLib.dll">`
I've now added it back as 
```xml
<Content Include="DLLs\Interop.QuartzTypeLib.dll">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
</Content>
```

but if it's not required at the runtime it should be reverted back as it was original

ivandrofly:
AppVeyor might need a manual trigger build I guess

niksedk:
Those dll files are still needed...

PR #7166, opened 2023-07-27
Introduce StringBuilderPool

ivandrofly:
The new StringBuilderPool class has been added to provide an implementation of an object pool pattern specifically for StringBuilders. This object pool assists with the use and re-use of StringBuilders, where more-often used string builders are returned from the pool and the least used ones are dropped. Limitations on pool size and capacity of string builders are also set to manage resource usage.

### DIFF
--- a/src/Test/Common/StringBuilderPoolTest.cs
+++ b/src/Test/Common/StringBuilderPoolTest.cs
@@ -1,0 +1,84 @@
+using System.Text;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Nikse.SubtitleEdit.Core.Common;
+
+namespace Test.Common
+{
+    [TestClass]
+    public class StringBuilderPoolTest
+    {
+        [TestMethod]
+        public void GetNotNullTest()
+        {
+            var sb = StringBuilderPool.Get();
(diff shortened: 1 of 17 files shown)


ivandrofly:
Add test specific for StringBuilderPool

ivandrofly:
For some reason it look like appveyor is complaining about the **internal access modifier** in StringBuilderPool
which it was before (see the commit history above)!

Also [InternalVisibleTo("Test)] is still present in LibSE project.

The image below shows that all the test are passing
image

niksedk:
I did a quick test... and it's a lot slower (probably better for the GC though).
I will have to do a bit more testing...

ivandrofly:
Consider that this is not meant to be the replacement of the StringBuilder in the entire system.
There are specific cases where pool is used, for example loop => caller => callee {create new stringbuilder)!
If the loop is running ~5000 times that is creating 5K instance to GC to deal with it!
Injecting stringbuidler 2/n level above just to use it somewhere isn't that great I think.

Note that due to the thread-safe capability, that might impact the performance a bit but it's future proof!

niksedk:
Definitively interesting... if we get some memory issue then let's look at it again. Until then, I prefer the performance.

PR #1899, opened 2016-08-12
Added new char-helper class

ivandrofly:

PR #7170, opened 2023-07-27
Use CharUtils.IsDigit method for enhanced consistency.

ivandrofly:
Ref #1899

### DIFF
--- a/src/libse/Common/StringExtensions.cs
+++ b/src/libse/Common/StringExtensions.cs
@@ -356,11 +356,10 @@ namespace Nikse.SubtitleEdit.Core.Common
                 return false;
             }
 
-            var max = s.Length;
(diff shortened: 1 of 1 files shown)


ivandrofly:
char.IsNumber is not the same as CharUtils.IsDigit as previously mentioend by @xylographe in https://github.com/SubtitleEdit/subtitleedit/pull/1899#discussion_r74571654

Another thing that if check the caller 

One of them has this statement before calling `  if (language == "en" && text.ContainsNumber())` which means it's only for English culture

and the other is `  if (language != null && "en-da-es-sv-de-nb-cz".Contains(language) && text.ContainsNumber())`

The expected digit range is 0-9

niksedk (approved):

PR #7174, opened 2023-07-28
[NikseComboBox] - Refactor & Optimization

ivandrofly:


src/ui/Controls/NikseComboBox.cs
```diff
@@ -403,8 +403,13 @@ private void HideDropDown()
                 _listViewShown = false;
             }
 
-            FindForm()?.Controls.Remove(_listView);
-            FindForm()?.Invalidate();
```
ivandrofly, line 407: Invoked 2x while we could just cache the found form if any!

src/ui/Controls/NikseComboBox.cs
```diff
@@ -620,6 +565,74 @@ protected override void OnMouseDown(MouseEventArgs e)
...
+            };
+
+            _listView.MouseClick += (sender, mouseArgs) =>
+            {
+                if (mouseArgs == null)
+                {
+                    return;
+                }
+
+                var cachedCount = _listView.Items.Count;
```
ivandrofly, line 617: cache the count as `_listView.Items.Count` is very expensive!

See source code

src/ui/Controls/NikseComboBox.cs
```diff
@@ -575,10 +520,10 @@ protected override void OnMouseDown(MouseEventArgs e)
                         totalY += ctl.Top;
                     }
 
-                    if (_listView.Items.Count > 0)
+                    if (listViewItems.Count > 0)
```
ivandrofly, line 523: Don't use `_listView.Count` see https://github.com/SubtitleEdit/subtitleedit/pull/7174/files#r1278103956

listViewItems is the source that is being used to init _lsitView and it's case to check the count from it
directly!

src/ui/Controls/NikseComboBox.cs
```diff
@@ -484,78 +489,18 @@ protected override void OnMouseDown(MouseEventArgs e)
                     _listViewShown = true;
                     _textBox.Focus();
 
-                    if (_listView == null)
```
ivandrofly, line 487: This is now moved to a method that will handle the initialization for the `listview`

PR #7176, opened 2023-07-29
Refactor GetVideoInfo method for code simplicity

ivandrofly:
The GetVideoInfo method in src/ui/Logic/UiUtil.cs was refactored to use a list of decoder functions instead of individual "try read". This change simplifies the code and makes it more DRY (Don't Repeat Yourself). Future decoder functions can now be easily added to the DecodersLazy list.


niksedk:
Hm, it is both simpler and more complex... I think the old code in somewhat easier to read.

PR #7182, opened 2023-07-29
"Refactor pixel width calculation in FixShortLinesPixelWidth"

ivandrofly:
The pixel width calculation for FixShortLinesPixelWidth has been directly incorporated into the constructor for FixShortLinesPixelWidth, replacing the previously used GetCustomCallbackData method. This change simplifies the process significantly by making the pixel width calculation a direct responsibility of the FixShortLinesPixelWidth class, and removing the associated logic from FixCommonErrors class and its interface reducing potential side effects. In addition, the test FixCommonErrorsTest has been updated to reflect these changes.


src/ui/Forms/FixCommonErrors.cs
```diff
@@ -681,18 +681,7 @@ public void LogStatus(string sender, string message)
                 }
             }
         }
-
-        public object GetCustomCallbackData(IFixCommonError sender, object input)
```
ivandrofly, line 685: This method doesn't belong here!

We should only methods/props that is common to all the fix-common-errors logics

PR #7191, opened 2023-07-31
New improvements to "Beautify time codes"

Flitskikker:
Here are some accumulated improvements I made to Beautify time codes while testing over the last weeks.
- Replaced LINQ queries by custom binary search functions to drastically increase performance
  - All queries in `TimeCodesBeautifier` and `ShotChangeHelper` have been replaced, except the initial conversions and some safe paragraph getters. Beautifying is now even faster.
  - I ran some (unofficial) speed tests (see below) and checks to ensure there are no differences between the old and new functions.
- When using exact time codes, actually use those for all msec <> frames conversions to improve accuracy
  - Before, I was using `SubtitleFormat.FramesToMilliseconds` and `SubtitleFormat.MillisecondsToFrame` and then snapping to the closest exact time code, but this led to rounding issues, especially on videos with variable framerates.
- Tweak for connected subtitles: if both cues are in red zones and one of them would go invalid (e.g. ≤0 duration), snap to the other shot change, even though it's not the closest
- Tweak for connected subtitles: if the left cue is in a red zone and the right cue is in the same shot change's green zone, let the green zone take priority (instead of the red zone) if the other subtitle would go invalid (e.g. ≤0 duration)
- Fix progress bar crash for generating shot changes and time codes
  - The progress bar's `Maximum` was set to the total video duration, but sometimes this was incorrect (e.g. with some XviD files) and the `Value` would exceed it. Added a `Math.Min` to prevent overflow.
- In Batch generate waveform: only update status if the step is actually executed
- Some other minor changes 

---

### LINQ vs. Binary Search

These tests where done on the shot change and time code files of a 1h53m movie file.
163367 time codes and 2674 shot changes.

| | LINQ | Binary Search |
|--------|--------|--------|
| 10000x find closest shot change | 379 ms | <1 ms |
| 10000x find closest time code | 23553 ms | 1 ms |
| 10000x find next shot change | 263 ms | 1 ms |
| 10000x find previous shot change | 468 ms | 1 ms |
| 10000x find first shot change between two points | DNF | 1 ms | 
| 100x find first shot change between two points | 4201 ms | <1 ms |


src/libse/Common/ArrayExtensions.cs
```diff
@@ -0,0 +1,181 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Nikse.SubtitleEdit.Core.Common
+{
+    public static class ArrayExtensions
```
ivandrofly, line 6: Maybe rename this to collection/list extension as all the method are for List<t>

ivandrofly (changes requested):

src/libse/Common/ArrayExtensions.cs
Flitskikker: Good point! Renamed the class and file.

PR #7192, opened 2023-07-31
Add secure password encryption and decryption

ivandrofly:
A package reference to 'System.Security.Cryptography.ProtectedData' has been added to implement secure password encryption using the 'Protect' and 'Unprotect' methods. This will enhance the security of stored passwords in the application. Password encoding and decoding methods in Settings.cs have been updated accordingly for secure password encryption/decryption. Moreover, error handling is improved in situations where provided password is null or empty.

### DIFF
--- a/src/libse/Common/Settings.cs
+++ b/src/libse/Common/Settings.cs
@@ -5,6 +5,7 @@ using System.Drawing;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Security.Cryptography;
 using System.Text;
 using System.Xml;
 using System.Xml.Serialization;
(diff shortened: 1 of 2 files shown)


ivandrofly:
Will prevent identity stolen when someone accidentally shares its setting with proxy credentials it it!
As it's very common for files to be shared here to help with debugging

niksedk:
Nice. Can you also do this without adding an package reference which would need additional dll in installer/beta?

ivandrofly:
> Nice. Can you also do this without adding an package reference which would need additional dll in installer/beta?

Libse is a multi-target library, see in Libse.csproj `\<TargetFrameworks>net48;netstandard2.1</TargetFrameworks\>` 
That means the package is required in order for the project to build due to the .netstandard2.1 presence in the *TargetFrameworks* element.

That doesn't mean `Subtitle Edit` must have dll in its install folder.
Subtitle Edit will always use the .net48 version of the Libse (which has dependency on the user's global installation of .net framework) 

If you want to see this in action, just build Subtitle Edit and go to `subtitleedit\src\ui\bin\Debug` there will be no `System.Security.Cryptography.ProtectedData.dll`  but if let say you try to reference **Libse** from a .net core 3.1 project then build and go to the output folder then you will see the `System.Security.Cryptography.ProtectedData.dll`

niksedk:
Has a dll + only Win

ivandrofly:
> Has a dll + only Win

Please take a look at https://github.com/ivandrofly/subtitleedit/commit/863a7bb50f2d7e8cee90470e1469bfe732aa8351

ivandrofly:
Please see this output binary
image


There is no `System.Security.Cryptography.ProtectedData.dll`

PR #7201, opened 2023-08-02
Add secure password encryption and decryption #2

ivandrofly:
See https://github.com/SubtitleEdit/subtitleedit/pull/7192 for context

### DIFF
--- a/src/libse/LibSE.csproj
+++ b/src/libse/LibSE.csproj
@@ -52,7 +52,6 @@
     <PackageReference Include="Microsoft.Win32.Registry" Version="5.0.0" />
     <PackageReference Include="System.Drawing.Common" Version="6.0.0" />
     <PackageReference Include="System.Net.Http" Version="4.3.4" />
-    <PackageReference Include="System.Security.Cryptography.ProtectedData" Version="7.0.1" />
     <PackageReference Include="UTF.Unknown" Version="2.5.1" />
     <PackageReference Include="zlib.net-mutliplatform" Version="1.0.6" />
   </ItemGroup>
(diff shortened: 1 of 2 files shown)


src/libse/LibSE.csproj
```diff
@@ -67,4 +67,12 @@
     </None>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net48'">
+    <Reference Include="System.Security" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.1'">
+    <PackageReference Include="System.Security.Cryptography.ProtectedData" Version="7.0.1" />
```
ivandrofly, line 75: Add `System.Security.Cryptography.ProtectedData` only when targeting `netstandard2.1`

src/libse/LibSE.csproj
```diff
@@ -67,4 +67,12 @@
     </None>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net48'">
+    <Reference Include="System.Security" />
```
ivandrofly, line 71: For `net48` use assembly reference instead

ivandrofly:
Please do not merge until https://github.com/SubtitleEdit/subtitleedit/issues/7202

If you agree with what was said in the issue, they this pr will change meet 
what I mentioned in the issue which I think it's the right way!

ivandrofly:
If we are not going with what was mentioned above https://github.com/SubtitleEdit/subtitleedit/pull/7201#issuecomment-1662981098 

Then I think this is safe to go in.

What changed?

Now we have a filter that will choose what to reference depending on the context.
If Libse is being targeted by `Subtitle Edit` (UI) project then it will use the .NET framework's library for Systme.Security
and no additional dll will be added (also no nuget package)

```xml
  <ItemGroup Condition="'$(TargetFramework)' == 'net48'">
    <Reference Include="System.Security" />
  </ItemGroup>
```

In order for project to build this is also required in case any other project is referencing to *libse* its .net standard target instead of the .net Framework
```xml
  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.1'">
    <PackageReference Include="System.Security.Cryptography.ProtectedData" Version="7.0.1" />
  </ItemGroup>
```

PR #7227, opened 2023-08-08
Improve UI responsiveness by adding async OnGotFocus

ivandrofly:
This commit replaces the previous synchronous focus handling in the NikseComboBox controller with an asynchronous one. It aims at enhancing the UI responsiveness by implementing an asynchronous handling of the OnGotFocus event and preventing the possible UI freeze due to the blocking call Application.DoEvents(). A Task.Delay is used to create a minor delay before focusing, without blocking the UI thread. A TaskScheduler is used to safely update UI from the async task.

See https://github.com/SubtitleEdit/subtitleedit/commit/da1bb89f753cb78262cf3538e312d4deca170f54#diff-f1a2a06af29f85e247ea02a221e6f732ab20d7fce985981b9ec70e1c70d09c55R231-R236 for more context

### DIFF
--- a/src/ui/Controls/NikseComboBox.cs
+++ b/src/ui/Controls/NikseComboBox.cs
@@ -6,6 +6,7 @@ using System.ComponentModel;
 using System.Drawing;
 using System.Drawing.Design;
 using System.Drawing.Drawing2D;
+using System.Threading.Tasks;
 using System.Windows.Forms;
 
 namespace Nikse.SubtitleEdit.Controls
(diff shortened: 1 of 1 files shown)


ivandrofly:
For the rest of the callers of `SubtitleEdit\Logic\SyncContextExtensions.cs:8` the same can be done

ivandrofly:
If you want changes to be done here please let me know otherwise feel free to close

src/ui/Controls/NikseComboBox.cs
```diff
@@ -236,20 +237,28 @@ public string SelectedText
...
+        private TaskScheduler _uiTaskScheduler;
+        protected override async void OnGotFocus(EventArgs e)
         {
             base.OnGotFocus(e);
             if (_textBox != null && DropDownStyle == ComboBoxStyle.DropDown)
             {
-                try
-                {
-                    Application.DoEvents();
-                    System.Threading.SynchronizationContext.Current.Post(TimeSpan.FromMilliseconds(25), () => _textBox.Focus());
```
ivandrofly, line 247: Btw, is there s specific reason why you want to delay 25ms before trigger the focus action?

niksedk:
OK, moved your code here: https://github.com/SubtitleEdit/subtitleedit/commit/5a292b0a820556b954787a3aa09c3b464e04896c

Seems to be faster too :)
(though I could not get Benchmark dot net to work... so I just measured with a timer)

ivandrofly:
Yup, but there is one thing - I don't think we need `System.Threading.SynchronizationContext.Current.Post` in order to scheduler the execution as it's also not being used inside the method.

Maybe refacting to something like Task.DelayExecuteOnUIContext(() => tb.Focuse(), 20);

ivandrofly:
Ah just refresh and see there is a refect :)
https://github.com/SubtitleEdit/subtitleedit/commit/900a85f724b37a8981150e6809b0cb451b13bdb7

PR #7233, opened 2023-08-12
Update/Fixes AdvancedTextBox 

ivandrofly:
Extracted the logic checking dictionary availability into a separate method "IsDictionaryAvailable". This aims to improve code readability and modularity, thereby promoting better maintainability. Instead of having the logic embedded in the InitializeLiveSpellCheck method, it's now in its own method that can be used elsewhere if needed.

Modified the LoadDictionariesAsync method to directly return a new Task. This eliminates the need for async/await and simplifies code while maintaining functionality.

An error was noticed in the language check within AdvancedTextBox.cs where it was checking for "en " (with an extra whitespace) which always returned false and thus the condition inside never executed. This commit removes the unwanted whitespace to fix the condition check for English language ("en").

### DIFF
--- a/src/ui/Controls/AdvancedTextBox.cs
+++ b/src/ui/Controls/AdvancedTextBox.cs
@@ -591,6 +591,19 @@ namespace Nikse.SubtitleEdit.Controls
             }
         }
 
+        private static bool IsDictionaryAvailable(string language)
+        {
+            foreach (var downloadedDictionary in Utilities.GetDictionaryLanguagesCultureNeutral())
+            {
+                if (downloadedDictionary.Contains($"[{language}]", StringComparison.OrdinalIgnoreCase))
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
         public async Task InitializeLiveSpellCheck(Subtitle subtitle, int lineNumber)
         {
             if (lineNumber < 0)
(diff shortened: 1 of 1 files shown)


src/ui/Controls/AdvancedTextBox.cs
```diff
@@ -848,7 +845,7 @@ private void SetWrongWords()
                         continue;
                     }
 
-                    if (CurrentLanguage == "en " && (currentWordText.Equals("a", StringComparison.OrdinalIgnoreCase) || currentWordText == "I"))
+                    if (CurrentLanguage == "en" && (currentWordText.Equals("a", StringComparison.OrdinalIgnoreCase) || currentWordText == "I"))
```
ivandrofly, line 848: Typo/bug here!

src/ui/Controls/AdvancedTextBox.cs
```diff
@@ -601,44 +614,28 @@ public async Task InitializeLiveSpellCheck(Subtitle subtitle, int lineNumber)
...
-                }
 
                 LanguageChanged = true;
                 CurrentLanguage = detectedLanguage;
             }
         }
 
-        private async Task LoadDictionariesAsync(string languageName) =>
-            await Task.Run(() => LoadDictionaries(languageName));
+        private Task LoadDictionariesAsync(string languageName) => new Task(() => LoadDictionaries(languageName));
```
ivandrofly, line 638: Do not force run right away here - by just returning Task is enough

PR #7264, opened 2023-08-19
Feature/ocr engine

ivandrofly:
Added a new method IsAllUppercase in the StringExtensions class. This method checks whether all letter-characters in a given string are uppercase. Created test methods for this new functionality in the StringExtensionsTest class. Also, a new class FormattableTextReader with Read() method is added to read the text correctly in IsAllUppercase method. FormattableTextReaderText class is added for testing FormattableTextReader class. The main purpose of these changes is to provide a way to validate if the string is in uppercase, ignoring HTML tags or any non-letter characters.

The code block that checks if a word is all uppercase has been refactored into a separate method, IsAllUppercase(), in the StringExtensions class. This change improves the readability and maintainability of the OcrFixEngine class.

### DIFF
--- a/src/libse/Common/FormattableTextReader.cs
+++ b/src/libse/Common/FormattableTextReader.cs
@@ -1,0 +1,82 @@
+namespace Nikse.SubtitleEdit.Core.Common
+{
+    public struct FormattableTextReader
+    {
+        private readonly string _text;
+        private int _current;
+
+        public FormattableTextReader(string text, int position = -1)
+        {
+            _text = text;
+            _current = position;
+        }
+
+        public bool Read()
+        {
+            var len = _text.Length;
+            if (_current + 1 < len)
+            {
+                _current++;
+
+                while (IsTagStartChar(GetCurrent()))
+                {
+                    var closeChar = GetClosingPair(GetCurrent());
+
+                    // skip tag
+                    var tagCloseIndex = _text.IndexOf(closeChar, _current + 1);
+                    // '<' found but it doesn't have a closing pair
+                    if (tagCloseIndex < _current + 2) // 1 tag name index
+                    {
+                        return true;
+                    }
+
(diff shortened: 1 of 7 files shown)


src/libse/Common/FormattableTextReader.cs
```diff
@@ -0,0 +1,82 @@
+namespace Nikse.SubtitleEdit.Core.Common
+{
+    public struct FormattableTextReader
```
ivandrofly, line 3: I already see several use case for this `FormattableTextReader`

ivandrofly: The struct is also used instead of class to avoid overheat that can be caused by class.

For example in method `IsAllUppercase`, let say is being called inside a loop to check if text in 
paragraphs are all uppercase then no allocation will happen due to struct being store in stack instead of heap!

src/libse/Common/StringExtensions.cs
```diff
@@ -756,6 +756,46 @@ public static bool HasSentenceEnding(this string value)
...
+        /// <param name="value">The string to evaluate.</param>
+        /// <remarks>
+        /// If the string is null or empty, the method will return false.
+        /// The method only checks letter-characters and ignores any non-letter characters.
+        /// If the string doesn't contain any letters, the method returns false.
+        /// </remarks>
+        /// <returns>
+        /// Returns true if all letter-characters in the string are uppercase. Returns false otherwise.
+        /// </returns>
+        public static bool IsAllUppercase(this string value)
```
ivandrofly, line 771: The documentation explain very well when to call this method to check if string is all uppercase!

One of the thing not to do is to call this method and try use its inverse value to say string is lowercase, 
the reason is that is *vaule* is null it will return `false` but that doesn't mean the string is all lowercase!

If we need `IsAllLowercase` it's better to create it basing on what it's done for `IsAllUppercase`

PR #7301, opened 2023-08-29
Refactor DarkTheme class to optimize sub control access

ivandrofly:
Optimize and remove unused method

### DIFF
--- a/src/ui/Logic/DarkTheme.cs
+++ b/src/ui/Logic/DarkTheme.cs
@@ -53,23 +53,12 @@ namespace Nikse.SubtitleEdit.Logic
             }
         }
 
-        public static IEnumerable<Control> GetAllControlByType(Control control, Type type)
-        {
-            var controls = control.Controls.Cast<Control>().ToList();
-
-            return controls.SelectMany(ctrl => GetAllControlByType(ctrl, type))
-                                      .Concat(controls)
-                                      .Where(c => c.GetType() == type);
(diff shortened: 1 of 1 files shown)


src/ui/Logic/DarkTheme.cs
```diff
@@ -53,23 +53,12 @@ public static void SetWindowThemeNormal(Control control)
...
-                                      .Where(c => c.GetType() == type);
-        }
-
-        private static List<T> GetSubControls<T>(Control c)
+        private static IEnumerable<T> GetSubControls<T>(Control c)
         {
             var type = c.GetType();
-            var fields = type.GetFields(BindingFlags.NonPublic | BindingFlags.Instance);
-            var contextMenus = fields.Where(f => f.GetValue(c) != null &&
-            (f.GetValue(c).GetType().IsSubclassOf(typeof(T)) || f.GetValue(c).GetType() == typeof(T)));
```
ivandrofly, line 70: Field type can be inferenced without the need to get the field value

src/ui/Logic/DarkTheme.cs
```diff
@@ -53,23 +53,12 @@ public static void SetWindowThemeNormal(Control control)
             }
         }
 
-        public static IEnumerable<Control> GetAllControlByType(Control control, Type type)
```
ivandrofly, line 56: This method is not being used

PR #7355, opened 2023-09-09
Feature/time code

ivandrofly:
- Zero TimeCode instance will now be created uzing TimeCode.Zero
- Implement `IEquatable<TimeCode>` to allow comparing TimeCode instance by value  also make the comparison between time code consistence

### DIFF
--- a/src/Test/Logic/ParagraphTest.cs
+++ b/src/Test/Logic/ParagraphTest.cs
@@ -56,7 +56,7 @@ namespace Test.Logic
         [TestMethod]
         public void TestAdjust3()
         {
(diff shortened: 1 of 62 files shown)


niksedk:
I like the  `IEquatable<TimeCode>` - could probably also be used a few places.

What is the reason for the `TimeCode.Zero`?

ivandrofly:
> What is the reason for the `TimeCode.Zero`?

Semantic - maybe TimeCode.Default?
Can also be reverted

ivandrofly:
> I like the `IEquatable<TimeCode>` - could probably also be used a few places.

yeah, can also have a overloading to deal with double directly
to save creating new instance of TimeCode just to perform comparison

PR #7374, opened 2023-09-12
Feature/http client

ivandrofly:


src/libse/Http/HttpClientAdapter.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        public HttpRequestHeaders DefaultRequestHeaders => _httpClient.DefaultRequestHeaders;
+
+        public Task<HttpResponseMessage> PostAsync(string uri, StringContent stringContent) => _httpClient.PostAsync(uri, stringContent);
+
+        public async Task<string> GetStringAsync(string url)
+        {
+            var buffer = await _httpClient.GetByteArrayAsync(url);
+            return _encoding.GetString(buffer, 0, buffer.Length);
+            // return Task.FromResult(_encoding.GetString(buffer, 0, buffer.Length - 1));
```
ivandrofly, line 42: ` // return Task.FromResult(_encoding.GetString(buffer, 0, buffer.Length - 1));`
I left this line because it was recently introduced and I don't get what is the point of having it?

src/libse/Http/LegacyHttpClient.cs
```diff
@@ -112,37 +112,7 @@ public void Dispose()
             _httpClient?.Dispose();
         }
 
-        public static void CopyTo(Stream source, Stream destination, int bufferSize)
```
ivandrofly, line 115: This method is removed as the framework as a similar one!
ps: The one in framework also perform validations

PR #7447, opened 2023-09-25
Feature/translate

ivandrofly:
- Fixes method name
- Remove unnecessary method 'FixInvalidCarriageReturnLineFeedCharacters' as the issue mentioned in https://github.com/SubtitleEdit/subtitleedit/compare/main...ivandrofly:feature/translate?expand=1#diff-410c4ad661319da5376718b483cf2f1d332ab8ff21c49721ff6acb5c2ff2a0ebL88 was being caused by replacing '\r' aka "SplitChar" with string.empty which will make the string from => foo\\r\\nbar => foo\\rbar (https://github.com/SubtitleEdit/subtitleedit/compare/main...ivandrofly:feature/translate?expand=1#diff-c5aba14dc988226e5beafa90797051a8a2e7967359d3b702d43316c6bb383058L66)
- Fix typo

### DIFF
--- a/src/libse/Translate/Service/GoogleTranslator1.cs
+++ b/src/libse/Translate/Service/GoogleTranslator1.cs
@@ -63,7 +63,8 @@ namespace Nikse.SubtitleEdit.Core.Translate.Service
                     input.Append(" " + SplitChar + " ");
                 }
 
-                var text = f.SetTagsAndReturnTrimmed(TranslationHelper.PreTranslate(p.Text.Replace(SplitChar.ToString(), string.Empty), sourceLanguage), sourceLanguage);
+                var text = TranslationHelper.ExpandContractions(p.Text, sourceLanguage);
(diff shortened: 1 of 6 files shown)


src/libse/Translate/Service/GoogleTranslator1.cs
```diff
@@ -63,7 +63,8 @@ public List<string> Translate(string sourceLanguage, string targetLanguage, List
                     input.Append(" " + SplitChar + " ");
                 }
 
-                var text = f.SetTagsAndReturnTrimmed(TranslationHelper.PreTranslate(p.Text.Replace(SplitChar.ToString(), string.Empty), sourceLanguage), sourceLanguage);
```
ivandrofly, line 66: This line is buggy `p.Text.Replace(SplitChar.ToString(), string.Empty)` will remove the \\n from \\r\\n which will make the line buggy!

PR #7489, opened 2023-10-07
Add support for TimeLineAscii in 'ToText' method

ivandrofly:
Added support for exporting subtitles to TimeLineAscii format in SubtitleFormat class. This would help to provide different formats for the subtitles depending on the user's preference. A new internal class called 'TimeLineAsciiTextWriter' has been introduced to handle the writing of subtitle data into the TimeLineAscii format. There's also a change in the placement of TimeLineAscii in the formats list and a 'TODO' comment has been added for future reference.

### DIFF
--- a/src/libse/SubtitleFormats/SubtitleFormat.cs
+++ b/src/libse/SubtitleFormats/SubtitleFormat.cs
@@ -220,6 +220,7 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
                     new TimedText200604CData(),
                     new TimedText200604Ooyala(),
                     new TimedText(),
+                    new TimeLineAscii(),
                     new TitleExchangePro(),
                     new Titra(),
                     new TmpegEncText(),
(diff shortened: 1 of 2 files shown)


src/libse/SubtitleFormats/SubtitleFormat.cs
```diff
@@ -666,7 +667,7 @@ public static SubtitleFormat[] GetTextOtherFormats()
                 new DlDd(),
                 new Ted20(),
                 new Captionate(),
-                new TimeLineAscii(),
+                new TimeLineAscii(), // TODO: remove from here?
```
ivandrofly, line 670: Should we keep the format in `GetTextOtherFormats` since it now support writing to text? @niksedk

src/libse/SubtitleFormats/TimelineAscii.cs
```diff
@@ -37,6 +37,7 @@ public TimeLineAsciiTextWriter(bool useWindowStyleNewline)
                 _stringWriter.NewLine = useWindowStyleNewline ? "\r\n" : "\n";
                 _stringWriter.WriteLine("<summary>");
                 _stringWriter.WriteLine("Timeline Ascii export - THE MOVIE TITRE EDITOR - http://www.pld.ttu.ee/~priidu/timeline/ by priidu@pld.ttu.ee");
+                _stringWriter.WriteLine();
```
ivandrofly, line 40: image

PR #7629, opened 2023-11-13
Refactor 'AddPeriods' in 'AudioToTextPostProcessor'

ivandrofly:
The logic within 'AddPeriods' method is optimized for better performance and enhanced readability. Addition of a helper function 'IsNonStandardLineTerminationLanguage' simplifies language determination. Also, unnecessary period addition in subtitle paragraphs has been prevented and conditions for adding periods have been cleaned up, providing a better structured and less repetitive code.

### DIFF
--- a/src/libse/AudioToText/AudioToTextPostProcessor.cs
+++ b/src/libse/AudioToText/AudioToTextPostProcessor.cs
@@ -215,9 +215,15 @@ namespace Nikse.SubtitleEdit.Core.AudioToText
             return s;
         }
 
+        private static bool IsNonStandardLineTerminationLanguage(string language)
+        {
+            return language == "jp" || language == "cn" || language == "yue";
+        }
+
         public Subtitle AddPeriods(Subtitle inputSubtitle, string language)
         {
-            if (language == "jp" || language == "cn" || language == "yue")
+            if (IsNonStandardLineTerminationLanguage(language) ||
+                inputSubtitle.Paragraphs.Count == 0)
             {
                 return new Subtitle(inputSubtitle);
             }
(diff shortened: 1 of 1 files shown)


src/libse/AudioToText/AudioToTextPostProcessor.cs
```diff
@@ -226,46 +232,48 @@ public Subtitle AddPeriods(Subtitle inputSubtitle, string language)
...
-            var last = subtitle.GetParagraphOrDefault(subtitle.Paragraphs.Count - 1);
-            if (last != null &&
-                !last.Text.EndsWith('.') &&
-                !last.Text.EndsWith('!') &&
-                !last.Text.EndsWith('?') &&
-                !last.Text.EndsWith(',') &&
-                !last.Text.EndsWith(':'))
-            {
-                subtitle.Paragraphs[subtitle.Paragraphs.Count - 1].Text += ".";
-            }
```
ivandrofly, line 267: This code is unnecessary as we can use "default paragraph" (see line 238) with start-time of the last paragraph + SetPeriodIfDistanceToNextIsMoreThanAlways+1 and let the for-loop goes to `subtitle.Paragraph.Count` instead of  `subtitle.Paragraph.Count-1`.

When we `subtitle.GetParagraphOrDefault(index + 1)` if null is returned then we can be sure that the current paragraph is the last and just return the default  `defaultParagraph`

src/libse/AudioToText/AudioToTextPostProcessor.cs
```diff
@@ -226,46 +232,48 @@ public Subtitle AddPeriods(Subtitle inputSubtitle, string language)
...
-                    else
-                    {
-                        var lastWord = GetLastWord(paragraph.Text);
-                        var nextFirstWord = GetFirstWord(next.Text);
-                        if (TwoLetterLanguageCode == "en" && (englishSkipLastWords.Contains(lastWord) || englishSkipFirstWords.Contains(nextFirstWord)))
-                        {
-                            continue;
-                        }
 
-                        paragraph.Text += ".";
```
ivandrofly, line 253: Previously the codes inside this `else` statement will never execute as line 240 would always evaluate to true!

PR #7633, opened 2023-11-15
Update NameList.cs

RocketFriday:
Automatically add Possessives for English language names. "Possessives in English, for example, can be marked with an apostrophe (‘s), as in the book’s cover or John’s car."

### DIFF
--- a/src/libse/Dictionaries/NameList.cs
+++ b/src/libse/Dictionaries/NameList.cs
@@ -1,4 +1,4 @@
-﻿using Nikse.SubtitleEdit.Core.Common;
+using Nikse.SubtitleEdit.Core.Common;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -258,13 +258,25 @@ namespace Nikse.SubtitleEdit.Core.Dictionaries
             }
 
             var de = nameListXml.DocumentElement;
-            if (de != null)
-            {
-                var node = nameListXml.CreateElement("name");
-                node.InnerText = name;
-                de.AppendChild(node);
-                nameListXml.Save(fileName);
+            if (de == null) 
+            { 
+                return true; 
             }
+
+            var node = nameListXml.CreateElement("name");
+            node.InnerText = name;
+            de.AppendChild(node);
+            if(LanguageName == "en")
+            {
+                if(!(name.EndsWith("'s") || name.EndsWith("'")))
+                {
+                    var nodeAlt = nameListXml.CreateElement("name");
+                    nodeAlt.InnerText = name.EndsWith("s") ? $"{name}'" : $"{name}'s";
+                    de.AppendChild(nodeAlt);
+                }
+            }
+
+            nameListXml.Save(fileName);
             return true;
         }
 


src/libse/Dictionaries/NameList.cs
```diff
@@ -1,4 +1,4 @@
-﻿using Nikse.SubtitleEdit.Core.Common;
+using Nikse.SubtitleEdit.Core.Common;
```
RocketFriday, line 1: I didn't change anything idk why this is showing up.

src/libse/Dictionaries/NameList.cs
```diff
@@ -258,13 +258,25 @@ public bool Add(string name)
             }
 
             var de = nameListXml.DocumentElement;
-            if (de != null)
+            if (de == null) 
+            { 
+                return true; 
+            }
```
RocketFriday, line 264: By 'early exiting' you avoid extra indentation, which minorly improves performance and readability.
Source: this stack exchange [post](https://softwareengineering.stackexchange.com/questions/299263/do-nested-conditionals-have-a-significant-performance-impact).
¯\\_(ツ)_/¯ don't have to keep it. Idc.

RocketFriday:
Also I don't have the proper .net version so I wasn't able to test this. Sorry.

src/libse/Dictionaries/NameList.cs
```diff
@@ -258,13 +258,25 @@ public bool Add(string name)
...
+            var node = nameListXml.CreateElement("name");
+            node.InnerText = name;
+            de.AppendChild(node);
+            if(LanguageName == "en")
             {
-                var node = nameListXml.CreateElement("name");
-                node.InnerText = name;
-                de.AppendChild(node);
-                nameListXml.Save(fileName);
+                if(!(name.EndsWith("'s") || name.EndsWith("'")))
```
ivandrofly, line 271: I don't think this is a good idea it will only increase the name file size
This type of process can easily be handled in code

RocketFriday: Yes, I've been thinking about this on and off for a while, 
this is definitely _not the correct solution_ but at the time I thought it was helpful.
I meant to take it down but forgot.
It also doesn't account for cases like, 
_"John'll be back soon."_

Using subedit regularly for the past few months has starkly pointed out the issues with name handling.
Like how names can't have spaces in them.
_"Zhu Li, do the thing!" - Varrick, The Legend of Korra_

RocketFriday: Or how you need to create multiple ocr fixes to account for errors in case.
_Zhu li -> Zhu Li
zhu Li -> Zhu Li
zhu li -> Zhu Li_

RocketFriday: It would also be nice to be able to have a generic name keyword. 
Example:
Find and Replace:
[Name]'ll
With: 
[Name] will

or

[Name]
with:
\<i>[Name]\</i>

idk ¯\_(ツ)_/¯ like you said, achievable through code but not everyone is a coder.

PR #7897, opened 2024-02-02
Fix: comment and and context was switched for some checks

PWillumsen:
For some checks in the NetflixQualityCheck the comment and context was the wrong way round.


niksedk:
Thx :)

PR #7930, opened 2024-02-14
Add "Verify completeness" feature

Flitskikker:
Another feature that first lived in a separate test application, now extended, improved and integrated into SE.

This feature allows you to verify the completeness of a subtitle against a reference/control subtitle. Easily find missing subtitles and insert them into the current subtitle with the click of a button.

---

It can be accessed from the File menu:

image

---

It will show the coverage/overlap of all the subtitles from the control subtitle. 0% coverage means that there is no subtitle during that timeframe in your current subtitle.
You can click on a row to navigate to there in the player (+ waveform and list). It's presented as a form instead of a dialog.

image

If a subtitle is really missing, you can insert it into your subtitle.
If deemed unnecessary, like names or simple "Yes" and "OK", you can dismiss it.

---

You can also sort by time by right-clicking (or clicking the columns).
Your preference will be saved in Settings.xml.

image

image


niksedk:
OK, cool idea :)

PR #7945, opened 2024-02-18
Update pt-PT.xml

hugok79:
For commit: https://github.com/SubtitleEdit/subtitleedit/pull/7942/commits/ffc3150a4490084ecf3727ac6b284662faacfd67


ivandrofly:
Good catch, copy past

PR #7946, opened 2024-02-18
Update FixMissingOpenBracket method description

ivandrofly:
The method description for FixMissingOpenBracket in Language.cs has been updated. It now includes fixing both missing square and round opening brackets, as opposed to just square brackets previously.


ivandrofly:
Wasn't showing the correct value at runtime
image

PR #7967, opened 2024-02-22
Add tests and improve 'MissingOpenBracket' fixer

ivandrofly:
The update introduces test cases for the 'MissingOpenBracket' fixer to ensure it behaves as expected. A significant re-work has also been done on the 'MissingOpenBracket' fixer, refining its handling of various scenarios to resolve potential issues more effectively. The updated fixer also checks for per line restoration and deals with 2 or more lines even if their adjacent line is not closed correctly.

### DIFF
--- a/src/Test/FixCommonErrors/FixCommonErrorsTest.cs
+++ b/src/Test/FixCommonErrors/FixCommonErrorsTest.cs
@@ -8,6 +8,7 @@ using Nikse.SubtitleEdit.Logic;
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 
 namespace Test.FixCommonErrors
 {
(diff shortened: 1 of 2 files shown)


src/Test/FixCommonErrors/FixCommonErrorsTest.cs
```diff
@@ -3532,5 +3533,55 @@ public void UnbreakShortLinesPixelWidthTags()
...
+        }
+
+        [TestMethod]
+        public void FixMissingOpenBracketThreeTest()
+        {
+            var engine = new FixMissingOpenBracket();
+            var sub = GetGenericSub();
+            sub.Paragraphs.First().Text = "- Foobar bar zzz).\n- Foo bar Zz";
+            engine.Fix(sub, new EmptyFixCallback());
+            Assert.AreEqual( "- (Foobar bar zzz).\n- Foo bar Zz", sub.Paragraphs.First().Text);
```
ivandrofly, line 3564: The idea is that, if you have two like dialog and the second line has the closing 
parentheses and e.g: - Foobar bar zzz.\\n- Foo bar Zz) the the right restoring mean adding missing
open parenthesis at the beginning of the line where the parenthesis is missing not at the beginning of the whole paragraphs itself.

PR #8007, opened 2024-03-06
Feature/audio text

ivandrofly:
Introduced a new method 'DistanceTo' for TimeCode objects to calculate the distance between two points in milliseconds, which replaced various manual calculations throughout multiple files. Additionally, a new method 'IsTerminated' was implemented to check if a string ends with specific punctuation, simplifying and reducing repetitive checks.

### DIFF
--- a/src/libse/AudioToText/AudioToTextPostProcessor.cs
+++ b/src/libse/AudioToText/AudioToTextPostProcessor.cs
@@ -227,8 +227,9 @@ namespace Nikse.SubtitleEdit.Core.AudioToText
                 return false;
             }
 
-            var ch = text[0];
-            return ch == '.' && ch == '!' && ch == '?' && ch == ',' && ch == ':' && ch == ')' && ch == ']' && ch == '}';
+            var lastCh = text[text.Length - 1];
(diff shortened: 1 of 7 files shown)


src/libse/AudioToText/AudioToTextPostProcessor.cs
```diff
@@ -235,45 +252,33 @@ public Subtitle AddPeriods(Subtitle inputSubtitle, string language)
...
+                if (AlwaysSetPeriodNoFurtherValidation(distanceToNext))
+                {
+                    paragraph.Text += ".";
+                }
+                else
                 {
-                    var gap = next.StartTime.TotalMilliseconds - paragraph.EndTime.TotalMilliseconds;
-                    if (gap > SetPeriodIfDistanceToNextIsMoreThanAlways)
+                    if (TwoLetterLanguageCode == "en" && (englishSkipLastWords.Contains(GetLastWord(paragraph.Text)) ||
+                                                          englishSkipFirstWords.Contains(GetFirstWord(next.Text))))
```
ivandrofly, line 271: Both `GetLastWord(paragraph.Text))` and  `GetFirstWord(next.Text)` are moved to where they are needed 
to avoid unnecessary method calls as only one of them needs to evaluated to true in order for the expression to be evaluated as true

PR #8021, opened 2024-03-12
Change the Gridline view

jeff11-1-1:
Change the gridline view so each second is divided into 25 equal parts when the video's framerate is 25, rather 10 parts which is the general case.


niksedk:
I can see this could give some value, but it should probably support all frame rates.

Also, lines will be on exactly the correct frames... would probably requires ffproble or ffmpeg to list all frame times.

PR #8065, opened 2024-03-25
Feature/context menu combobox

ivandrofly:
Fixes a very sneaky issue with **NikseCombobox** and **ContextMenuStrip**

Currently there for example in Find Dialog, if you select "Regular expression" and right click 
inside the textarea it won't show the context menu for that regular expression,
In order to display the context menu, you have to right click in border or the dropbox icon area

This patch fixes that issue as it will capture any click in made in inner textbox and notify the **NikseCombobox** to display the context menu

It also remove the need of having duplicated codes .

PS: I can also add a gift showing the behavior of both in case the explanation above is not clear

### DIFF
--- a/src/ui/Forms/MultipleReplace.cs
+++ b/src/ui/Forms/MultipleReplace.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Text.RegularExpressions;
 using System.Windows.Forms;
 using System.Xml;
+using Nikse.SubtitleEdit.Controls.Adapters;
 using MessageBox = Nikse.SubtitleEdit.Forms.SeMsgBox.MessageBox;
 
 namespace Nikse.SubtitleEdit.Forms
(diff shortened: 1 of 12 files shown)


ivandrofly:
https://github.com/SubtitleEdit/subtitleedit/assets/6579818/e45bbb22-4d4e-4a83-9f27-cc2892ad5679

src/ui/Forms/MultipleReplace.cs
```diff
@@ -207,7 +208,14 @@ private void MultipleReplace_KeyDown(object sender, KeyEventArgs e)
 
         private void RadioButtonCheckedChanged(object sender, EventArgs e)
         {
-            textBoxFind.ContextMenuStrip = sender == radioButtonRegEx ? FindReplaceDialogHelper.GetRegExContextMenu(textBoxFind) : null;
+            // remove regex context menu from find text box
+            textBoxFind.ContextMenuStrip = null;
+
+            // only get the regex context menu if we select the regex option
+            if (sender == radioButtonRegEx && radioButtonRegEx.Checked)
```
ivandrofly, line 215: Checking if ` radioButtonRegEx.Checked`  is important as if the **regex radio button** was previously selected 
and you click to select another radio button, this method will still be invoked twice which will then call `FindReplaceDialogHelper.GetRegExContextMenu(textBoxFind)` and generate non-necessary allocation!

ivandrofly:
> It deletes existing text now clicking on an item in the context menu in Find... could you fix that too?

Look like you have already patched it here https://github.com/SubtitleEdit/subtitleedit/commit/53a6e817caa9bdb7937beceb59cee9e4bcc95c8a but will take a look to confirm the fix

PR #8082, opened 2024-03-27
Refactor OnCopyData event invocation in MessageHandlerWindow

ivandrofly:
The event invocation code in MessageHandlerWindow.cs file specifically for OnCopyData has been updated. Instead of creating a new instance of EventArgs, the code now uses the static EventArgs.Empty instance, which is more efficient and adheres to common C# best practices.


ivandrofly:
Not sure if MPC-HC is still being used out there but as WndProc is a method that is likely to be called several times 
within a second it's better to user the static empty EventArg instance to avoid putting pressure on GC to clean up those generated instance of EventArgs

PR #8089, opened 2024-03-28
Fixes #7958

### LINKED ISSUE #7958: History function of "Search and Replace"
I'm slowly starting to ditch Notepad++ as the extra editor to do replacements.
The spelling check is similar in SE and a few things even better (I like the Google search) and replace all.

But the history function in "search and replace" in Notepad++ also reacts on the up and down keys.
This way you don't need to take your hands of the keyboard to use the mouse and go through the drop-down.

Can you make it similar in SE ???

The conversion of "Hearing impaired" => "normal" has improved a lot and I'm now relying completely on it.
It may miss some things, but doing it in Notepad++ is so much more work.

Thanks for constantly improving this!!

ivandrofly:
User can now press up/down to navigate through the history items just like in Notepad++ :)

### DIFF
--- a/src/ui/Controls/NikseComboBox.cs
+++ b/src/ui/Controls/NikseComboBox.cs
@@ -424,6 +424,30 @@ namespace Nikse.SubtitleEdit.Controls
             return base.IsInputKey(keyData);
         }
 
+        private void NavigateUp() => Navigate(_selectedIndex - 1);
+
+        private void NavigateDown() => Navigate(_selectedIndex + 1);
+        
+        private void Navigate(int index)
+        {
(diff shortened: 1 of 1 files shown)


src/ui/Controls/NikseComboBox.cs
```diff
@@ -560,48 +584,15 @@ public NikseComboBox()
 
             _textBox.KeyDown += (sender, e) =>
             {
-                if (DropDownStyle != ComboBoxStyle.DropDown)
```
ivandrofly, line 563: This line was blocking this feature on NikseCombobox

Issue #7947, opened 2024-02-18
Karaoke Effect issue


Langrisser91:
Karaoke effect is not working good , the text flickers constantly and the color pick is wrong.

i made this video to show whats goin on
https://github.com/SubtitleEdit/subtitleedit/assets/160432467/d074e3e9-a04f-4834-bbc5-5b7c017836eb

Thanks ♥

niksedk:
Closed with #8100

Issue #7909, opened 2024-02-07
[Feature Request] Adding New Effect


M4Mahdi:
I know that there is Karaoke Effect and Typewriter Effect but their time is static and there is lack of a more flexible option. I want to automatically highlight the word that is currently speaking while the other words are in normal format in that subtitle line. I think it can be implemented with "Audio to text" feature.

As you can see, the word "went" is highlighted here exactly when the narrator is speaking it:
![image](https://github.com/SubtitleEdit/subtitleedit/assets/59066635/9481892e-1ac7-433f-bcfa-519c265e7158)

And the word "to" will be highlight after that:
![image](https://github.com/SubtitleEdit/subtitleedit/assets/59066635/f01a3746-aed8-4e8b-967f-492af7672993)

M4Mahdi:
OK. I think i need to share something here.
Immediately after opening this issue, i tried to do it in Python to have a solution. And i wrote a step by step guidance beside that Python script for myself. I didn't share it anywhere until now. This is that. And it works well for me. I hope it be understandable for anybody read this:

### Subtitle Edit - Step by Step Current Word Highlight Style:

1. Open "Subtitle Edit" and drop your sound or video to its internal player.

2.  Go to "Video -> Audio to text (Vosk/Kaldi)" -> Choose model -> English (very large) {or medium}

3. There is a checkbox with this label: "Use post-processing (line merge, fix casing, punctuation, and more)".
A. Once "select" this checkbox and press "Generate" button to have "Sentences Subtitle" file.
B. Once "unselect" this checkbox and press "Generate" button to have "Words Subtitle" file.
C. Save each file after it generated and before another generate.

4. Before using Python script to get the final output subtitle file, you must know or do these:
A. Words of final output file that Python will give us, are from "Sentences Subtitle" file. Times of output file are from "Words Subtitle" file.
B. If a word needed refinement, just change it in the "Sentences Subtitle" file.
Refinements without adding or removing "space character" have no problem but refinements with that, need more advanced efforts.
For refinements with adding or removing "space character", you need to {change both "Sentences Subtitle" file and "Words Subtitle" file}:Easier or {change only the final output file}:Harder(because it is an styled subtitle and there are many repetitive sentences and words than a normal subtitle file).
Same word count of "Sentences Subtitle" file and "Words Subtitle" file is critically necessary. Words are separated with "space character" or "newline character"; I am using split() function in the Python 3.x and it is doing it.
[...]

PR #8100, opened 2024-03-30
Fixes #6234 #7947 #7909  Add karaoke effect for text transformation

### LINKED ISSUE #6234: [Request and Bug] Karaoke Effect
1. Kindly add support for word-level Karaoke Effect?
2. When I add the Karaoke effect to a video, the subtitles tend to flicker. Is this normal and can it be fixed?

ivandrofly:
Added classes for karaoke effect text transformation with support for splitting by character or word. Also implemented a strategy pattern to handle the different transformations. Additionally, added UI components to let users choose the type of effect and made adjustments to existing classes for the new features.

- Karaoke: Legacy strategy is preserved (Karaoke effect)
- Karaoke: Is formatting aware (tag existing tag in text will be keep*)
- Karaoke: Has both per word and per char coloring

Note: The new **Typewriter effect class**  also fixes tag issue but it's not yet integrated with the UI.

@niksedk if you could also test on your side and provide feedback.
Unit-tests will be added in upcoming days


ivandrofly:
For #7909 if user choose option for **per-word karaoke effect** it fits* with what the user has described

src/libse/Common/TextEffect/KaraokeEffect.cs
```diff
@@ -0,0 +1,60 @@
...
+            }
+
+            var len = text.Length;
+            for (var i = 0; i < len; i++)
+            {
+                var ch = text[i];
+                if (HtmlUtil.IsStartTagSymbol(ch))
+                {
+                    var closeIdx = text.IndexOf(HtmlUtil.GetClosingPair(ch), i + 1);
+                    if (closeIdx < 0) return i;
```
niksedk, line 50: Please always use brackets

src/libse/Common/TextEffect/KaraokeEffect.cs
```diff
@@ -0,0 +1,60 @@
...
+                }
+                else return i;
```
niksedk, line 52: Please always use brackets

niksedk:
Hi @ivandrofly ,

Looks good :)

Suggestions:
- Please do remove the "legacy" version + remove the "New effect/legacy effect" (perhaps an setting for white space).
- The gap between lines should be zero (was also 1 in the old version), but start/end time should be the same to avoid flickering (or a setting).
- The `tab-stop order` should be fixed (was also broken in old version). 
- Tabbing out of 'End delay in seconds' changes the value.
- 'End delay in seconds' combobox does not work?

ivandrofly:
> * 'End delay in seconds' combobox does not work?

Combobox? You mean the numeric up/down? It does work, try increasing the value and click preview

> * Tabbing out of 'End delay in seconds' changes the value.

Yes, but you have to click Preview button to play/replay the effect. Do you want it to be reactive as soon as the value changed?

niksedk:
Let me know when it's ready :)

ivandrofly:
> Let me know when it's ready :)

Sure, currently on Mac and Unit-test project doesn't play nice with it :D, maybe we need start considering making unit-test project target multi-platform testing framework as LibSE .

niksedk:
> > Let me know when it's ready :)
> 
> Sure, I'm currently on Mac and Unit-test project doesn't play nice with it :D, maybe we need start considering making unit-test project target multi-platform testing framework as LibSE .

So, it's ready besides unit tests? (they can always be added later)

ivandrofly:
> > > Let me know when it's ready :)
> > 
> > 
> > Sure, I'm currently on Mac and Unit-test project doesn't play nice with it :D, maybe we need start considering making unit-test project target multi-platform testing framework as LibSE .
> 
> So, it's ready besides unit tests? (they can always be added later)

yes

ivandrofly:
1/3 Issues was closed, can you close the remaining @niksedk

niksedk:
Is think #7909 is something else...

ivandrofly:
Oh I think I got it now... something like e.g: 

Input: **Foo bar zz**

Output:
\<font..>foo\</font> bar zz
foo \<font..>bar\</font> zz
foo bar \<font..>zz\</font>

Maybe char/second can be used to try to estimated what word will is to be highlighted.

niksedk:
I think this requires something like:

image

ivandrofly:
> I think this requires something like `--highlight_words true` in Faster Whisper:
> 
> image

Okay, Will take a look - haven't played with Whisper yet

niksedk:
> Okay, Will take a look - haven't played with Whisper yet

It's pretty good, and it's getting even better slowly.... the  `--highlight_words true` actually gives a pretty large block...
The timestamps are sometimes a little off. And a word here and there are also often off.

M4Mahdi:
> Oh I think I got it now... something like e.g:
> 
> Input: **Foo bar zz**
> 
> Output: <font..>foo</font> bar zz foo <font..>bar</font> zz foo bar <font..>zz</font>
> 
> The font open tag will be moving instead of staying in fixed location like in karaoke effect.
> 
> Maybe char/second can be used to try to estimated what word will is to be highlighted.
> 
> Ref #7909

You can check this comment too:
https://github.com/SubtitleEdit/subtitleedit/issues/7909#issuecomment-2067876022

PR #8101, opened 2024-03-30
Update HtmlUtilTest for formattable text check

ivandrofly:
The commit changes some of the test cases for text formattability in HtmlUtilTest.cs. Notably, it modifies a repeated assertion with a string "<i>!?.", replacing it with a fresh assertion testing "!?.</i>".

### DIFF
--- a/src/Test/Logic/HtmlUtilTest.cs
+++ b/src/Test/Logic/HtmlUtilTest.cs
@@ -68,15 +68,15 @@ namespace Test.Logic
 
             Assert.AreEqual("Hallo!", c);
         }
-        
+
         [TestMethod]
         public void IsTextFormattableFalse()
         {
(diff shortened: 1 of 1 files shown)


src/Test/Logic/HtmlUtilTest.cs
```diff
@@ -68,15 +68,15 @@ public void RemoveFontName4()
...
             Assert.AreEqual("Hallo!", c);
         }
-        
+
         [TestMethod]
         public void IsTextFormattableFalse()
         {
             Assert.IsFalse(HtmlUtil.IsTextFormattable("<i></i>"));
             Assert.IsFalse(HtmlUtil.IsTextFormattable("<i>!?."));
-            Assert.IsFalse(HtmlUtil.IsTextFormattable("<i>!?."));
```
ivandrofly, line 77: Was a duplication with line 76

PR #8118, opened 2024-04-02
Switch MouseDown event to MouseUp in NikseUpDown control

ivandrofly:
The code changes correct an event handling in the NikseUpDown custom control. Previously, the base's OnMouseDown event was called, but the correct behavior should be to call the base's OnMouseUp event. This switch ensures the control responds properly to mouse interactions.

Short: `OnMouseDown` base is being called on `OnMouseUp` event

### DIFF
--- a/src/ui/Controls/NikseUpDown.cs
+++ b/src/ui/Controls/NikseUpDown.cs
@@ -532,7 +532,7 @@ namespace Nikse.SubtitleEdit.Controls
                 _buttonLeftIsDown = false;
                 Invalidate();
             }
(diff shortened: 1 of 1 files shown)


src/ui/Controls/NikseUpDown.cs
```diff
@@ -532,7 +532,7 @@ protected override void OnMouseUp(MouseEventArgs e)
                 _buttonLeftIsDown = false;
                 Invalidate();
             }
-            base.OnMouseDown(e);
```
ivandrofly, line 535: Wrong base

PR #8121, opened 2024-04-02
Fixes wrong WinForm timer usage

ivandrofly:
Fixes wrong usage of WinForm timer.

WinForm timer are already single threaded which mean only the main-thread will process the events it doesn't matter if you
have N-timers, main-thread will have to switch between them to process the event. This is not to be confused with several threads handling the events.

Also the interval like 5 is invalid or will always be Floored up to 5.

see: https://learn.microsoft.com/en-us/dotnet/api/System.Windows.Forms.Timer?view=netframework-4.8

### DIFF
--- a/src/ui/Controls/SubtitleListView.cs
+++ b/src/ui/Controls/SubtitleListView.cs
@@ -81,8 +81,6 @@ namespace Nikse.SubtitleEdit.Controls
 
         private readonly List<SyntaxColorLineParameter> _syntaxColorList = new List<SyntaxColorLineParameter>();
         private readonly List<SetStartAndDurationParameter> _setStartAndDurationList = new List<SetStartAndDurationParameter>();
-        private static readonly object SyntaxColorListLock = new object();
-        private static readonly object SetStartTimeAndDurationLock = new object();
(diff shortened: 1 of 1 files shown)


src/ui/Controls/SubtitleListView.cs
```diff
@@ -367,10 +365,10 @@ public SubtitleListView()
                 ShowGapColumn(LanguageSettings.Current.General.Gap);
             }
 
-            _syntaxColorLineTimer = new Timer { Interval = 41 };
+            _syntaxColorLineTimer = new Timer { Interval = 55 };
             _syntaxColorLineTimer.Tick += SyntaxColorLineTimerTick;
 
-            _setStartAndDurationTimer = new Timer { Interval = 3 };
```
ivandrofly, line 373: Timer component is single-threaded, and is limited to an accuracy of 55 milliseconds

See: https://learn.microsoft.com/en-us/dotnet/api/System.Windows.Forms.Timer?view=netframework-4.8

niksedk:
30ms works perfectly fine here....

It might be possible that just one timer would have less overhead, so all timers merged to one might give performance benefits. Needs testing good though...

PR #8123, opened 2024-04-02
Fixes #8122

### LINKED ISSUE #8122: Nikse Combobox: Dropdown closes too soon
If you click to show the drop down and let say you scrolling to find a item in middle of the
list then accidentally move the mouse out of the area of the drop down, the dropdown will close
and you will have to re-do the whole operation again.

Maybe give 2/3 seconds before closing to give the time to return/re-enter the area again?

Note: Things to consider, always close the drop-down whenever a click happen mouse-click happen outside
 the area ignoring the close delay mentioned above.

ivandrofly:
See issue #8122 for context

### DIFF
--- a/src/ui/Controls/NikseComboBox.cs
+++ b/src/ui/Controls/NikseComboBox.cs
@@ -615,29 +615,13 @@ namespace Nikse.SubtitleEdit.Controls
             BackColorDisabled = Color.FromArgb(240, 240, 240);
             BackColor = SystemColors.Window;
 
-            _mouseLeaveTimer = new Timer();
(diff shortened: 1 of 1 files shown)


ivandrofly:
Ready for review @niksedk

niksedk:
Hm, seems to behave a bit weird... closes when over original box, and pergaps more

ivandrofly:
> Hm, seems to behave a bit weird... closes when over original box, and pergaps more

Target combobox bound wasn't included on purpose - maybe we should just leave the dropdown open until the user set the focus to other control? That is the also the behavior in notepad++

PR #8132, opened 2024-04-04
Move MKV Bluray parsing from BinEdit.cs into libse for use by third-party code.

davepusey:
I'm currently writing a small utility for myself that counts and reports how many subtitles (forced and total) are present in each track of a .mkv file that was created using MakeMKV.

Some of the MKV Bluray parsing code required to do this is locked away inside the UI project, and at present I have to duplicate it into my project to be able to use it. Whilst this does work, I now need to monitor this repo for any potential changes to that code to then manually carry over.

This change extracts the relevant section of the code from BinEdit.cs and moves it into libse, enabling direct use by third-party code via libse.

### DIFF
--- a/src/libse/BluRaySup/BluRaySupParser.cs
+++ b/src/libse/BluRaySup/BluRaySupParser.cs
@@ -18,6 +18,7 @@
  */
 
 using Nikse.SubtitleEdit.Core.Common;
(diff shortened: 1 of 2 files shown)


niksedk:
The package [libse 4.0.5](https://www.nuget.org/packages/libse/4.0.5) was recently published on NuGet Gallery by niksedk.

PR #8156, opened 2024-04-09
Add grammar check functionality to subtitles

ivandrofly:
A new feature has been implemented that adds a grammar check functionality in the UI under the new 'Grammar Check' menu item. This leverages the LanguageToolService to check the grammar of selected subtitles, notifying the user about any potential errors or confirm that the check has passed successfully.

image

Hi Nikse, is this something you are interested in so I can continue with the dev @niksedk?
The idea is to have it for now to work only for selected line from main-listview but since it also provide fix-suggestion maybe it can be other version of spell check.

Couple of thing I would like to implement on this current pull request
- Display the offset and length per mistakes found
- If possible highlight found mistakes in textbox (if it's rich-text-box color the offset+length range to red)
- Use language detected from subtitle. (**auto** query param can give false positive if text is too short)
- Localization

If you can also give it a try and provide inputs :)

Steps to setup docker LanguageTool docker image (link below)
https://github.com/meyayl/docker-languagetool

### DIFF
--- a/src/ui/Forms/Main.cs
+++ b/src/ui/Forms/Main.cs
@@ -47,6 +47,7 @@ using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Forms;
+using Nikse.SubtitleEdit.Core.Grammar;
 using Nikse.SubtitleEdit.Forms.Tts;
 using static System.Windows.Forms.VisualStyles.VisualStyleElement.TrackBar;
 using CheckForUpdatesHelper = Nikse.SubtitleEdit.Logic.CheckForUpdatesHelper;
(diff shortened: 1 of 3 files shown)


src/ui/Forms/Main.cs
```diff
@@ -652,6 +653,21 @@ public Main()
...
 
                 ListViewHelper.RestoreListViewDisplayIndices(SubtitleListview1);
+
+                toolStripMenuItemGrammarCheck.Click += async (sender, args) =>
+                {
+                    var langaugeToolClient = new LanguageToolService();
+                    var isServiceAvailable = await langaugeToolClient.IsAvailableAsync();
+                    if (!isServiceAvailable)
+                    {
+                        MessageBox.Show("Service not availble", MessageBoxIcon.Exclamation);
```
ivandrofly, line 663: Ups typo :)

PR #8161, opened 2024-04-10
Fix D-Cinema SMPTE 2014/png

ischeriad:
* XML file generated was not valid. (It looks like it was previously directly adapted from D-Cinema InterOp.)
  * Fixed timecode format for TimeIn/TimeOut. SMPTE DCP needs frames, not milliseconds.
  * Image file names are now UUIDs.

I verified this format against easyDCP Creator+ 4.2.0 (current version).

There are still some things that could be improved:
  * It would be nice to integrate this export format with the "D-Cinema SMPTE 2014 properties..." dialog for things like movie title, reel number, language etc.
  * The frame rate should be taken from the project, or the at least comboBoxFrameRate should have the project framerate preselected. At the moment, the timecode list does not reflect the selected framerate in the PNG export dialog. This can lead to mistakes, if the project is set to e.g. 25, and the export dialog is set to 24, there can be illegal timecodes like xx:xx:xx:24.


niksedk:
Thx :)

The frame rate issue would be nice to get fixed too ;)

PR #8183, opened 2024-04-14
TTS on non Windows systems? (Fix build on Linux)

Gzushgshsh:
System.Speech.Synthesis is only available on Windows, I couldn't think of anything better than disabling it on other systems ;-;


niksedk:
Could it be done in TextToSpeech.cs without disabling text to speech entirely?

Gzushgshsh:
If you isolate all the code that depends on System.Speech.Synthesis, there should be no problems at runtime or in compilation

PR #8303, opened 2024-05-03
Implement IDisposable in translators and properly dispose HttpClient …

ivandrofly:
…instances

Several classes, particularly the ones related to translation services, have been modified to implement IDisposable interface. Additionally, all instances of HttpClient are being disposed properly now using "using" keyword. This will help ensure that system resources are promptly released when they're no longer required, potentially improving application performance and stability.

### DIFF
--- a/src/libse/AutoTranslate/GoogleTranslateV2.cs
+++ b/src/libse/AutoTranslate/GoogleTranslateV2.cs
@@ -17,7 +17,7 @@ namespace Nikse.SubtitleEdit.Core.AutoTranslate
     /// <summary>
     /// Google translate via Google Cloud V2 API - see https://cloud.google.com/translate/
     /// </summary>
-    public class GoogleTranslateV2 : IAutoTranslator
+    public class GoogleTranslateV2 : IAutoTranslator, IDisposable
     {
         private string _apiKey;
         private IDownloader _httpClient;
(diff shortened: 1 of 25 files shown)


niksedk:
Nice :)

niksedk:
Hm, could this have caused the "Purfview faster whisper" model download to fail? There seems to be a problem...

niksedk:
Message: DownloadAsync failed - https://huggingface.co/Systran/faster-whisper-tiny.en/resolve/main/config.json

System.InvalidOperationException
Message: This instance has already started one or more requests. Properties can only be modified before sending the first request.
StackTrace:    at System.Net.Http.HttpClient.CheckDisposedOrStarted()
   at System.Net.Http.HttpClient.set_Timeout(TimeSpan value)
   at Nikse.SubtitleEdit.Core.Http.HttpClientDownloader.<DownloadAsync>d__9.MoveNext() in C:\git\subtitleedit\src\libse\Http\HttpClientDownloader.cs:line 41

ivandrofly:
Taking a look

ivandrofly:
Is any progress done here https://github.com/SubtitleEdit/subtitleedit/commit/4a02923c8024210ec972a257e71be6e1c0d845e3

ivandrofly:
> Message: This instance has already started one or more requests. Properties can only be modified before sending the first request.

Seems like it's not possible to change HttpClient it's time out once you start using it. An easy fix would have been setting the time-out property at the constructor instead of `HttpClientDownloader.DownloadAsync` method which is invoked several on the same instance.

PR #8316, opened 2024-05-05
Feature/ifo parser

ivandrofly:
Three commits have been described.

 Commit 1: A logic condition within the method in the IfoParser class has been revisited to improve readability and ensure correct order of execution.

Commit 2: Enhancements have been made to the GetAllLanguages function to take into account the count of three different lists when computing a minCount variable, likely to prevent out of range errors.

Commit 3: Further optimizations were made to GetAllLanguages method where bounds checking is now outside the loop which may potentially improve performance. The minimum count between two lists is computed prior to entering the loop which simplifies the logic.

### DIFF
--- a/src/libse/Common/IfoParser.cs
+++ b/src/libse/Common/IfoParser.cs
@@ -39,37 +39,39 @@ namespace Nikse.SubtitleEdit.Core.Common
             public List<string> GetAllLanguages()
             {
                 var list = new List<string>();
-                for (int i = 0; i < Subtitles.Count; i++)
+                var minCount = Math.Min(SubtitleIDs.Count, SubtitleTypes.Count);
+                for (int i = 0; i < minCount; i++)
                 {
-                    if (i < SubtitleIDs.Count && i < SubtitleTypes.Count)
+                    var ids = SubtitleIDs[i].Split(',');
+                    var types = SubtitleTypes[i].Split(',');
+                    if (ids.Length == 2 && ids[0].Trim() == ids[1].Trim() || ids.Length == 3 && ids[0].Trim() == ids[1].Trim() && ids[1].Trim() == ids[2].Trim())
                     {
-                        var ids = SubtitleIDs[i].Split(',');
-                        var types = SubtitleTypes[i].Split(',');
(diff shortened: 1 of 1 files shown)


src/libse/Common/IfoParser.cs
```diff
@@ -39,37 +39,39 @@ public class VtsVobs
...
+                for (int i = 0; i < minCount; i++)
                 {
-                    if (i < SubtitleIDs.Count && i < SubtitleTypes.Count)
+                    var ids = SubtitleIDs[i].Split(',');
+                    var types = SubtitleTypes[i].Split(',');
+                    if ((ids.Length == 2 && ids[0].Trim() == ids[1].Trim()) || (ids.Length == 3 && ids[0].Trim() == ids[1].Trim() && ids[1].Trim() == ids[2].Trim()))
                     {
-                        var ids = SubtitleIDs[i].Split(',');
-                        var types = SubtitleTypes[i].Split(',');
-                        if (ids.Length == 2 && ids[0].Trim() == ids[1].Trim() || ids.Length == 3 && ids[0].Trim() == ids[1].Trim() && ids[1].Trim() == ids[2].Trim())
```
ivandrofly, line 48: Bug in this line, the statement should've been grouped with `\( \)` because the "or" performed for ` ids[0].Trim() == ids[1].Trim() || ids.Length == 3`

The correct is `(ids.Length == 2 && ids[0].Trim() == ids[1].Trim()) || (ids.Length == 3 && ids[0].Trim() == ids[1].Trim() && ids[1].Trim() == ids[2].Trim())`

src/libse/Common/IfoParser.cs
```diff
@@ -39,37 +39,39 @@ public class VtsVobs
             public List<string> GetAllLanguages()
             {
                 var list = new List<string>();
-                for (int i = 0; i < Subtitles.Count; i++)
+                var minCount = Math.Min(Subtitles.Count, Math.Min(SubtitleIDs.Count, SubtitleTypes.Count));
```
ivandrofly, line 42: The loop only need to go to min between the three Subtitles,  SubtitleIDs and SubtitleTypes

PR #8338, opened 2024-05-08
Feature: Hot keys to the Ocr-Character-Inspect-Form

trienow:
This pull request allows the user to not have to rely on their mouse as much when mapping a bitmap to a character.

It definitely helped me speed up the OCR-process. It might be useful for others, but that is up to you.

VOB OCR Character Edit

In the list box:
- `Enter` in the list box puts focus onto the text box
- `Ctrl+Up`/`Ctrl+Down` jumps to the next missing match

In the text box:
- `Enter` in the text box adds or updates the character
- `Up`/`Down` navigates up and down in the list box
- `Ctrl+Up`/`Ctrl+Down` jumps to the next missing match

To be honest, I wasn't sure how to make it clear to the user, that these hot keys exist. I was going to add a label with a hint, but I didn't want to do too much and it not being wanted to begin with. Feedback is welcome.

### DIFF
--- a/LanguageBaseEnglish.xml
+++ b/LanguageBaseEnglish.xml
@@ -3353,6 +3353,13 @@ Keep changes?</KeepChangesMessage>
     <InspectItems>Inspect items</InspectItems>
     <AddBetterMatch>Add better match</AddBetterMatch>
     <Add>Add</Add>
+    <AddBetterMultiMatch>Add better multi match</AddBetterMultiMatch>
+    <AddOrUpdateMatch>Add or update match</AddOrUpdateMatch>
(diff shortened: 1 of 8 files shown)


niksedk:
You could a context menu strip for the list box and one for the text box.

"Enter/up/down" can probably not be set as a shortcut key, so just set in the text of the context menu - like "Update character (Enter)".

trienow:
The context menu suggestion was a good idea. The `ShortcutKeyDisplayString` is free text, so I was able to add the "forbidden" keys there too.

Showing off the context menus in the character inspect form

PR #8351, opened 2024-05-10
Added binary format check to CsvNuendo

ivandrofly:
A new method, HasKnownBinaryFormatExtension, was added to the CsvNuendo class. This method checks whether a file has a known binary format extension. Then, it's used in the IsMine method to prevent wrongly identifying binary files as CsvNuendo format.


ivandrofly:
Csv Nuedo hangs for too long if a Cavena890 format is dropped (which is a binary subtitle format)

image

Sample
[Heart of the Hunter 2024-hi_engcp.zip](https://github.com/SubtitleEdit/subtitleedit/files/15279270/Heart.of.the.Hunter.2024-hi_engcp.zip)

PR #8409, opened 2024-05-21
Update ro-RO.xml

MediaExpres:
Added the <WordLists>Lista de cuvinte...</WordLists> translation for Romanian. Since it was missing, the English one "Word lists..." was shown by default on the Romanian language user interface.


MediaExpres:
> I think the ro.RO file is missing a lot of tags...

Thanks for pointing it out, we got a client with subtitles to translate and discovered your nice work with Subtitle Edit, and while using it we bumped into those small problems. I can not make promises now, but will keep this on the list for help and improvements. 

Keep in touch,
Mihai

PR #8448, opened 2024-05-29
Remove "'" prefix before "REG QUERY" command

ivandrofly:
Typo?

### DIFF
--- a/build.bat
+++ b/build.bat
@@ -263,7 +263,7 @@ FOR %%G IN (7za.exe) DO (SET "SEVENZIP_PATH=%%~$PATH:G")
 IF EXIST "%SEVENZIP_PATH%" (SET "SEVENZIP=%SEVENZIP_PATH%" & EXIT /B)
 
 FOR /F "tokens=2*" %%A IN (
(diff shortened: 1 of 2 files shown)


niksedk:
I'm not really sure what this `'` means... but the .bat files does not work if I remove it...

ivandrofly:
Nvm it's not a typo, it's a opening that is closed later in the line bellow!

```bat
'REG QUERY "HKLM\SOFTWARE\7-Zip" /v "Path" 2^>NUL ^|^|
REG QUERY "HKLM\SOFTWARE\Wow6432Node\7-Zip" /v "Path" 2^>NUL'
```

One thing that I find weird is the escape of | (^|) 

Normally without the caret symbol (^) the whole command is just saying:
Try the first one `REG QUERY "HKLM\SOFTWARE\7-Zip" /v "Path" 2^>NUL` if fails then try the second `REG QUERY "HKLM\SOFTWARE\Wow6432Node\7-Zip" /v "Path" 2^>NUL`

The escape is normally used when eching e.g: `echo foo ^| bar` => foo | bar

PR #8484, opened 2024-06-05
Feature/http downloader

ivandrofly:
Allow reusing same httpclient instance to avoid port exhaustion (mainly inside for loop)
- Rename `MakeHttpClient` => `CreateProxiedHttpClient` (as this method alreays create httpclient with proxy)
- Now there are two overloads for *CreateProxiedHttpClient*
    - CreateProxiedHttpClient() => Creates client with infinity timeout
    - CreateProxiedHttpClient(Timeout) => Take timeout

Related to https://github.com/SubtitleEdit/subtitleedit/pull/8303

### DIFF
--- a/src/ui/Forms/AudioToText/WhisperModelDownload.cs
+++ b/src/ui/Forms/AudioToText/WhisperModelDownload.cs
@@ -108,9 +108,10 @@ namespace Nikse.SubtitleEdit.Forms.AudioToText
                     var pct = (int)Math.Round(progress * 100.0, MidpointRounding.AwayFromZero);
                     labelPleaseWait.Text = LanguageSettings.Current.General.PleaseWait + "  " + pct + "%";
                 });
-                foreach (var url in LastDownloadedModel.Urls)
+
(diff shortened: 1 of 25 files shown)


src/ui/Forms/AudioToText/WhisperModelDownload.cs
```diff
@@ -108,9 +108,10 @@ private void MultiFileDownload()
                     var pct = (int)Math.Round(progress * 100.0, MidpointRounding.AwayFromZero);
                     labelPleaseWait.Text = LanguageSettings.Current.General.PleaseWait + "  " + pct + "%";
                 });
-                foreach (var url in LastDownloadedModel.Urls)
+
+                using (var httpClient = DownloaderFactory.CreateProxiedHttpClient())
                 {
-                    using (var httpClient = DownloaderFactory.MakeHttpClient())
```
ivandrofly, line 113: This line was the issue* as it's creating a new instance of http-client for every url!

PR #8501, opened 2024-06-07
Optimize comboBox item selection

ivandrofly:
The previous implementation of the comboBox resulted in the processing of all items every time. This commit optimizes this process by only dealing with visible items, thus improving efficiency. Calculations for the visible range are now included, starting from the top item and only iterating over visible items.

### DIFF
--- a/src/ui/Controls/NikseComboBox.cs
+++ b/src/ui/Controls/NikseComboBox.cs
@@ -1029,8 +1029,11 @@ namespace Nikse.SubtitleEdit.Controls
                     return;
                 }
 
-                var cachedCount = _listView.Items.Count;
-                for (var i = 0; i < cachedCount; i++)
+                var topItem = _listView.TopItem;
+                // calculate visible range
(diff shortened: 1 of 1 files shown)


src/ui/Controls/NikseComboBox.cs
```diff
@@ -1029,8 +1029,11 @@ private void EnsureListViewInitialized()
                     return;
                 }
 
-                var cachedCount = _listView.Items.Count;
-                for (var i = 0; i < cachedCount; i++)
+                var topItem = _listView.TopItem;
+                // calculate visible range
+                var count = Math.Min(topItem.Index + _listView.Bounds.Height / topItem.Bounds.Height, _listView.Items.Count);
+                
+                for (var i = topItem.Index; i < count; i++)
```
ivandrofly, line 1036: For example in subtitle format combobox, instead of looping through 300+ item we can just do it to 14/15 (visible items)

PR #8523, opened 2024-06-12
Update csproj files to the new Project Sdk format

ivandrofly:
Simplified and cleaned up the structure of multiple csproj files. This includes removing unnecessary tags and attributes, and adopting the new Project Sdk format. This change should have no functional impact but makes the project files streamlined and easier to read.

- Will make it easier to switch to the new .NETs

Note: The unit-test project is the only one complaining due to the unit-test runner having dependency on Visual Studio (it was left untouched for now)


niksedk:
Cool, that's surely a lot simpler :)

ivandrofly:
> Cool, that's surely a lot simpler :)

For sure :)

PR #8549, opened 2024-06-19
Refactor AutoTranslate code to use async/await

ivandrofly:
Removed the BackgroundWorker in the `nikseComboBoxEngine_SelectedIndexChanged` method, replacing it with async/await for better readability and performance. Extracted a `DownloadOllamaModelsAsync` method to download models, which also replaced the synchronous approach in `UpdateLocalModelsToolStripMenuItem_Click`.

Related to https://github.com/SubtitleEdit/subtitleedit/commit/c695ed7b351bb3b434f329b14fa81ea02fa69eb2

### DIFF
--- a/src/ui/Forms/Translate/AutoTranslate.cs
+++ b/src/ui/Forms/Translate/AutoTranslate.cs
@@ -6,7 +6,6 @@ using Nikse.SubtitleEdit.Forms.Options;
 using Nikse.SubtitleEdit.Logic;
 using System;
 using System.Collections.Generic;
-using System.ComponentModel;
 using System.Diagnostics;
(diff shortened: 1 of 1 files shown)


niksedk:
The UI was a bit better before.... can you fix this, so it does not block the UI?

ivandrofly:
> The UI was a bit better before.... can you fix this, so it does not block the UI?

I will take a look

ivandrofly:
> The UI was a bit better before.... can you fix this, so it does not block the UI?

Maybe you also mean not to disable the **combobox** while trying to get the models from ollama?
It was disable to prevent spamming with new request.

PR #8552, opened 2024-06-20
Implement async actions in spell checker

ivandrofly:
Refactored the spell-check actions and related functions in the SubtitleEdit application to be asynchronous, allowing for smoother user experience and improved application performance. All actions related to spelling correction have been turned into asynchronous tasks, leading to a more responsive user interface as now operations won't block the UI thread. Also, converted synchronous Hunspell suggestions to an async operation with timeout, reducing the risk for application freezing in case of delay in suggestions retrieval.

In short words, use thread-pool instead of creating a new thread every-time you are requesting word suggestion
as already known created new thread is very expensive!

### DIFF
--- a/src/ui/Forms/SpellCheck.cs
+++ b/src/ui/Forms/SpellCheck.cs
@@ -650,7 +650,7 @@ namespace Nikse.SubtitleEdit.Forms
         {
             try
             {
-                return await Task.Run(() => _hunspell.Suggest(word), NewAutoCancelTokenAfter3Sec().Token);
+                return await Task.Run(() => _hunspell.Suggest(word), NewAutoCancelTokenAfter3Sec().Token).ConfigureAwait(false);
             }
             catch (Exception e)
             {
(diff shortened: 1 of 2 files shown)


src/ui/Forms/SpellCheck.cs
```diff
@@ -659,40 +646,39 @@ public bool DoSpell(string word)
...
+        private async Task<List<string>> DoSuggestAsync(string word)
         {
-            var parameter = new SuggestionParameter(word, _hunspell);
-            var suggestThread = new System.Threading.Thread(DoWork);
-            suggestThread.Start(parameter);
-            suggestThread.Join(3000); // wait max 3 seconds
-            if (!parameter.Success)
+            try
+            {
+                return await Task.Run(() => _hunspell.Suggest(word), CancellationToken.None).ConfigureAwait(false);
```
ivandrofly, line 658: It's safe to resume on a thread-pool thread here since no interaction with the UI was done from this point!

PR #8588, opened 2024-06-29
Resources: Update

ivandrofly:
Update resource

### DIFF
--- a/src/ui/Forms/Ocr/GetTesseract302Dictionaries.cs
+++ b/src/ui/Forms/Ocr/GetTesseract302Dictionaries.cs
@@ -61,15 +61,12 @@ namespace Nikse.SubtitleEdit.Forms.Ocr
             {
                 comboBoxDictionaries.Items.Clear();
                 var doc = new XmlDocument();
-                using (var rdr = new StreamReader(stream))
-                using (var zip = new GZipStream(rdr.BaseStream, CompressionMode.Decompress))
+                using (var zip = new GZipStream(stream, CompressionMode.Decompress))
+                using (var reader = XmlReader.Create(zip, new XmlReaderSettings { IgnoreProcessingInstructions = true }))
                 {
-                    var data = new byte[195000];
-                    var bytesRead = zip.Read(data, 0, data.Length);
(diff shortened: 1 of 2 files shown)


src/ui/SubtitleEdit.csproj
```diff
@@ -242,15 +242,6 @@
...
     <Content Include="Icons\WaveformToggle.png" />
-    <Content Include="Resources\TesseractDictionaries.xml">
-      <SubType>Designer</SubType>
-    </Content>
-    <Content Include="Resources\HunspellBackupDictionaries.xml">
-      <SubType>Designer</SubType>
-    </Content>
-    <Content Include="Resources\HunspellDictionaries.xml">
-      <SubType>Designer</SubType>
-    </Content>
```
ivandrofly, line 253: These files are never part of the build, they are just used to update the `HunspellBackupDictionaries.xml.gz` and `HunspellDictionaries.xml` which are used as EmbeddedResource, so can be removed from the csproj as content.

src/ui/Forms/Ocr/GetTesseract302Dictionaries.cs
```diff
@@ -61,15 +61,12 @@ private void LoadDictionaryList(string xmlResourceName)
...
                 comboBoxDictionaries.Items.Clear();
                 var doc = new XmlDocument();
-                using (var rdr = new StreamReader(stream))
-                using (var zip = new GZipStream(rdr.BaseStream, CompressionMode.Decompress))
+                using (var zip = new GZipStream(stream, CompressionMode.Decompress))
+                using (var reader = XmlReader.Create(zip, new XmlReaderSettings { IgnoreProcessingInstructions = true }))
                 {
-                    var data = new byte[195000];
-                    var bytesRead = zip.Read(data, 0, data.Length);
-                    var s = System.Text.Encoding.UTF8.GetString(data, 0, bytesRead).Trim();
```
ivandrofly, line 69: Avoid cumbersome reading of the gzip file by using raw byte and also wrapping with `StreamReader`
This same reading technic was also used in `GetDictionary.cs`

PR #8621, opened 2024-07-12
Update async calls to use ConfigureAwait(false)

ivandrofly:
The code updates are primarily related to the usage of async calls across multiple translation services files. The async calls have been adjusted to use ConfigureAwait(false) for performance reasons and to avoid deadlocks. Changes are also made to several methods to make them asynchronously and thus improve the responsiveness of the operations.


niksedk:
Have you test all this?
I think this older http client is a bit strange sometimes...

ivandrofly:
> Have you test all this?

Tested with google api

ivandrofly:
> I think this older http client is a bit strange sometimes...

Old httpclient?

PR #8635, opened 2024-07-14
Refactor code to use GetCharactersPerSecond method in Paragraph class

ivandrofly:
Removed GetCharactersPerSecond method from the Utilities class and refactored all its references to use the equivalent method newly added in the Paragraph class. This refactoring resulted in cleaner, more object-oriented code and consolidated related logic in the Paragraph class.


ivandrofly:
The failing check is not related to this PR

See: https://github.com/SubtitleEdit/subtitleedit/commit/d30f7fafb0afea6d6b6270ed47bde0fb970ee335

PR #8669, opened 2024-07-25
Refactor FixCasingAfterTitles to use class fields.

ivandrofly:
Extracted the arrays `_titles` and `_notChangeWords` as readonly class fields to reduce redundancy and improve readability. Adjusted the associated logic to reference these fields instead of creating new arrays within the method.

### DIFF
--- a/src/libse/Common/FixCasing.cs
+++ b/src/libse/Common/FixCasing.cs
@@ -159,24 +159,25 @@ namespace Nikse.SubtitleEdit.Core.Common
             return text;
         }
 
+        private readonly string[] _titles = { "Mrs.", "Miss.", "Mr.", "Ms.", "Dr." };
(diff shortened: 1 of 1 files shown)


src/libse/Common/FixCasing.cs
```diff
@@ -159,24 +159,25 @@ private static string FixEnglishAloneILowerToUpper(string input)
             return text;
         }
 
+        private readonly string[] _titles = { "Mrs.", "Miss.", "Mr.", "Ms.", "Dr." };
+        private readonly string[] _notChangeWords = { "does", "has", "will", "is", "and", "for", "but", "or", "of" };
```
ivandrofly, line 163: Moved outside 'cause `FixCasingAfterTitles` is being callow inside a foreach-paragraph
which would alloc `n*_titles + n*_notChangeWords`

PR #8752, opened 2024-08-20
Fixes edge cases for parsing milliseconds.

PoneyClairDeLune:
Discovered by @Jirehlov, this PR aims to fix edge cases where parsing the millisecond portion of the time codes doesn't have three digits.

Cannot install Visual Studio Community on Linux, and cannot compile SubtitleEdit on Linux due to a lack of WinForm, so this PR is not tested.


niksedk:
Thx for the info/fix :)

PR #8821, opened 2024-09-13
Improves/Add documentation

ivandrofly:
Enhanced the `FileUtil` and HtmlUtil classes by adding documentation


ivandrofly:
Since libse is packable it's good to have public methods documented

PR #8853, opened 2024-09-24
Use Tesseract 5.5.0 api instead of command line (pass bmp in memory, not in files)

ArturAlekseev:
Change TesseractDownloadUrl in DownloadTesseract5.cs to: "http://www.arturalekseev.ru/Files/tesseract550.zip" before compile (move it to your/github location).
Compiled build for testing: https://disk.yandex.ru/d/UGOQdjV_080Ljw
Source of tesseractapi.dll: https://github.com/ArturAlekseev/tesseractapi


ArturAlekseev:
I'v updated zip file with tesseract (not all dlls were there). It worked for me because I had tesseract 5.4.1 installed. Now I don't and you don't need to have it installed.

ArturAlekseev:
I'v updated it with tesseract 5.5.0.

PR #8888, opened 2024-10-07
QoL additions for Waveform and others

### LINKED ISSUE #8779: [QoL] - Make splitting subtitles not split the text as an option
As title says.

When you split a subtitle (in Spectrogram) it splits the text in the selected subtitle as well.
Would love to know if it is possible to make the subtitles not split as an option
The result would be, that upon splitting, both subtitles would have the same text

Reason:
I usually split subtitles, because its an easy way to add another subtitle right at the same spot for formatting purposes. (karaoke text as an example), so this would be a nice QoL change since i dont have to copy and paste the original subtitle into every splitted subtitle.

Current release:
```
This is a text subtitle, it is long because there are a lot of words
(turns into)
This is a text subtitle, it is long 
because there are a lot of words
```

My suggestion:
```
This is a text subtitle, it is long because there are a lot of words
(turns into)
This is a text subtitle, it is long because there are a lot of words
This is a text subtitle, it is long because there are a lot of words
```

MitchSirco:
I've gone ahead and implemented the following things:
* visible \k and \t indicator in Waveform
* ability to move \k around in the waveform ui
* bumped up subtitle cards render limit from 99 to 199 and reduced startTime of that (so that the first 15 seconds wouldn't have too much real estate)
* added a toggleable option for duplicating text when splitting a subtitle card

Original Comment:
this would fix #8779 and is implemented as a togglable option
There might be some edgecases that i haven't covered yet.
Im also noticing, that it only works when setting the "Unbreak Subtitles shorter than" option to a high number


MitchSirco:
Preview of the "moving the \k point around"
@niksedk

https://github.com/user-attachments/assets/b71777b6-b483-481b-9c41-eb0d6d45f340

PR #8896, opened 2024-10-10
Feature/minor updates

ivandrofly:


ivandrofly:
Swap async <=> void position (which was causing build issue)

PR #8909, opened 2024-10-15
Refactor and test StartsAndEndsWithTag method

ivandrofly:
Simplified the logic in StartsAndEndsWithTag by using arrays for prefix and suffix formats. Added a unit test to ensure the method handles various tag formats correctly.

### DIFF
--- a/src/libse/Common/Utilities.cs
+++ b/src/libse/Common/Utilities.cs
@@ -1215,40 +1215,26 @@ namespace Nikse.SubtitleEdit.Core.Common
                 return false;
             }
 
-            if (!text.Contains(startTag) || !text.Contains(endTag))
+            text = text.FixExtraSpaces();
+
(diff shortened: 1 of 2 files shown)


src/libse/Common/Utilities.cs
```diff
@@ -1215,40 +1215,26 @@ public static bool StartsAndEndsWithTag(string text, string startTag, string end
...
-
-            while (text.Contains("  "))
-            {
-                text = text.Replace("  ", " ");
-            }
-
-            var s1 = "- " + startTag;
-            var s2 = "-" + startTag;
-            var s3 = "- ..." + startTag;
-            var s4 = "- " + startTag + "..."; // - <i>...
```
ivandrofly, line 1231: this was always false too... as [line 1228](https://github.com/SubtitleEdit/subtitleedit/pull/8909/files#diff-895f77470248dd0801fef50338bcd93ce27acfb2ad5de1da85485cd599e1c2cfL1228) would always match first

PR #8950, opened 2024-10-27
Remove packages.config from UI project

ivandrofly:
Remove packages.config as the project is already using PackageReference


ivandrofly:
For some reason appveyor is still looking for `src/ui/packages.config`....

PR #8961, opened 2024-10-31
DIscord RPC

leexey:
It basically shows Subtitle Edit on Discord + the name of the subtitle

image
image

### DIFF
--- a/src/ui/Program.cs
+++ b/src/ui/Program.cs
@@ -3,8 +3,6 @@ using System;
 using System.Diagnostics;
 using System.Threading;
 using System.Windows.Forms;
-using DiscordRPC;
-using System.Timers;
 
 namespace Nikse.SubtitleEdit
 {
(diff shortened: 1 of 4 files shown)


src/ui/Program.cs
```diff
@@ -3,33 +3,70 @@
 using System.Diagnostics;
 using System.Threading;
 using System.Windows.Forms;
+using DiscordRPC;
+using System.Timers;
```
ivandrofly, line 7: Is this been used?

src/ui/Program.cs
```diff
@@ -3,33 +3,70 @@
...
+
+                    // Extract name from the window
+                    var fileName = windowTitle.Contains(" - ") ? windowTitle.Split(new string[] { " - " }, StringSplitOptions.None)[0] : "No file opened";
+
+                    discordRpcMain.UpdatePresence($"Editing: {fileName}");
+                }
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine("Error updating Discord RPC: " + ex.Message);
```
ivandrofly, line 64: Maybe on recursive fail just disable & dispose the timer?

leexey: god i completely forgot about that sorry

src/ui/Program.cs
leexey: was testing other things and committing with that... it's already been removed 👍 thanks for noticing

PR #8987, opened 2024-11-09
Refactor: Consolidate word boundary characters

ivandrofly:
Updated the OCR fix engine to use a single constant for word boundary characters. This simplifies the code and improves maintainability by avoiding repetition of the same character set in different places.

### DIFF
--- a/src/ui/Logic/Ocr/OcrFixEngine.cs
+++ b/src/ui/Logic/Ocr/OcrFixEngine.cs
@@ -523,9 +523,10 @@ namespace Nikse.SubtitleEdit.Logic.Ocr
             sb = new StringBuilder();
             var word = new StringBuilder();
             string lastWord = null;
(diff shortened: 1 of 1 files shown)


src/ui/Logic/Ocr/OcrFixEngine.cs
```diff
@@ -523,9 +523,10 @@ public string FixOcrErrors(string input, Subtitle subtitle, int index, string pr
             sb = new StringBuilder();
             var word = new StringBuilder();
             string lastWord = null;
+            const string wordBoundaryChars = "\r\n ";
             for (var i = 0; i < text.Length; i++)
             {
-                if (text[i] != '.' && "\r\n ".Contains(text[i]))
+                if (wordBoundaryChars.Contains(text[i]))
```
ivandrofly, line 529: `text[i] != '.'` always true if `"\r\n ".Contains(text[i])` is true

PR #9043, opened 2024-11-27
Remove redundant regular expressions loading loop

ivandrofly:
Prevent re-adding previously removed regex.


ivandrofly:
The 'removed' prefix is intended to indicate that these words should be excluded, following a similar approach used in the <blacklist> element of names.xml.

Is that correct?

PR #9119, opened 2024-12-21
Allow build/running on WSL/Mono

ivandrofly:
This commit ensures that the `System.Resources.Extensions` library is included in both the build process and the installer scripts to address potential runtime issues from the absence of this dependency. Specifically, the library is added to the files copied during the build (`build.bat` and `build_beta.bat`) and to the application's installer configuration to ensure it is properly distributed with the application. 

Related to
https://github.com/SubtitleEdit/subtitleedit/pull/9107


niksedk:
Would this make it possible to run the unit tests as a Github action (on Linux)?

ivandrofly:
> Would this make it possible to run the unit tests as a Github action (on Linux)?

I think so, haven't test it though.
As long as **mono** is included in **github action** Linux image I think it should work just fine

_I was able to run tests using Ubuntu  on WSL with mono.
There are couple test failing mostly due to trivial error e.g expecting \n instead of \r\n_

There are the failing tests:
image


Merry Christmas 🎄 and Happy New Year 🎆

/cc @niksedk  @diomed

diomed:
Happy New Year and blessed holidays and days to come to you as well Ivandro and Nikolaj!

ivandrofly:
Links to https://github.com/SubtitleEdit/subtitleedit/pull/9175

PR #9165, opened 2025-01-02
Feature/strippable

ivandrofly:


src/libse/Common/StrippableText.cs
```diff
@@ -4,8 +4,11 @@
 
 namespace Nikse.SubtitleEdit.Core.Common
 {
-    public class StrippableText
+    public struct StrippableText
```
ivandrofly, line 7: This change should bring huge improvement

ivandrofly:
What about this one? 

This should bring nice improvement as as StrippableText will not longer be class which means no allocation/garbage collector.

Also if you take a look on how strippable text is being used, it's mostly without a method body there is no much passing around the ref.

@niksedk

PR #9182, opened 2025-01-06
Feature/strippable text optimization

ivandrofly:
Faster and better algorithm for Strippable Text.

Less allocations


src/Tests/Logic/StrippableTextTest.cs
```diff
@@ -57,9 +57,9 @@ public void StrippableTextItalic3()
         public void StrippableTextFontDontTouch()
         {
             var st = new StrippableText("{MAN} Hi, how are you today!");
-            Assert.AreEqual("", st.Pre);
+            Assert.AreEqual("{", st.Pre);
```
ivandrofly, line 60: I think this test was wrong.

We should treat '{' as '[' When it hold hearing impaired word/sentence

PR #9220, opened 2025-01-14
Piper Speed and Voice Over Video

Chris-656:
Hi, i am fascinated from this tool. I am experimenting with piper text2speech. I found some minor issues with the automatic speed adaption using ffmpg.

I made some adaptions so that your tool can be used as semiautometd workflow for translating video with a sythetic voice.

image
- added 2 checkboxes (no translation at the moment) 
Changes:
- Adapt speed (ffmpeg): 
  here you can switch off the ffmpeg speed calculation. ffmpeg is not that good as piper can do. I also chenged the speed 
  adaption slower and also faster to a certain limit. 20% Max in both directions

- 2 pass speed correction (piper)
  Since piper can user length_scale parameter for slower or faster speed by using the pauses in between the words the results are much better. So in the first pass the length is calulated and in the second pass the correction is made to the wav output.

- use voice over
 This creates a new voice over channel by decompressing the original channel when the voice channel is active. This gives an automatic voice over where music or anything in the background survices when the voice over is silent.

I have not changed language files or  inputmasks at the moment. Maybe we cab do that together.

would be greatto get in contact
regards Chris


niksedk:
Hi @Chris-656,
some good ideas :)

Could you make two separate PRs?
One for the the voice over, and one for the adjust-speed-via-piper.

The adjust via speed could perhaps just always use piper first (only when using Piper) when the audio is tool fast, and then ffmpeg as a last resort?

Chris-656:
will separate as suggested.
regards Chris

PR #9241, opened 2025-01-18
separation between pt-pt and pt-br

Blackspirits:
I separated them because there are names of countries that we write differently


Blackspirits:
@niksedk  please accept this pull

niksedk:
Seems to work :)

image

Issue #9250, opened 2025-01-21
[Feature request] Add PaddleOCR standalone executable for simpler usage


timminator:
Hi there!
It's me again :-)

The last week while you (@niksedk) were working on implementing PaddleOCR into Subtitle Edit I worked on a new little site project of mine.
I was curious if I could create a standalone executable of paddleocr that could be run without setting up python and installing the required packages.
It took quite a bit of tinkering but I managed to get it working, my repository can be found [here](https://github.com/timminator/PaddleOCR-Standalone). I've created a CPU and GPU version.

It should work exactly like the command line version before.
I also fixed the bug in this release that prevented implementing the batch mode. :-)

So in theory Subtitle Edit could just prompt the user to download this version now.

And don't be confused when it takes a longer time to start when running it the first time. This is the normal behavior. Afterwards it runs normally.

We would only need to detect if an Nvidia GPU is present in the users system to decide which version should be used. I think we could do this by checking for [example](https://stackoverflow.com/questions/67504079/how-to-check-if-an-nvidia-gpu-is-available-on-my-system) if the command "nvidia-smi" exists.

Edit: The GPU version falls back to using the CPU if no CUDA device is present so it also works in this case, just for your info.

PR #9283, opened 2025-01-28
Add PaddleOCR-Standalone and Batch mode support

timminator:
This PR adds support for the PaddleOCR-Standalone version I have also made.

This enables the usage of PaddleOCR without requiring the user to install python, paddlepaddle and paddleocr. Instead it prompts the user to download the standalone version, similar to when using tesseract 5.5.0 for the first time.
The PR still detects if the user installed PaddleOCR beforehand and does not prompt the user in this case to download PaddleOCR again.
Furthermore it checks if the system supports the GPU version or not. If not, the CPU version is used instead.

This PR also adds one of two desired features mentioned in #9250. 

Edit: The second commit also adds Batch mode support when using PaddleOCR.

Fixes: #9250

### DIFF
--- a/src/ui/Forms/Ocr/DownloadPaddleOCR.Designer.cs
+++ b/src/ui/Forms/Ocr/DownloadPaddleOCR.Designer.cs
@@ -1,0 +1,78 @@
+﻿namespace Nikse.SubtitleEdit.Forms.Ocr
+{
+    sealed partial class DownloadPaddleOCR
+    {
+        /// <summary>
+        /// Required designer variable.
+        /// </summary>
+        private System.ComponentModel.IContainer components = null;
+
+        /// <summary>
+        /// Clean up any resources being used.
+        /// </summary>
+        /// <param name="disposing">true if managed resources should be disposed; otherwise, false.</param>
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing && (components != null))
+            {
+                components.Dispose();
+            }
+            base.Dispose(disposing);
+        }
+
+        #region Windows Form Designer generated code
+
+        /// <summary>
+        /// Required method for Designer support - do not modify
+        /// the contents of this method with the code editor.
+        /// </summary>
+        private void InitializeComponent()
+        {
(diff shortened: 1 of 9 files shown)


timminator:
Hi! @niksedk:
I think you pushed this commit in the wrong repo...
I am actually kinda puzzled that you could push a commit into my fork, but thats no problem.
I want to update my commit/add a commit and I am not sure if your added commit is now also in the Subtitle Edit repo. I saw that one change was also done in another commit there, but this change:

https://github.com/SubtitleEdit/subtitleedit/blob/a679c60b3e462f41655fd1046565aba246326c10/src/libse/SubtitleFormats/Pac.cs#L2772-L2773

is not in the main repo. Are you aware of that?
Before I update/add another commit to my branch I wanted to ask you this...

niksedk:
Yes, I did commit to the wrong repro... I did not know I could do that, LOL
The commit is also in the main SE branch.

I will try to look at your PR tomorrow.

Also, do you know what this is; https://github.com/PaddlePaddle/PaddleOCR/blob/main/deploy/cpp_infer/readme.md ?

timminator:
> I will try to look at your PR tomorrow.

That would be really cool. :-)

I've also added now batch mode support for PaddleOCR.
It is backwards compatible, so with the current version of paddleocr via pip (2.9.1) it is not active, but with the next version it will automatically activate it. With my standalone version it is used already.
I've also added a check that the paddleocr version has to be version 2.7 or greater due to us using the model PP-OCRv4 which is not supported on older versions. It prompts the user to upgrade paddleocr in this case.

I've had to change quite a few thing for this, but I tried to test everything to make sure that it works correctly.
I was not sure how to test the function DoBatch() - I think it is used when using SubtitleEdit from the command line? I've updated it to the best of my knowledge - but maybe you could test this one?

I'm quite happy how this turned out - I hope you like it aswell! :-)

timminator:
Thank you for your quick merge! I really appreciate it! :-)

niksedk:
Thank you for making PaddleOCR a lot better in SE :)

PR #9329, opened 2025-02-07
Add HunspellDictionaries.xml.gz as an embedded resource

ivandrofly:
Wasn't showing dictionaries from "Get Dictionaries"


niksedk:
Good catch :)

PR #9348, opened 2025-02-10
Refactor OCR capitalization logic and enhance ellipses handling.

ivandrofly:
Renamed and optimized `FixLowercaseIToUppercaseI` to `CapitalizeInitialI` for clarity and maintainability. Improved `HasSentenceEnding` to handle ellipses properly, preventing false positives. Added corresponding unit tests to ensure coverage and correctness.

### DIFF
--- a/src/Tests/Core/StringExtensionsTest.cs
+++ b/src/Tests/Core/StringExtensionsTest.cs
@@ -453,5 +453,12 @@ namespace Tests.Core
             Assert.IsTrue("foobar;".HasSentenceEnding(greekCultureTwoLetter));
         }
 
+        [TestMethod]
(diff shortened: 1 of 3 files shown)


ivandrofly:
Why are the `Dictionaries` items from Tests\Dictionaries stored as `Embedded Resources`

image


Are you aware that these files will never be copied to the Test output folder which will cause test dependent on the dictionary to fail? Was it done intentionally?

Ps: I currently have 30 failing unit-test running from latest change on main (without including the changes from this pull request)

ivandrofly:
See https://github.com/SubtitleEdit/subtitleedit/pull/9349

PR #9349, opened 2025-02-10
Fixes broken unit-tests

ivandrofly:
Well, I wonder why they are passing in app-veyor.. maybe due to cache settings?

### DIFF
--- a/src/Tests/Logic/Forms/RemoveTextForHearImpairedTest.cs
+++ b/src/Tests/Logic/Forms/RemoveTextForHearImpairedTest.cs
@@ -965,11 +965,10 @@ namespace Tests.Logic.Forms
             var target = GetRemoveTextForHiLib();
             target.Settings.RemoveIfAllUppercase = false;
             target.Settings.RemoveInterjections = false;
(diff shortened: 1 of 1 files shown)


src/Tests/Logic/Forms/RemoveTextForHearImpairedTest.cs
```diff
@@ -1322,7 +1322,7 @@ public void RemoveTextForHiInDialogue1()
         public void RemoveTextForHiInDialogue2()
         {
             var target = GetRemoveTextForHiLib();
-            target.Settings.RemoveTextBetweenBrackets = true;
+            target.Settings.RemoveTextBetweenSquares = true;
             string text = "-[Ronnie laughs]" + Environment.NewLine + "-[sighs] Wow, Ronnie.";
```
ivandrofly, line 1326: The mistake can clearly be noticed here... we are expecting to remove `[` but the settings was configured to `{`... why are the test passing? 😕

ivandrofly:
Any updates? There are still 30 failing test in the solution!

niksedk:
The unit tests works fine here... even when cloning...

PR #9362, opened 2025-02-14
Refactor text removal logic to simplify tag handling

ivandrofly:
Removed unnecessary intermediate tag patterns (e.g., "]:" and "}:"), keeping only the primary tag pairs. This change improves code clarity and ensures consistent behavior for text removal settings.


src/libse/Forms/RemoveTextForHI.cs
```diff
@@ -1405,22 +1405,18 @@ public string RemoveHearImpairedTags(string text)
 
             if (Settings.RemoveTextBetweenSquares)
             {
-                text = RemoveTextBetweenTags("[", "]:", text);
```
ivandrofly, line 1408: logic in `RemoveTextBetweenTags` already handles this

Issue #9381, opened 2025-02-17
Formats Help Text Is Not Correct or Add Code Format (Blu-raysup)


AYColumbia:
The formats help states
`- For Blu-ray .sup output use: 'Blu-raysup'`

However, if you use "Blu-raysup", it won't work because the code doesn't support it.  So not sure if this is misspelled in the format help text or was accidently missed in the conditionals for blu-ray in CommandLineConverter.cs class, i.e.,
`   else if (targetFormat == "sup" || targetFormat == "bluray" || targetFormat == "blu-ray" || targetFormat == "bluraysup" || targetFormat == "bluray-sup" || targetFormat == "bdsup")`

timminator:
Works fine for me here?

```
PS C:\Users\GPUVM\Downloads\subtitleedit-main (3)\subtitleedit-main\src\ui\bin\Release\net48>
.\SubtitleEdit.exe /convert "C:\Users\GPUVM\Desktop\test.sub" blu-raysup

Subtitle Edit 0.0.0 - Batch converter

Found VobSub subtitle format
Loading subtitles from file "C:\Users\GPUVM\Desktop\test.sub"
Using OCR via tesseract to extract subtitles
OCR... : 100%
Extracted subtitles from file "C:\Users\GPUVM\Desktop\test.sub"
Converted subtitle
1: test.sub -> C:\Users\GPUVM\Desktop\test.sup... done.

1 file(s) converted in 00:14:23.0825542
```

The code you quoted is just there to convert not matching inputs to the correct one in case people use shortcuts or abbreviations. It does not need to contain the correct argument itself.

AYColumbia:
I guess I don't get the why the help format says to use a format that is not even checked.  Since the conditional doesn't set the output format based on the help text isn't that going to be a problem?  Anyhow, I just wanted to point out that the help text is not part of the conditional is all.  Thanks.

timminator:
The specified argument is assigned to the targetformat right above the if-else statement:

https://github.com/SubtitleEdit/subtitleedit/blob/b70d6a197529ede4cf1a7f5d177aaced5c03679f/src/ui/Logic/CommandLineConvert/CommandLineConverter.cs#L208

As I already mentioned, the code you quoted is just there to convert not matching inputs to the correct one in case people use shortcuts or abbreviations.
[...]

PR #9388, opened 2025-02-17
Enhancement: add -h and --help options for command line help

timminator:
While I was looking at Issue #9381 I tried to run the help command and intuitively typed `-h` and because that didnt work I tried `--help` - also without success.

I thought adding these would be a nice enhancement. They are part of the GNU coding standard and are quite often used in other project aswell.


GrampaWildWilly:
Just to be complete & consistent, I'd be tempted to add -? and /h.  But I'm not a coder . . .

PR #9393, opened 2025-02-19
Update pt-PT.xml

hugok79:
@Blackspirits Always save the file in **UTF-8-BOM** Recommend using **Notepad++**
@niksedk for commit https://github.com/SubtitleEdit/subtitleedit/commit/50bf25405af07563875dd4787f34158a271d0f15

### DIFF
--- a/src/ui/Languages/pt-PT.xml
+++ b/src/ui/Languages/pt-PT.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Language Name="Português (Portugal)">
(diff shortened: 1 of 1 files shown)


hugok79:
@Blackspirits When I opened the pt-PT.xml the file it wasn't in UTF-8-BOM it was only in UTF-8

hugok79:
After
Captura de ecrã 2025-02-19 195105

hugok79:
After

Captura de ecrã 2025-02-19 195155

Blackspirits:
ok, I hadn't even realized. I usually save everything in UTF-8-BOM, I'll be more careful. Thanks ;)

PR #9424, opened 2025-03-02
Normalize language codes to lowercase when retrieving them

ivandrofly:
The two-letter language code is now explicitly converted to lowercase when retrieved to ensure consistency. This change eliminates redundant `.ToLowerInvariant()` calls and ensures `installedLanguages` contains uniformly formatted codes.

### DIFF
--- a/src/ui/Forms/Translate/AutoTranslate.cs
+++ b/src/ui/Forms/Translate/AutoTranslate.cs
@@ -792,10 +792,11 @@ namespace Nikse.SubtitleEdit.Forms.Translate
                 {
                     continue;
                 }
-                var iso639 = Iso639Dash2LanguageCode.GetTwoLetterCodeFromEnglishName(layoutName);
+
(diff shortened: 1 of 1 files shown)


src/ui/Forms/Translate/AutoTranslate.cs
```diff
@@ -792,10 +792,11 @@ public static string EvaluateDefaultTargetLanguageCode(string defaultSourceLangu
                 {
                     continue;
                 }
-                var iso639 = Iso639Dash2LanguageCode.GetTwoLetterCodeFromEnglishName(layoutName);
+
+                var iso639 = Iso639Dash2LanguageCode.GetTwoLetterCodeFromEnglishName(layoutName)?.ToLowerInvariant();
                 if (!string.IsNullOrEmpty(iso639) && !installedLanguages.Contains(iso639))
```
ivandrofly, line 797: The check was done without lowering `iso639.ToLowerInvariant()` but then when adding to `installedLanguages.Add(iso639.ToLowerInvariant());`  `.ToLowerInvariant()` was called.

This PR 🔗 fixes the inconsistency!

PR #9464, opened 2025-03-14
Fix Common Errors: Add functionality to split 3+ lines into multiple paragraphs

ivandrofly:
Introduced a new feature to handle subtitles with over three lines. The "Split 3+ lines into multiple paragraphs" fix was implemented, with corresponding updates in the UI and language structure. This enhances subtitle formatting flexibility and improves display readability.

Todo:

~Unit-test coming soon...~

This feature will allow breaking very complicated to split paragraph that contains more than 3 lines into individual paragraphs, 
but it can also still split any paragraph that contains more then 3 lines into individual paragraph.


ivandrofly:
For a quick manual test

```
text
1
01:31:01,793 --> 01:31:11,334
"The Adventure of..."
"The Invisible Army"?
"The League of Furies"?
"The Monstrous Regiment"?
```

PR #9466, opened 2025-03-15
Feature/ollama ocr

ivandrofly:
Fetch available models from ollama instead of manually typing them!

image

This pull request also contains some clean up if you want them separately I can split it into 2 pull requests maybe!

But the clean up are obvious and small!


ivandrofly:
A filter could have also be set to only display models containing vision "vision" but you never know what the upcoming models will be named :D so I think it's better we just display them all

PR #9484, opened 2025-03-20
Refactor OCR delay logic to use `Task.Delay` directly

ivandrofly:
Replaced `Utilities.TaskDelay` with `Task.Delay` in `StartOcrFromDelayed` for cleaner and more modern asynchronous programming. Removed the custom `TaskDelay` implementation from `Utilities` as it is no longer needed. This simplifies the code and reduces dependency on custom utilities.

### DIFF
--- a/src/libse/Common/Utilities.cs
+++ b/src/libse/Common/Utilities.cs
@@ -2657,19 +2657,6 @@ namespace Nikse.SubtitleEdit.Core.Common
             return text;
         }
 
-        /// <summary>
-        /// Creates a task that will complete after a time delay.
-        /// </summary>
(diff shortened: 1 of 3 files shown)


src/libse/Common/Utilities.cs
```diff
@@ -2657,19 +2657,6 @@ public static string RemoveSpaceBeforeAfterTag(string input, string openTag)
             return text;
         }
 
-        /// <summary>
-        /// Creates a task that will complete after a time delay.
-        /// </summary>
-        /// <param name="millisecondsDelay">The number of milliseconds to wait before completing the returned task.</param>
-        /// <returns>A task that represents the time delay.</returns>
-        public static Task TaskDelay(int millisecondsDelay)
```
ivandrofly, line 2665: Legacy code

PR #9513, opened 2025-03-30
Fixes wrong usage of ControlStyles flags

ivandrofly:
See: https://learn.microsoft.com/en-us/dotnet/api/System.Windows.Forms.ControlStyles?view=netframework-4.8

> **AllPaintingInWmPaint**: 

    If true, the control ignores the window message WM_ERASEBKGND to reduce flicker. This style should only be applied if the 
    [UserPaint](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.controlstyles?view=netframework- 
    4.8#system-windows-forms-controlstyles-userpaint) bit is set to true.


niksedk:
Both worked fine as far as I could see, but yours look nicer without the "this."

PR #9527, opened 2025-04-02
New Spectrogram Appearance

taylanunver:
I have added a new spectrogram appearance called "Heat". It is a modified version of the color map inferno. It is made to resemble the audio tools in the music industry. To achieve this with minimal performance hit I have created a method to import 1x256 images as palettes. Here are some example spectrograms.


taylanunver:
I have also added a unified view of spectrogram and waveform with an opacity slider at the bottom. Spectrogram now also scales vertically to fill container.

image
image
image

niksedk:
Looks good :)

Could you move the opacity setting to Options - Settings - Waveform?
The main interface should only contain stuff used often, and I think most people will probably just use the default value (if it's sensible).

The selected paragraph also needs the opacity...

taylanunver:
Here's the newly added option in the settings menu.

https://github.com/user-attachments/assets/df091b30-ac64-46a7-ab50-5045349028a5

Flitskikker:
@taylanunver @niksedk 

Can you please add an option to make the merging of the waveform + spectrogram optional?
I just updated to 4.0.12 and I find the merged view very confusing with my color settings (Classic), I prefer to have them one above the other.

Unfortunately, I don't have spare time to add it myself, so I'll have to downgrade to 4.0.11 in the meantime.

Thank you.

PR #9591, opened 2025-04-19
Improve VogSub parsing

C0bra5:
This PR adds better handling of .sub files without an IDX file accompanying them.

I encountered a sub file that somehow didn't get fully read by SE while my video players were handling it fine. So, I dug into the format and SE and found that SE made some bad assumptions while parsing these files.

image

### DIFF
--- a/src/Tests/Logic/VobSub/VobSubTest.cs
+++ b/src/Tests/Logic/VobSub/VobSubTest.cs
@@ -24,7 +24,7 @@ namespace Tests.Logic.VobSub
             reader.Open(fileName);
             var list = reader.MergeVobSubPacks();
 
-            Assert.IsTrue(list.Count == 2);
+            Assert.AreEqual(2, list.Count);
         }
 
     }
--- a/src/libse/VobSub/VobSubParser.cs
+++ b/src/libse/VobSub/VobSubParser.cs
@@ -4,10 +4,8 @@ using System.Collections.Generic;
 using System.Drawing;
 using System.IO;
 
-namespace Nikse.SubtitleEdit.Core.VobSub
-{
-    public class VobSubParser
-    {
+namespace Nikse.SubtitleEdit.Core.VobSub {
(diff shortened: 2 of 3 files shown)


niksedk:
thx, looks good to me :)

If you can upload a sample vobsub with this issue, that would also be nice.

C0bra5:
> thx, looks good to me :)
> 
> If you can upload a sample vobsub with this issue, that would also be nice.

Since the original is from a commercial DVD, I probably can't upload the whole file, but here is a sample from that file that would have caused tissues.

I'll make a PR with some test cases that would have caused issues previously.

[weird-vodsubs.zip](https://github.com/user-attachments/files/19822828/weird-vodsubs.zip)

### odd-packet-size.sub
The sub should have 2 identical images in sequence (the time codes will differ, though). The old code won't recognize the second line because the last packets of both images are 2047 characters long.

image

### mpeg-header-stuffing-byte.sub
It only contains one line, the same as the first line from the other file. The final packet of the image has been modified to have a single stuffing byte. The contents of the private stream data remain the same.

image

C0bra5:
> thx, looks good to me :)
> 
> If you can upload a sample vobsub with this issue, that would also be nice.

Since the original is from a commercial DVD, I probably can't upload the whole file, but here is a sample from that file that would have caused tissues.

I'll make a [PR](https://github.com/SubtitleEdit/subtitleedit/pull/9594) with some test cases that would have caused issues previously.

[weird-vodsubs.zip](https://github.com/user-attachments/files/19822828/weird-vodsubs.zip)

### odd-packet-size.sub
The sub should have 2 identical images in sequence (the time codes will differ, though). The old code won't recognize the second line because the last packets of both images are 2047 characters long.

image

### mpeg-header-stuffing-byte.sub
It only contains one line, the same as the first line from the other file. The final packet of the image has been modified to have a single stuffing byte. The contents of the private stream data remain the same.

image

Issue #2434, opened 2017-06-10
High DPI Support


stax76:
In case your screen is Full HD and you have good enough eyesight to use 96 DPI you might not know that your app looks poor on systems with 144 or more DPI. Some reasons people use high DPI is they have large 4K screens and/or poor eyesight, in my case both. I can help you if you have problems or questions regarding High DPI win forms development.

evolfox:
Hello.

I use a 4K screen and the preview video is not displayed correctly. https://i.imgur.com/pC0UV5f.jpg

nyuszika7h:
There's still no high DPI support, this would be nice.

niksedk:
Sorry, I don't know how to test this... any help is welcome. Also, I'm not sure what the problem is.

nyuszika7h:
The interface looks blurry on my 1080p screen with 125% scaling:

![Screenshot of Subtitle Edit with scaling performed by system](https://user-images.githubusercontent.com/482367/43587417-7bb0fad4-966a-11e8-8c8b-bc5cc275a7b2.png)

If I enable "Override high DPI scaling behavior. Scaling performed by: Application", some spacing is not exactly right:

![Screenshot of Subtitle Edit with scaling performed by application](https://user-images.githubusercontent.com/482367/43587507-ad5566b0-966a-11e8-8b28-56d07d5e6316.png)

PannenkoekenNL:
> Sorry, I don't know how to test this... any help is welcome. Also, I'm not sure what the problem is.
You can test it by setting the DPI in Windows higher. Setting > Display > Change the size of text, apps, and other items.

![test](https://user-images.githubusercontent.com/16383282/53661486-b2743d00-3c60-11e9-84a3-bb76242bc86b.png)

The problems is that Subtitle Edit looks blurry when having a higher DPI, a higher DPI is needed when using high resolution tablets/notebooks, 4K monitors etc. Or just when you've bad eye sight.

dabisu:
I see this bug report is open from 2017. Most of applications are DPI aware and scale properly. Any chance it will be fixed in SE? I really like SE, but blurry fonts... Nowadays a growing number of users are buying bigger monitors, and if you're don't have good sight it's a must to set the dpi scaling to 125% or more.

lavadrop:
This is still a thing and it's really impacting my use of the program. For example, the configuration menus are almost unreadable in Linux at 200% scale using mono.

dabisu:
[...]

PR #9637, opened 2025-05-03
Make UI more DPI-scaling friendly

eugenesvk:
I've updated some of the UI to be more DPI scaling friendly

- For all **non-main** forms, I've simply replaced Font scaling mode with DPI scaling mode, it seems to work mostly fine, though there are some minor overlaps of e.g. text labels and the fields positioned below them (didn't fix them). Also haven't tested all of the forms as there are too many of them :)

- For the **main** form using DPI scaling mode didn't work, some elements like Video player could disappear altogether at some sizes. So I had to do manual tweaks like adding button size constraints to avoid overlaps. This is rather crude as I guess might not work well in all different DPI and scaling factors (I've only briefly tested 125,150,175)

- I've also made some of the elements denser as the current version wastes way too much space, so on smaller window sizes some elements/buttons wouldn't even fit. So I've removed a bunch of margins/paddings and gaps

- Note that I've only tested on Windows 11 and only with DPI scaling on "regular" monitor (so not big high resolution high DPI one)

This version also restores DPI awareness setting in the app config file, so poor users don't need to do any manual compatibility configs of the exe file

Some screenshots for comparison

- 1 Old noDPI
SubtitleEdit 1 Old noDPI

- 2 Old DPI-scaled
SubtitleEdit 2 Old DPI-scaled

- 3 New DPI-scaled (no main) and denser UI
SubtitleEdit 3 New DPI-scaled (no main) and UI

A few open questions:
- Text field can't be increased to the same size as in the old version. My guess is that there is some internal dynamic override that prevents the top listview from being too small (or the middle field from becoming too large). Is there a simple way to list this restriction and let the user do whatever sizing he wants?
- When testing label placement, I've noticed that there some bugs where text field is displayed on top of the label, so had to manually move labels to the bottom, but not they sometimes overlap the field border (see "single line length"). But in a .NET 8 branch this issue is resolved. But the PR for Net8 was closed. Do you ever plan to update .NET 8? Or is there a simpler solution in Net 4.8 (I found some horizontal position overrides for the labels, but not vertical ones, and trying to manually set `.top` didn't work for me, but maybe I was just missing something simple
- `F10` labels in create/adjust do NOT fit after switching tabs, though on app launch they do fit if the corresponding tab is opened. Do you know why?
- The bottom translate section can be reduced in size to make Google and other buttons go offscreen. While I think it's fine since if you don't use translate, you can have more space to fit the actual subtitles, still curious, is it due to some internal dynamic height calculations or simply an issue with static designer constraints (trying to set min height didn't help)?

More general Visual Sudio issues:

- When editing in Visual Sudio forms editor with the simplest of "move this label by 1px to the right", I often get whole sections removed, leading to compile errors, e.g., 
  ```c#
    // toolStripSeparatorToggle
    // 
    this.toolStripSeparatorToggle.Margin = new System.Windows.Forms.Padding(0, 3, 0, 3);
    this.toolStripSeparatorToggle.Name = "toolStripSeparatorToggle";
    this.toolStripSeparatorToggle.Size = new System.Drawing.Size(6, 28);
    // 
  ```
  do you have any idea why this happens and how to prevent it?
- Similarly the dump Visual Studio editor keeps chaning index numbers of `Nikse.SubtitleEdit.Core.Common.TimeCode timeCode1 = new Nikse.SubtitleEdit.Core.Common.TimeCode();` to `4` then `8` etc. on changes. Is there a way to stop that as well and maintain fixed indices as these changes generate too much noise that has to be manually removed?

partially addresses https://github.com/SubtitleEdit/subtitleedit/issues/2434 and similar


eugenesvk:
The waveform issue can be "resolved" by reducing the dynamic width by `225` (with dpi scaling 150%) to the 
`audioVisualizer.Width = groupBoxVideo.Width - (audioVisualizer.Left + 10 + 225);` line in `TabControlModes_SelectedIndexChanged`, though this wastes too much space.

When chaning tabs, I've noticed that the waveworm first jumps to the right-most window border before being resized. My guess is that without `-225` it "overflows" the tabs on the left, and that's what's causing it to expand to the full window width.

But I still don't understand the fundamental issue why this glitches
Is there maybe a way to first reduce the width so that it fits and then move it so it never overflows?
One workaround that works dynamically is to remove the "right" anchor of the waveform, then it doesn't overlap the left tabs when they change, but then on launch the waveform is smaller than it needs be , guess that could also be fixed by repositioning it even if no tabs are changed

https://github.com/user-attachments/assets/599eebf5-9b54-4b45-8a0b-ab427f491fa7

abc16361:
I have tried SubtitleEdit_high_dpi_win.zip, and it looks quite clear on my 4k monitor with 150% scale. But under dark theme, some line heights on Settings are not large enough, and it could look better if more spaces are given to both sides of the Text Box area . 

margin_textbox

lineheight_darktheme

eugenesvk:
No idea why dark changes the line height, will wait til niksedk graces us with his deep knowledge.

Margins are partially intentional - I want as much space as possible to be given to the text itself, especially useful if you're not going full screen or using a smaller screen. The lack of margin symmetry isn't intentional, I think I had some issues with making the text box wider or maybe the start time smaller, don't rememeber, so left it as is, there were a bunch of tiny things I couldn't properly control in the layout. `0` not fully visible is anothe bug I couldn't fix

So on your screenshot instead of more useless margins on the left I'd make the start time smaller, move duration to the left and increase **Text** field width so it can easier fit longer subtitle lines with bigger text size and then also have small symmetrical margins on the right

(one of the things that bugged me in the old version were all those multiple cascading margins and lines from nested containers eating away at precious subtitle content)

abc16361:
The margins on the light mode are also too small but are better than the dark mode, for me this margin is not a big deal, just leave it as what it is if it's not easy to manipulate. But I find that the size of the dropdown from key chooser is more important. Right now it is too narrow and is even not wide enough originally on old version. That would be good if it can be resized so that keys can be easy to view. There are also some bugs on Choose color pop-up and Controls. I leave them here as references if it can be solved. I understand it is not easy and hopefully niksedk has time to work it out and adopt these modifications. Subtitle Edit is really a good tool. I have tried different video players with waveform but only Subtitle Edit is a real player to me. 
lightTheme
keyChoose
choosecolor
Controls

PR #9680, opened 2025-05-24
Enhance video file detection

### LINKED ISSUE #9679: [Feature Request] Enhance video file detection
I noticed that when opening srt files that are named like "test (1).mp4" or "test - Copy.srt" the corresponding video "test.mp4" is not found for example.

"(1)" and "- Copy" or common ways of programs to name files in case the original one already exists.

I would like it if SubtitleEdit would attempt to find the original video file aswell in this case.

timminator:
I implemented a solution for enhancing the video file detection logic. We still try to find a video file with the full title of the subtitle for example and when this does not return any matches, we try to clean the filename for duplicate naming styles and try again to open the corresponding video file.

The only thing that's not so clean is that Windows by default names copys in the system language like "test - Copy.srt" for english or "test - Kopie.srt" for german. I added a dictionary that contains these words for all supported languages by SubtitleEdit.

I'm interested in your feedback, in case you would like to handle this differently.

Fix #9679

### DIFF
--- a/src/libse/Common/Utilities.cs
+++ b/src/libse/Common/Utilities.cs
@@ -3000,6 +3000,66 @@ namespace Nikse.SubtitleEdit.Core.Common
             return fileName;
         }
 
+        private static readonly HashSet<string> CopyWords = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+        {
+            "copy",      // en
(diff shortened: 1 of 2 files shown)


niksedk:
Could you build it into "TryToFindAndOpenVideoFile" - could possible make it simpler?

PR #9791, opened 2025-07-12
adds horizontal panning using mousepad by overriding WndProc

bjorn-ali-goransson:
Hello there,

Just a behavior I have from using Audacity.

Mousepad horzontal scrolling is very important to me personally 😄


niksedk:
How does this work? How can I test this?

PR #9848, opened 2025-07-31
OCRSpellCheck form uses defined subtitle font

BookOfBeasts:
This makes the UI more consistent with other forms that use the subtitle font.

It also allows for easier differentiation between similar characters by letting the user choose a font they find easier to read.

### DIFF
--- a/src/ui/Logic/UiUtil.cs
+++ b/src/ui/Logic/UiUtil.cs
@@ -1,4 +1,4 @@
-﻿using Nikse.SubtitleEdit.Controls;
+using Nikse.SubtitleEdit.Controls;
 using Nikse.SubtitleEdit.Core.Common;
 using Nikse.SubtitleEdit.Core.Enums;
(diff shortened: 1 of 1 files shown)


src/ui/Logic/UiUtil.cs
```diff
@@ -530,6 +524,18 @@ public static void InitializeSubtitleFont(Control control)
             }
         }
 
+        private static void SetControlSubtitleFont(Control control, string subtitleFontName, int subtitleFontSize, bool subtitleFontBold)
+        {
+            if (subtitleFontBold)
```
ivandrofly, line 529: You can single line this if you want.

`control.Font = new Font(subtitleFontName, subtitleFontSize,  subtitleFontBold ? FontStyle.Bold : FontStyle.Regular)`

ivandrofly (changes requested):

PR #9933, opened 2025-09-03
Add support for Google Lens OCR

timminator:
This PR adds support for the Google Lens API.

It is now possible to interact with the Google Lens API available in the Chrome Browser or on Android/Iphone directly.
More info can be found here:
https://github.com/timminator/chrome-lens-py

I made a standalone version again and integrated it into SubtitleEdit now.
The OCR accuracy and capability from Google Lens is highly impressive. It's the best I've ever seen.
It is not as fast as PaddleOCR though and of course all images are send to Google - the OCR process is not performed locally.


niksedk:
Looks good, but I seem to get some of these:

````
- (18/844) Error for: 017.png - HTTP ошибка: 500 (Status Code: 500) Response Body (partial): Internal error encountered.
````

Perhaps some retry or ?

PR #9975, opened 2025-09-28
Add support for Unicode characters in MCC

DavisR-GitHub:
Hi,

I propose adding support for Unicode characters in the MCC format.

According to _ANSI-CTA-708-E+S-2023+++Errata+FINAL.pdf_, section 7.1.1:

> The code space may be extended further (for future use) to handle other character sets not included in
> the language sets listed in Section 7.1.7. This extension capability anticipates code sets which require 16-
> bit character code addressing, such as characters to implement Chinese and Japanese. The P16 code in
> the C0 Miscellaneous Control Code set is used for this purpose. When a decoder encounters the P16
> code, it uses the succeeding two bytes to address characters in a 16-bit code set.

Based on this specification, I’ve implemented processing of the P16 sequence, enabling proper handling of Unicode characters.


niksedk:
Cool, do you have any sample files?

Issue #10002, opened 2025-10-15
Copying, searching for and pasting two lines of text causes a crash


darnn:
I'm on the latest beta, but this also happened in the previous build I was using, from July.
To reproduce this, open a subtitle file that contains a subtitle with two lines of text text. My sample file looks like this:

<img width="977" height="700" alt="Image" src="https://github.com/user-attachments/assets/ff73df43-be81-4ff6-9527-5b72549d8d9c" />

Focus the textbox for the subtitle and select its contents. Use ctrl+C to copy the text. Now press the keyboard shortcut for `Find`. Use ctrl+V to paste the text (though it's already there). You'll see the following:

<img width="977" height="699" alt="Image" src="https://github.com/user-attachments/assets/472009b2-af8e-496e-a97f-83bb3ee96414" />

Note that both lines of text are in the input field. Press enter or click `Find next`:

<img width="976" height="699" alt="Image" src="https://github.com/user-attachments/assets/73921aa1-9c50-49ba-8a26-806973e130d4" />

Note that the status bar is now two lines tall rather than one, to accommodate both lines of text.
Close the `Find` window by pressing escape. You'll now be in the textbox again with the text selected:

<img width="976" height="699" alt="Image" src="https://github.com/user-attachments/assets/7adf3ef5-7475-4de7-af49-0921fc8e93f9" />

Now press ctrl+V again. The program will crash.

I suspect this has to do with what happens to the status bar, since if you do anything to reset the status bar to a single line, no crash occurs. Also, it's best to test this with a large or maximized window, since it apparently doesn't happen if the window is very small.

Here's everything together as a gif:

![Image](https://github.com/user-attachments/assets/38ae1c5e-0762-4139-8949-bfd240f2cff6)

niksedk:
Thx for the nicely explained bug report! :)

Hopefully fixed here: https://github.com/SubtitleEdit/subtitleedit/releases/download/4.0.13/SubtitleEditBeta.zip

Also, I'm mostly working on SE 5 now - still in the early preview phase, but it's getting better slowly: https://github.com/niksedk/subtitleedit-avalonia/releases

darnn:
Still happens, unfortunately.

And since you asked:
General: Merge with next, Merge original and translation, Switch original and translation, Go to current line start/end, Cycle audio tracks, Choose profile
Save all, Import -> Plain Text, Right to left
[...]

PR #10015, opened 2025-10-21
Feature/se textboxes

ivandrofly:
This refactor centralizes all active textbox operations through a new GetActiveTextBox helper, removing duplicated null checks and branching across many properties and methods. 

Property accessors (Text, Font, selection, multiline, enabled, read-only, lines, etc.) and actions (SelectAll, Clear, Undo, Cut/Copy/Paste, Focus, ScrollToCaret) now delegate via a single call, adding null-safety. 

ScrollBars handling is consolidated with type checks for the underlying control. Dark theme application and rollback are simplified to operate on the active textbox with safer early-returns. Additionally, an unused constructor was removed to reduce surface area.


src/ui/Controls/SETextBox.cs
```diff
@@ -13,21 +13,6 @@ namespace Nikse.SubtitleEdit.Controls
...
-        // ReSharper disable once InconsistentNaming
-        public new event MouseEventHandler MouseClick;
-        // ReSharper disable once InconsistentNaming
-        public new event EventHandler Enter;
-        // ReSharper disable once InconsistentNaming
-        public new event KeyEventHandler KeyUp;
-        // ReSharper disable once InconsistentNaming
-        public new event EventHandler Leave;
-        // ReSharper disable once InconsistentNaming
-        public new event MouseEventHandler MouseMove;
```
ivandrofly: Any other control that want to hook to these event can do it via the parent class 
and if we want to invoke these event we can just use for example "OnEnter(EventArgs)" so no need for this!

ivandrofly (commented):
@niksedk  Can you take a look at this?

It  should solve the bug #10002

PR #10024, opened 2025-10-25
Fix incorrect remaining time in Whisper dialog

pintelt:


pintelt:
Send video file path to GetVideoInfo instead of audio file path.

pintelt:
It doesn't work all the time. GetVideoInfo tries to get the info from the video files header, wav files returned an empty VideoInfo and then the estimated time was incorrect.

PR #10078, opened 2025-11-13
Refactor GetTimeCode method

AaryanHazCompter:
I was doing some subtitles and noticed when using the Export Custom window that when I tried the format `ss.zzzz` it rounded up a second, ie, when I tried using `{start}`, I expected `08.990` and instead got `09.990`. Looking into the code I noticed that there was also some redundancy and wanted to future proof it, as well as fix my own problem.

Solved my problem, tests run fine, and now instead of making several redundant / manual calls, this will dynamically handle replacement as required for seconds and milliseconds. Negative timecode checking is better, total time calculations behave properly and continues to store the essential replacements. 

ETA: I want to add I noted that you are in bug-fix mode and this is 100% a bug-fix imo

### DIFF
--- a/src/ui/Forms/ExportCustomTextFormat.cs
+++ b/src/ui/Forms/ExportCustomTextFormat.cs
@@ -177,14 +177,68 @@ namespace Nikse.SubtitleEdit.Forms
                 return wholeSeconds.ToString().PadLeft(length, '0');
             });
 
+            // Detect pure-z mode (template contains ONLY z's)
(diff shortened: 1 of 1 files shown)


niksedk:
Try `z` and `zz` and `zzz` alone.

niksedk:
thx :)
Also added for SE 5: https://github.com/niksedk/subtitleedit-avalonia/commit/9e28f21d59aea3b2e1f3a74be46d4dfc7391a80d

PR #10100, opened 2025-11-27
Add gzip error message to pt-PT.xml

Blackspirits:


niksedk:
Also, I'm working a lot on SE 5 now, which is almost ready for translation...

The language files are now json - I've got a new tool here:
https://github.com/niksedk/json-content-translator/releases

Read more here: https://github.com/niksedk/subtitleedit-avalonia/discussions/79

PR #10437, opened 2026-03-26
Add horizontal scrolling to the audio visualizer

derzz:
This commit adds horizontal scrolling to the audio visualizer, making it more friendly to use on a laptop.

I'll also note that I am on a Mac and I have natural scrolling turned on so I have to turn on "invert mouse wheel" in settings. To me when I either swipe up or left, the visualizer should go right, and vice versa.


GrampaWildWilly:
OK.  I think I understand.  When I hover my mouse over the waveform, I have no way of just magically scrolling.  I repeat, my mouse does NOT have a scroll wheel.  This means that you have excluded me from the target audience for this code change.  That's a shame.  Oh well.  I guess I'll just go back to waiting for the implementation of the scroll bar I've already asked for a few times.

PR #10481, opened 2026-03-29
Fix native video resize flicker and mpv preview refresh

Ironship:
## Summary
This PR fixes two native-video regressions without pulling in the unrelated changes from the broader test branch:

1. Avoid native video resize flicker on Windows
2. Refresh the mpv subtitle preview after opening or recreating the video player

## What changed
- Cherry-picked the two minimal Windows-native video commits from the fork's resize-flicker branch
- Added a one-line follow-up in `MainViewModel.VideoOpenFile(...)` to mark the mpv preview dirty after `_mpvReloader.Reset()`

## Why this is minimal
- Excludes unrelated changes from `fix/mpv-native-resize-flicker` such as the Qwen asset casing commit
- Keeps the subtitle-preview fix to a single state update instead of adding new abstractions or broader refresh plumbing
- Limits Windows-specific native-window handling behind OS checks
- Preserves the existing Linux resize workaround and leaves the macOS path untouched

## Validation
- Built `src/UI/UI.csproj` successfully in `Release` on Linux using .NET 10
- Reviewed the resulting diff to ensure it only touches:
  - `src/UI/Logic/VideoPlayers/LibMpvDynamic/LibMpvDynamicNativeControl.cs`
  - `src/UI/Logic/VideoPlayers/LibVlcDynamic/LibVlcDynamicNativeControl.cs`
  - `src/UI/Features/Main/MainViewModel.cs`

## Regression being fixed
When subtitles were already loaded and the video player was opened/recreated, `_mpvReloader.Reset()` cleared mpv preview state but nothing marked the preview dirty again. As a result, subtitles stayed invisible in the video preview until another action such as changing subtitle format triggered `RefreshMpv(...)`.

Setting `_mpvPreviewDirty = true` immediately after opening the player restores the expected refresh path.

### DIFF
--- a/src/UI/Logic/VideoPlayers/LibMpvDynamic/LibMpvDynamicNativeControl.cs
+++ b/src/UI/Logic/VideoPlayers/LibMpvDynamic/LibMpvDynamicNativeControl.cs
@@ -108,6 +108,7 @@ public class LibMpvDynamicNativeControl : NativeControlHost
     protected override IPlatformHandle CreateNativeControlCore(IPlatformHandle parent)
     {
         _nativeHandle = CreateRenderTargetHandle(parent.Handle);
+        var handleDescriptor = _ownedChildHandle != IntPtr.Zero ? "HWND" : parent.HandleDescriptor;
 
         if (!_isInitialized && _mpvPlayer != null)
         {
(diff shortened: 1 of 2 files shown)


Ironship:
Addressed the two Copilot review comments in follow-up commit `fb3d5ed37`.

What changed:
- `LibMpvDynamicNativeControl.CreateNativeControlCore(...)` now returns `parent.HandleDescriptor` when we reuse or fall back to the parent native handle, and only uses `HWND` for the owned child-window path.
- `LibVlcDynamicNativeControl.CreateNativeControlCore(...)` now does the same.

This keeps the Windows child-host fix intact while avoiding hard-coded `HWND` descriptors on Linux/macOS or on Windows fallback paths.

Validation:
- `dotnet build src/UI/UI.csproj -c Release` passes on Linux after the change.

PR #10571, opened 2026-04-08
Fix OCR user dictionary add-actions

HellbringerOnline:
## Summary
This PR fixes two related OCR dictionary issues in the unknown-word flow.

1. `Add to user dictionary` did write the word to the runtime `*_user.xml`, but OCR could still ask for the same word again in the same session because the OCR runtime dictionary state was stale.
2. Some OCR add-actions used the raw OCR token (`Word.Word`) instead of the displayed/corrected word (`FixedWord` or the edited popup text), so the saved value could differ from what the user actually confirmed.

These two bugs are related in the UI, but they have different root causes:
- fix 1 updates OCR runtime state immediately after adding a user word
- fix 2 makes OCR add-actions submit the displayed word instead of the raw OCR token

### 1. Stale OCR runtime state after `Add to user dictionary`
`OcrFixEngine2` loaded user words into `SpellCheckWordLists2` during initialization and then kept using that in-memory state for the current OCR run.

When the user pressed `Add to user dictionary`, the XML file was updated, but the current OCR engine state was not. The unknown-word list that had already been built for the session was also not cleaned up. As a result, words such as `recuérdalo` could already be present in the runtime dictionary file and still be prompted again immediately.

### 2. Raw OCR token used instead of displayed word
In several OCR add-action paths, the value sent to the user dictionary or names list was `Word.Word`.

That is the internal OCR token, not necessarily the same word shown to the user in the popup. In cases where OCR had already normalized or corrected the token for display, the saved value could be different from the visible word the user approved.

### Fix 1: refresh OCR session state after adding a user word
- Added `AddUserWord(string word)` to `IOcrFixEngine2` and `OcrFixEngine2`.
- Routed OCR `Add to user dictionary` through `_ocrFixEngine.AddUserWord(...)` instead of only updating XML directly.
- Reused `SpellCheckWordLists2.AddUserWord(...)` so one call both:
  - writes the word to `Se.DictionariesFolder`
  - updates the current OCR engine's in-memory word lists
- Removed the just-added word from the current OCR unknown-word state and from the pending popup traversal list so it is not shown again in the same OCR run.

### Fix 2: use the displayed/edited OCR word for add-actions
- Added `DisplayWord` and `ResolveSubmittedWord(...)` to `UnknownWordItem`.
- Changed side-panel `Add to user dictionary` / `Add to names` to initialize from the displayed OCR word instead of the raw token.
- Changed popup `Add to user dictionary` / `Add to names list` to use:
  - the edited popup text first
  - otherwise the displayed fixed word
  - and only then fall back to the raw token
- Left raw-token-specific paths alone where that behavior is still intentional, such as OCR pair and Google search actions.

## User Impact
After this change:
- adding a word to the user dictionary in OCR takes effect immediately in the current OCR session
- the same word is not prompted again from stale OCR state in that run
- words saved through OCR add-actions match the word the user sees and confirms in the UI

## Validation
Ran locally:

```powershell
dotnet build SubtitleEdit.sln -c Debug --no-restore
dotnet test tests\UI\UITests.csproj --filter "FullyQualifiedName~UITests.Features.Ocr.FixEngine" --no-restore
dotnet test tests\UI\UITests.csproj --filter "FullyQualifiedName~UITests.Features.Ocr" --no-restore
```

## Tests Added
- `OcrFixEngineUserDictionaryTests`
  - verifies that `AddUserWord(...)` makes a new word valid without reinitializing the OCR engine
- `UnknownWordItemTests`
  - verifies that OCR add-actions prefer the displayed/fixed word and edited popup text over the raw OCR token

## Example Scenario
Reported example:

`Escucha lo que te digo, chico, y recuérdalo bien.`

Confirmed behavior before fix:
- `recuérdalo` could already be written correctly to `es_ES_user.xml`
- OCR still prompted it again because the current OCR runtime state had not been refreshed

With this PR:
- the add-action updates both the runtime dictionary file and the OCR in-memory state
- the current unknown-word session state is cleaned up immediately
- the saved word matches the displayed/confirmed word

### DIFF
--- a/src/UI/Features/Ocr/FixEngine/OcrFixEngine2.cs
+++ b/src/UI/Features/Ocr/FixEngine/OcrFixEngine2.cs
@@ -20,6 +20,7 @@ public interface IOcrFixEngine2
     void ChangeAll(string from, string to);
     void SkipAll(string word);
     void AddName(string name);
+    bool AddUserWord(string word);
 }
 
 public partial class OcrFixEngine2 : IOcrFixEngine2, IDoSpell
@@ -501,6 +502,16 @@ public partial class OcrFixEngine2 : IOcrFixEngine2, IDoSpell
         }
 
         _spellCheckWordLists.AddName(name);
+    }
+
+    public bool AddUserWord(string word)
+    {
+        if (string.IsNullOrWhiteSpace(word) || !_isLoaded)
+        {
+            return false;
+        }
+
+        return _spellCheckWordLists.AddUserWord(word);
     }
 }
 
--- a/src/UI/Features/Ocr/OcrViewModel.cs
+++ b/src/UI/Features/Ocr/OcrViewModel.cs
@@ -484,7 +484,7 @@ public partial class OcrViewModel : ObservableObject
         }
 
         var result = await _windowService.ShowDialogAsync<AddToNamesListWindow, AddToNamesListViewModel>(Window,
-            vm => { vm.Initialize(selectedWord.Word.Word, Dictionaries.ToList(), SelectedDictionary); });
+            vm => { vm.Initialize(selectedWord.DisplayWord, Dictionaries.ToList(), SelectedDictionary); });
         _isCtrlDown = false;
     }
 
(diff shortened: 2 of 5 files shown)


niksedk:
Thx, added now with some changes.

Issue #10486, opened 2026-03-29
Binding single-key shortcuts for use in the textbox


darnn:
When binding a function (I tried `Video, one second back`) to one of the numpad keys - again, Home in this case - it doesn't work. But Control + Home does work, for instance. The square brackets (Oem4, OemCloseBrackets), the ` key (Oem3) and the equals key (OemPlus) can't be bound as single keys either, at least not for use within the textbox. There are probably others I haven't checked.

Issue #10485, opened 2026-03-29
Binding numpad keys


darnn:
When binding keyboard shortcuts, there doesn't currently seem to be a way of distinguishing between numpad and non-numpad keys (e.g. the Home key binds both the numpad and non-numpad one).

ivandrofly:
AI:

  Root cause: Avalonia's Key enum is a logical key — when NumLock is off, numpad 7 fires Key.Home, identical to the regular Home key. There's no way to distinguish them using Key alone.
                                                                                                                                                                                                                    
  Fix: Use KeyEventArgs.PhysicalKey, which reflects the actual hardware key. Numpad physical keys all start with "NumPad", so:
  - Regular Home: PhysicalKey.Home → stored as "Home"                                                                                                                                                               
  - Numpad 7 (NumLock on or off): PhysicalKey.NumPad7 → stored as "NumPad7"

PR #10576, opened 2026-04-09
Fix numpad vs non-numpad key ambiguity in shortcut binding

ivandrofly:
## Summary
- Numpad navigation keys with NumLock off (e.g. NumPad7 → `Key.Home`) were indistinguishable from their non-numpad counterparts, making it impossible to bind different shortcuts to the two keys
- Added `GetEffectiveKeyName(Key, PhysicalKey)` helper in `ShortcutManager` that returns the `PhysicalKey` name when the physical key is a numpad key (all Avalonia numpad physical keys are prefixed with `"NumPad"`), falling back to the logical `Key` name otherwise
- `ShortcutManager` now tracks a parallel `_activeEffectiveNames: HashSet<string>` using the physical override; shortcut hash lookup in `CheckShortcuts` iterates this set instead of `_activeKeys`, while `_activeKeys` is preserved so existing callers (count checks, `AllowedSingleKeyShortcuts` filtering) are unaffected
- `GetKeyViewModel` uses `GetEffectiveKeyName()` when recording a key press so that numpad 7 (either NumLock state) is saved as `"NumPad7"` rather than `"Home"`, matching what `CheckShortcuts` sees at runtime

## Test plan
- [ ] Open Options → Shortcuts, click a shortcut, press the regular **Home** key — confirm it is recorded as `Home`
- [ ] With NumLock **off**, press **NumPad 7** — confirm it is recorded as `NumPad7` (not `Home`)
- [ ] With NumLock **on**, press **NumPad 7** — confirm it is still recorded as `NumPad7`
- [ ] Bind one shortcut to `Home` and a different shortcut to `NumPad7`; confirm each fires its own action independently
- [ ] Verify shortcuts using non-numpad keys (letters, F-keys, arrows) still work as before

### DIFF
--- a/src/UI/Features/Options/Shortcuts/GetKeyViewModel.cs
+++ b/src/UI/Features/Options/Shortcuts/GetKeyViewModel.cs
@@ -3,6 +3,7 @@ using Avalonia.Controls;
 using Avalonia.Input;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
+using Nikse.SubtitleEdit.Logic;
 using Nikse.SubtitleEdit.Logic.Config;
 
 namespace Nikse.SubtitleEdit.Features.Options.Shortcuts;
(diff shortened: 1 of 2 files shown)


ivandrofly:
https://github.com/SubtitleEdit/subtitleedit/issues/10485
https://github.com/SubtitleEdit/subtitleedit/issues/10486

niksedk:
thx, added

PR #10578, opened 2026-04-10
Clean up Program.cs startup code

ivandrofly:
## Summary
- Extract `AppName` constant to deduplicate the `\"Subtitle Edit\"` magic string used in two places
- Consolidate six repeated `b.Instance != null` null checks in `ConfigureApplication` into a single early return
- Remove redundant base URI argument from `StyleInclude` constructors (base URI is unused when `Source` is already an absolute `avares://` URI)
- Fix `UnhandledException` handler to preserve error info when the thrown object is not an `Exception` (previously used `?? new Exception()` which silently discarded native exception details)

## Test plan
- [x] Application starts and displays the main window with title \"Subtitle Edit\"
- [x] Fluent theme, DataGrid, AvaloniaEdit, and ColorPicker styles load correctly (no visual regressions)
- [ ] On macOS, the native menu \"About Subtitle Edit\" entry opens the About dialog
- [ ] On macOS, opening a file via Finder \"Send to\" / file association loads the subtitle
- [ ] Trigger an unhandled exception and confirm it is logged correctly

### DIFF
--- a/src/UI/Program.cs
+++ b/src/UI/Program.cs
@@ -31,10 +31,16 @@ namespace Nikse.SubtitleEdit
             try
             {
                 // Global exception handling
-                AppDomain.CurrentDomain.UnhandledException += (sender, e) =>
+                AppDomain.CurrentDomain.UnhandledException += (_, e) =>
                 {
-                    var exception = e.ExceptionObject as Exception;
-                    Se.LogError(exception ?? new Exception(), "Unhandled AppDomain Exception");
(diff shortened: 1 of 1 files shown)


niksedk:
thx, fixed in https://github.com/SubtitleEdit/subtitleedit/commit/37f4e0057cdc83380088716630c5d30b2570964d

We're on Avalonia UI 12 now... I wonder how that goes ;)

PR #10594, opened 2026-04-13
Fix OCR user dictionary add-actions

HellbringerOnline:
## Summary
- use the displayed or confirmed OCR word when adding unknown words to the user dictionary
- refresh the in-memory OCR dictionary immediately after popup add-actions so the current OCR run sees the new user word right away
- keep the side-panel unknown word add-action UI-only and avoid mutating the current unknown word list there

## Problem
Some OCR add-actions used the raw OCR token instead of the displayed or corrected word. In addition, popup flows did not refresh the current OCR dictionary state immediately after adding a user word, so the just-added word could still be treated as unknown until a later refresh.

## Changes
- add `AddUserWord(string word)` to the OCR fix engine and update the runtime dictionary state immediately
- add `DisplayWord` and `ResolveSubmittedWord(...)` so add-actions consistently submit the visible or confirmed word
- keep popup traversal cleanup where it belongs, but remove unknown-word list cleanup from the side-panel add-action
- cover the behavior with OCR-focused tests

## Validation
- `dotnet test tests/UI/UITests.csproj --filter "FullyQualifiedName~UnknownWordItemTests|FullyQualifiedName~OcrFixEngineUserDictionaryTests"`
- `dotnet build src/UI/UI.csproj -c Debug --no-restore`


niksedk:
Looks good :)

`RemoveUnknownWordsFromCurrentState` shouldn’t be called here, as the UI unknown word list actions are only related to the UI.

Also, it would be great to keep PRs focused on a single concern. Could you move the Unicode normalization changes into a separate PR?

HellbringerOnline:
> Looks good :)
> 
> `RemoveUnknownWordsFromCurrentState` shouldn’t be called here, as the UI unknown word list actions are only related to the UI.
> 
> Also, it would be great to keep PRs focused on a single concern. Could you move the Unicode normalization changes into a separate PR?

Ok, I'll try.

niksedk:
> > Looks good :)
> > `RemoveUnknownWordsFromCurrentState` shouldn’t be called here, as the UI unknown word list actions are only related to the UI.
> > Also, it would be great to keep PRs focused on a single concern. Could you move the Unicode normalization changes into a separate PR?
> 
> Ok, I'll try.

thx :)

niksedk:
I think this does too much - and the remove stuff is not wanted.

PR #10641, opened 2026-04-17
fix(flatpak): keep nuget-sources.json in sync with UI.csproj in CI (fixes #10625)

### LINKED ISSUE #10625: GitHub action build flatpack failed since release Subtitle Edit v5.0.0-beta14
Unfortunately, the flatpak build step in the CI/CD pipeline failed when building Subtitle Edit release v5.0.0-beta14 and newer.

The action appears to be failing and producing the following error message:

Error: Build failed: Error: The process '/app/bin/xvfb-run' failed with exit code 1

Ironship:
Fixes #10625

## Problem

The flatpak CI build has been failing since **v5.0.0-beta14** with:

```
/run/build/subtitleedit/src/UI/UI.csproj : error NU1102:
  Unable to find package Google.Api.Gax.Grpc with version (>= 4.13.1 && < 5.0.0)
  - Found 1 version(s) in /run/build/subtitleedit/nuget-sources [ Nearest version: 4.12.1 ]
Error: module subtitleedit: Child process exited with code 1
Error: Build failed: Error: The process '/app/bin/xvfb-run' failed with exit code 1
```

### Root cause

`flatpak-builder` runs **offline** — it consumes only the NuGet packages pre-listed in `installer/flatpak/nuget-sources.json`. When `UI.csproj` was bumped to `Google.Cloud.TextToSpeech.V1 3.18.0` (and `SharpCompress 0.47.4`) in commit 6455601, only the two *directly referenced* packages were added to `nuget-sources.json`. Their **transitive** dependencies were not regenerated — in particular `Google.Api.Gax.Grpc` stayed at `4.12.1`, while `Google.Cloud.TextToSpeech.V1 3.18.0` requires `>= 4.13.1`. `dotnet restore` inside the sandbox therefore failed.

This is a foot-gun: every future bump in `UI.csproj` requires someone to remember to run `generate-nuget-sources.sh` and commit the result, otherwise the flatpak pipeline silently breaks.

## Fix

Two complementary changes:

### 1. Regenerate `nuget-sources.json` to match current `UI.csproj`

Ran `dotnet restore` for both `linux-x64` and `linux-arm64` (same flow as the fallback path in `generate-nuget-sources.sh`) and refreshed:
- `installer/flatpak/nuget-sources.json` — now includes `google.api.gax 4.13.1`, `google.api.gax.grpc 4.13.1` and the full consistent transitive closure.
- `src/UI/packages.lock.json` — re-locked accordingly.

### 2. Make the CI pipeline self-healing

Split `.github/workflows/build-flatpak.yml` into two jobs so this class of bug cannot come back:

- **`regenerate-nuget-sources`** (plain `ubuntu-latest` runner): checks out the repo, sets up .NET 10 via `actions/setup-dotnet@v4`, runs `installer/flatpak/generate-nuget-sources.sh`, uploads the fresh JSON as an artifact. Emits a `::notice::` if the committed file was stale, so drift is still visible.
- **`build-flatpak`** (flatpak container, `needs: regenerate-nuget-sources`): downloads the artifact into `installer/flatpak/` before the build, overwriting the committed copy. The rest of the job is unchanged.

Why two jobs: the `ghcr.io/flathub-infra/flatpak-github-actions:freedesktop-24.08` container does not ship the .NET SDK, and `container:` forces every step inside it. Doing the restore on a plain runner and passing the result via artifact is the cleanest split.

### Result

Any future bump in `UI.csproj` will now produce a consistent `nuget-sources.json` at build time — the flatpak pipeline cannot silently go out of sync with the project's dependency graph again. The committed JSON is still useful as a snapshot for local `flatpak-builder` invocations.

## Verification

- `dotnet restore src/UI/UI.csproj -r linux-x64` and `-r linux-arm64` both succeed locally with the regenerated JSON.
- `Google.Api.Gax.Grpc 4.13.1` is now present alongside `4.12.1`, satisfying the `>= 4.13.1 && < 5.0.0` constraint of `Google.Cloud.TextToSpeech.V1 3.18.0`.

### DIFF
--- a/.github/workflows/build-flatpak.yml
+++ b/.github/workflows/build-flatpak.yml
@@ -10,7 +10,42 @@ on:
         type: boolean
 
 jobs:
+  # Regenerate nuget-sources.json from the current UI.csproj so that the
+  # pre-fetched NuGet packages consumed by flatpak-builder always match the
+  # project's (possibly updated) dependency graph. Runs on a plain runner
+  # because the flatpak container image does not ship the .NET SDK.
+  regenerate-nuget-sources:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Set up .NET SDK
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: '10.0.x'
+
+      - name: Regenerate nuget-sources.json
+        run: bash installer/flatpak/generate-nuget-sources.sh
+
+      - name: Verify nuget-sources.json is consistent with UI.csproj
+        run: |
+          if ! git diff --quiet -- installer/flatpak/nuget-sources.json; then
+            echo "::notice::nuget-sources.json was stale and has been regenerated for this build."
+            git --no-pager diff --stat -- installer/flatpak/nuget-sources.json
+          else
+            echo "nuget-sources.json is already up to date."
+          fi
+
+      - name: Upload regenerated nuget-sources.json
+        uses: actions/upload-artifact@v5
+        with:
+          name: nuget-sources
+          path: installer/flatpak/nuget-sources.json
+          if-no-files-found: error
+          overwrite: true
+
   build-flatpak:
+    needs: regenerate-nuget-sources
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/flathub-infra/flatpak-github-actions:freedesktop-24.08
@@ -18,6 +53,12 @@ jobs:
 
     steps:
       - uses: actions/checkout@v5
+
+      - name: Download regenerated nuget-sources.json
+        uses: actions/download-artifact@v5
+        with:
+          name: nuget-sources
+          path: installer/flatpak/
 
       - name: Update metainfo version/date from Se.cs
         run: bash installer/flatpak/update-metainfo-version.sh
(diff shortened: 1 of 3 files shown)


Ironship:
Thanks Copilot, all four points addressed in 3a41653:

1. **win-x64 runtime packs in `nuget-sources.json`** (lines 337, 540) — reran `dotnet restore` for linux-x64 + linux-arm64 only in a clean tempdir and added a filter in the generator that drops `win-*`, `osx-*`, `android-*`, `ios-*`, `tvos-*`, `freebsd-*`, `windowsdesktop.*` and `maccatalyst-*` entries. `microsoft.aspnetcore.app.runtime.win-x64` and `microsoft.windowsdesktop.app.runtime.win-x64` are gone. The two remaining `*windows*` entries (`avalonia.angle.windows.natives`, `system.security.principal.windows`) are normal managed transitive deps of `Avalonia.Desktop` that NuGet requires to resolve even on Linux — not runtime packs.

2. **`packages.lock.json` contained a net10.0/win-x64 target** — reverted to main. The flatpak manifest uses `-p:RestoreLockedMode=false`, so this PR should not touch the lock file; the win-x64 target was an artefact of me running `dotnet restore` on a Windows host while preparing the patch.

3. **Unpinned `flatpak-dotnet-generator.py` download** — pinned to commit `3fc0620788a1dda1a3a539b8f972edadce8260ab` (current tip of `dotnet/flatpak-dotnet-generator.py` on `flatpak/flatpak-builder-tools`) with a comment explaining how to bump it. Also made the download conditional: the fallback path (used by the new CI job) never needs the generator, so we no longer fetch it just to throw it away.

PR #10706, opened 2026-04-26
Add visual waveform split lanes for ASS/SSA metadata

Ironship:
## Summary

Replaces the earlier ASS/SSA multi-track editing prototype with a smaller visual-only waveform split mode.

This keeps the subtitle model and saved ASS/SSA data unchanged. Instead of adding real `TrackIndex` state, the waveform can be visually split into lanes based on existing subtitle metadata:

- none
- actor
- style
- layer
- ASS position/alignment

The intent is to make complex ASS/SSA files easier to inspect in the waveform without introducing generated track styles or changing tools such as Fix Common Errors, merge, bridge gaps, or duration adjustment.

## Behavior

- Adds a `Split: None` menu button to the waveform toolbar.
- Draws subtitle blocks in visual lanes for the selected split mode.
- Shows lightweight lane dividers and labels.
- Keeps hit testing within the clicked lane.
- Keeps movement/resizing constrained to neighboring subtitles in the same visual lane.
- Does not change actor/style/layer/positioning metadata when dragging. Moving between virtual lanes is not a data-editing operation.

## Why this direction

This follows the feedback that ASS/SSA already has actors, styles, layers, and positioning, so a visual grouping mode fits the format better than adding real multi-track state to `Paragraph`.

The previous larger prototype is preserved in my fork at:

`backup/pr-10706-multitrack-original-20260427`

## Verification

- `dotnet build .\src\ui\UI.csproj -c Debug -v minimal`
- `dotnet test .\tests\UI\UITests.csproj -c Debug -v minimal` (94 passed)
- Manual smoke test of the waveform split menu and visual lanes.

## Notes

Compared to previous more complex change we modify 3x less code.


niksedk:
I'm not really sure ASSA/SSA makes sense for multi-track - the mostly have 1 or like 20+ styles/actors.
Multi-track does make sense for when translating with two tracks - or when using Video - more - open secondary subtitle.

(sorry, renamed the "UI" folder to "ui")

Ironship:
@niksedk  Good point. I had also been considering a smaller **visual-only** variant with fake tracks in parallel, and this was the kind of feedback I needed to decide which direction made more sense.

I have replaced this PR with that second prototype: Now the subtitle model is unchanged, and ASS/SSA is split only visually in the waveform by existing metadata (actor, style, layer, or ASS position/alignment). Dragging/resizing still edits timing only; moving between virtual lanes is not a metadata-editing operation, so actor/style/layer/position should still be edited in the normal text/style UI as it was dane before :)

The previous larger TrackIndex-based prototype is preserved in my fork as `backup/pr-10706-multitrack-original-20260427` in case we want to compare or recover anything from it.


## New / simplier version adds simple splitting view
mcmrsJAFQQ

This should make multi actors editing much easier. At least in my cases where I have 1 to 6 actors. I had lots of trouble when i had to add notes on top of video, and two actors were trying to speak at once. Posibility to change view improves readability.

Ironship:
Addressed the active Copilot review feedback in 49cf52ed0: cached the visual lane map for pointer-move/hit-test paths, moved Split waveform labels into Se.Language.Waveform, made Split button ordering deterministic, and changed Style splitting to use SubtitleLineViewModel.Style without falling back to Extra.

niksedk:
OK, I tried this a bit and it looks cool, but I was unable to see anything useful for the ASSA files I tried with... either a lot of styles/actors/effects or just one style and zero actors.

It could probably be useful for "Open second subitle" or subtitles with more than one language.

Ironship:
@niksedk

> It could probably be useful for "Open second subitle" or subtitles with more than one language.

Well in my use case with translations from war in Ukraine or translations for charities its usefull when I have 2-3 actors + additional notes.

Another niche is Fansubs for anime. Usually they have 2-6 actors + double subs.

### Update: ###
I’ve added logic so that the Split button in the waveform only appears when:
- the current format supports styled/metadata subtitles (HasStyleSupport)
and at least one split mode actually provides more than one lane
- I have hidden individual menu options if they are not relevant for the current file
- I reset the split to None if the selected mode becomes unavailable after a change in format or data

Ironship:
No interest in change No merge for more than week. Deployed it on my fork then

PR #10718, opened 2026-04-26
Distinguish numpad keys in shortcuts

### LINKED ISSUE #10485: Binding numpad keys

Ironship:
## Summary
- Map physical numpad keys to numpad shortcut keys when detecting shortcuts.
- Use the same mapping when matching shortcuts at runtime.
- Add regression coverage for numpad navigation keys like NumPad7 vs Home.

Fixes #10485

## Tests
- `dotnet test tests/UI/UITests.csproj`


niksedk:
Something should probably also be fixed in: 
- ShortcutsViewModel.cs / ShortcutsWindows.cs (list psychical keys or are we already doing that?)
- GetKeyViewModel.cs + GetKeyWindow.cs (check box?)

Ironship:
@niksedk  I checked these. ShortcutsViewModel already lists keys from Avalonia Key, which includes NumPad* values, and the shortcut model/runtime matching still expects Key names rather than PhysicalKey names. GetKeyViewModel already uses ShortcutManager.GetShortcutKey(e), so capturing a physical numpad key stores NumPad* even when Avalonia reports the navigation key.

I don't think GetKeyWindow needs an extra checkbox for this case. What you thinik?

niksedk:
OK, let's start with this, and if users want to choose mode in GetKeyWindow we can add this later.

PR #10726, opened 2026-04-27
refactor: convert AutoTranslate Translate methods to async/await

ivandrofly:
## Summary

- Replace blocking `.Result` calls with `async`/`await` in 7 translator classes
- `GoogleTranslateV1`, `GoogleTranslateV2`, `MicrosoftTranslator`, `MyMemoryApi`: promoted `Translate` from sync-wrapping-Task to a proper `async Task<string>`, replacing `.Result` on HTTP calls and removing `Task.FromResult` on the return
- `LibreTranslate`: method was already `async` but was still calling `.Result` on the initial `PostAsync`; fixed
- `NoLanguageLeftBehindServe`, `SeamlessM4TTranslate`: same partial-async issue, now fully awaited and `CancellationToken` forwarded to the HTTP call

Note: `MicrosoftTranslator.GetAccessToken` and `GetTranslationPairs` call HTTP from sync interface methods (`Initialize`, `GetSupportedSourceLanguages/TargetLanguages`) and are left unchanged.

## Test plan

- [x] Build `LibSE.csproj` with no errors or warnings
- [ ] Translate a subtitle using Google Translate V1 and verify the result returns correctly
- [ ] Translate a subtitle using Google Translate V2 (API key required) and verify it succeeds
- [ ] Translate a subtitle using Microsoft Translator and verify it succeeds
- [ ] Translate a subtitle using LibreTranslate and verify it succeeds
- [ ] Translate a subtitle using MyMemory API and verify it succeeds
- [ ] Confirm cancellation (e.g. stop translation mid-way) propagates correctly and doesn't deadlock

### DIFF
--- a/src/libse/AutoTranslate/GoogleTranslateV1.cs
+++ b/src/libse/AutoTranslate/GoogleTranslateV1.cs
@@ -45,7 +45,7 @@ namespace Nikse.SubtitleEdit.Core.AutoTranslate
             return GetTranslationPairs();
         }
 
-        public Task<string> Translate(string input, string sourceLanguageCode, string targetLanguageCode, CancellationToken cancellationToken)
+        public async Task<string> Translate(string input, string sourceLanguageCode, string targetLanguageCode, CancellationToken cancellationToken)
         {
             string jsonResultString;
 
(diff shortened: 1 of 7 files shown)


niksedk:
Nice :)

PR #10865, opened 2026-05-12
Fix PaddleOCR CUDA extraction on Linux

po5:
Extraction would hang because of incorrect base path within the 7z file.


niksedk:
thx :)

PR #10975, opened 2026-05-16
Register Alt+Space Windows system menu via Window class handler

niksedk:
## Summary
- Replace the 14 per-window `UiUtil.SetupWindowsSystemMenu(Window)` calls with a single class-level `KeyDown` handler on `Window`, registered once at app startup in `Program.ConfigureApplication`.
- Every top-level `Window` (current and future) now responds to Alt+Space on Windows automatically — no per-window opt-in required. No-op on macOS/Linux.
- Renames `SetupWindowsSystemMenu` to `RegisterWindowsSystemMenuClassHandler`; idempotent via a static guard.

## Test plan
- [ ] Build is clean (verified locally: 0 warnings, 0 errors).
- [ ] On Windows, Alt+Space opens the standard system menu (Move/Size/Min/Max/Close) on:
  - [ ] Main window
  - [ ] Settings, Shortcuts, Settings Import/Export
  - [ ] OCR, Binary OCR character add, NOCR character add
  - [ ] Multiple Replace, BinaryEdit, ShowImage
  - [ ] Undocked audio visualizer, undocked video player
  - [ ] ASSA Draw, ASSA Apply Advanced Effect
  - [ ] A window that previously did **not** opt in (e.g. Batch Convert, Fix Common Errors, Auto-Translate) — should now also respond to Alt+Space.
- [ ] On macOS/Linux, no behavior change.

### DIFF
--- a/src/ui/Features/Assa/AssaApplyAdvancedEffect/AssaApplyAdvancedEffectViewModel.cs
+++ b/src/ui/Features/Assa/AssaApplyAdvancedEffect/AssaApplyAdvancedEffectViewModel.cs
@@ -303,7 +303,6 @@ public partial class AssaApplyAdvancedEffectViewModel : ObservableObject
     internal async void OnLoaded()
     {
         UiUtil.RestoreWindowPosition(Window);
-        UiUtil.SetupWindowsSystemMenu(Window);
 
         if (string.IsNullOrEmpty(_videoFileName))
         {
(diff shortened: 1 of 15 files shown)


GrampaWildWilly:
Since I griped about this so long ago during the Previews, I suppose I ought to come in here to say this does appear to be working.  I still find it a bit odd that when I Alt+Spacebar on my video player on monitor #2 that the menu pops up on monitor #1 because that's where my mouse happens to be positioned.  The menu pops up near the mouse cursor.  But I tried it with a few windows (settings, Speech to Text, audio visualizer ...) & they all worked.

PR #10987, opened 2026-05-16
Add workflow_dispatch trigger to stale workflow

niksedk:
## Summary
- Adds a `workflow_dispatch` trigger to the stale workflow so it can be kicked off on demand instead of only on the weekly Tuesday cron.

## Test plan
- [ ] After merge, run the workflow manually via `gh workflow run stale.yml` (or the Actions tab) and confirm it executes.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -2,6 +2,7 @@ name: "Close stale issues"
 on:
   schedule:
     - cron: '0 0 * * 2'
+  workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


GrampaWildWilly:
Just thought I'd let you know this appears to have worked.  I suddenly got almost 200 of these stinkers in my inbox.  Not complaining.  It will happen only once.  Cleaning up moribund junk is a good idea.  Just a bit startling.  I pity those who don't have high speed connections.

PR #11009, opened 2026-05-17
seconv: add --input-encoding-fallback for non-UTF-8 input

### LINKED ISSUE #11002: [Feature Request] A CLI parameter for seconv.exe to deal with input encoding
I'm the one who [requested a "keep source encoding" option for old SE's CLI](https://github.com/SubtitleEdit/subtitleedit/issues/6707). To get that out of the way, does seconv.exe support `--encoding:source`? That would be the contingency plan, as this request is about "doing the right thing" and converting everything to UTF8.

The reasoning behind my request was that encoding detection heuristics sometimes failed. The exact case I've encountered is "Central European" getting detected as "Western European" and resulting in bad UTF8 subs.

I'm proposing an `--assumed-input-encoding` parameter for seconv that should work the way video players do it:
- if seconv is 100% confident it detected the encoding, it should use it's detected input encoding. From what i gather, UTF8 detection, with or without BOM, should be basically infallible.
- if seconv has to make assumptions, fall back to the input encoding set by `--assumed-input-encoding`

Most people will deal with a mixture of UTF8 and a specific encoding for their region, so this should allow bulk conversion with confidence to UTF8.

Thank you!

niksedk:
## Summary
- Adds `--input-encoding-fallback <name>` to seconv. The ANSI codepage heuristic in `LanguageAutoDetect.GetEncodingFromFile` occasionally misdetects Central European content as Western European; the new flag lets the user name an encoding to use **only** when BOM detection and UTF-8 byte-pattern detection both come up empty.
- BOM still wins (UTF-8, UTF-16 LE/BE, UTF-32 LE/BE) and valid UTF-8 without BOM still wins. The fallback only replaces the heuristic step, so bulk conversion of mixed UTF-8 / region-specific files lands consistently on UTF-8.
- Ignored when `--encoding` is set (that already forces input encoding too).

Closes #11002.

## Example
```
seconv *.srt subrip --input-encoding-fallback:windows-1250
```

## Test plan
- [x] Unit tests added in `tests/seconv/Core/InputEncodingFallbackTest.cs` covering: UTF-8 BOM wins over fallback; UTF-8 no-BOM with valid multi-byte wins over fallback; CP1250 bytes (no BOM, not valid UTF-8) → fallback; pure ASCII → fallback (decoded text identical either way); end-to-end `ConvertAsync` round-trip from CP1250 input to UTF-8 output with correct `Cześć ąęłóż.`.
- [x] Full seconv test suite: 134/135 pass. The one failure (`ContainerLoaderTest.ConvertAsync_MkvWithImageTracks_SkipsAndReportsZeroSuccess`) is pre-existing on `main`, unrelated to this change.
- [x] CLI smoke test: a CP1250-encoded `.srt` converts to UTF-8 BOM with all Polish diacritics correct.

### DIFF
--- a/src/seconv/Commands/ConvertCommand.cs
+++ b/src/seconv/Commands/ConvertCommand.cs
@@ -276,6 +276,17 @@ internal sealed class ConvertCommand : AsyncCommand<ConvertCommand.Settings>
                 return 1;
             }
 
+            // Fail fast on a typo in --input-encoding-fallback so we don't silently substitute
+            // UTF-8 and decode the input incorrectly without the user noticing.
+            if (!string.IsNullOrWhiteSpace(settings.InputEncodingFallback) &&
+                !LibSEIntegration.TryGetEncoding(settings.InputEncodingFallback, out _))
+            {
+                AnsiConsole.MarkupLine(
+                    $"[red]Error: Unknown encoding '{settings.InputEncodingFallback.EscapeMarkup()}' for --input-encoding-fallback.[/]");
+                AnsiConsole.MarkupLine("[dim]Use 'seconv list-encodings' to see supported encodings.[/]");
+                return 1;
+            }
+
             // Load --settings:path.json overrides into libse Configuration before any conversion.
             // --profile <name> selects a named overlay from the same file.
             if (!string.IsNullOrWhiteSpace(settings.SettingsPath))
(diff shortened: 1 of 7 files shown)


Hlsgs:
@niksedk  Thank you for this! I'll be testing the next beta.

Couple of quick questions:
1. Is `--encoding:source` still supported as it was with old SE? Would like this to be the case for when dealing with whatever edge cases or frankenstein subs one may encounter.
2. 

> * Ignored when `--encoding` is set (that already forces input encoding too).

Why would `--encoding` force input as well, especially if `--encoding:source` and `--input-encoding-fallback` exist? If `--encoding` forces input, one basically can't bulk convert to anything but UTF8.

PR #11010, opened 2026-05-17
Tail the speech-to-text console log in batch mode too

niksedk:
## Summary
- Follow-up to #10974. That PR made the inner-`ScrollViewer.ScrollToEnd()` call work for single-mode users, but batch-mode users still saw the log freeze at the top.
- Root cause: `SpeechToTextWindow` builds two console views — `MakeConsoleLogAndBatchView` (visible when `IsBatchMode`) and `MakeConsoleLogOnlyView` (visible otherwise) — each with its own `TextBox`. Both assigned `vm.TextBoxConsoleLog = textBoxConsoleLog`, so the second assignment (single-mode) won. In batch mode the *visible* TextBox was the other one and `LogToConsole` was scrolling the hidden TextBox.
- Split the VM field into `TextBoxConsoleLogBatch` / `TextBoxConsoleLogSingle`; `LogToConsole` calls `ScrollToEnd` on both. The hidden one is a harmless no-op so the visible one always tails regardless of mode.

## Test plan
- [ ] Start a transcription in **batch** mode (multiple files) and confirm the console log tails new ffmpeg / whisper lines without needing to focus the box or press PageDown.
- [ ] Start a transcription in **single** mode and confirm the existing single-mode tail behavior still works.
- [ ] Toggle batch ↔ single between runs and confirm both still tail.

### DIFF
--- a/src/ui/Features/Video/SpeechToText/SpeechToTextViewModel.cs
+++ b/src/ui/Features/Video/SpeechToText/SpeechToTextViewModel.cs
@@ -106,7 +106,12 @@ public partial class SpeechToTextViewModel : ObservableObject
     public Subtitle TranscribedSubtitle { get; private set; }
     public List<AudioClip> ResultAudioClips { get; private set; }
     public string? LastBatchSubtitleFileName { get; private set; }
-    public TextBox TextBoxConsoleLog { get; internal set; }
+    // Two views host the console log — the batch-mode grid layout and the single-mode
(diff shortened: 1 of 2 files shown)


GrampaWildWilly:
A visible console & a hidden console?  This sounds laughably complicated.  But on the plus side, CoPilot seems to like what you did.  I can hardly wait for beta 29.

PR #11145, opened 2026-05-24
Fix #11144: translate Mac Cmd shortcut to Ctrl on Windows import

### LINKED ISSUE #11144: [BETA-31] Importing settings form MacOS to Windows version sets  the WIN key for shortcuts instead of CTRL.
Importing shortcuts settings form MACOS to Windows version sets  the WIN key instead of CTRL for default shortcuts.

niksedk:
## Summary
- Fixes #11144 — importing a shortcuts/settings file exported from macOS into the Windows build left default shortcuts bound to the Windows/Meta key instead of Ctrl.
- The cause: macOS stores the Cmd/⌘ modifier as `"Win"` (see `ShortcutsMain.GetCommandOrWin`), while Windows/Linux store the same default as `"Ctrl"`. The importer copied keys verbatim, so on Windows the Mac `"Win"` modifier was interpreted as the actual Windows key.
- Fix: in `SettingsImportExportViewModel.ImportSettings`, translate the modifier to the host OS's expected default on import — `"Win"` → `"Ctrl"` on Windows/Linux, `"Ctrl"` → `"Win"` on macOS.

## Test plan
- [ ] Export settings on macOS (default shortcuts use Cmd/⌘), import on Windows: verify default shortcuts use Ctrl (e.g. Ctrl+S Save, Ctrl+O Open, Ctrl+Z Undo) rather than the Windows key.
- [ ] Export settings on Windows, import on macOS: verify default shortcuts use Cmd/⌘ rather than Ctrl.
- [ ] Round-trip on a single platform leaves shortcuts unchanged.

### DIFF
--- a/src/ui/Features/Options/Settings/SettingsImportExport/SettingsImportExportViewModel.cs
+++ b/src/ui/Features/Options/Settings/SettingsImportExport/SettingsImportExportViewModel.cs
@@ -6,6 +6,8 @@ using CommunityToolkit.Mvvm.Input;
 using Nikse.SubtitleEdit.Logic;
 using Nikse.SubtitleEdit.Logic.Config;
 using Nikse.SubtitleEdit.Logic.Media;
+using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Text.Json;
 using System.Threading.Tasks;
@@ -355,6 +357,7 @@ public partial class SettingsImportExportViewModel : ObservableObject
         {
             if (importData.Shortcuts != null)
             {
+                NormalizeShortcutModifiersForCurrentOs(importData.Shortcuts);
                 Se.Settings.Shortcuts = importData.Shortcuts;
             }
         }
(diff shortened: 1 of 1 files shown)


niksedk:
Thanks @copilot — good catch. The original patch would have corrupted same-OS round-trips (e.g. a user-customized Ctrl shortcut on macOS, or a legitimate Win/Meta binding on Windows).

Pushed 390b97f11 to address it:
- Export now stamps the JSON with a top-level `exportSourceOs` marker (`Se` has no such field, so System.Text.Json silently ignores it on read — no schema change).
- Import only calls `NormalizeShortcutModifiersForCurrentOs` when the marker is present **and** differs from the host OS. Same-OS round-trip leaves shortcuts verbatim.
- Legacy files exported before this change have no marker and are also imported verbatim. Users hitting the original bug will need to re-export from macOS once on the fixed build for the cross-OS translation to take effect — chose this over a heuristic to preserve same-OS round-trip semantics for everyone.

PR #11150, opened 2026-05-24
Fix missing JSON encoding for prompt in DeepSeekTranslate

89799969:
When using a multi-line or custom prompt in DeepSeek translation, the request fails (404/400) because the `prompt` variable in the JSON payload was missing `Json.EncodeJsonText()`.
     
     This PR wraps the prompt variable with `Json.EncodeJsonText(prompt)`, aligning the logic with `ChatGptTranslate` and others.


niksedk:
thx :)

PR #11177, opened 2026-05-25
Pre-RC1 cleanup: resource leaks, plugin Update-All cancel, dup-row bug

niksedk:
## Summary

Fixes 7 issues found in a pre-RC1 review of recent changes (TTS Cast, TTS Review window, STT chunking, plugin manager, SE 4 shortcut import).

- **Resource cleanup on close** in `ReviewSpeechViewModel`, `ActorVoiceMappingViewModel`, `DownloadTtsViewModel` — dispose `CancellationTokenSource` + `Timer`, unsubscribe `PropertyChanged` on the waveform mirror VMs, and drain the multi-GB download `MemoryStream`s when the download window closes mid-flight.
- **Review window dup-row bug**: `ReviewSpeechViewModel.Ok()` had a leftover loop that re-added every step result to `Lines`, doubling the grid on OK. Removed.
- **STT chunked upload**: each ~23 MB chunk WAV is now deleted as soon as the upload finishes (instead of accumulating in temp for the whole run). `OpenAiSttService` now `using`s the `HttpResponseMessage` so the SSE path doesn't leak the connection.
- **Plugin Update-All**: now uses a real `CancellationTokenSource` through the loop (was `CancellationToken.None`, so cancel was a no-op). Added one more `ThrowIfCancellationRequested` between the targetPath delete and Move in `PluginDownloadService` so cancelling mid-install can't leave a user with no plugin where one existed.
- **Cross-OS shortcut import**: `NormalizeShortcutModifiersForCurrentOs` now also maps `"Control"` → `"Win"` on macOS. The SE 4 importer normalises to `"Control"` (Avalonia token) and historical SE 5 settings can hold either spelling; `ShortcutManager` treats them as the same modifier but the cross-OS rewrite was only matching `"Ctrl"`.
- **`ReviewSpeechViewModel.TrimAndAdjustSpeed`** temp WAVs (trim / VAD / time-stretched) are tracked and deleted on window close.

No behavior change for the happy path of any feature; everything here is leak/race/cancellation hardening plus one user-visible bug (review-window OK doubling the row count).

## Test plan
- [x] `dotnet test tests/UI/UITests.csproj` (touched-area filter) — 63/63 pass
- [ ] Manual smoke: open Cast dialog, test a voice, close → no leaked CTS
- [ ] Manual smoke: TTS Review window, regenerate a few rows, OK → grid not doubled, `_waveFolder` cleaned
- [ ] Manual smoke: STT against an OpenAI-compatible endpoint with a >25 MB file → temp/ stays clean
- [ ] Manual smoke: Plugin manager → Update All → cancel mid-loop → no half-installed plugin

### DIFF
--- a/src/ui/Features/Options/Plugins/PluginManagerViewModel.cs
+++ b/src/ui/Features/Options/Plugins/PluginManagerViewModel.cs
@@ -36,6 +36,7 @@ public partial class PluginManagerViewModel : ObservableObject
     private readonly IFolderHelper _folderHelper;
     private readonly IWindowService _windowService;
     private CancellationTokenSource? _checkCts;
+    private CancellationTokenSource? _updateAllCts;
 
     public PluginManagerViewModel(IPluginCatalog pluginCatalog, IPluginDownloadService downloadService, IFolderHelper folderHelper, IWindowService windowService)
     {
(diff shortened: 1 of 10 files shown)


niksedk:
@copilot review

PR #11178, opened 2026-05-25
Fix #11176: dispose mpv off the UI thread to unstick BurnIn dialogs

### LINKED ISSUE #11176: Adding a logo to the output video freezes the interface and it cannot be closed in any way.
When I open the "Add logo to output video" tab, the interface freezes, whether I add an image by saving the action and clicking "OK," or even without adding anything by simply clicking "Cancel." It cannot be closed in any way; it becomes completely unresponsive, forcing me to close it from the Windows Task Manager.

Image

niksedk:
Fixes #11176.

## Diagnosis

`LibMpvDynamicSoftwareControl.OnDetachedFromVisualTree` disposes its `LibMpvDynamicPlayer` synchronously on the UI thread:

```csharp
_mpvPlayer.Dispose();   // → mpv_terminate_destroy()
```

`mpv_terminate_destroy` stops every mpv worker and blocks until they exit. When mpv is idle it returns in milliseconds; when a file load is still in flight on a slow / stuck path (network mount, weird codec, demuxer hung in I/O) it can stall the calling thread for many seconds. That's the UI thread, on every close of the BurnIn logo / effect / select-position dialogs (each spins up a fresh `LibMpvDynamicPlayer`).

Matches the reporter's symptoms exactly: dialog opens, freezes, refuses to close, only Task Manager helps. Doesn't reproduce on machines where mpv terminates quickly.

## Fix

Detach the render callback synchronously (so the control becomes collectible immediately) but hand the native dispose to `Task.Run`. The OS reclaims the mpv resources at its own pace without holding the UI hostage.

Also wrap `BurnInLogoViewModel.OnLoaded` (`async void`) in try/catch so a throw from `Open` / `WaitForPlayersReadyAsync` doesn't bubble into the dispatcher's unhandled-exception path.

## Caveats
I can't reproduce the freeze locally, so this diagnosis is inferred from reading the close path against the reporter's symptoms. That said, the change is a strict win for any mpv state where `terminate_destroy` isn't instant — the UI thread shouldn't be blocking on native worker shutdown regardless.

## Test plan
- [ ] Manual smoke: open + close BurnIn logo dialog several times on a normal video → no regression
- [ ] Manual smoke: same on the dialog that triggered #11176 → reporter to confirm the freeze is gone

### DIFF
--- a/src/ui/Features/Video/BurnIn/BurnInLogoViewModel.cs
+++ b/src/ui/Features/Video/BurnIn/BurnInLogoViewModel.cs
@@ -255,42 +255,51 @@ public partial class BurnInLogoViewModel : ObservableObject
 
     internal async void OnLoaded()
     {
-        if (!string.IsNullOrEmpty(BurnInLogo.LogoFileName) && File.Exists(BurnInLogo.LogoFileName))
+        // async void: any throw here lands in the dispatcher's unhandled-exception
+        // path. Catch so a bad logo file or a stuck video open can't crash the app.
+        try
         {
-            LogoFileName = BurnInLogo.LogoFileName;
-            await LoadLogo();
+            if (!string.IsNullOrEmpty(BurnInLogo.LogoFileName) && File.Exists(BurnInLogo.LogoFileName))
+            {
+                LogoFileName = BurnInLogo.LogoFileName;
+                await LoadLogo();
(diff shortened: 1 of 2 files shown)


niksedk:
@copilot review

PR #11180, opened 2026-05-25
Fix 3 RC1 bugs: negative video offset, DCinema clamp, F4Text rounding

niksedk:
Three independent bugs found in a pre-RC1 sweep.

## 1. `AudioVisualizer` waveform labels ignore negative video offsets
`AudioVisualizer.cs:1711, 1723` — `if (Se.Settings.General.CurrentVideoOffsetInMs > 0.00001)` silently dropped every negative offset. `CurrentVideoOffsetInMs` is signed (users can set negative offsets when the audio leads the video), so the waveform's timeline labels and frame-aligned gridlines were wrong for those files. Introduced in 9bc6afaff (frame-aligned gridlines). Fix: `Math.Abs(...) > 0.00001`.

## 2. `DCinemaSmpte{2010,2014}.MsToFramesMaxFrameRate` clamps against the wrong frame rate
The overflow guard compared `frames >= Configuration.Settings.General.CurrentFrameRate` instead of the method's `frameRate` **parameter**. Whenever the export frame rate differed from the active project setting (e.g., a 24 fps DCinema export with the project at 25 fps), the clamp either fired when it shouldn't or failed to fire when it should — producing an out-of-range frame number on the last frame of every second. Fix: compare against the parameter.

## 3. `F4Text` writes invalid timecodes for 950–999 ms
`F4Text.cs:49` used `Math.Round(time.Milliseconds / 100.0, 0):0`. With banker's rounding, 950..999 ms rounds to 10 deciseconds, and `:0` formats that as `"10"`. The format regex on L15 requires exactly **one** decisecond digit — so the writer was emitting files that the parser silently dropped on reload. Affected the last 50 ms of every second. Fix: clamp to 9 before formatting.

## Tests
- `tests/libse/SubtitleFormats/DCinemaSmpteTest.cs` — 2 cases (2010 + 2014). Each sets a global frame rate that differs from the parameter, calls the helper, asserts the clamp uses the parameter.
- `tests/libse/SubtitleFormats/F4TextTest.cs` — 5 `[InlineData]` cases (0, 499, 949, 950, 999). Encodes through `ToF4Text` and asserts every emitted timecode matches the single-digit format regex.

Bug 1 lives in a private static helper on a UI control with no UI test harness in this repo, so it's covered by code review only.

## Test plan
- [x] `dotnet test tests/libse/LibSETests.csproj` — 352/352 pass (7 new)
- [ ] Manual: set a negative video offset, open waveform → timeline labels reflect the offset

### DIFF
--- a/src/libse/SubtitleFormats/DCinemaSmpte2010.cs
+++ b/src/libse/SubtitleFormats/DCinemaSmpte2010.cs
@@ -1054,7 +1054,10 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
         public static int MsToFramesMaxFrameRate(double milliseconds, double frameRate)
         {
             var frames = (int)Math.Round(milliseconds / (TimeCode.BaseUnit / frameRate));
-            if (frames >= Configuration.Settings.General.CurrentFrameRate)
+            // Clamp against the parameter, not the global CurrentFrameRate setting.
+            // Using the global produced the wrong cap when exporting at a frame
+            // rate that differed from the active project setting.
+            if (frames >= frameRate)
             {
                 frames = (int)(frameRate - 0.01);
             }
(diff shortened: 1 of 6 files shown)


niksedk:
@copilot review

PR #11182, opened 2026-05-25
Defensive: BluRaySup OdsData + Matroska short-read on truncated files

niksedk:
Two unrelated parser bugs surfaced while pre-RC1 fuzzing.

## 1. `BluRaySupParser.ParseOds` crashes on malformed/truncated PGS

`ParseOds` did `new byte[segment.Size - 11]` (or `- 4` for a continuation fragment) without lower-bound checking `segment.Size`. A malformed / truncated PGS whose ODS segment advertised a `Size` smaller than the 11-byte header (or 4-byte continuation header) produced a negative length, and `new byte[..]` rejected it with `OverflowException` — aborting the whole SUP parse instead of skipping the bad segment. Now we return an empty fragment and let the caller move on.

## 2. `MatroskaFile.ReadUIntAsInt` / `ReadUIntAsLong` ignore short reads

Both methods called `_stream.Read(_buffer, 0, length)` but ignored the return value. `Stream.Read` may legitimately return fewer bytes than requested (truncated file, network stream, slow pipe). The old code folded stale `_buffer` bytes from prior reads into the integer — **silent garbage track numbers / pixel dimensions / durations on truncated MKV files**. Now we return 0 when the read comes up short.

## Tests
Couldn't add a regression test for either, and shipping a misleading test is worse than no test:
- BluRay: `ParseOds` is only dispatched once a valid PCS has been processed earlier in the stream; constructing the full PCS prelude is a lot of scaffolding for one assertion.
- Matroska: `ReadUIntAsInt` is private; testing requires a `Stream` wrapper that fakes short reads, which is its own surface area.

Both fixes are pure defensive bounds-checking around existing invariants.

## Test plan
- [x] `dotnet test tests/libse/LibSETests.csproj` — 352/352 pass
- [x] `dotnet build src/libse/LibSE.csproj` — clean

### DIFF
--- a/src/libse/BluRaySup/BluRaySupParser.cs
+++ b/src/libse/BluRaySup/BluRaySupParser.cs
@@ -720,6 +720,16 @@ namespace Nikse.SubtitleEdit.Core.BluRaySup
             var info = new ImageObjectFragment();
             if (first)
             {
+                // First-fragment ODS layout: 11 bytes of header (object id+ver+seq+
+                // 3-byte length + 2-byte width + 2-byte height) before the image
+                // bytes. A malformed / truncated PGS file can advertise a Size
+                // smaller than that — `new byte[Size - 11]` then throws
+                // OverflowException and aborts the whole SUP parse. Bail out with
+                // an empty fragment instead so the rest of the file is still read.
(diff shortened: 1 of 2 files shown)


niksedk:
@copilot review

PR #11184, opened 2026-05-25
Async fixes: flow CancellationToken, drop dup timer sub, guard OnClosing

niksedk:
Three independent async correctness bugs found in a pre-RC1 sweep.

## 1. AutoTranslate providers ignore `CancellationToken` (3 files)
`MyMemoryApi`, `NoLanguageLeftBehindApi`, and `LaraTranslate` all accept a `CancellationToken` in their `Translate(...)` signature but never flow it into the actual HTTP call. User clicks Cancel mid-translation → the HTTP request keeps running to completion.

Caveat: `libse` targets `netstandard2.1`, where `ReadAsStringAsync(CancellationToken)` / `ReadAsByteArrayAsync(CancellationToken)` / `GetStringAsync(string, CancellationToken)` don't exist. So we flow the token through `SendAsync` / `PostAsync` (the long phase) and leave the body-read calls bare. `MyMemoryApi` was using `GetStringAsync` — switched to an explicit `SendAsync` so the token has a place to go.

## 2. `DownloadSpeechToTextModelsViewModel`: timer handler subscribed twice
The constructor wires `_timer.Elapsed += OnTimerOnElapsed;`, and so did `Download()`. Cancel-and-retry on a download caused the state-machine tick to fire **2x per timer interval** (then 3x, 4x...). The doubled invocations race the file-move / index-advance and produce flaky behaviour — double "download failed" toasts, files moved twice, progress jumping back. Dropped the redundant re-subscription.

## 3. `MainViewModel.OnClosing` async-void event handler with no try/catch
`OnClosing` is `async void` with two awaits (`MessageBox.Show`, `SaveSubtitle`). Both can throw — disk full, permission denied, OneDrive sync conflict. With `async void`, an unhandled exception bypasses dispatcher flow and the user is left staring at a half-closed window with `e.Cancel = true`. Wrapped in try/catch + log + force-close. Same shape as the `BurnInLogoViewModel.OnLoaded` guard from PR #11178.

## Out of scope (kept narrow)
- `MicrosoftTranslator.cs:136, :158` use `.Result` on async (sync-over-async deadlock risk). Fixing them properly cascades through several private static call sites — separate PR if anyone wants to chase it.

## Test plan
- [x] `dotnet build src/libse/LibSE.csproj` + `dotnet build src/ui/UI.csproj` — clean, 0 warnings
- [x] `dotnet test tests/libse/LibSETests.csproj` — 352/352 pass
- [ ] Manual: cancel an in-flight MyMemory / NLLB / Lara translation → request actually stops
- [ ] Manual: download a Whisper model, cancel, re-download → no "doubled" UI updates
- [ ] Manual: close main window with unsaved changes and trigger a save failure → window still closes cleanly

### DIFF
--- a/src/libse/AutoTranslate/LaraTranslate.cs
+++ b/src/libse/AutoTranslate/LaraTranslate.cs
@@ -84,7 +84,7 @@ namespace Nikse.SubtitleEdit.Core.AutoTranslate
             // netstandard2.1: ReadAsByteArrayAsync has no CancellationToken
             // overload; the token is flowed through SendAsync, which is where
             // the long blocking happens.
-            var response = await _httpClient.SendAsync(request, cancellationToken);
+            using var response = await _httpClient.SendAsync(request, cancellationToken);
 
             var bytes = await response.Content.ReadAsByteArrayAsync();
             var json = Encoding.UTF8.GetString(bytes).Trim();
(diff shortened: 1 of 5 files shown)


niksedk:
@copilot review

PR #11185, opened 2026-05-25
Perf: _slowTimer layer-filter — RemoveAll instead of Where+Clear+AddRange

niksedk:
Tiny perf cleanup in `MainViewModel`'s 400 ms slow timer.

## Change
The Hide-Layers-From-Video-Preview branch used `Where().ToList() + Paragraphs.Clear() + AddRange()` to keep only paragraphs on visible layers. That allocates an extra `List`, walks the collection three times, and was duplicated for the LibMpv and LibVlc paths (each re-reading `Se.Settings.Assa.HideLayersFromVideoPreview`).

Hoisted the setting read once, then `subtitle.Paragraphs.RemoveAll(p => !_visibleLayers!.Contains(p.Layer))` — single in-place walk, zero extra `List`.

## Validation
Ran BenchmarkDotNet (0.14.0, .NET 10, ~80 % of paragraphs retained — realistic for a layer filter):

| Subtitles | `Where + ToList + Clear + AddRange` | `RemoveAll` | Speedup | Alloc savings |
|---|---|---|---|---|
| 100  |  1.55 µs |  0.75 µs | **2.0×** | 1760 B → 0 B |
| 1000 | 12.30 µs |  7.17 µs | **1.7×** | 14584 B → 8056 B |
| 5000 | 69.43 µs | 40.87 µs | **1.7×** | 72256 B → 0 B (ServerGC) |

(Two other candidate optimizations were also benchmarked but rejected — one was actually slower in the realistic input range, one was too cosmetic to ship alone.)

## Test plan
- [ ] Manual: play a video with HideLayersFromVideoPreview enabled and multiple ASSA layers — verify the visible layers still filter correctly

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -16485,31 +16485,37 @@ public partial class MainViewModel :
         {
             UpdateTitleStatus();
             UpdateGaps();
+
             var vp = GetVideoPlayerControl();
-            if (_mpvPreviewDirty && vp?.VideoPlayer is LibMpvDynamicPlayer mpv)
+            if (!_mpvPreviewDirty || vp == null)
+            {
+                return;
+            }
+
+            // Filter once instead of: Where().ToList() + Clear + AddRange (which
+            // allocates an extra List and walks the paragraphs three times).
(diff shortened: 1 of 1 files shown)


niksedk:
@copilot review

PR #11187, opened 2026-05-25
feat(mp4): decode CEA-708 (DTVCC) captions from H.264 SEI cc_data

niksedk:
Before this PR, MP4 files with CEA-708 captions (DTVCC packets in H.264 SEI cc_data) came in as "no subtitle track found" — even though SE already ships a working `Cea708` decoder for MCC / MacCaption files. The MP4 parser detected and counted the type-2/3 triplets, then dropped them.

Wire the existing decoder into the MP4 path.

## Changes

**`GetCcDataHelper`** (`src/libse/Cea608/GetCcDataHelper.cs`)
- `IsCcType` now accepts types 2 and 3 (was 0/1 only); callers filter by `CcData.Type` to route 608 vs 708.
- The "non-empty cc data" pair filter (low-7-bits of both > 0) is a CEA-608 parity convention. For CEA-708 packet data, `0x80` / `0x00` are legitimate payload bytes (window-bitmap arguments, padding) and dropping them corrupts the DTVCC packet. Scope the filter to types 0/1.

**`Mp4Parser`** (`src/libse/ContainerFormats/Mp4/Mp4Parser.cs`)
- New `TrunCea708Subtitle` property mirroring `TrunCea608Subtitle`.
- `CheckForMoovVideoCea608` splits `sortedCcData` by type and runs both decoders; the `cea608Parser` only sees 608 entries so its state machine isn't corrupted by 708 triplets.
- New `DecodeMoovVideoCea708` groups `cc_type==2` service-block bytes by PTS frame and feeds them to `Cea708.Decode`, mirroring the `CcDataSection.GetText` pattern that `MacCaption10` has used for years (decoder is tolerant of starting partway through a packet; the `cc_type==3` second byte is treated as packet header and skipped).

**`MainViewModel`** — fall back to `TrunCea708Subtitle` when no CEA-608 or text-subtitle track was found.

## Drive-by bug fix
`Mdia.cs:36` had missing parens — `Name == "minf" && IsTextSubtitle || IsVobSubSubtitle || IsClosedCaption || IsVideo` parsed as `(minf && TextSubtitle) || VobSub || ClosedCaption || IsVideo`. On a video track the `IsVideo` term made the branch fire for the `mdhd` box too, hijacking mdhd parsing and leaving `Mdhd` null. That forced the default `timescale = 90000` for video tracks — caption timing was scaled by `90000:actual`. Required for the new CEA-708 timing assertion, and quietly fixes timing for the existing CEA-608 path too.

## Tests
- New `Mp4Cea708H264Test` builds a 2-sample H.264-in-MP4 with `cc_type==2` triplets carrying `'H' 'i' '!' + HideWindows`, verifies the decoder produces a single `"Hi!"` paragraph spanning frame 0 → frame 1.
- All 353 existing libse tests still pass.

## Limitations
This is a best-effort first pass:
- Service block parsing is loose (skips the `cc_type==3` header byte, like MacCaption). Real-world service-1-only captions work; multi-service or non-standard service numbering is not yet supported.
- Timing per caption is frame-based (from PTS); a real CEA-708 implementation would track per-window display/hide commands for fine-grained start/end. Good enough for a usable subtitle round-trip.
- DVB-style `clcp/c708` tracks in `stsd` are still not handled (separate from H.264 SEI).

These can be follow-ups. The current implementation covers the common case (US broadcast MP4s with H.264 SEI captions).

## Test plan
- [x] `dotnet test tests/libse/LibSETests.csproj` — 353/353 pass
- [ ] Manual: open a real CEA-708 MP4 (US broadcast capture, no tx3g track) → captions surface in SE

### DIFF
--- a/src/libse/Cea608/GetCcDataHelper.cs
+++ b/src/libse/Cea608/GetCcDataHelper.cs
@@ -93,7 +93,15 @@ namespace Nikse.SubtitleEdit.Core.Cea608
                             {
                                 var ccData1 = buffer[i + 1];
                                 var ccData2 = buffer[i + 2];
-                                if (IsNonEmptyCcData(ccData1, ccData2))
+                                // The "non-empty" filter is a CEA-608 convention
+                                // (high bit is parity; both bytes need non-zero
+                                // low-7-bits to be a meaningful char pair). For
+                                // CEA-708 packet data (types 2/3), 0x80 / 0x00
+                                // are legal payload bytes — e.g. a window-bitmap
+                                // argument — and dropping them corrupts the
+                                // DTVCC packet. Scope the filter to CEA-608.
+                                var isCea608 = ccType == 0 || ccType == 1;
+                                if (!isCea608 || IsNonEmptyCcData(ccData1, ccData2))
                                 {
                                     fieldData.Add(new CcData(ccType, ccData1, ccData2));
                                 }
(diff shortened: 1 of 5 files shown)


niksedk:
@copilot review

PR #11210, opened 2026-05-27
Fix FindNext crash: null ScrollIntoView, stale _findService state, missing ClearKeys

mjuhasz:
##### Problem
      
`FindNext()` had four bugs that caused crashes and incorrect behavior when using the Find Next keyboard shortcut:
      
1. **`ScrollIntoView` crash** — After `SubtitleGrid.SelectedIndex = idx`, the code called `SubtitleGrid.ScrollIntoView(SubtitleGrid.SelectedItem, null)`. In Avalonia, setting the index does not synchronously update `SelectedItem`; it remains `null` until the control processes the change, causing a null-reference crash.
      
2. **Stale `_findService` state in the UI callback** — The `Dispatcher.UIThread.Post` lambda read `_findService.CurrentTextFound / .CurrentLineNumber / .CurrentTextIndex` *after* being posted. If the user pressed Find Next a second time before the callback ran, `_findService` would already hold results for the newer search, causing the wrong subtitle to be highlighted and the wrong text range to be selected.
      
3. **Missing `ClearKeys()` on the not-found path** — When `idx < 0` (search term not found), the method returned early without calling `_shortcutManager.ClearKeys()`. This left the shortcut key state stuck, so the shortcut could not re-trigger after a failed search.
      
4. **`EditTextBox` not populated when empty** — Selection indices were applied even when `EditTextBox.Text` was still empty (subtitle not yet loaded into the box), so nothing was visually selected.
      
##### Fix
      
- Look up the subtitle via `Subtitles.GetOrNull(idx)` before entering the dispatcher callback; add a null guard and set both `SelectedIndex` and `SelectedItem` explicitly, then pass the subtitle directly to `ScrollIntoView` — matching the pattern already used in `HandleFindResult()`.
- Capture `CurrentTextFound`, `CurrentLineNumber`, and `CurrentTextIndex` into local variables *before* `Post`, so the lambda closes over stable values.
- Add `ClearKeys()` before the early return on not-found.
- Add a guard to populate `EditTextBox.Text` from `subtitle.Text` when the box is empty — same defensive check already in `HandleFindResult()`.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -9272,24 +9272,40 @@ public partial class MainViewModel :
         if (idx < 0)
         {
             ShowStatus(string.Format(Se.Language.General.XNotFound, _findService.SearchText));
+            _shortcutManager.ClearKeys();
             return;
         }
 
+        var foundText = _findService.CurrentTextFound;
+        var foundLine = _findService.CurrentLineNumber;
+        var foundIndex = _findService.CurrentTextIndex;
+
         Dispatcher.UIThread.Post(() =>
         {
-            SubtitleGrid.SelectedIndex = idx;
-            SubtitleGrid.ScrollIntoView(SubtitleGrid.SelectedItem, null);
+            var subtitle = Subtitles.GetOrNull(idx);
+            if (subtitle == null)
+            {
+                return;
+            }
 
-            ShowStatus(string.Format(Se.Language.General.FoundXInLineYZ, _findService.CurrentTextFound, _findService.CurrentLineNumber + 1, _findService.CurrentTextIndex + 1));
+            SubtitleGrid.SelectedIndex = idx;
+            SubtitleGrid.SelectedItem = subtitle;
+            SubtitleGrid.ScrollIntoView(subtitle, null);
+
+            ShowStatus(string.Format(Se.Language.General.FoundXInLineYZ, foundText, foundLine + 1, foundIndex + 1));
 
             // wait for text box to update
-            Task.Delay(50);
+            Task.Delay(75);
 
-            EditTextBox.CaretIndex = _findService.CurrentTextIndex;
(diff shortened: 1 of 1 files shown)


niksedk:
Can you test if we can remove the ClearKeys calls?

mjuhasz:
> Can you test if we can remove the ClearKeys calls?

We could probably remove it for FindNext/FindPrevious specifically, but I'd say it's not worth it.

Key-up events already clean up state one key at a time. ClearKeys is a bulk defensive reset for cases where key-up events are missed: most commonly when a dialog opens, steals focus, and the main window never receives the key-up events.

I'm sending another PR your way, #11220 with more keyboard shortcut fixes and in there stale key states were also causing jarring issues, like dialogs opening up out of nowhere.

FindNext and FindPrevious don't open dialogs, so their key-up events should arrive normally. That means ClearKeys is theoretically redundant there. But:

- The cost is zero (a HashSet.Clear and two bool assignments)
- Removing it gains nothing
- Edge cases are hard to enumerate: OS-level focus grabs, accessibility layers, external screen readers

To actually test removal, we'd need to verify these scenarios after stripping the calls:
1. Hold a modifier key (e.g. Ctrl), trigger FindNext via F3 shortcut, release Ctrl. Then press an unrelated shortcut. Does it misfire?
2. Trigger FindNext via key repeat (hold shortcut down). Does it keep firing?
3. Trigger FindNext, alt-tab away, alt-tab back, press another shortcut. Does stale F3 remain in _activeKeyNames and corrupt a lookup?

_Scenario 3 is the real one._ When you alt-tab, F3-up goes to the OS, not to OnKeyReleased. Without ClearKeys, F3 stays in _activeKeyNames until the next key event clears it. This could spuriously match a shortcut that includes F3.

Bottom line is, let's keep the ClearKeys calls. Low risk, zero cost, and the pattern exists for good reason across the entire ViewModel.

PR #11226, opened 2026-05-27
Restore focus correctly when Find/Replace dialog is dismissed

mjuhasz:
## Problem

  Dismissing the Find or Replace dialog (via Escape or the close button) left focus on the now-invisible dialog window. The user had to click back into the editor or the subtitle list before keyboard shortcuts or typing would work again.

This is inconsistent with SE 4 behavior, and usability suffers because of the extra clicking required.

 Also, focus edit text box after Find Next/Previous shortcut.

  ## Behaviour after this change

  **If a match was found and navigated to** (Find Next, Replace):
  → focus goes to the text editor, where the found/replaced text is selected.

  **If nothing was found, or the dialog was closed without searching, or Replace All was used** (which replaces in-place without navigating):
  → focus returns to whichever control was active before the dialog was opened — the subtitle list, a time field, the text editor, etc.

  The previously-active control is also refreshed when an already-open dialog is re-activated via the keyboard shortcut, so the restored focus always reflects the most recent invocation.

  For the Find Next/Previous shortcut, when text is found, call FocusEditTextBox() so the subtitle edit box receives focus and the highlighted match is immediately editable.

  ## Changes

  - `FindViewModel` / `ReplaceViewModel` — add `ResultFound` flag, set synchronously when a match is confirmed (before any dispatcher post, to avoid a race with the window-closed event).
  - `MainViewModel.ShowFind` / `ShowReplace` — capture the focused control before opening (and on each re-activation), attach a `window.Closed` handler that restores focus based on `ResultFound`.
  - call `FocusEditTextBox()` for the Find Next/Previous shortcut, when text is found

### DIFF
--- a/src/ui/Features/Edit/Find/FindViewModel.cs
+++ b/src/ui/Features/Edit/Find/FindViewModel.cs
@@ -24,6 +24,7 @@ public partial class FindViewModel : ObservableObject
 
     public bool FindNextPressed { get; private set; }
     public bool FindPreviousPressed { get; private set; }
+    public bool ResultFound { get; set; }
 
     private IFindService? _findService;
     private List<string> _subs = new List<string>();
--- a/src/ui/Features/Edit/Replace/ReplaceViewModel.cs
+++ b/src/ui/Features/Edit/Replace/ReplaceViewModel.cs
@@ -28,6 +28,7 @@ public partial class ReplaceViewModel : ObservableObject
     public bool FindNextPressed { get; private set; }
     public bool ReplacePressed { get; private set; }
     public bool ReplaceAllPressed { get; private set; }
+    public bool ResultFound { get; set; }
 
     private IFindService? _findService;
     private List<string> _subs = new List<string>();
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -284,6 +284,7 @@ public partial class MainViewModel :
     VideoPlayerUndockedViewModel? _videoPlayerUndockedViewModel;
     AudioVisualizerUndockedViewModel? _audioVisualizerUndockedViewModel;
     FindViewModel? _findViewModel;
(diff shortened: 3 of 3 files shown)


GrampaWildWilly:
Hmmmm.......  All this talk of getting focus right makes me want to remind you guys of this:

1. Undock the controls.
2. Alt+Tab to either the Audio visualizer or the video player.
3. Alt+Tab to the main SE window.

If you could fix that, it would be utterly wonderful.

niksedk:
> Hmmmm....... All this talk of getting focus right makes me want to remind you guys of this:
> 
>     1. Undock the controls.
> 
>     2. Alt+Tab to either the Audio visualizer or the video player.
> 
>     3. Alt+Tab to the main SE window.
> 
> 
> If you could fix that, it would be utterly wonderful.

When you Alt+Tab back to main, what exactly is wrong? Are keyboard shortcuts dead? Is focus on the wrong control? Is the window not really in front?

GrampaWildWilly:
Oops.  This one is so old I forgot I ought to include a proper description.

Focus simply refuses to return to the main window.  It stays on whichever of the other 2 windows you were on.  You can Alt+Tab between those other 2 windows to your heart's content.  But the main window refuses to take focus back via the keyboard.  Via the mouse, no problem.  But via the keyboard, doesn't happen.  It's very specific to the other 2 undocked SE windows.  You can Alt+Tab from the main window to any other task (Firefox, Notepad++, whatever), & Alt+Tab will quite happily return focus to the main window.  But Alt+Tab to either of the other 2 SE windows makes that main window unreachable by Alt+Tab.  Once that happens, Alt+Tab to any non-SE window still doesn't get focus to return to the main SE window via Alt+Tab.  I think I reported this in my first problem report on the first Preview I ever ran.  It's been a while since I nagged you about it.  I thought it was an Avalonia thing.  But what you've done here with the Find dialogs gives me hope.

PR #11229, opened 2026-05-28
Show undocked video/waveform windows as independent top-levels

niksedk:
## Summary
Addresses the long-standing focus complaint raised in #11226:
> Undock the controls → Alt+Tab to either the audio visualizer or the video player → Alt+Tab to the main SE window doesn't work. The main window refuses to take focus back via the keyboard.

**Root cause:** `WindowsService.ShowWindow<T,VM>(owner, …)` calls `window.Show(owner)`. On Windows that makes the new window an *owned* top-level — the OS groups owner + owned windows in Alt+Tab. Once the user activates an owned window, the OS remembers it as the group's last-active member, so Alt+Tab back from any non-SE window returns to the owned window, not the main one.

**Fix:** the two undocked windows shouldn't be owned by main — they're peer top-levels, not modal sidekicks. Adds `IWindowService.ShowIndependentWindow<T,VM>` which calls `Show()` with no owner, and switches the video-player and audio-visualizer undocked windows to use it. They now appear in Alt+Tab as independent top-levels.

Owner-cascade close is no longer available, but `MainViewModel.CleanUp()` already explicitly closes both undocked windows (sets `AllowClose = true` first, then `Close()`), so the existing shutdown path keeps working. To make sure ApplicationShutdown closes (e.g. OS shutdown, last-window-close) aren't accidentally trapped in a minimize loop, the `OnClosing` handlers on both VMs are tightened to only intercept `WindowCloseReason.WindowClosing` — explicit user closes via the title-bar X.

## Test plan
- [ ] Undock the controls. Alt+Tab from main → audio visualizer → main (works).
- [ ] Alt+Tab from main → video player → main (works).
- [ ] With the undocked window active, Alt+Tab to Firefox/Notepad → back to main (works).
- [ ] Click X on an undocked window → it minimizes (existing behavior preserved).
- [ ] Redock controls → both undocked windows close cleanly, main resumes normally.
- [ ] Close main window with both undocked open → app exits, no zombie windows.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -5214,13 +5214,13 @@ public partial class MainViewModel :
                 _audioVisualizerUndockedViewModel = null;
             }
 
-            _windowService.ShowWindow<VideoPlayerUndockedWindow, VideoPlayerUndockedViewModel>(Window, (window, vm) =>
+            _windowService.ShowIndependentWindow<VideoPlayerUndockedWindow, VideoPlayerUndockedViewModel>((window, vm) =>
             {
                 _videoPlayerUndockedViewModel = vm;
                 vm.Initialize(_videoFileName ?? string.Empty, position, volume, this);
             });
 
-            _windowService.ShowWindow<AudioVisualizerUndockedWindow, AudioVisualizerUndockedViewModel>(Window, (window, vm) =>
+            _windowService.ShowIndependentWindow<AudioVisualizerUndockedWindow, AudioVisualizerUndockedViewModel>((window, vm) =>
             {
                 _audioVisualizerUndockedViewModel = vm;
                 vm.Initialize(AudioVisualizer, this);
--- a/src/ui/Features/Shared/Undocked/AudioVisualizerUndockedViewModel.cs
+++ b/src/ui/Features/Shared/Undocked/AudioVisualizerUndockedViewModel.cs
@@ -36,7 +36,12 @@ public partial class AudioVisualizerUndockedViewModel : ObservableObject
 
     internal void OnClosing(object? sender, WindowClosingEventArgs e)
     {
-        if (!AllowClose && e.CloseReason != WindowCloseReason.OwnerWindowClosing)
+        // Only intercept explicit user closes (clicking the title-bar X). Owner-window
+        // and app-shutdown closes must be allowed through; otherwise the app gets stuck
+        // when the main window is closing or the OS is shutting down. Since this window
+        // is now an independent top-level (no owner), the cascade comes from
+        // MainViewModel.CleanUp() which sets AllowClose=true before calling Close().
+        if (!AllowClose && e.CloseReason == WindowCloseReason.WindowClosing)
         {
             e.Cancel = true;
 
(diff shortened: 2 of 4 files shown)


GrampaWildWilly:
I am so looking forward to this.  Thank you!  This has probably been my biggest annoyance with SE 5.  I am in no way qualified to evaluate code but a couple of thoughts occurred to me.

If I'm reading this correctly, you've made the 2 child windows more like peers of the main window.  But there is still a level of subordination that should be preserved.  I'm thinking that it should not be possible to close either the video player or the Audio visualizer because there's no way to open them again (short of cycling the whole app).  Seems to me at a certain point back in the Preview days that you did something so that "closing" them actually just minimized them.  I'm hoping that's still true.

Another thing you added that wasn't there initially was to make the audio visualizer float over the main window so it's always visible despite possibly not having focus.  I hope this is preserved.

The thought crossed my mind that closing the main window should still close the 2 child windows.  I believe you covered that so there's no worry there.

Otherwise, I am most grateful for this fix.  I had totally given up hope, believing this was just another of those annoying artifacts of Avalonia that couldn't be addressed.  So I'm quite eager to see beta 33.

niksedk:
> I am so looking forward to this...

Beta 33 is up now: https://github.com/SubtitleEdit/subtitleedit/releases
Pretty huge change log: https://raw.githubusercontent.com/SubtitleEdit/subtitleedit/refs/heads/main/change-log.txt

GrampaWildWilly:
Alt+Tab goes to the main window!  HURRAY!

But the audio visualizer doesn't stay visible, floating on top of the main window.  The only way it stays on top is if it actually is the foreground window.  Alt+Tab to the main window puts the Audio visualizer behind the main window, which with my sizes/positions puts the Audio visualizer out of sight.

Closing the children does still minimize them instead of closing.  One oddity with the video window is that a close/minimize then Alt+Tab to refocus it doesn't restore it in the fullscreen mode I had it in before the "close."  A quick Alt+Enter fixes that but it would be nice if it remembered I previously had it in fullscreen.

Closing the main window does properly take down the 2 children.

GrampaWildWilly:
I have these shortcuts:



Alt+1 works.  Alt+2 has never worked & still doesn't.  It's like a no-op.  Here's something really weird.  When I do have the Audio visualizer in foreground with the main window behind it not in focus, Alt+1 does skip back & forth between the text box & the grid.  After doing Alt+1, the Arrow keys do move up & down the subtitles in the grid.  I can even apply edits in the text box.  Alt does activate the menu bar.  This is weird because the Audio visualizer retains foreground status, something I can tell by the color of the window title bar.  The Audio visualizer is in focus but I'm still able to operate the main window despite it not having focus, not being in foreground.  It's like Alt+1 does half the function of Alt+Tab.  If I drop down any menu in the main window -- strictly via the keyboard, avoiding the mouse -- when things are in this state, focus doesn't change.  But as soon as I activate any menu entry, focus does go to the main window & the Audio visualizer disappears behind the main window.

GrampaWildWilly:
Perhaps this is a clue.  Only you can decide.  When I open SE, the Audio visualizer appears for a fraction of a second before it disappears behind the main window.  It's as if SE is trying to float the Audio visualizer on top of the main window, & then something comes along that prevents that from succeeding.

PR #11233, opened 2026-05-28
Add setting to hide Plugins menu

niksedk:
## Summary
- New Appearance setting **Show Plugins menu** (default `true`) lets users hide the `Plugins` entry in the main menu.
- Wired through `SeAppearance`, `SettingsViewModel`, `SettingsPage` (Appearance section), `InitMenu`, and `MainViewModel.ApplySettings` so the change takes effect immediately after Apply/OK.
- Added `LanguageSettings.ShowPluginsMenu` for the label (no hardcoded string).
- Updated `docs/plugin.md` to note that the Plugins menu must be enabled in Settings.

## Test plan
- [ ] Open Subtitle Edit → menu shows `Plugins` by default.
- [ ] Options → Settings → Appearance → uncheck **Show Plugins menu** → Apply → `Plugins` menu disappears without restart.
- [ ] Re-check the setting → `Plugins` menu reappears.
- [ ] Restart the app with the setting unchecked → `Plugins` menu stays hidden.

### DIFF
--- a/docs/plugin.md
+++ b/docs/plugin.md
@@ -15,6 +15,10 @@ compatible with SE5 and must be ported.
 > and update plugins from the index) are implemented. Installing, enabling, disabling,
 > and removing plugins update the **Plugins** menu live (no restart needed). A startup
 > update notification is still to come.
+>
+> The **Plugins** menu is shown by default but can be hidden via
+> **Options → Settings → Appearance → Show Plugins menu**. The menu must be enabled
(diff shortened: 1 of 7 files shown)


GrampaWildWilly:
The Plugins entry on the menu bar does not have an underlined letter.  That means there is no method of opening the menu strictly via keyboard actions.  You are forced to use the mouse.  It's the only menu item on the menu bar that does not have an underlined letter.

PR #11242, opened 2026-05-28
Fix SHIFT+arrow selection and DataGridRow hover highlight

mjuhasz:
This PR brings the behavior in line with SE 4.

  Two related fixes to the subtitle grid: keyboard selection with SHIFT+arrow and the hover highlight on rows.
  
  ## SHIFT+arrow selection
  
  Avalonia's DataGrid does not properly clear the `:selected` visual state when reversing SHIFT+arrow direction — rows that should be deselected remain highlighted. Fixed by intercepting SHIFT+Down/Up in the key handler and manually managing the selection range, mirroring the existing `DragSelectSubtitleGridToIndex` pattern.

  The scenario is as follows:
  1. Open a subtitle file with several rows.
  2. Click row 3 to set it as the selection anchor.
  3. Hold SHIFT and press Down four times — rows 3–7 are now selected and highlighted.
  4. Still holding SHIFT, press Up three times to shrink the selection back toward the anchor.
  5. Expected: rows 5, 6, 7 are deselected and their highlight clears as you arrow back up.
  6. Actual (before fix): rows 5, 6, 7 stay visually highlighted.
  
  ## DataGridRow hover highlight
  
  Suppressing the Fluent theme's default gray hover background turned out to require several iterations. This applies to all DataGrids in the app.

  Hover highlight on unselected rows
  1. Open a subtitle file.
  2. Without clicking, move the mouse over any row.
  3. Expected: no visual change — the row background stays neutral.
  4. Actual (before fix): the row gets a gray hover background from the Fluent theme.
  
  Selected row changing color on hover
  1. Click a row to select it — it turns blue.
  2. Move the mouse away, then hover back over the same selected row.
  3. Expected: the row stays the same blue.
  4. Actual (before fix): the row darkens noticeably on hover as Avalonia increases the background opacity from 0.6 to 0.8 on top of the selection color.
  
  Some implementation details:
  - **Styles.axaml was never loaded.** The file was excluded from AXAML compilation in `UI.csproj` and never referenced, making all overrides dead code. Fixed by removing the exclusion and adding a `StyleInclude` in `Program.cs` after the DataGrid theme so our rules take precedence. Also corrected the `FocusVisual`/`CurrencyVisual` selectors to use the element types required by the AXAML compiler.
  - **`DataGridRow.Background` is ignored by Avalonia** for individual pointer states. The hover color is painted via `Fill` on `Rectangle#BackgroundRectangle` inside the row template, so the override targets that element directly.
  - **The `:pointerover` override was too broad**, making selected rows go transparent on hover and preventing the selection color from appearing immediately on click. Narrowed the selector to `:not(:selected)` so it only suppresses the hover background on unselected rows.
  - **Selected rows still darkened on hover** because the Avalonia theme sets `Opacity` separately from `Fill`. Fixed by also resetting `Opacity` to the non-hovered value for both focused and unfocused selected states.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -12306,6 +12306,9 @@ public partial class MainViewModel :
             return;
         }
 
+        _shiftSelectAnchorIndex = -1;
+        _shiftSelectCurrentIndex = -1;
+
         lock (_scrollLock)
         {
             _pendingScrollIndex = index;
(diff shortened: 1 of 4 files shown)


GrampaWildWilly:
The Shift+Up/DownArrow misbehavior is something I reported back in the Preview days.  THANK YOU for fixing it!

PR #11243, opened 2026-05-29
Make Find/Replace topmost only while main window has focus

niksedk:
## Summary
- Find/Replace dialogs now toggle `Topmost` dynamically based on owner/child window activation instead of being unconditionally topmost.
- Avoids the dialogs floating above other applications on macOS (where Avalonia maps `Topmost` to `NSWindowLevel.Floating` process-wide), without needing an `OperatingSystem.IsMacOS()` branch.
- Alternative to #11234.

## Implementation
- New `WindowService.KeepTopmostWhileOwnerActive(child, owner)` helper subscribes to `Activated`/`Deactivated` on both windows and recomputes `child.Topmost = owner.IsActive || child.IsActive`. Handlers are removed on `child.Closed`.
- `ShowFind` and `ShowReplace` in `MainViewModel` call the helper instead of setting `window.Topmost = true`.
- `ShowWrapAroundDialog` no longer needs to manually flip `Topmost` around `MessageBox.Show` — the message box stealing focus already triggers the helper to drop the dialog out of topmost.

## Test plan
- [ ] macOS: open Find, switch to another app — Find no longer floats above it.
- [ ] macOS: switch back to Subtitle Edit — Find pops back above the main window.
- [ ] macOS: click the Find window, switch to another app — Find drops behind.
- [ ] macOS: trigger wrap-around search so the MessageBox appears over the Find dialog — message box appears in front, Find dialog returns to topmost after closing.
- [x] Windows: existing behavior unchanged (Find on top while main active, drops when main loses focus).
- [ ] Linux: existing behavior unchanged.
- [ ] Same checks for Replace.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -9288,7 +9288,7 @@ public partial class MainViewModel :
         var subs = Subtitles.Select(p => p.Text).ToList();
         var result = _windowService.ShowWindow<FindWindow, FindViewModel>(Window!, (window, vm) =>
         {
-            window.Topmost = true;
+            WindowService.KeepTopmostWhileOwnerActive(window, Window!);
             _findViewModel = vm;
 
             var selectedText = string.Empty;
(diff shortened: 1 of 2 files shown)


GrampaWildWilly:
This business of windows hovering over others sounds a lot like it's related to a comment I added to #11229.  I don't know if you noticed it.  It seems to me like what you're dealing with here, with certain windows allowed to & not allowed to hover over certain other windows is very closely related to the issue I raised in the other pull request.  I'm speculating that the work you're doing here is applicable there as well.

PR #11253, opened 2026-05-29
Fix Shift+Page Up/Down and Shift+Home/End selection in subtitle grid

mjuhasz:
This PR is the continuation of DataGrid selection usability improvements I started in #11242

Avalonia's DataGrid does not handle Shift+PageDown/PageUp/Home/End correctly — it uses its own broken range selection logic that bypasses the anchor-based tracking introduced for Shift+Arrow. The result was that Shift+PageDown added rows incorrectly and Shift+PageUp failed to deselect them.

  ## Changes

  - Intercept Shift+PageDown, Shift+PageUp, Shift+Home, and Shift+End in the grid's key handler and route them through the existing `HandleShiftArrowSelection` machinery, which correctly manages the anchor index and rebuilds `SelectedItems` as a contiguous range.
  - `HandleShiftArrowSelection` now clamps the target index instead of bailing at boundaries, so it can handle page-sized and full-extent jumps in addition to single-row steps.
  - Add `GetSubtitleGridPageSize()` which computes the visible row count from `DataGridRowsPresenter.Bounds.Height / DataGrid.RowHeight` (with `Math.Ceiling` so a partially-visible bottom row is included). Falls back to counting rendered `DataGridRow` instances for variable-height grids.

  ## Behaviour

  - **Shift+PageDown**: extends selection by one page (last visible row included), with one row of overlap on the next press.
  - **Shift+PageUp**: correctly shrinks the selection back.
  - **Shift+Home / Shift+End**: extends selection to the first / last row.
  - Plain PageDown/PageUp/Home/End: unaffected — anchor resets via the existing `SelectionChanged` handler.

  ## A Note on Variable Row Height in DataGrid

Avalonia DataGrid's row virtualizer requires uniform row heights to correctly estimate the scroll extent. When rows have different heights (e.g. 1-line vs 2-line subtitle text rendered via LineBreak inlines), the virtualizer's position estimates drift on every layout pass, causing a visible oscillation between scroll-to-top and scroll-to-bottom on each trackpad gesture. References I found point in the direction that this is a known issue and not going to be fixed in Avalonia.

https://github.com/user-attachments/assets/0e8a2a89-322b-4c19-b9d2-875266b4a62a

Options are using fixed row height, or replacing DataGrid. The latter being very involved, I've been using Subtitle Edit with fixed row height for a while and I like the spacious layout. With all rows having the same height, the virtualizer gets a stable, accurate scroll extent.

 I am still thinking about the best ui design choice here. In this PR, I'm handling both variable and fixed row height. Should we switch to fixed row height to work around the Avalonia DataGrid issue, this PR could be slightly simplified.

https://github.com/user-attachments/assets/cb5eb440-09d8-44e9-8bc0-c28b3bb1a632

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -16425,6 +16425,30 @@ public partial class MainViewModel :
                     _shiftSelectAnchorIndex = -1;
                     _shiftSelectCurrentIndex = -1;
                 }
+                else if (keyEventArgs.Key == Key.PageDown && keyEventArgs.KeyModifiers == KeyModifiers.Shift && Subtitles.Count > 0)
+                {
+                    keyEventArgs.Handled = true;
+                    HandleShiftArrowSelection(GetSubtitleGridPageSize());
+                    return;
+                }
+                else if (keyEventArgs.Key == Key.PageUp && keyEventArgs.KeyModifiers == KeyModifiers.Shift && Subtitles.Count > 0)
+                {
+                    keyEventArgs.Handled = true;
+                    HandleShiftArrowSelection(-GetSubtitleGridPageSize());
+                    return;
+                }
+                else if (keyEventArgs.Key == Key.Home && keyEventArgs.KeyModifiers == KeyModifiers.Shift && Subtitles.Count > 0)
+                {
+                    keyEventArgs.Handled = true;
+                    HandleShiftArrowSelection(-Subtitles.Count); // clamps to 0
+                    return;
+                }
+                else if (keyEventArgs.Key == Key.End && keyEventArgs.KeyModifiers == KeyModifiers.Shift && Subtitles.Count > 0)
+                {
+                    keyEventArgs.Handled = true;
+                    HandleShiftArrowSelection(Subtitles.Count); // clamps to Count - 1
+                    return;
+                }
                 else if (keyEventArgs.Key == Key.Home && keyEventArgs.KeyModifiers == KeyModifiers.None && Subtitles.Count > 0)
                 {
                     keyEventArgs.Handled = true;
@@ -16710,8 +16734,8 @@ public partial class MainViewModel :
             _shiftSelectCurrentIndex = anchor;
         }
 
-        var newCurrent = _shiftSelectCurrentIndex + direction;
-        if (newCurrent < 0 || newCurrent >= Subtitles.Count)
+        var newCurrent = Math.Clamp(_shiftSelectCurrentIndex + direction, 0, Subtitles.Count - 1);
+        if (newCurrent == _shiftSelectCurrentIndex)
         {
             return;
         }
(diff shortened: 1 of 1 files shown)


niksedk:
Let's stay with variable height gridview for now... 

A new control could make sense. I think the Avalonia teams has mentioned a new control (tableview) a few times, but we might have to look elsewhere...

GrampaWildWilly:
Your work here & in the related Pulls forces one to ask why would Avalonia have pioneered their own keyboard controls when Windows (& presumably Linux & Mac, neither of which I have any experience with) already does a perfectly workable job with it?  No doubt, a rhetorical question.

One comment you made makes me ask: Do you handle a case of a starting point lower down & then Shift+UpArrow/PageUp?  Marking should progress upward from there.  And then the associated unmarking by moving down.  This would be kind of the mirror image of what you've been explicitly mentioning.

Since you are exhibiting such superlative expertise with circumventing Avalonia GUI quirks, perhaps I could coax you into tackling the way the menu bar is handled, at least in Windows.  Here's the test.  Alt (or F10).  Then any arrow key.  Then any other arrow key.  Example.  Alt, then RightArrow, then LeftArrow.  (Not hold down Alt while typing the Arrow key.  Separate keyboard operations for each key.)  Then Esc.  At the moment, these don't follow long-established Windows GUI behavior that every user expects from any app.

mjuhasz:
@GrampaWildWilly

> No doubt, a rhetorical question.

I am still happy to answer that, stating mostly my opinion and observations.

Avalonia does have real quirks, but a lot of what people call “quirks” are usually edge cases around virtualization, custom layouts, and cross-platform input behavior rather than random instability. So exactly the kind of things we are running into here. It is solid for many desktop apps, but some controls, especially DataGrid, have long-standing rough edges and workaround-heavy behavior, like what I'm dealing with here.

Most of the loud complaints cluster around DataGrid, not Avalonia UI as a whole. Keyboard selection differences (like my PR here), fixed-height/layout requirements, and scroll behavior with mixed row heights (I mentioned this as well in the PR description and attached two videos.) One could argue that the jumpy scrolling is “the price for being virtualized,” which is a candid way of saying the control prioritizes performance and scalability over perfectly smooth variable-height behavior. Some of the past design decisions are being questioned by the Avalonia team as well. These can make the control feel less polished than what users would expect, especially if they come from WPF. On the other hand this grid is way faster than the one in SE 4 with thousands of rows, so yeah, it's a trade-off.

If an app relies heavily on complex grids, mixed row heights, or precise keyboard interaction (and yes, we need these all), we’re more likely to hit the framework’s sharp edges. If an app is mostly forms, navigation, and standard controls, it's pretty smooth sailing.

> One comment you made makes me ask: Do you handle a case of a starting point lower down & then Shift+UpArrow/PageUp? Marking should progress upward from there. And then the associated unmarking by moving down. This would be kind of the mirror image of what you've been explicitly mentioning.

I believe the behavior is correct now but here is a screen recording (I still have fixed row height enabled, ignore that) and you can check. I made keystrokes visible. I started from row 50, and exercised SHIFT+UP/DOWN, SHIFT+PGUP/PGDOWN, and SHIFT+HOME/END.
 
https://github.com/user-attachments/assets/b50f7cf7-4ebd-41a3-9f01-d39cbc1fd180

> At the moment, these don't follow long-established Windows GUI behavior that every user expects from any app.

I'll need to spin up a Windows virtual machine for that so development/testing is not super straightforward for me but these kind of quirks can usually be worked around, at the expense of maintainability of the application.

If I find the time and access to a Windows host I will look into this, or perhaps @niksedk will.

GrampaWildWilly:
That little movie answers my doubts.  I raised the issue because you had talked about only downward marking & upward unmarking.  But it looks like it works fine in both directions so I'm putting that to bed in my mind.

As for the standard menu bar, I hope that can be addressed.  The menu bar behavior is something else I raised almost in my first review of a Preview a few months ago.  Nik told me it was "how Avalonia works" so I had given up.  But your work here to remove these quirks in the grid give me hope there are yet more ways around Avalonia.  I hope your emulation in your virtual environment does not mask some things that would happen on an actual Windows machine.  If you want somebody to test some candidate change on a real (& up to date) Windows 11 system, I am a ready volunteer standing by.  I'm not prepared to build the code myself, but I can execute a package.  I'm imagining that I ought to create yet a third directory.  I already have one directory for SE 4, which I actually haven't even used in a couple of months.  And a second directory for SE 5.  Both are just portable versions, not formally installed versions.  And they are not in C:\Users\whatever or C:\Program Files\whatever.  They are "private," dedicated directories.  So I am assuming I'd set up a third directory where I would just unzip this testing candidate & execute it from there.  Maybe copy in my Settings.json to make things simpler, more familiar for me?  Wait.  Maybe not.  There's directory pointers in there that would connect to my current SE 5.  OK.  No copying settings.  Start from scratch.  Still.  I'm ready.

PR #11270, opened 2026-05-30
Fix Merge Continuation Lines on mixed-script (CJK+Latin) files (#11267)

### LINKED ISSUE #11267: [5 beta] 'Merge Continuation Lines' doesn't work (at all) with certain subtitles
In 5.0.0 beta, 'Merge Continuation Lines' doesn't work at all if the subtitle contains certain combinations of words.
If you replace some words with certain words, the tool will ignore the entire lines in the file. (It will say "No continuation candidates found".)

The 4.0.15 build works properly.

Here are srt files I tested:
https://drive.google.com/file/d/1jdoRU_QNvGY0rs3VBerNXQ2zRapYO0Ha/view?usp=sharing

niksedk:
## Problem

Fixes #11267 — **Merge Continuation Lines** returns "No continuation candidates found" and merges nothing on certain files. This is a regression from 4.0.x.

The cause: `MergeContinuationLinesHelper.Detect` skips the **entire file** when the whole-file auto-detected language is CJK (`zh/ja/th/lo/my/km/ko/bo`). The reported file opens with a handful of Japanese lines over an otherwise English body, so `AutoDetectGoogleLanguage` returns `ja` and the whole feature bails out — none of the English continuation lines merge.

## Fix

Detect CJK **per line** instead of per file. The guard now runs on each candidate's current line, mirroring the exact CJK check already inside `Utilities.QualifiesForMerge` (`CalcCjk.IsCjk` on the last visible character). This keeps CJK lines noise-free (a CJK ending isn't an English-style unfinished sentence) while letting the non-CJK body merge normally.

- Pure-CJK files behave exactly as before (every pair is skipped → no candidates).
- Mixed-script files now merge their non-CJK continuation lines.

`IsLanguageSkipped` is retained (still used by the Text-to-Speech merge prompt).

## Tests

Added `MergeContinuationLinesHelperTests` (3 cases, all passing):
- English continuation lines produce candidates.
- Mixed file detected as `ja` still merges the English body, and no candidate starts from a Japanese line.
- Pure-CJK lines produce no candidates.

### DIFF
--- a/src/ui/Features/Tools/MergeContinuationLines/MergeContinuationLinesHelper.cs
+++ b/src/ui/Features/Tools/MergeContinuationLines/MergeContinuationLinesHelper.cs
@@ -1,4 +1,5 @@
 using Nikse.SubtitleEdit.Core.Common;
+using Nikse.SubtitleEdit.Core.Common.TextLengthCalculator;
 using Nikse.SubtitleEdit.Features.Main;
 using System;
 using System.Collections.Generic;
@@ -26,7 +27,7 @@ public static class MergeContinuationLinesHelper
         int maxTotalLength)
     {
         var result = new List<MergeContinuationLinesCandidate>();
-        if (subtitles == null || subtitles.Count < 2 || IsLanguageSkipped(language))
+        if (subtitles == null || subtitles.Count < 2)
         {
             return result;
         }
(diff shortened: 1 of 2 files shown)


acc4github:
I'm not a programming expert, but this patch description feels like it will skip and ignore the non-English (CJK) languages.

PR #11276, opened 2026-05-30
update to slnx

WeihanLi:
use slnx


niksedk:
Sorry, not right now as we are in the rc phase.
Perhaps later.

Did the whole build pipe line still work (yml files etc.)?

WeihanLi:
> Sorry, not right now as we are in the rc phase.
> Perhaps later.

Never mind, thanks for the amazing project.

> Did the whole build pipe line still work (yml files etc.)?

verified build as expected locally

PR #11291, opened 2026-05-31
Fix OpenAI-compatible STT for "Speech to text selected lines"

dkakaie:
Two fixes to the OpenAI-compatible STT engine when transcribing selected lines:
  - **Empty transcription text:** the result was matched back to the audio clip by the transcoded temp file name instead of the original clip, so it never attached and the selected line came back empty. Now matched by `_videoFileName`, like the whisper engines.
  - **Multi-segment results not split:** a selected line whose audio the engine split into several segments was only split into multiple subtitle lines when the line was longer than 10 s; otherwise the segments were concatenated into one line. Removed the 10 s gate so any multi-segment result splits into one line per segment.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -5567,9 +5567,13 @@ public partial class MainViewModel :
             var transcribedLine = resultSpeechToText.ResultAudioClips[i];
             if (transcribedLine != null)
             {
-                if (selectedLine.Duration.TotalSeconds > 10 && transcribedLine.Transcription.Paragraphs.Count > 1)
+                if (transcribedLine.Transcription.Paragraphs.Count > 1)
                 {
-                    // generate new subtitles
+                    // The engine split this clip into several segments — replace the
+                    // selected line with one new line per segment instead of folding
+                    // them all into the original. Segment times are relative to the
+                    // clip, so offset each by the selected line's start (the StartTime
+                    // setter preserves Duration, so EndTime follows automatically).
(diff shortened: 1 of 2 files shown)


niksedk:
Thanks for the PR! Just a quick heads up regarding the logic here:

The condition selectedLine.Duration.TotalSeconds > 10 is used to switch the logic from single line mode to one-selection-to-many-lines mode. Your changes might break this distinction I think...

dkakaie:
Thanks for the reply. What user scenario was the `Duration.TotalSeconds > 10` check originally intended to handle? Was it mainly to avoid creating many short subtitle segments from a single selection?

My thinking was that modern ASR models are generally quite good at segmentation and timestamping. When a transcription contains multiple timestamped segments, concatenating them back into a single subtitle line—especially for longer clips—can discard some of that structure.

If the 10-second threshold was added as a UX choice rather than due to limitations of the transcription engine, perhaps we could consider relying on the model's segmentation directly and remove the single line mode branching. Another option would be to make the threshold configurable, although that may be unnecessary if the segmentation quality is consistently good.

I'm interested in the original rationale and would be curious to hear your thoughts on whether the model's segmentation should drive subtitle splitting here.

niksedk:
If a line is more than 10 seconds (could also be like 10 minutes), SE will create may lines from the selection.

If less than 10 seconds, SE will treat the lines as a fixed line, and only update the text (not touch timings or create more lines)

dkakaie:
I just reverted to also check for a minimum of 10 seconds source duration. Would appreciate any further points you might have.

Issue #11037, opened 2026-05-18
Findings after testing seconv.exe


Hlsgs:
1. `seconv info` reports some ANSI subs as beeing UTF8 without BOM. This is also the case with SE5 GUI, while SE4 opens these as windows-1252, even if they are windows-1250(wrong, but not unexpected). While some errors in ANSI detection are to be expected, it was my understanding that UTF8 is not easily "confused". I'm attaching an [example sub](https://github.com/user-attachments/files/27972614/test.srt.txt).

2. `--input-encoding-fallback:windows-1250` works

3. Default `--FixCommonErrors` applies Spanish inverted question/exclamation marks. Using `--fix-common-errors-rules:all,-FixSpanishInvertedQuestionAndExclamationMarks` works, but, I'm not seeing the `FixSpanishInvertedQuestionAndExclamationMarks` equivalent listed in the SE5 GUI in it's respective place, nor does the GUI app apply that fix, even if selecting spanish as the language. Not sure what exactly is going on here, but, beeing familiar with the settings in the GUI, I suspect many users will be scratching their heads after runnign the default config.

4. `--encoding:NONSENSE` does not generate an error, which it should

5. `--encoding:source` does not work as it did with old SE, and probably falls within 4. Even with `--input-encoding-fallback` beeing available, this is still a useful option to have, just in case.

6. How do these differ and which is recommended? `--fix-common-errors-rules:FixOverlappingDisplayTimes` vs `--apply-min-gap:0`

7. Is seconv supposed to be able to ingest the `Settings.json` file that SE5 generates? That doesn't seem to be the case. For example, your [documentation](https://subtitleedit.github.io/subtitleedit/reference/command-line.html) specifies "removeTextForHearingImpaired", while SE5 uses "RemoveTextForHi" in it's settings file. Seeing as SE4 CLI actually used (some of) the settings set in the batch convert section of the GUI, it would probably make sense for seconv to be able so use it's GUI counterpart's settings file.

8. After trying 7., I followed the [documentation](https://subtitleedit.github.io/subtitleedit/reference/command-line.html), and tried and failed to remove an all uppercase line with `--settings:Settings.json --RemoveTextForHI`, with the following contents of said setting file:
```
{
  "removeTextForHearingImpaired": {
    "IsRemoveTextUppercaseLineOn": true
  }
}
```
Used the same example sub I've attached in 1.

niksedk:
[...]

PR #11300, opened 2026-06-01
seconv: gate Spanish FCE rule on language + clarify --settings / --apply-min-gap docs (#11037)

niksedk:
## Summary
Addresses three actionable items from #11037 in one PR. Items 1, 2, 4, 5 are either already fixed in the new seconv or out-of-scope (UTF-8 detection); items 3, 6, 7, 8 are handled here.

### Item 3 — code: `FixSpanishInvertedQuestionAndExclamationMarks` was applying to non-Spanish content
The GUI's *Fix Common Errors* window only adds this rule to the rule list when the selected language is `es` ([`FixCommonErrorsViewModel.cs:377`](../blob/main/src/ui/Features/Tools/FixCommonErrors/FixCommonErrorsViewModel.cs#L377)). seconv was running it unconditionally in the default *all rules* pass, so e.g.

```
"Are you coming home tonight?" → "¿Are you coming home tonight?"
```

`FixCommonErrorsRunner.Run` now skips language-conditional rules in the implicit-all pass when the auto-detected language doesn't match. Explicit opt-in via `--FixCommonErrorsRules:FixSpanishInvertedQuestionAndExclamationMarks` still applies the rule regardless of detection (this matches what the user expects when they name a rule).

Three new tests cover all three branches: skip on English, apply on Spanish, opt-in restores on non-Spanish.

### Item 6 — docs: `--apply-min-gap` vs `FixOverlappingDisplayTimes`
`--apply-min-gap` wasn't even in the README's Time/Frame table. Added the row + a short explainer of when to use which.

### Items 7 & 8 — docs: `--settings:<path.json>` ≠ SE 5 GUI's `Settings.json`
The user tried to feed seconv the GUI's `Settings.json` with the GUI's key names (`IsRemoveTextUppercaseLineOn`). seconv expects its own small libse-shaped schema (`RemoveIfAllUppercase`). Added an explicit callout under the `--settings` row in the README with a minimal working example.

## What's not in this PR
- **Item 1 (UTF-8 misdetection)** — real libse bug, but isolating the trigger needs more digging. User has `--input-encoding-fallback` as a workaround.
- **Items 2, 4, 5** — already fixed in the current seconv (`--encoding:NONSENSE` errors out, `--encoding:source` works).

## Test plan
- [x] 16/16 `FixCommonErrorsRunnerTest` pass (including the 3 new ones for the language gate)
- [x] Build clean; only the pre-existing unrelated `MkvWithImageTracks` failure (fixed separately in #11299) remains in the full suite
- [ ] Manually verify on issue reporter's `test.srt.txt`: defaults no longer add `¿` / `¡` to English content; `--FixCommonErrorsRules:FixSpanishInvertedQuestionAndExclamationMarks` still applies them

Refs #11037.

### DIFF
--- a/src/seconv/Commands/ConvertCommand.cs
+++ b/src/seconv/Commands/ConvertCommand.cs
@@ -322,13 +322,18 @@ internal sealed class ConvertCommand : AsyncCommand<ConvertCommand.Settings>
 
             // Resolve FCE rule selection. Passing --FixCommonErrorsRules implicitly
             // enables --FixCommonErrors so users don't have to specify both.
+            // We also parse out which rules the user named by hand — that's a separate
+            // signal from the resolved set so language-conditional rules can bypass
+            // their gate when explicitly requested (see FixCommonErrorsRunner.Run).
             IReadOnlyList<string> fceRules = [];
+            IReadOnlyList<string> fceExplicitlyNamed = [];
             var fceRequested = settings.FixCommonErrors || !string.IsNullOrWhiteSpace(settings.FixCommonErrorsRules);
             if (fceRequested)
             {
                 try
                 {
                     fceRules = FixCommonErrorsRunner.ResolveRuleIds(settings.FixCommonErrorsRules);
+                    fceExplicitlyNamed = FixCommonErrorsRunner.ParseExplicitlyNamedRules(settings.FixCommonErrorsRules);
                 }
                 catch (ArgumentException ex)
                 {
@@ -423,6 +428,7 @@ internal sealed class ConvertCommand : AsyncCommand<ConvertCommand.Settings>
                 Overwrite = settings.Overwrite,
                 Operations = operations,
                 FixCommonErrorsRules = fceRules,
+                FixCommonErrorsExplicitlyNamedRules = fceExplicitlyNamed,
                 DeleteFirst = settings.DeleteFirst,
                 DeleteLast = settings.DeleteLast,
                 DeleteContains = settings.DeleteContains,
--- a/src/seconv/Core/FixCommonErrorsRunner.cs
+++ b/src/seconv/Core/FixCommonErrorsRunner.cs
@@ -60,6 +60,17 @@ internal static class FixCommonErrorsRunner
                 continue;
             }
 
+            // Language-conditional rules mirror what the GUI's Fix Common Errors
+            // window does: it only surfaces FixSpanishInvertedQuestionAndExclamationMarks
+            // when the detected language is "es" (FixCommonErrorsViewModel.cs:377).
+            // Running it on non-Spanish content would insert ¿ / ¡ on questions and
+            // exclamations in any language, which surprises users (issue #11037 item 3).
+            // The user can still opt in explicitly via --FixCommonErrorsRules.
+            if (wanted == null && IsLanguageOnlyRule(id, language))
+            {
+                continue;
+            }
+
             try
             {
                 factory().Fix(subtitle, callbacks);
@@ -69,6 +80,22 @@ internal static class FixCommonErrorsRunner
                 // A rogue rule shouldn't kill the conversion. Skip and continue.
             }
         }
+    }
+
+    /// <summary>
+    /// Returns true when <paramref name="ruleId"/> is a language-conditional rule
+    /// that should not run in the default ("all rules") pass because the detected
+    /// language doesn't match. Skipped here, included by name in an explicit
+    /// <c>--FixCommonErrorsRules</c> list.
+    /// </summary>
+    private static bool IsLanguageOnlyRule(string ruleId, string language)
+    {
+        return ruleId switch
+        {
+            "FixSpanishInvertedQuestionAndExclamationMarks"
+                => !"es".Equals(language, StringComparison.OrdinalIgnoreCase),
+            _ => false,
+        };
     }
 
     /// <summary>
--- a/src/seconv/Core/FixCommonErrorsRunner.cs
+++ b/src/seconv/Core/FixCommonErrorsRunner.cs
@@ -30,11 +30,30 @@ internal static class FixCommonErrorsRunner
     public static void RunAll(Subtitle subtitle) => Run(subtitle, null);
 
     /// <summary>
-    /// Runs the specified rules. Pass <c>null</c> or an empty collection to run all rules.
-    /// Rules execute in canonical order, not caller order, to keep behaviour stable
-    /// across invocations.
+    /// Back-compat overload. Pass <c>null</c> or an empty collection to run all rules
+    /// (gates language-conditional rules to their language). When a non-empty list is
+    /// passed, every entry is treated as both <em>wanted</em> and <em>explicitly named</em> —
+    /// language gates are bypassed for those rules. Rules execute in canonical order,
+    /// not caller order, to keep behaviour stable across invocations.
     /// </summary>
     public static void Run(Subtitle subtitle, IReadOnlyCollection<string>? ruleIds)
+        => Run(subtitle, ruleIds, explicitlyNamedRules: ruleIds);
+
+    /// <summary>
+    /// Runs the specified rules.
+    /// <para><paramref name="ruleIds"/> selects which rules execute (<c>null</c>/empty = all).</para>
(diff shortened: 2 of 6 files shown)


Hlsgs:
> `FixCommonErrorsRunner.Run` now skips language-conditional rules in the implicit-all pass when the auto-detected language doesn't match. Explicit opt-in via `--FixCommonErrorsRules:FixSpanishInvertedQuestionAndExclamationMarks` still applies the rule regardless of detection (this matches what the user expects when they name a rule).

My personal expectation when adding a language specific feature to `--fix-common-errors-rules`, like this spanish one or the english `FixAloneLowercaseIToUppercaseI`, is that it would be applied only when dealing with a sub in that specific language. Otherwise, processing both english and my native language would require switching to using the `--fix-common-errors-rules:all,-` pattern. Which I guess would work, but is not ideal, especially if ever a new rule gets added in an update. Put differently, I see the entire`--fix-common-errors-rules` as whitelist: things that are allowed to run, if they apply. Not an enforcement list.

> ### Item 6 — docs: `--apply-min-gap` vs `FixOverlappingDisplayTimes`
> 
> `--apply-min-gap` wasn't even in the README's Time/Frame table. Added the row + a short explainer of when to use which.
> ### Items 7 & 8 — docs: `--settings:<path.json>` ≠ SE 5 GUI's `Settings.json`
> 
> The user tried to feed seconv the GUI's `Settings.json` with the GUI's key names (`IsRemoveTextUppercaseLineOn`). seconv expects its own small libse-shaped schema (`RemoveIfAllUppercase`). Added an explicit callout under the `--settings` row in the README with a minimal working example.

Apologies if I'm being daft, but I can't find the documentation/README changes meantioned in these pull requests. Do these refer to [this documentation](https://subtitleedit.github.io/subtitleedit/reference/command-line.html#settings-json) or the tool's --help? Also, it would be a great help if we had a full seconv settings.json file listing all libse's defaults, so we can use it as starter.

Regarding both of the above: can the rules used by `--fix-common-errors` be specified via a settings json rather than from cli?

niksedk:
Not daft at all — these questions found real gaps, thanks for the careful read. #12745 addresses most of it.

**Where the docs live.** The published reference is [`command-line.html`](https://subtitleedit.github.io/subtitleedit/reference/command-line.html#settings-json) (source: `docs/reference/command-line.md`) — not the tool's `--help`. The `--settings` / `--apply-min-gap` additions from this PR originally landed in the seconv `README.md`, and a later cleanup consolidated everything into the published reference — but in the process the specific "don't feed seconv the GUI's `Settings.json`" callout got dropped, which is exactly why you couldn't find it. It's restored now.

**A full defaults file to start from.** Added `seconv dump-settings` — it prints the entire settings schema populated with the current libse defaults, so you can do:

```bash
seconv dump-settings > my-settings.json
```

and edit from a known-good, correctly-keyed file (e.g. it shows `removeIfAllUppercase`, which is what tripped up the original report). It round-trips through the loader with zero unknown keys.

**Can FCE rules go in the settings file?** Not currently — rule *selection* is CLI-only (`--fix-common-errors-rules`); the `--settings` JSON shapes rule *behaviour* (line length, min gap, dialog/continuation style, CPS). That's now stated explicitly in the docs. If you'd find a `fixCommonErrors.rules` section in the settings/profile file useful, say so — it's a reasonable addition, especially alongside the next point.

**On `--fix-common-errors-rules` as a whitelist vs. an enforcement list.** This is the interesting one, and I think you're making a fair point. Today, naming a language-specific rule forces it on regardless of detected language — which, as you note, breaks the mixed-language batch case and pushes you toward the fragile `all,-` pattern. Your "allowed to run *if it applies*" model is arguably the better default.

The catch is that naming a rule is currently also the only way to *override* a wrong auto-detection. So if we make the whitelist respect the language gate, I'd want to add a `--language:<code>` override to cover the "force it, detection is wrong" case cleanly. I left that behaviour change out of #12745 (it reverses what the merged PR deliberately did, so it deserves its own discussion) — but if that combination (gate-respecting whitelist + `--language` escape hatch) matches what you'd want, I'll do it as a follow-up. Does it?

Hlsgs:
> **Can FCE rules go in the settings file?** Not currently — rule _selection_ is CLI-only (`--fix-common-errors-rules`); the `--settings` JSON shapes rule _behaviour_ (line length, min gap, dialog/continuation style, CPS). That's now stated explicitly in the docs. If you'd find a `fixCommonErrors.rules` section in the settings/profile file useful, say so — it's a reasonable addition, especially alongside the next point.

The current architecture of keeping rule selection in the CLI and behaviour in the settings json makes more sense and probably avoids some headaches by having a single source of truth.

> The catch is that naming a rule is currently also the only way to _override_ a wrong auto-detection. So if we make the whitelist respect the language gate, I'd want to add a `--language:<code>` override to cover the "force it, detection is wrong" case cleanly. I left that behaviour change out of #12745 (it reverses what the merged PR deliberately did, so it deserves its own discussion) — but if that combination (gate-respecting whitelist + `--language` escape hatch) matches what you'd want, I'll do it as a follow-up. Does it?

To answer directly, yes, a gate-respecting whitelist and a language enforcement parameter was my thinking too, but, I refrained from suggesting the latter because of what it might mean, or, at least, be perceived to mean, for other language-aware functionalities I don't use and therefore have zero grasp over, like OCR, auto-translation etc. I guess, all that could be avoided by having it be `--fce-language` instead?

One remaining idea about `--fix-common-errors-rules` rules would be to have, either in the docs or `seconv list-fce-rules` a 1:1 matching between the CLI name and the GUI semantic equivalent. I think I've managed to create the list I'm interested in, but it was a little daunting. And since most users would probably test in the GUI before commiting to scripting, I suspect some head-scratching would ensue, as was the case for me.

**EDIT:**
> **Where the docs live.** The published reference is [`command-line.html`](https://subtitleedit.github.io/subtitleedit/reference/command-line.html#settings-json) (source: `docs/reference/command-line.md`) — not the tool's `--help`. The `--settings` / `--apply-min-gap` additions from this PR originally landed in the seconv `README.md`, and a later cleanup consolidated everything into the published reference — but in the process the specific "don't feed seconv the GUI's `Settings.json`" callout got dropped, which is exactly why you couldn't find it. It's restored now.

Not sure if that's supposed to be the case, but I'm still not seeing the `--apply-min-gap` vs `FixOverlappingDisplayTimes` stuff in the command-line.md.

niksedk:
Thanks, this is all really useful feedback — follow-up in #12752.

**`--fce-language` (the language escape hatch).** Already shipped, and named exactly as you suggested to keep the scope obvious: `--fce-language:<code>` (accepts `es` / `spa` / `Spanish`; unknown values warn and fall back to auto-detect). It only affects FCE language gating and the FCE OCR-fix pass — it does **not** touch OCR, auto-translation, or anything else — so the "what does this mean for functionality I don't use" worry doesn't apply. On the "gate-respecting whitelist" half: rather than flip what naming a rule does, I kept explicit-naming as a hard force *and* added `--fce-language` for the "genuinely Spanish but mis-detected" case, which covers the same need without a behaviour reversal.

**1:1 CLI ↔ GUI mapping.** Agreed the naming mismatch is a real papercut when you prototype in the GUI first. #12752 adds a **GUI equivalent** column to `seconv list-fce-rules` and a full mapping table to the reference docs, both fed from one source-of-truth map so they can't drift:

```
╭───────────────────────────────┬──────────────────────────────┬───────────────╮
│ Rule ID                       │ GUI equivalent               │ Language gate │
├───────────────────────────────┼──────────────────────────────┼───────────────┤
│ FixEmptyLines                 │ Remove empty lines/unused    │ —             │
│                               │ line breaks                  │               │
│ FixSpanishInvertedQuestionAnd │ Fix Spanish inverted         │ es only       │
│ ExclamationMarks              │ question and exclamation     │               │
│                               │ marks                        │               │
╰───────────────────────────────┴──────────────────────────────┴───────────────╯
```

**`--apply-min-gap` vs `FixOverlappingDisplayTimes`.** Good catch — you were right, it never made it into the reference; it got dropped when the README content was consolidated (same thing that happened to the `Settings.json` callout). Restored in #12752 under *Operations*.

Docs page (the mapping table + the min-gap explainer land here once #12752 is merged): https://subtitleedit.github.io/subtitleedit/reference/command-line.html

PR #11373, opened 2026-06-03
Fix last inconsistent toggle label in native macOS menu bar

mjuhasz:
## Summary

   The native macOS menu bar uses checkmark ticks to convey on/off state for toggle items, so labels don't need to say "Toggle" — the tick already communicates that. The four toggle items in the native Mac menu are:

   | Menu | Item | Label |
   |---|---|---|
   | Edit | Right-to-left mode | ✅ No "Toggle" prefix |
   | Video | Select subtitle while playing | ✅ Fixed in #11240 |
   | Video > More | SMPTE timing (non-integer frame rate) | ✅ No "Toggle" prefix |
   | Video > More | ~~Toggle waveform toolbar~~ → **Waveform toolbar** | ✅ Fixed here |

   "Toggle waveform toolbar" was the last item still using the "Toggle" prefix, making it inconsistent with the other three. This PR applies the same fix: add a plain `WaveformToolbar` language string to `LanguageMainMenu` and use it in `InitNativeMacMenu`. The cross-platform Avalonia menu is unaffected and continues to use the existing `ToggleWaveformToolbar` shortcut string.


niksedk:
thx for the PRs :)

Issue #11357, opened 2026-06-03
SE 5 RC 1: Editing text in the text box causes the video to change position


GrampaWildWilly:
I have defined these keyboard shortcuts:

<img alt="#01" src="https://github.com/user-attachments/assets/3d323ea3-68b8-47fe-a70a-47839e58b74d" />

When I am editing text in the text box, I want to use Ctrl+Right/LeftArrow to skip to the next/previous word.  This is a global Windows standard keyboard shortcut that works in every other app.  But in SE, since I have those shortcuts defined as video skips, there's a conflict.

Now, I do have this in my SE settings:

<img alt="#02" src="https://github.com/user-attachments/assets/fe588264-7de4-4da1-bcdc-a8345ae874f0" />

But that's irrelevant.  Ctrl+Right/LeftArrow are not single-letter shortcuts.  But we need some way similar to that setting to let the text box act like a Windows standard text entry area, suppressing the recognition of colliding SE shortcuts so long as focus is in the text box.  Just the text box, not the grid or anywhere else in SE, just in the text box.  I have found that navigating up & down the subtitles via Alt+Up/DownArrow moves the video position.  Focus can be in the text box or in the grid, doesn't matter.  Alt+Up/DownArrow act the same either way.  This is correct behavior.  Alt+Up/DownArrow are not standard Windows text editing keyboard shortcuts.  So these shortcuts should act like this when focus is in the text box.  I mention this to contrast Alt+Up/DownArrow with Ctrl+Right/LeftArrow.  The former mean nothing to Windows; the latter most definitely do mean something to Windows.  SE is interfering with this standard Windows behavior.

This temporary suppression of SE shortcuts should probably apply in any text entry field, like in Find, Speech to Text Advanced Settings, anywhere the user might type text.

Curiously, Ctrl+Shift+Right/LeftArrow work fine to highlight single words.  Maybe that's an accident.  I happen not to have defined any shortcuts with those particular key sequences.  Maybe if I did, they might not work as I would like.  In fact, I did go ahead & add a definition of Shift+Ctrl+Right to skip forward 100ms.  Sure enough, the text editing keystroke got usurped by the SE shortcut.  I reset the shortcut & the keystroke went back to proper text marking behavior.

[...]

PR #11428, opened 2026-06-05
Also ignore Ctrl+Left/Right in text input for word-jump navigation (#11357)

niksedk:
## Summary
Follow-up to #11403 / #11357.

PR #11403 bailed on bare Left/Right inside a focused text input so caret navigation was no longer overridden by shortcuts. The reporter noted in [this comment](https://github.com/SubtitleEdit/subtitleedit/issues/11357#issuecomment-4633186581) that **Ctrl+Left/Right** (word-jump) still fires shortcuts instead of moving the caret one word.

## Change
Extend the existing guard in `MainViewModel.OnKeyDownHandler` to also return early when `KeyModifiers.Control` is the only modifier:

```csharp
if ((keyEventArgs.Key == Key.Left || keyEventArgs.Key == Key.Right)
    && (keyEventArgs.KeyModifiers == KeyModifiers.None
        || keyEventArgs.KeyModifiers == KeyModifiers.Control))
{
    return;
}
```

Ctrl+Shift+Left/Right (extend selection by word) is unaffected — those still go through normal shortcut resolution. Only the pure Ctrl+Left/Right combination is protected.

## Test plan
- [ ] Bind Ctrl+Left to a video shortcut. Click into the subtitle text box, press Ctrl+Left. Confirm the caret jumps back one word and the video shortcut does **not** fire.
- [ ] Same for Ctrl+Right.
- [ ] Ctrl+Shift+Left/Right still extends the selection (or fires a bound shortcut), unchanged.
- [ ] Bare Left/Right still work as before (from #11403).

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -16923,11 +16923,12 @@ public partial class MainViewModel :
 
             if (IsTextInputFocused())
             {
-                // Bare Left/Right are fundamental caret navigation in any text
-                // input — never override them with shortcuts even when
+                // Bare and Ctrl+Left/Right are fundamental caret navigation in any
+                // text input — never override them with shortcuts even when
                 // "allow single-letter shortcuts in text box" is on (#11357).
                 if ((keyEventArgs.Key == Key.Left || keyEventArgs.Key == Key.Right)
-                    && keyEventArgs.KeyModifiers == KeyModifiers.None)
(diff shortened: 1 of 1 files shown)


GrampaWildWilly:
I have my fingers crossed.  Wait.  That might make it a bit tough to type.  OK.  So I'll uncross my fingers when RC4 or a general availability release comes, whichever you do next.

PR #11459, opened 2026-06-07
Write frames as two digits in custom export

OmrSi:
HH:MM:SS:10 should not be written as HH:MM:SS:1


OmrSi:
Should also probably be applied to SE 5...

PR #11503, opened 2026-06-08
Restore Fix Continuation Style settings

mjuhasz:
Continuation style related functionality is now fully present in SE5, used in the Fix Common Errors rule, and I also fixed some related bugs. The PR looks large because I salvaged most of the localization and it contains tests as well.

  ### Changes

  **New: Fix Continuation Style Settings dialog**
  A dedicated settings sub-dialog accessible from Settings → Rules, containing:
  - Heuristic filter flags: uncheck inserts in all-caps / italic / lowercase, hide candidates without a name, ignore lyrics
  - Continuation pause threshold (ms)
  - Shortcut to edit the custom continuation style

  **Sync continuation settings into libse on startup**
  `SetLibSeSettings` now copies all continuation fields (`ContinuationPause`, 13 `CustomContinuationStyle*` fields, 5 `FixContinuationStyle*` flags) into `Configuration.Settings.General` so libse consumers (`ContinuationUtilities`, `FixContinuationStyle`) use the correct values at runtime.

  **Bug fixes**
  - Fixed invalid character (`\uFFFD`) in custom continuation style dropdown items
  - Fixed "Edit custom style" button not appearing when the settings page was reopened
  - Fixed continuation settings not being reset when choosing Reset Rules
  - Fixed `ColorDurationTooLong` not being reset when choosing Reset Syntax Coloring (self-assignment no-op)

  **Localization**
  Added `fixContinuationStyleSettings`, `customContinuationStyleNote`, and six supporting strings to all 25 language files. Hungarian and Portuguese have full translations.

  
Settings - Tools

Settings for fixing continuation style

Edit custom continuation style


niksedk:
thx for the PR :)

I think we should wait with the UI stuff for now, and keep with the SE 4 style.

The bugs + per profile handling included in this PR: https://github.com/SubtitleEdit/subtitleedit/pull/11510

mjuhasz:
A fix was missed; it's a trivial change. I opened a PR for that: #11513

mjuhasz:
@niksedk One more thing: ContinuationPause (the gap threshold used to decide whether to apply the "long pause" variant of a style) is consumed by **all** "None/mixed" continuation styles via ContinuationUtilities.GetMinimumGapMs(), not just Custom.

In your PR, it is only exposed in the UI through the Custom style editor, so users who never select Custom will always get the hardcoded default of 300 ms and have no way to change it. This may be an acceptable design choice.

However, it also means the value persists across style switches: if a user sets it to e.g. 2000 ms while on Custom, then switches to "None, ellipsis for pauses", the 2000 ms threshold silently carries over and suppresses most ellipsis insertions. That's unexpected: the same (non-Custom) continuation style behaves differently based on past configuration of another (the Custom) style.

I am not exactly sure how you want to design the rest of the continuation style UI/behavior so I'm going to stop here with my changes for the time being and let you decide: make non-Custom profiles always use the default 300ms (fairly easy), expose the setting globally or per continuation style, or leave things as is for the time being. I'm happy to help with the change if we align on the direction. If I see it correctly I can replicate any of the "None/mixed" styles in Custom so the choice of defaulting to 300ms sounds reasonable, and if one wants a different value, go to Custom and set one there.

PR #11546, opened 2026-06-11
Make bookmark add/toggle shortcuts work from the text editor

mjuhasz:
Move AddOrEditBookmark and ToggleBookmarkSelectedLinesNoText from ShortcutCategory.SubtitleGrid to General so they fire regardless of focus, consistent with GoToNextBookmark, GoToPreviousBookmark, and ListBookmarks which were already General.

 ## Context

  Bookmark shortcuts were split across two categories in an inconsistent way:
  - GoToNextBookmark, GoToPreviousBookmark, ListBookmarks → General (work everywhere)
  - AddOrEditBookmark, ToggleBookmarkSelectedLinesNoText → SubtitleGrid (grid focus only)

  This means a user editing text in the subtitle text box could navigate to the previous/next bookmark but could not add or toggle one — a confusing inconsistency with no technical justification. One use case is when I navigate between subtitles using Option(Alt)-Up/Down, making minor edits as I go, but in order to bookmark a line I have to grab the mouse to change the focus to the subtitle grid.

  The two SubtitleGrid-only commands both use SubtitleGrid.SelectedItems, which still reflects the currently-selected subtitle when focus is in the text editor, so they work correctly when triggered from either context.

### DIFF
--- a/src/ui/Logic/ShortcutsMain.cs
+++ b/src/ui/Logic/ShortcutsMain.cs
@@ -430,8 +430,8 @@ public static class ShortcutsMain
         AddShortcut(shortcuts, vm.DoAlignmentAn7Command, nameof(vm.DoAlignmentAn7Command), ShortcutCategory.SubtitleGrid);
         AddShortcut(shortcuts, vm.DoAlignmentAn8Command, nameof(vm.DoAlignmentAn8Command), ShortcutCategory.SubtitleGrid);
         AddShortcut(shortcuts, vm.DoAlignmentAn9Command, nameof(vm.DoAlignmentAn9Command), ShortcutCategory.SubtitleGrid);
-        AddShortcut(shortcuts, vm.AddOrEditBookmarkCommand, nameof(vm.AddOrEditBookmarkCommand), ShortcutCategory.SubtitleGrid);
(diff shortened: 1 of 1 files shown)


GrampaWildWilly:
> I have to grab the mouse to change the focus to the subtitle grid.

Well . . . not necessarily . . .



I have been greatly impressed by your evident skill in improving SE.  I am hoping that pointing out that shortcut & praising you will inspire your gratitude & you will make yet another improvement:  Make that other shortcut shown there, `Toggle focus between subtitle grid and waveform/spectrogram`, work, something it has never done.

mjuhasz:
@GrampaWildWilly Thank you for your kind words.

Thank you for pointing out the shortcut as well. I believe it's still important to address the inconsistency of some bookmark shortcuts not functioning in the edit box, even if a workaround exists. The workaround adds friction: it requires pressing an extra keyboard shortcut to refocus the subtitle grid and another to return focus to the edit box. This results in three keystrokes instead of one, and all three are likely complex (multi-key combinations.)

I fixed the reported shortcut in #11550 🎉

GrampaWildWilly:
The kind words are well earned.  But you do realize you have now proven that flattery works.  That's a tool I'm keeping in my toolbox.  You have been warned. 😁

PR #11550, opened 2026-06-11
Fix: Toggle focus grid↔waveform not working in waveform→grid direction

mjuhasz:
## Problem

  The "Toggle focus between subtitle grid and waveform/spectrogram" shortcut only worked in one direction. Pressing it from the subtitle grid moved focus to the waveform correctly, but pressing it again from the waveform did nothing — focus stayed on the waveform.

  ## Root cause

  `ToggleFocusGridAndWaveform()` called `SubtitleGrid.Focus()` synchronously during key-event dispatch while the `AudioVisualizer` still held focus. Avalonia reverts focus back to the previously focused control before the event is fully resolved, so the focus change was silently lost.

  The analogous `ToggleFocusTextBoxAndWaveform` command works in both directions because its waveform→textbox path defers the focus change via `Dispatcher.UIThread.Post` inside `FocusEditTextBox()`.

  ## Fix

  Introduced a `FocusSubtitleGrid()` helper that mirrors `FocusEditTextBox()`: it defers the `SubtitleGrid.Focus()` call via `Dispatcher.UIThread.Post` and re-evaluates the `SkipNextPointerEntered` flag at execution time (preventing `FocusOnMouseOver` from stealing focus back if the mouse is hovering over the waveform). `ToggleFocusGridAndWaveform()` now uses this helper for the waveform→grid direction.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -11420,6 +11420,19 @@ public partial class MainViewModel :
         });
     }
 
+    private void FocusSubtitleGrid()
+    {
+        Dispatcher.UIThread.Post(() =>
+        {
+            if (AudioVisualizer != null && AudioVisualizer.IsFocused)
+            {
+                AudioVisualizer.SkipNextPointerEntered = true;
+            }
+
+            SubtitleGrid.Focus();
+        });
+    }
+
     [RelayCommand]
     private void WaveformInsertAtPositionAndFocusTextBox()
     {
@@ -12060,12 +12073,7 @@ public partial class MainViewModel :
         }
         else
         {
-            if (AudioVisualizer.IsFocused)
-            {
-                AudioVisualizer.SkipNextPointerEntered = true;
-            }
-
-            SubtitleGrid.Focus();
+            FocusSubtitleGrid();
         }
 
         _updateAudioVisualizer = true;


GrampaWildWilly:
😀

GrampaWildWilly:
Just got RC 4.  Further to this . . .



My Alt+2 shortcut still does nothing.  I was so optimistic.  You seemed to grasp the issue & even though I can't make heads or tails of the code, it looked like you nailed it.  Sadly, not.

mjuhasz:
@GrampaWildWilly Fear not, we'll get to the bottom of this.

The fix I made solved the issue (it's called a race condition) when everything is in the same window. I strongly suspect that you have these windows

1. undocked, and
2. on Windows.

Windows behaves differently from macOS: asking for focusing a window implicitly activates the window on macOS but not on Windows.

So the earlier fix fixes the issue on macOS (both the docked and undocked case) but misses the _undocked_ case on _Windows_.

Seemingly simple tasks, such as transferring focus between windows or creating a Find dialog, can be quite complicated. But as I said, we'll fix this: if one iteration was not enough, then in another one. 

Fix is here:  #11559

GrampaWildWilly:
> @GrampaWildWilly Fear not, we'll get to the bottom of this.

That has been your track record.

> Windows behaves differently from macOS

Well of course.  Why wouldn't it?  Entire communities of engineers & users have to be kept on their toes.

> Fix is here: #11559

Yes, I saw that.  Waiting as patiently as possible for the next release, whether that's another RC or the first GA, whichever Nik decides.

GrampaWildWilly:
It's been a few releases & I'd forgotten about this.  But in SE 5.1 beta 1 my Alt+2 shortcut FINALLY WORKS!!!

Issue #12, opened 2014-02-23
Character spacing in vobsub/sup export


rebawest:
I'm not sure how difficult this would be to program. By analysing DVD's and blurays subtitles by major film distributors, I've noticed that they use character spacing when generating their subtitles.
This allows them to have more characters on screen without reducing the font size.

Here's an example from a Warner Bros. DVD. By setting character spacing to "condensed 2 points", the font size could be set to 36 instead of 32.

![image](https://f.cloud.github.com/assets/6587015/2241748/8ec2fe9c-9ce1-11e3-9e4b-cb45d1b29725.png)

And here's another example from a Disney bluray. Again, a "condensed 2" is used and the font size could be set to 45 instead of 41.

![image](https://f.cloud.github.com/assets/6587015/2241750/c78ed958-9ce1-11e3-8a3d-5982e812c30c.png)

I hope I was clear enough. Please ask for any clarifications.
Thank you very much, Kind regards...

Adrián

Issue #11, opened 2014-02-23
Ability to choose specific audio channel/track for Wave generation.


LeonCheung:
I know this will make little sense to most users, but currently I do meet such need when working on a project. The video clip got both Left/Right channels. However, the wave seems messed up, because the L channel is full of noise and no voice.

AFAIK, on some video files, one of the
stereo channels (typically left) is used to encode LTC time code. So, here I think the R channel would be the ideal one for wave generation, instead of the mix or stereo channels. 

So I just wonder if it would be possible for users to decide which channel to use, L, R, or stereo. FYI. Thank you!

niksedk:
I think this is working now

Issue #10, opened 2014-02-19
Alignment tags lose "{\" in JSON formats


zatch:
When using the Alignment feature and any of the JSON formats, the leading "{\" is lost in the Source View.  This also causes the file to be saved without the leading "{\", and so reopening the file carries the issue over to the List View.

Take this line from List View, for example:

```
{\an8}This whole line will appear on the top of the screen.
```

The JSON in Source View will read like so:

```
{"start":4.468,"end":9.428,"text":"an8}This whole line will appear on the top of the screen."}
```

And after saving/reopening, the line will read in List View:

```
an8}This whole line will appear on the top of the screen.
```

PR #11557, opened 2026-06-12
Fix several SE5 RC issues from OmrSi's report (#10766)

niksedk:
Addresses several items from OmrSi's RC test report in discussion #10766. Each is a focused commit.

## Fixes

- **Numpad `+ - * /` shortcut keys** — the capture path prefixed every numpad key with `NumPad`, so the four arithmetic operators became `NumPadAdd`/`NumPadMultiply`/etc. and stopped matching the bare Avalonia `Key` names used by the shortcut dropdown, SE4 import, and the default bindings (the `Shift+Add`/`Shift+Subtract` waveform-zoom defaults were silently broken too). Stop prefixing those four; migrate any tokens stored during RC builds back to the bare names. Digits, `Decimal` and `Enter` keep the prefix. *(report #2)*
- **Go to bookmark seeks the video** — "go to next/previous bookmark" only moved the grid selection; SE4 also moved the video. *(report #3)*
- **Reopen seeks the video to the last line** — opening from the Reopen menu restored the last line but left the video at 0:00; now mirrors the startup file-restore path. *(report #11)*
- **"Toggle dialog dashes" works from the text boxes** — was registered as `SubtitleGrid` (grid-focus only). Moved to `SubtitleGridAndTextBox` and routed that category for the original/translation box too (previously only the normal box was routed), so it works from whichever text box has focus, like SE4. *(report #5)*
- **Windows dark-mode title bar** — SE never set the Win32 immersive dark-mode flag, so the OS title bar stayed light. Added a `DwmSetWindowAttribute` helper (attribute 20, falling back to 19 for older Windows 10) applied via the universal per-window hook, so every window gets it on open and on theme switches. No-op off Windows. *(report #9)*
- **Hidden setting to disable automatic waveform generation** — `Waveform.WaveformAutoGenerate` (Settings.json only, no UI yet, default `true`). When `false`, opening a video no longer kicks off ffmpeg extraction; cached peaks still load. *(report #10)*
- **Keep video position when entering fullscreen** — fullscreen opens a fresh player and seeks back to the saved position, but the seek was applied immediately after the player reported ready and didn't reliably stick — leaving fullscreen at 0:00, most visibly when paused on an empty stretch of the timeline. Settle briefly and re-apply, the same pattern as the file-restore/Reopen paths. (Tier 1: fixes the "resets to beginning"; the brief 0:00 flash from re-opening a second player is a separate, larger change.) *(report #1)*

## Testing

- `dotnet build src/ui/UI.csproj` — clean (0 warnings / 0 errors).
- **Not runtime-verified** — I don't have a Windows box, so the numpad-key, dialog-dashes, video-seek, fullscreen, and especially the dark-title-bar P/Invoke paths should be checked on a real Windows 10 install. Notes per item:
  - Dark title bar: on some Win10 builds the caption may not repaint until the window is re-activated; confirm live dark↔light switching updates open windows.
  - Numpad: capture a numpad key (combobox should now show it), import an SE4 file with numpad shortcuts, and verify `Shift+numpad +/-` zooms the waveform.
  - Fullscreen: the `Task.Delay(200)` settle value is empirical and may need tuning.

Remaining items from the report not addressed here: #4 (find scroll/highlight), #6/#7 (Find layout / grid separator), #8 (unfocused-selection color), #12 (missing SE4 shortcuts).

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -1821,6 +1821,23 @@ public partial class MainViewModel :
 
             await SubtitleOpen(recentFile.SubtitleFileName, recentFile.VideoFileName, recentFile.SelectedLine, desiredAudioTrackId: recentFile.AudioTrack);
 
+            // Seek the video to the restored line, like SE 4 did — otherwise Reopen leaves it
+            // at 0:00. Mirrors the startup file-restore path: the video opens asynchronously, so
+            // wait for the players to be ready before seeking, then re-apply after layout settles.
+            var vp = GetVideoPlayerControl();
+            if (!string.IsNullOrEmpty(_videoFileName) && SelectedSubtitle != null && vp != null)
+            {
+                await vp.WaitForPlayersReadyAsync();
+                await Task.Delay(200);
+                var s = SelectedSubtitle;
+                if (vp != null && s != null)
+                {
+                    vp.Position = s.StartTime.TotalSeconds;
+                    Dispatcher.UIThread.Post(() => { vp.Position = SelectedSubtitle.StartTime.TotalSeconds; });
+                    CenterAudioVisualizerOnCurrentVideoPosition();
+                }
+            }
+
             if (!string.IsNullOrEmpty(recentFile.SubtitleFileNameOriginal) &&
                 File.Exists(recentFile.SubtitleFileNameOriginal))
             {
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -17287,7 +17287,10 @@ public partial class MainViewModel :
                 }
             }
 
-            if (EditTextBox.IsFocused)
+            // Route SubtitleGridAndTextBox shortcuts when either edit box is focused — the
+            // original/translation box should behave the same as the normal one (e.g. so
+            // "Toggle dialog dashes" works in whichever text box has focus, like SE 4).
+            if (EditTextBox.IsFocused || EditTextBoxOriginal.IsFocused)
             {
                 var relayCommand = _shortcutManager.CheckShortcuts(keyEventArgs, ShortcutCategory.SubtitleGridAndTextBox.ToString());
                 if (relayCommand != null)
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -15671,7 +15671,12 @@ public partial class MainViewModel :
         var trackNumber = _audioTrack?.FfIndex ?? -1;
         var peakWaveFileName = WavePeakGenerator2.GetPeakWaveFileName(videoFileName, trackNumber);
         var spectrogramFileName = WavePeakGenerator2.SpectrogramDrawer.GetSpectrogramFileName(videoFileName, trackNumber);
-        if (!File.Exists(peakWaveFileName) || (Se.Settings.Waveform.GenerateSpectrogram && !File.Exists(spectrogramFileName)))
+        var needToGenerate = !File.Exists(peakWaveFileName) || (Se.Settings.Waveform.GenerateSpectrogram && !File.Exists(spectrogramFileName));
+
+        // Hidden setting: when WaveformAutoGenerate is off, never kick off ffmpeg extraction on
+        // video open. Cached peaks still load via the branch below, so existing waveforms keep
+        // showing; the user generates new ones on demand.
+        if (needToGenerate && Se.Settings.Waveform.WaveformAutoGenerate)
         {
             if (FfmpegHelper.IsFfmpegInstalled())
             {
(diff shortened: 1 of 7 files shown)


mjuhasz:
@niksedk Are you already looking into the not addressed items?

I was gonna look into some of these but would like to avoid ending up both of us fixing the same issue. I was gonna start looking at the scrolling issue (has been high on my list), I can look at the Find dialog and grid separator, and the unfocused-selection color has also been high on my list.

Let me know if I should continue or you're already on top of these.

mjuhasz:
Regarding [6353c0e](https://github.com/SubtitleEdit/subtitleedit/pull/11557/commits/6353c0e4bf0acd2d63d2672dc34f96f68e392cb4):

This can be quite slow when moving quickly between bookmarks and the video is a large one on rotational hard drive.

I suggest that we gate it behind a toggle, similar to Go to line number. It would be consistent that way, and I'd argue that in this case the option is even more important. You can't jump to line numbers in quick succession but you can do it with bookmarks.

Screenshot 2026-06-12 at 11 03 40 AM

mjuhasz:
Grid focus tweak is here: #11561

mjuhasz:
Find/Replace layout tweak is here: #11565

mjuhasz:
Unfocused selection items for Listboxes: #11566

With this, data grids and list boxes are aligned, and at this point no other component requires such a tweak.

mjuhasz:
Grid separation and some related visual tweaks: #11568

mjuhasz:
@niksedk  FWIW, I gave another try to the scrolling issue, to no avail. And then there's the unbearable oscillation I'm getting on scroll, too (on macOS at least.)

The root cause is Virtualization. Avalonia's DataGrid only renders rows currently visible in the viewport. Off-screen rows have no DataGridRow element in the visual tree. To scroll to a row that isn't rendered yet, the DataGrid needs to know its Y offset in the content, which requires knowing the cumulative height of all rows above it. With variable heights, those unmeasured rows have unknown heights, so Avalonia can't calculate the correct offset.

The two reliable solutions: 

  - Fixed row height - This works like a charm. Avalonia can pre-calculate every row's Y position exactly, so ScrollIntoView works correctly and centering math is straightforward (index × rowHeight). Scrolling issue, oscillation, all gone. This is what I've always been using in my ow private build: I can't live with the oscillation and unreliable scroll, and have not had any success yet to solve it reliably, variable row height is just not supported in these virtualized data grids.
  - Disable virtualization - all rows are realized at once, so their actual heights are always known and TranslatePoint / BringIntoView work reliably. The trade-off is memory and initial render time for large subtitle files. I have not explored this avenue yet since I've been quite happy with fixed row height and there were many other things to look at in SE5.

niksedk:
@mjuhasz: I am away from home atm, so I will be somewhat slow at replying.
The scrolling seems mostly okay on Windows, so let's keep it for now. We can investigate other controls and/or hope for the table view control.
Disabling virtualization for less than 1000 rows or so could also be worth a shot.

niksedk:
@mjuhasz: ~~We somewhere lost the multi-line grid, can you revert that?~~
(Ups, had the one-line-setting enabled)

OmrSi:
@niksedk @mjuhasz Thank you for all the fixes!

The only thing remaining is number 4, which is basically two parts:
1. Using `Find next`, the line isn't always in focus, I might have to scroll down a bit to see it. This only happens when `Show subtitle text as single line in grid` is disabled.

2. The text I'm trying to find isn't always highlighted in the textbox, the cursor is just set to the end of it.

mjuhasz:
@OmrSi

1. The first issue, when a selected line is not scrolled into the viewport automatically, is not straightforward to fix.

It can happen when the subtitle grid rows have different height, and we programmatically (e.g. via search) select a row. As you noticed, when the option "Show subtitle text as single line in grid" is disabled, we have variable row heights and then this issue occurs.

This affects all grids with variable heights where we select a row and try to scroll it into view, such as the Find common errors or the OCR windows.

If the grid contains rows of equal height, depending on the direction of search, the first or last visible row will be selected (I may tweak this later - there is something to be said for centering the selection like in typewriter mode to see some context: previous and next rows). Just to prove my point that variable row height is the culprit, here's how the main window works with equal row heights:


https://github.com/user-attachments/assets/725a7998-9441-477e-8cb1-d56e7716f44f



There are several ways to go about this, but it will require thorough assessment of those options: memory usage, maintenance burden (does the fix survive upgrades, will it be consistent across all grids etc.), usability etc.

Both the author and myself are on the same opinion that this scrolling glitch should be eliminated but we'll need some more time to come up with a viable solution.

2. I can't replicate your second issue. If you take a video of how you reproduce it, and attach your Settings file (I can help you where to find it - it's easier than telling me all the relevant settings you have), I will look into it. At the very least, besides the video, I will need to know if you have color tags enabled under Settings. If you have the footage just open a Github issue and ping me (@mjuhasz) and I'll take a look. You can open a Github issue by clicking the New Issue button here: https://github.com/SubtitleEdit/subtitleedit/issues

OmrSi:
@mjuhasz Got it. The second issue happens only when this is disabled:
image

Also, two more things to add:
1. Is it normal that the scrollbar isn't clickable? Like, you can't click on the bar itself, you have to hold the slider and take it up and down, and you can't click the arrows at the top and bottom either, nothing happens.

2. If you open SE without any file already being open, you get an asterisk in the title, even though nothing was actually changed, and it goes away if you click `New`.

image

mjuhasz:
@OmrSi 

* I managed to replicate and fix the find highlight issue, it's already merged: #11579
* I fixed the asterisk after opening, also merged: #11578
* I improved the scrollbar handling, that's in the review queue: #11591

PR #11568, opened 2026-06-12
Add bordered UI sections to main editor and Fix Common Errors dialog

mjuhasz:
## Summary

  Addresses feedback that the subtitle list and edit area visually blend together.

  ### Main window
  - Wraps the subtitle DataGrid and the edit box (with timing, formatting, and info controls) in separate gray-bordered panels, replacing the plain horizontal separator between them.

main window

  ### Fix Common Errors — Step 2
  - Each grid (fixes and subtitles) now has its own gray border.
  - The "Select all / Invert / Refresh / Apply" buttons sit in the visible gap between the two grids.
  - The edit panel is left-aligned with the subtitles grid at a fixed width, with navigation buttons (Back to fix list, Done, Cancel) anchored to its bottom-right.
  - The fixes-applied count moves next to the step title, separated by a vertical divider that hides when no fixes have been applied.
  - `MinWidth` raised to prevent button clipping on narrow windows.

Fix common errors

### DIFF
--- a/src/ui/Features/Main/Layout/InitListViewAndEditBox.cs
+++ b/src/ui/Features/Main/Layout/InitListViewAndEditBox.cs
@@ -545,8 +545,10 @@ public static partial class InitListViewAndEditBox
             Source = vm,
         };
 
-        Grid.SetRow(dropHost, 0);
-        mainGrid.Children.Add(dropHost);
+        var dropHostBorder = UiUtil.MakeBorderForControlNoPadding(dropHost);
+        dropHostBorder.Margin = new Thickness(UiUtil.SplitterWidthOrHeight, UiUtil.SplitterWidthOrHeight, UiUtil.SplitterWidthOrHeight, 0);
+        Grid.SetRow(dropHostBorder, 0);
+        mainGrid.Children.Add(dropHostBorder);
 
         // Create a Flyout for the DataGrid
         var flyout = new MenuFlyout();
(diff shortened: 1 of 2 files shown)


mjuhasz:
@niksedk I see that you have not merged this one. I'm always open to feedback if there is something you don't like or would do differently, or if you would like me to explain my thinking.

niksedk:
> @niksedk I see that you have not merged this one. I'm always open to feedback if there is something you don't like or would do differently, or if you would like me to explain my thinking.

I've tried to remove most of the boxes from SE4 on purpose, so I will close this for now.

OmrSi:
@niksedk Especially in the main window, the textbox area doesn't look good without a border/separator at all. It looks very blended with the listview.
It needs at least a separator here, or something to give it its own space and defined borders. Even a disabled GridSplitter might work.

Here is the case I'm talking about:
image

mjuhasz:
There is also some inconsistency here that would be good to resolve, in whatever interface guidelines the app wants to follow.

For example, the OCR window grid is boxed (see screenshot below), but the main window grid is not, and it looks out of place to my eyes, similar to the Fix common errors grids: when the lowermost rows are cut (see FCE screenshot below) it looks like we dropped some buttons / a textbox on top of the grid and they are floating, and I get the same depth perception on the main window grid.

OCR window:

OCR

Fix Common Errors window:

FCE

PR #11581, opened 2026-06-13
Fix burn-in passing 0x0 resolution to ffmpeg for audio-only input

### LINKED ISSUE #11570: Bug Report 2: Pure Audio to Video Burn-In Passes 0x0 Resolution to FFmpeg
Description:

When attempting to use the Video -> Burn-In Subtitles feature with only a subtitle file and a pure audio file (.wav/.mp3) loaded in the main workspace (no background video file preset), it fails to pass user-defined canvas constraints to the FFmpeg execution string.

Even if the resolution fields are manually changed to 1920x1080 in the UI panel, clicking Generate reveals that Subtitle Edit passes -vf "scale=0:0..." and -s 0x0 to the command line.

Steps to Reproduce:

    Load an .ass subtitle file.
    Load a pure audio track into the waveform visualizer.
    Open Generate Video with Burn-In Subtitles.
    Set the resolution fields to 1920x1080.
    Set audio to mp3 320kbps
    Check the FFmpeg command line parameters.

Generated Command Line (Incorrect):
ffmpeg -i "input.wav" -vf "scale=0:0,ass=temp.ass" -g 30 -bf 2 -s 0x0 -c:v libx264 -preset slow -crf 23 -pix_fmt yuv420p -c:a mp3 "output.mkv"

Expected Command Line (Correct):
ffmpeg -i "input.wav" -f lavfi -i color=c=black:s=1920x1080 -vf "ass='input.ass'" -c:v libx264 -preset slow -crf 23 -pix_fmt yuv420p -c:a mp3 -shortest "output.mp4"

Technical Notes

    Virtual Canvas Generation: Replaces the broken -s 0x0 and -vf "scale=0:0..." arguments with -f lavfi -i color=c=black:s=1920x1080 to dynamically generate a valid background layout matching the user's resolution choice if working from pure audio and subtitle file.

    Bypassing temp ass Cache: References the live, active user script (input.ass) directly inside the working folder rather than pointing to a temp.ass clone, ensuring timeline modifications aren't lost on rerun.

    Stream Containment: Appends the -shortest flag to automatically end the video processing matrix when the input audio file terminates, outputting a standard, compliant .mp4 video.

Ideally the option to choose mp4 as the output and extension instead of only mkv option would be nice.

Thank you for your great software BTW <David>

muaz978:
Fixes #11570

## Problem

With only a subtitle and a pure audio file (.wav/.mp3) loaded (karaoke workflow), *Video -> Generate video with burned-in subtitles* passed `-vf "scale=0:0,..."` and `-s 0x0` to ffmpeg - even when the user typed a valid resolution like 1920x1080 in the fields.

## Root cause

1. The parsed media dimension (0x0 for an audio file) unconditionally overwrote both the UI resolution fields (`Initialize`, `GetCurrentVideoAsJobItems`) and the job item at generate time (`InitAndStartJobItem`), discarding whatever the user entered.
2. Even with correct numbers, an audio file has no video stream for `-vf` to burn subtitles into.

## Fix

- Only take width/height from parsed media info when non-zero, so the user-chosen resolution survives for audio-only input.
- When the input has audio but no video stream (detected via the parsed track list), synthesize a black canvas with `-f lavfi -i color=c=black:s=WxH:r=25`, burn the subtitles onto it, and add `-shortest` so encoding stops when the audio ends. This is the same lavfi approach the mpv audio-only preview already uses (`LibMpvDynamicPlayer`), matching the behaviour @dask-music described in the issue.
- Logo overlay still works: the `filter_complex` stream labels shift to `[1:v]`/`[2:v]` when the canvas input is present.
- Progress frame count for audio input derived from duration x 25 fps (the canvas rate).

Generated command now (audio-only input):
```
ffmpeg -i "input.wav" -f lavfi -i color=c=black:s=1920x1080:r=25 -vf "scale=1920:1080,ass=sub.ass" -g 30 -bf 2 -s 1920x1080 -c:v libx264 -preset medium -crf 25 -pix_fmt yuv420p -c:a aac -ar 44100 -use_editlist 0 -movflags +faststart -shortest "out.mkv"
```
Video-file inputs are unchanged - dimensions still come from the source when non-zero.

## Verification

- `dotnet build`: 0 errors / 0 warnings (full solution).
- End-to-end with ffmpeg on Windows: a 5 s sine `.wav` + a minimal `.ass` produces a 1920x1080 H.264/AAC file with duration matching the audio (5.02 s - `-shortest` terminates the otherwise-infinite color source).
- Subtitle rendering confirmed via signalstats luma (YAVG 16.52 on a subtitle frame vs 16.00 black floor).
- Logo-overlay variant with the remapped stream labels verified the same way.

## Out of scope

The issue's secondary request for an mp4 output option is intentionally left out to keep this a focused fix.

### DIFF
--- a/src/ui/Features/Video/BurnIn/BurnInJobItem.cs
+++ b/src/ui/Features/Video/BurnIn/BurnInJobItem.cs
@@ -25,6 +25,12 @@ public partial class BurnInJobItem : ObservableObject
     public double TotalSeconds { get; set; }
     public string VideoBitRate { get; set; }
 
+    /// <summary>
+    /// Input has audio but no video stream (e.g. karaoke from a .wav + .ass), so the
+    /// subtitles are burned onto a generated black canvas at the chosen resolution.
+    /// </summary>
+    public bool InputIsAudioOnly { get; set; }
+
     [ObservableProperty] private string _size;
 
     [ObservableProperty] private string _status;
--- a/src/ui/Features/Video/BurnIn/BurnInViewModel.cs
+++ b/src/ui/Features/Video/BurnIn/BurnInViewModel.cs
@@ -261,8 +261,14 @@ public partial class BurnInViewModel : ObservableObject
                 _mediaInfo = FfmpegMediaInfo2.Parse(videoFileName);
                 Dispatcher.UIThread.Post(() =>
                 {
-                    VideoWidth = _mediaInfo.Dimension.Width;
-                    VideoHeight = _mediaInfo.Dimension.Height;
+                    // Audio-only files have no video stream (0x0) - keep the configured
+                    // resolution instead of overwriting the fields with zeros (issue #11570).
+                    if (_mediaInfo.Dimension.Width > 0 && _mediaInfo.Dimension.Height > 0)
+                    {
+                        VideoWidth = _mediaInfo.Dimension.Width;
+                        VideoHeight = _mediaInfo.Dimension.Height;
+                    }
+
                     UseSourceResolution = false;
                 });
             });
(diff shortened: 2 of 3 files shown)


dask-music:
Thank you for this. I will test everything on the next release candidate and open any new issues strictly as single issues in future.

Issue #2, opened 2014-02-04
(Suggestion) CPS and separation in subtitle list


rebawest:
Let me suggest to add CPS and Separation (or Blank, Gap, Pause) to the subtitle list.
I attach a capture of Softni subtitler.
I know you can't have 2 lines in a listview, so two more columns would be needed is this is done.

Notice that by having more info in a single cell (in Softni) the subtitle list is narrower, and so there is more space for the video on the right. Also, each subtitle in the list appears in two lines which makes them easier to read.
So maybe a in a future version the list could be improved, maybe using a textbox inside each cell; I can't think of any other way of having line breaks inside a cell.

Thank you very much. Regards...

niksedk:
CPS available in list view in latest beta

Issue #1, opened 2014-02-04
Ebu Stl - Italics


rebawest:
When exporting to EBU STL, italics do not work properly. When a subtitle has both lines in italics, only the first one gets the formatting.

(This old issue #155. https://code.google.com/p/subtitleedit/issues/detail?id=155 )

PR #11591, opened 2026-06-13
Fix scrollbar interactivity and appearance in data grids and list views

mjuhasz:
## Summary

  Three related scrollbar bugs fixed in this pass:

  ### 1. Scrollbar clicks not working in the main subtitle grid

  Clicking the scrollbar track or arrow buttons had no effect — only thumb dragging worked. The root cause was `EndSubtitleGridDragSelect` calling `e.Pointer.Capture(null)` unconditionally in the tunnel phase, releasing pointer capture before the scrollbar's `RepeatButton` controls could process `PointerReleased` in the bubble phase. `RepeatButton` depends on captured pointer state to manage its interaction lifecycle.

  **Fix:** Guard `Pointer.Capture(null)` with `if (_dragSelectStartIndex >= 0)` so it only fires during actual drag-select operations. Also early-return from `SubtitleGrid_PointerPressed` when the click originates inside a `ScrollBar`.

  ### 2. Scrollbar clicks triggering double-tap actions

  Holding a scrollbar arrow button fires `RepeatButton` rapidly; two firings within Avalonia's double-tap threshold generate a spurious `DoubleTapped` event that bubbles up to the `DataGrid`. This caused the OCR window's subtitle grid to open the inspection window when scrolling with the arrow buttons.

  **Fix:** Added `UiUtil.IsScrollBarSource(RoutedEventArgs)` helper and applied it as an early-return guard to all at-risk `DoubleTapped` handlers (OCR, Shortcuts, Shot Changes, Find Double Lines/Words, Review Speech, Embedded Subtitles Edit windows).

  ### 3. Scrollbars not showing expanded appearance on macOS "Always" setting

  When macOS System Settings → Appearance → Show scroll bars is set to **Always**, scrollbars in `DataGrid` controls still showed the compact (2 px) overlay style instead of the full expanded appearance.

  **Root cause:** `DataGrid.OnApplyTemplate` sets `AllowAutoHide=true` on its internal `ScrollBar` instances at LocalValue priority (0). No `Application.Styles` setter, resource override, or template selector can beat LocalValue priority.

  **Fix:** A `Control.LoadedEvent` class handler registered via `Control.LoadedEvent.AddClassHandler<DataGrid>` fires after `OnApplyTemplate` and sets `AllowAutoHide=false` at LocalValue priority, overriding DataGrid's value. This triggers Avalonia's `UpdateIsExpandedState()` → `IsExpanded=true`, which makes the Fluent `[IsExpanded=True]` ControlTheme apply the correct expanded appearance (full thumb size, colour, visible arrow buttons).

  The macOS scroll bar preference is read from `defaults read -g AppleShowScrollBars`.

  
> [!NOTE]
> On Windows and Linux, `AllowAutoHide=false` is applied unconditionally (scrollbars are always visible). That is the correct native Windows expectation, and I am not aware of this being universally configurable. On macOS, it matches the OS preference. The non-macOS preference could be set differently, or have a setting for it (not macOS -> reveal option to always show expanded scrollbars or auto-hide them).

  ## Files changed

  - `src/ui/Features/Main/MainViewModel.cs` — pointer capture guard + ScrollBar early-return
  - `src/ui/Logic/UiUtil.cs` — `IsScrollBarSource()` helper
  - `src/ui/Logic/UiTheme.cs` — `ApplyScrollBarStyle()` with DataGrid `Loaded` class handler
  - `src/ui/Logic/MacHelper.cs` — `GetShowScrollBarsPreference()`
  - `src/ui/Program.cs` — call `UiTheme.ApplyScrollBarStyle()` at startup
  - 9 window/viewmodel files — `DoubleTapped` guards using `IsScrollBarSource`

   ## Screenshots

The scrollbar has the following parts:
  - Thumb — the draggable bar that moves
  - Track — the full channel the thumb slides in (the clickable area between the buttons)
  - Arrow buttons (or step buttons) — the small buttons at each end
  - Page up / Page down area — the track region between the thumb and an arrow button (clicking here scrolls by one page)

  #### Expanded scrollbar

   Thumb (draggable bar) is functioning as it should, so do the arrow buttons, and the page up/down area of the track can be clicked as well. Note: Replicating the macOS setting to configure the page up/down area to jump to the next page or the spot clicked would be very involved and imho it's not worth the complexity.

visible

 #### Compact scrollbar

  Also fully functional once its components get revealed.

hidden

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -17814,7 +17814,9 @@ public partial class MainViewModel :
     {
         StopSubtitleGridDragSelectAutoScroll();
         if (_dragSelectStartIndex >= 0)
+        {
             e.Pointer.Capture(null);
+        }
         _dragSelectStartIndex = -1;
         _dragSelectLastIndex = -1;
         _dragSelectAppliedIndex = -1;
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -17407,6 +17407,11 @@ public partial class MainViewModel :
                     return;
                 }
 
+                if (current is ScrollBar)
+                {
+                    return;
+                }
+
                 current = current.Parent as Control;
             }
 
@@ -17808,7 +17813,8 @@ public partial class MainViewModel :
     private void EndSubtitleGridDragSelect(PointerEventArgs e)
     {
         StopSubtitleGridDragSelectAutoScroll();
-        e.Pointer.Capture(null);
+        if (_dragSelectStartIndex >= 0)
+            e.Pointer.Capture(null);
         _dragSelectStartIndex = -1;
         _dragSelectLastIndex = -1;
         _dragSelectAppliedIndex = -1;
--- a/src/ui/Features/Ocr/OcrWindow.cs
+++ b/src/ui/Features/Ocr/OcrWindow.cs
@@ -459,7 +459,7 @@ public class OcrWindow : Window
         };
         dataGridSubtitle.Bind(DataGrid.SelectedItemProperty, new Binding(nameof(vm.SelectedOcrSubtitleItem)) { Source = vm });
         dataGridSubtitle.KeyDown += vm.SubtitleGridKeyDown;
-        dataGridSubtitle.DoubleTapped += (s, e) => { if (!UiUtil.IsScrollBarSource(e)) vm.SubtitleGridDoubleTapped(); };
+        dataGridSubtitle.DoubleTapped += (s, e) => vm.SubtitleGridDoubleTapped();
         dataGridSubtitle.AddHandler(InputElement.PointerPressedEvent, vm.DataGridSubtitleMacPointerPressed, Avalonia.Interactivity.RoutingStrategies.Tunnel);
         dataGridSubtitle.AddHandler(InputElement.PointerReleasedEvent, vm.DataGridSubtitleMacPointerReleased, Avalonia.Interactivity.RoutingStrategies.Tunnel);
         dataGridSubtitle.SelectionChanged += vm.SubtitleGridSelectionChanged;
--- a/src/ui/Features/Ocr/OcrWindow.cs
+++ b/src/ui/Features/Ocr/OcrWindow.cs
@@ -459,7 +459,7 @@ public class OcrWindow : Window
         };
         dataGridSubtitle.Bind(DataGrid.SelectedItemProperty, new Binding(nameof(vm.SelectedOcrSubtitleItem)) { Source = vm });
         dataGridSubtitle.KeyDown += vm.SubtitleGridKeyDown;
-        dataGridSubtitle.DoubleTapped += (s, e) => vm.SubtitleGridDoubleTapped();
+        dataGridSubtitle.DoubleTapped += (s, e) => { if (!UiUtil.IsScrollBarSource(e)) vm.SubtitleGridDoubleTapped(); };
         dataGridSubtitle.AddHandler(InputElement.PointerPressedEvent, vm.DataGridSubtitleMacPointerPressed, Avalonia.Interactivity.RoutingStrategies.Tunnel);
         dataGridSubtitle.AddHandler(InputElement.PointerReleasedEvent, vm.DataGridSubtitleMacPointerReleased, Avalonia.Interactivity.RoutingStrategies.Tunnel);
         dataGridSubtitle.SelectionChanged += vm.SubtitleGridSelectionChanged;
--- a/src/ui/Features/Options/Shortcuts/ShortcutsViewModel.cs
+++ b/src/ui/Features/Options/Shortcuts/ShortcutsViewModel.cs
@@ -1123,6 +1123,7 @@ public partial class ShortcutsViewModel : ObservableObject
 
     internal void ShortcutsDataGridDoubleTapped(object? sender, TappedEventArgs e)
     {
+        if (UiUtil.IsScrollBarSource(e)) return;
         _ = ShowGetKey();
     }
 
--- a/src/ui/Features/Options/Shortcuts/ShortcutsViewModel.cs
+++ b/src/ui/Features/Options/Shortcuts/ShortcutsViewModel.cs
@@ -1123,7 +1123,6 @@ public partial class ShortcutsViewModel : ObservableObject
 
     internal void ShortcutsDataGridDoubleTapped(object? sender, TappedEventArgs e)
     {
-        if (UiUtil.IsScrollBarSource(e)) return;
         _ = ShowGetKey();
     }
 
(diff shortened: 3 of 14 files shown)


niksedk:
Solid, well-researched PR. The problem analysis (especially the Avalonia `AllowAutoHide` / LocalValue-priority explanation) is accurate, and fix #1 is correct. Main concern: fix #2 is applied inconsistently.

### ✅ Fix #1 — pointer-capture guard (correct)

Traced the capture lifecycle. Pointer capture is acquired in exactly one place — `MainViewModel.cs:17542`, inside the `_dragSelectHasMoved` branch, which only runs after `_dragSelectStartIndex` was set during a drag. So coupling the *release* to `if (_dragSelectStartIndex >= 0)` correctly mirrors the *acquire* condition. A plain scrollbar click never captured, so skipping the release is right. The `current is ScrollBar` early-return in `SubtitleGrid_PointerPressed` sits in the same hit-test walk that already handles `DataGridColumnHeader`, so it's a consistent spot. No issues.

### ⚠️ Fix #2 — double-tap guard is incomplete (main concern)

The `IsScrollBarSource` guard was added to ~9 handlers, but `src/ui` has 20+ `DoubleTapped` handlers on `DataGrid`/`ListBox`, and the same RepeatButton-fires-`DoubleTapped` bug applies to all of them. Unguarded ones include:

- **Bookmark grids** → `OnBookmarksGridDoubleTapped`: `FindDoubleLinesWindow.cs:71`, `FindDoubleWordsWindow.cs:72`, `ErrorListWindow.cs:72`, `BookmarksListWindow.cs:83` (VM methods weren't touched)
- `BinaryEditWindow:372`, `FindTextWindow:128`, `CutVideoWindow:150`, `PickTsTrackWindow:130`, `PickMatroskaTrackWindow:138`, `PickVobSubLanguageWindow:112`, `PickOnlineSubtitleWindow:113`, `FindRuleWindow:128`, `PickRuleProfileWindow:109`, `ExportCustomTextFormatWindow:100,122`, `MultipleReplaceWindow:286`
- ListBox handlers (also scrollable): `PromptUnknownWordWindow:261`, `SpellCheckWindow:280`, `PickSubtitleFormatWindow:64`, `AssaTagHistoryWindow:47`

Notably `FindDoubleLines`/`FindDoubleWords` got the *main* grid guarded (line 95/106) but their *bookmark* grid (line 71/72) left unguarded — so even the touched windows are only half-covered.

**Recommendation:** replace the per-handler guards with a single global class handler, the same pattern already used for the `LoadedEvent` fix:

```csharp
DataGrid.DoubleTappedEvent.AddClassHandler<DataGrid>((_, e) =>
{
    if (UiUtil.IsScrollBarSource(e)) e.Handled = true;
});
// + same for ListBox / ItemsControl base
```

DRY, covers every current and future grid, and removes the need to remember the guard in each new `DoubleTapped` wiring. It also lets you revert the 9 call-site edits to keep the diff small. (`e.Handled` on a scrollbar source is always safe — double-tap is never a legitimate gesture there.)

### ⚠️ `ApplyScrollBarStyle` re-invocation asymmetry (latent)

The style part is idempotent (removes + re-adds `_scrollBarStyle`), but the `Control.LoadedEvent` class handler is registered once, *unconditionally* sets `AllowAutoHide = false`, and is never removed. If the method is ever called a second time with `allowAutoHide = true` (macOS pref = *WhenScrolling*), DataGrids would stay forced-expanded while every other control auto-hides — inconsistent. Only called once at startup today, so not a live bug, but the method *reads* as re-invokable (it bothers to remove/re-add the style). Either capture the current preference in the handler body, or add a comment that it's startup-only.

### 🔧 Minor

- **Unused usings** in `UiTheme.cs`: `Avalonia.Controls.Shapes`, `Avalonia.Markup.Xaml.Styling`, `Avalonia.Threading`, and likely `System.Collections.Generic` don't appear to be referenced by the new code — please drop them if so.
- **Brace style**: `if (_dragSelectStartIndex >= 0) e.Pointer.Capture(null);` and `if (current is ScrollBar) return true;` omit braces, deviating from the file's braces-on-every-`if` convention.
- **Tree consistency**: `IsScrollBarSource` walks the logical `.Parent` chain, while the new DataGrid handler uses visual `GetVisualDescendants`. The `.Parent` walk matches the existing working code at `MainViewModel.cs:17404`, so it's probably fine — but worth a quick check that a click on a *track* RepeatButton (not just the thumb) reports an `e.Source` whose logical chain reaches the `ScrollBar`. Visual-ancestor walking would be more robust across template boundaries.

### Verdict

Direction is good; fixes #1 and #3 are sound. The one change I'd ask for before merge: **make the double-tap guard global** (a class handler) rather than per-handler, since the current set leaves the majority of grids — including half of the ones it touches — still vulnerable. The unused usings and brace nits are trivial cleanups.

mjuhasz:
Thank you for the thorough review. All the main points are addressed. Here's what changed and what remains.

  **Fixed in follow-up commit:**

  - **Fix `#2` — global class handlers**: replaced the per-handler `IsScrollBarSource` guards with two global class handlers in `Program.cs`:

    ```csharp
    DataGrid.DoubleTappedEvent.AddClassHandler<DataGrid>((_, e) => { if (UiUtil.IsScrollBarSource(e)) e.Handled = true; });
    ListBox.DoubleTappedEvent.AddClassHandler<ListBox>((_, e) => { if (UiUtil.IsScrollBarSource(e)) e.Handled = true; });
  All 9 per-handler call-site edits reverted. Covers all current and future DataGrid/ListBox handlers.

  - Unused usings: removed Avalonia.Controls.Shapes, Avalonia.Markup.Xaml.Styling, Avalonia.Threading, System.Collections.Generic from UiTheme.cs. - they were leftovers from a merge conflict resolution.
  - Brace style: both single-line if blocks now have braces.
  - ApplyScrollBarStyle asymmetry: the LoadedEvent handler now reads _scrollBarAllowAutoHide as a static field (not a captured local), so a hypothetical second call with a different preference would correctly affect future DataGrid loads. Added a doc comment noting that already-open DataGrids won't update (the Loaded event won't re-fire), and that the method is designed to be called once at startup.
  - GetShowScrollBarsPreference — stderr: removed RedirectStandardError = true since stderr was never read; the unread pipe was a latent deadlock risk.

  Tree walk consistency: IsScrollBarSource uses .Parent (logical) — same as the existing SubtitleGrid_PointerPressed handler. A RepeatButton (track) click has RepeatButton → Track → ScrollBar in its logical parent chain, so it's covered. GetVisualDescendants in the LoadedEvent handler uses the visual tree, which is necessary to find the scrollbars inside the DataGrid's template. Both are appropriate for their context.

  handledEventsToo: true note: SubtitleGridDropHost_DoubleTapped fires even after the global class handler marks the event handled. It returns early harmlessly because GetDataGridRowIndexFromPoint returns -1 for scrollbar hit-tests. The handledEventsToo: true was pre-existing for drag-select behavior and isn't related to this fix.

Issue #10068, opened 2025-11-09
Command-line BDN-XML export forcibly performs OCR and embeds OCR errors into generated PNG images


MavericksNightmare:
**Description:**

#### Introduction  
First of all, thank you very much for providing and maintaining Subtitle Edit — it’s an excellent tool and an essential part of many video workflows.  
I’ve been using it for archiving and converting Blu-ray subtitle tracks, and I truly appreciate the amount of work and care that has gone into both the GUI and the command-line interface.  

While doing this, I’ve run into a small but rather unfortunate limitation in the command-line conversion feature that seems to make it unusable for certain automation scenarios. I’d like to describe it here in case it’s an oversight or could be improved in a future version.  

---

#### Environment  
- Subtitle Edit version: **4.0.14.0**  
- Operating System: **Windows 11 (64-bit)**  

---

#### Summary  
When exporting Blu-ray PGS/SUP subtitles to graphics-based formats such as BDN-XML from the command line, Subtitle Edit appears to *always* perform OCR, even when it is neither required nor desired.  The resulting PNG images will hence not contain original subtitle texts but OCR texts, modified by OCR recognition errors. This makes it difficult to use the exported BDN-XML format since OCR errors are already firmly embedded into the images, for anything, including external OCR processing. 

---

#### Expected behaviour  
As long as not specifically requested by the user otherwise, the command-line export to BDN-XML and other image-based formats should produce PNG images that represent the correct **original subtitle graphics**, without any OCR processing or text modification.  
If OCR is desired, it should be optional and controlled via a clear CLI parameter (for example `/ocr:none`, `/ocrengine:tesseract`, etc.).  

---

#### Actual behaviour  
Even when using the command line with:  
```
SubtitleEdit /convert "Venom_Eng.sup" BDN-XML /ocrengine:nOCR /outputfolder:tmp
```
the tool appears to perform OCR and then embed the (faulty) recognized text into the output PNG files.

Example (taken from the English subtitle track of *Venom*):  
- Original subtitle text (as seen in the SubtitleEdit GUI):  
  > I’m glad that you like it  
- Text shown in the PNG exported via command line:  
  > L’m glad that you like it  

[...]

PR #11592, opened 2026-06-13
Qwen3 TTS (CrispASR): add real voice cloning + fix CustomVoice speakers

niksedk:
## Problem

The **Qwen3 TTS (CrispASR)** "1.7B CustomVoice" model was presented as voice cloning, but it never used the cloned voice.

Root cause: the crispasr `qwen3-tts-1.7b-customvoice` backend does **not** clone — it selects one of nine **fixed speakers** baked into the GGUF, chosen by name (`vivian`, `ryan`, …). SE was sending the imported reference WAV's **absolute path** as the request's `voice` field. The backend treats a `.wav` value as an invalid speaker name and silently falls back to its first speaker (`aiden`), so synthesis produced a generic voice, not the clone.

The actual voice-clone-from-WAV backend is `qwen3-tts-1.7b-base` (ICL cloning from a reference WAV + its transcript), which SE didn't wire up.

## Changes

Both features are now available, each mapped to the correct backend:

- **New "1.7B Voice clone" model** → `qwen3-tts-1.7b-base`. Real ICL cloning from an imported reference WAV + transcript.
  - Download wiring, byte-size guard, and SHA-256 for `qwen3-tts-12hz-1.7b-base-q8_0.gguf` (reuses the existing 12 Hz codec, so it's just the one talker file).
  - Sends the **bare voice name** (not the absolute path) so the server resolves `<voice-dir>/<name>.wav` and auto-loads `<name>.txt` as ref-text.
  - Cloning errors without a transcript, so the import flow now **prompts for the transcription** (sibling-`.txt` autofill + speech-to-text helper, matching CosyVoice3/F5) and writes the sidecar. `Speak()` surfaces a clear up-front error if the WAV or transcript is missing.
- **CustomVoice fixed** → now lists the nine real built-in speakers (`aiden`, `dylan`, `eric`, `ono_anna`, `ryan`, `serena`, `sohee`, `uncle_fu`, `vivian`) and sends the speaker name instead of a WAV path.
- Settings dialog shows install status for the new Base talker.

## Notes / migration

- Existing users on "1.7B CustomVoice" stay on it (now the built-in-speaker model). To clone an imported voice, pick **"1.7B Voice clone"** — this downloads the ~1.9 GB base talker on first use.
- Voices imported before this change have no transcript sidecar; re-import (or add a `<name>.txt`) to use them for cloning.

Related: SubtitleEdit/subtitleedit#10068

## Testing

- `dotnet build src/ui/UI.csproj` — clean (0 warnings, 0 errors).
- Backend semantics, the `voice`-field resolution, the `--voice-dir`/`.txt` auto-load, and the base GGUF size/SHA were verified against the CrispASR source (`crispasr_backend_qwen3_tts.cpp`, `crispasr_model_registry.cpp`) and HuggingFace metadata.

### DIFF
--- a/src/ui/Features/Video/TextToSpeech/Engines/Qwen3TtsCrispAsr.cs
+++ b/src/ui/Features/Video/TextToSpeech/Engines/Qwen3TtsCrispAsr.cs
@@ -266,12 +266,15 @@ public class Qwen3TtsCrispAsr : ITtsEngine
     private static bool _voiceSeedAttempted;
 
     /// <summary>
-    /// One-time best-effort copy of WAV reference voices from the existing qwen3-tts.cpp
-    /// engine's voices folder. The same files work for both engines (just 24 kHz mono
-    /// reference audio with optional .txt sidecars), so users who already downloaded the
-    /// qwen3-tts.cpp voice pack get them for free here without a second ~10 MB download.
-    /// Users who never installed qwen3-tts.cpp get voices.zip pulled by the
-    /// DownloadTtsViewModel flow that chains after model download.
+    /// One-time best-effort seeding of reference voices from the existing qwen3-tts.cpp engine's
+    /// voices folder, so users who already have that voice pack get the WAVs here for free. The
+    /// Base (Voice clone) backend strictly requires a 24 kHz mono reference, so each WAV is
+    /// resampled on the way in rather than plain-copied (the qwen3-tts.cpp pack ships 22.05/48 kHz
+    /// clips). The .txt sidecars in that pack are Wikimedia attribution blurbs, NOT spoken
+    /// transcriptions — feeding them as ref-text produces runaway, off-voice output — so they are
+    /// deliberately NOT copied. Cloning a seeded voice prompts for the real transcript via the
+    /// Speak pre-check / re-import flow. Users who never installed qwen3-tts.cpp get voices.zip
+    /// pulled by the DownloadTtsViewModel flow that chains after model download.
     /// </summary>
     private static void SeedVoicesFromQwen3TtsCppIfEmpty(string voicesFolder)
     {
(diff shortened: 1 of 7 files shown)


niksedk:
## End-to-end verification (live against CrispASR 0.7.1, Metal)

Ran the actual `crispasr` binary in `--server` mode exactly as SE launches it and hit `/v1/audio/speech`.

**Binary support** — `--list-backends` confirms `qwen3-tts-1.7b-base`, `qwen3-tts-1.7b-customvoice`, and `qwen3-tts-1.7b-voicedesign` are all compiled in, so the fix works with the currently-shipped CrispASR (no runtime update needed).

**Base GGUF download wiring** — downloaded `qwen3-tts-12hz-1.7b-base-q8_0.gguf`: size `2066258176` and SHA-256 `bbb93ab1…520d` both match the values committed to `Qwen3TtsCrispAsr.ExpectedFileSizes` / `DownloadHashManager` exactly.

**CustomVoice (built-in speakers)** — `voice:"vivian"` → HTTP 200, valid 24 kHz WAV, backend log `selected CustomVoice speaker 'vivian'`. The model's nine speakers match the hardcoded list byte-for-byte: `aiden, dylan, eric, ono_anna, ryan, serena, sohee, uncle_fu, vivian`. The old behavior (sending a WAV path) was reproduced: `expects a speaker NAME … falling back to first speaker` → `aiden` (the original bug).

**Voice clone (Base)** — bare name resolves `<voice-dir>/<name>.wav` + auto-loads `<name>.txt`; backend log shows the ECAPA speaker encoder + `ICL prefill` running (cloning engaged, not a default voice). With a correctly-matched 24 kHz reference + transcript: **HTTP 200 in 3 s, 1.12 s of audio for a short line**. The old absolute-path form fails with `--ref-text was not set`, confirming the bare name is required.

### Issues found and fixed during verification
- Base requires **exactly 24 kHz** mono. The import path already resamples; the **seeding** path plain-copied 22.05/48 kHz clips → now resampled on seed (2nd commit).
- The qwen3-tts.cpp pack's `.txt` sidecars are **attribution blurbs, not transcripts**; using one as ref-text caused a ~240 s runaway render. Seeding no longer copies them, so the transcript prompt / pre-check drives a correct transcription (2nd commit).

Build clean (0/0). Temporary test artifacts removed; the downloaded base GGUF left in place (it's the real model, SHA-verified).

Issue #11590, opened 2026-06-13
[Feature Request] Please add an auto-fit setting for the subtitle grid, so the visible subtitle text length can automatically adapt to the window size.


lsmxpftz-cmyk:
<img width="722" height="540" alt="Image" src="https://github.com/user-attachments/assets/600725ff-41c9-4fd4-83e7-fea685fdf95d" />

 ⬆️Just like in the gif above.

Also, thank you so much for all your hard work on this software. I really appreciate everything you’ve done to make it better.💗

muaz978:
I took a look at this. Today the grid Text/Original columns are hardcoded to no-wrap, so long lines get clipped at the column edge and multi-line subtitles make very tall rows. I put together a setting so the visible text can adapt to the window size.

**Settings -> Appearance -> "Subtitle text fit in grid"**, three modes:
- **Clip (single line)** - current behaviour, unchanged (default).
- **Wrap to fit window** - text wraps to the column width and rows grow to show the full text (this is what your screenshot shows).
- **Single line with ellipsis** - one compact line, truncated with an ellipsis that adapts to the window width.

PR: SubtitleEdit/subtitleedit#11594 (default stays Clip, so nothing changes unless you pick a mode).

If you'd like to try it before it's reviewed, here are test builds from that branch - pick the mode and resize the window to see the difference:
- Windows (portable x64): https://github.com/muaz978/subtitleedit/releases/tag/win-test-20260613-203441
- macOS (Apple Silicon + Intel): https://github.com/muaz978/subtitleedit/releases/tag/macos-test-20260613-203408

These are unofficial test builds from a fork, for trying the feature only. Feedback welcome on naming/defaults.

PR #11594, opened 2026-06-13
Add subtitle grid text fit setting: clip / wrap / ellipsis (#11590)

muaz978:
Implements #11590 - an auto-fit setting for the subtitle grid so the visible text adapts to the window size.

## Background

The grid Text and Original columns were hardcoded to `TextWrapping.NoWrap`, so long lines are clipped at the column edge (you can't see the rest without widening) and multi-line subtitles produce very tall rows.

## Change

Adds an appearance setting **"Subtitle text fit in grid"** (Settings -> Appearance) with three modes:

- **Clip (single line)** - default, unchanged: `NoWrap`, long lines clipped at the edge.
- **Wrap to fit window** - `TextWrapping.Wrap`; text reflows to the column width and rows grow to show the full text (this is what the screenshot in the issue shows).
- **Single line with ellipsis** - `NoWrap` + `MaxLines=1` + `CharacterEllipsis`; one compact line, truncated with an ellipsis that adapts to the column (window) width.

It is modelled on the existing grid-lines-visibility dropdown: a `SubtitleGridTextDisplayMode` enum + display list, persisted in `SeAppearance.SubtitleGridTextDisplay` (stored as the mode name, default `Clip`). The mode is applied where the Text/Original cell templates are built, so it takes effect when the grid is rebuilt on Settings OK - the same path `GridCompactMode` / grid-lines use. It composes with the existing `SubtitleGridTextSingleLine` join setting.

## Notes

- Default `Clip` preserves current behaviour exactly; no migration needed (missing/old setting parses to `Clip`).
- English strings added in `LanguageSettings.cs`; other language JSONs fall back to those until translated.
- Builds clean (0 warnings / 0 errors).
- I don't have a Windows/macOS GUI here to judge the look, so I also produced portable Windows and macOS test builds from this branch so the feel of each mode (especially row growth in Wrap) can be confirmed. Happy to adjust defaults, naming, or where the setting lives.

### DIFF
--- a/src/ui/Features/Main/Layout/InitListViewAndEditBox.cs
+++ b/src/ui/Features/Main/Layout/InitListViewAndEditBox.cs
@@ -11,6 +11,7 @@ using Avalonia.Media;
 using Avalonia.Styling;
 using AvaloniaEdit;
 using Nikse.SubtitleEdit.Controls;
+using Nikse.SubtitleEdit.Features.Options.Settings;
 using Nikse.SubtitleEdit.Features.Shared.TextBoxUtils;
 using Nikse.SubtitleEdit.Logic;
 using Nikse.SubtitleEdit.Logic.Config;
@@ -118,6 +119,9 @@ public static partial class InitListViewAndEditBox
         var nullToOpacityConverter = new NullToOpacityConverter();
         var syntaxHighlightingConverter = new TextWithSubtitleSyntaxHighlightingConverter();
         vm.SubtitleDataGridSyntaxHighlighting = syntaxHighlightingConverter;
+        // How the Text/Original cells fit their text to the window (feature #11590). Read once here;
+        // the grid is rebuilt when settings are applied, so a changed mode takes effect then.
+        var gridTextDisplayMode = SubtitleGridTextDisplayModeDisplay.FromSettings();
         var gapConverter = new DoubleToDisplayShortConverter();
         var inverseBooleanConverter = new InverseBooleanConverter();
         var textOneLineShortConverter = new TextOneLineShortConverter();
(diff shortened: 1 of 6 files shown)


niksedk:
New language string will have to wait for 5.0.1

muaz978:

niksedk:
Please add screenshots - also how does this work with the current one-line-setting with string separator?

muaz978:
Screenshots coming shortly.

On how this composes with the existing "show subtitle text as single line" + separator setting: the two are independent layers and mostly compose cleanly.

`SubtitleGridTextSingleLine` / `SubtitleGridTextSingleLineSeparator` decide whether the break between a subtitle's line 1 and line 2 is rendered as a literal separator token (e.g. " ⏎ ") or a real line break, in `TextWithSubtitleSyntaxHighlightingConverter.AppendLineSeparator`.

The new display mode (Clip / Wrap / Ellipsis) is applied afterward, purely as `TextWrapping` / `MaxLines` / `TextTrimming` on the resulting text block, in `SubtitleGridTextDisplayModeDisplay.ApplyTo`. It has no knowledge of whether the break it is wrapping/trimming came from a real line break or a separator token.

One real edge case from that: with Ellipsis mode and the separator setting OFF (so subtitle lines keep a real line break), `MaxLines = 1` cuts at the first physical line, so a 2-line subtitle shows only its first line plus a trailing ellipsis, line 2 is fully hidden rather than just visually cut off. With the separator setting ON there is no real line break to cut at, so this does not happen; the whole joined string (with the separator token inline) is what gets truncated.

If you would rather Ellipsis always show something from every original line (e.g. truncate each line's tail instead of dropping line 2 outright), that is a small follow-up change I am happy to make, just say the word.

muaz978:
Please see the attached GIF.

muaz978:
Any thought on this dear, did you have the chance to check it out?

niksedk:
He, I could not see anything in the .gif file as the resolution was too low... I need to download the repro a test a bit...

muaz978:
Is this clearer?


https://github.com/user-attachments/assets/f0b6fcfe-5f3a-43a5-88ca-d4c97aeb926d

PR #11597, opened 2026-06-14
Add a couple of `Extend` shortcuts from SE4

OmrSi:
Add two useful Extend shortcuts that were in SE4:
1. Extend previous line's end to selected's start
2. Extend next line's start to selected's end


niksedk:
New language tags will have to wait for 5.0.1

PR #11598, opened 2026-06-14
Work on video player labels

OmrSi:
Having the timecode on the video and the video slider as just a short string of the video is really confusing; it's typical in places like these to have the full timecode, not just a part of it.

I can't intuitively understand what I'm seeing while I'm trying to read the timecode at some random point in time.

Also, even when showing the remaining time only, it's typical to show the full duration. This is the case in most video players.
Plus, it looks weird and empty having one timecode under the slider.

In addition, the tooltip of the volume showing only the word "Volume" seemed weird, so I added the level to that.

### DIFF
--- a/src/ui/Controls/VideoPlayer/VideoPlayerControl.cs
+++ b/src/ui/Controls/VideoPlayer/VideoPlayerControl.cs
@@ -802,6 +802,7 @@ namespace Nikse.SubtitleEdit.Controls.VideoPlayer
 
                 SetPositionDisplayOnly(pos);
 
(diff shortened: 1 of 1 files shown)


OmrSi:
One thing to add, we need a way to go to a video position and copy the current video position.
In SE4, this was possible from the `Video position` NumericUpDown, now I don't see a way to do it.

muaz978:
I will try to handle it today, hopefully, I am working on 3 issues, as soon as I finish them, I will start working on this.

Issue #11584, opened 2026-06-13
[Feature Request] Support for Remote llama.cpp Server via URL Endpoint


noname8556:
Hi Subtitle Edit Team,

Currently, the `llama.cpp` integration in Subtitle Edit appears to be limited to a locally managed instance controlled directly by the application. While having Subtitle Edit manage the local process is convenient for simple setups, it lacks the flexibility required for high-performance workflows. Allowing users to connect to an external, existing `llama.cpp` instance would enable GPU-accelerated translations, providing much higher speed.

I suggest adding an option to connect to a `llama.cpp` server via its API Endpoint (URL), similar to how the Ollama implementation works.

muaz978:
Took a look at this. Good news: the llama.cpp translate engine already POSTs to a configurable OpenAI-compatible endpoint internally - it's functionally identical to the LM Studio engine - it just wasn't exposed in the UI (it downloaded a model and managed a local server pointed at localhost).

**Workaround you can use right now:** select the **LM Studio (local LLM)** translate engine and set its URL to your remote llama.cpp server, e.g. `http://your-gpu-box:8080/v1/chat/completions`. It will translate against the remote GPU server immediately.

**Proper fix:** I opened SubtitleEdit/subtitleedit#11599, which adds a **"Use external server (URL)"** checkbox to the llama.cpp engine itself:
- off (default): unchanged - local model + managed server.
- on: shows an editable API URL field and posts directly to your remote server (no local download/management).

If you'd like to try the fix before review, test builds from that branch (pick llama.cpp, tick the box, enter your server URL):
- Windows (portable x64): https://github.com/muaz978/subtitleedit/releases/tag/win-test-20260614-063933
- macOS (Apple Silicon + Intel): https://github.com/muaz978/subtitleedit/releases/tag/macos-test-20260614-063856

These are unofficial test builds from a fork, for trying the feature only.

ladrex:
Thanks a lot! I just tried the LM Studio workaround and it works perfectly. I'm also looking forward to the official update once your fix is merged.

muaz978:
That's great to know.
It'd be great if you kindly tried the test build so we can confirm it working before adding it to the main build and before @niksedk merge it.

ladrex:
I tested the Windows test build in a virtual machine and can confirm that it works as expected.

muaz978:
Thank you so much, I really appreciate it.

@niksedk, rest is yours.

[...]

PR #11599, opened 2026-06-14
Add remote/external llama.cpp server (URL) option to translate (#11584)

muaz978:
Implements #11584 - lets the llama.cpp translate engine connect to an external/remote server via URL (e.g. a GPU box), instead of only managing a local instance.

## Background

`LlamaCppTranslate` and `LmStudioTranslate` are functionally identical - both POST to an OpenAI-compatible `/v1/chat/completions` endpoint read from a settings URL. The translate core for llama.cpp already targets a configurable `LlamaCppApiUrl`; the engine just never exposed it, instead downloading a model and managing a local server pointed at localhost.

## Change

Adds a **"Use external server (URL)"** checkbox to the llama.cpp engine in Auto-translate:

- **off (default)** - unchanged: local model download + SE-managed server.
- **on** - shows an editable API URL field; translation posts directly to that endpoint and skips the local download / `EnsureServerRunningAsync` path.

New setting `AutoTranslate.LlamaCppUseRemoteServer` (default `false`, so existing behaviour is untouched). The toggle swaps the panel live (a small guard prevents the programmatic re-setup from recursing). Mirrors how the Ollama/LM Studio engines present a URL field.

## Notes

- Builds clean (0 warnings / 0 errors).
- Until this is merged, users can already do this today by selecting the **LM Studio (local LLM)** engine and pointing its URL at the remote llama.cpp server - same endpoint shape. This PR just makes it discoverable under llama.cpp.
- I don't have a Windows/macOS GUI here, so I also produced portable Windows and macOS test builds from this branch to verify the toggle and a real remote translation. Happy to adjust naming/placement.

### DIFF
--- a/src/ui/Features/Translate/AutoTranslateViewModel.cs
+++ b/src/ui/Features/Translate/AutoTranslateViewModel.cs
@@ -92,6 +92,8 @@ public partial class AutoTranslateViewModel : ObservableObject
     [ObservableProperty] private LlamaCppModelDisplay? _selectedLlamaCppModel;
     [ObservableProperty] private bool _llamaCppModelComboIsVisible;
     [ObservableProperty] private bool _llamaCppButtonsAreVisible;
+    [ObservableProperty] private bool _llamaCppRemoteToggleIsVisible;
+    [ObservableProperty] private bool _llamaCppUseRemoteServer;
     [ObservableProperty] private string _llamaCppServerButtonText = "Start server";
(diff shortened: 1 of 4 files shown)


niksedk:
PRs with new language tags will have to wait for SE 5.0.1

muaz978:
It is completely up to you and how you see things best fit for the sake of development of the SE, please tell me if there is anything I can do more when needed.

PR #11608, opened 2026-06-14
Add a Source View button to the main toolbar

### LINKED ISSUE #11605: Source View has no toolbar button (toggle gone from SE4)
In SE4 there was a Source View toggle on the toolbar. In SE5 Source View still exists (it opens as a window via the command), but there is no toolbar button for it, so the quick toggle is gone.

**Request:** add a Source View button to the main toolbar (it can open the existing Source View window), with a Settings toggle to show/hide it like the other toolbar icons.

I have a PR for this.

muaz978:
Fixes #11605

Source View only existed as a command in SE5 (no toolbar entry), so the toggle button from SE4 was missing.

- Adds a toolbar button using the existing `SourceView` theme icon that runs `ShowSourceViewCommand`.
- Gated by a new `ToolbarShowSourceView` appearance flag (default on) with a matching "Show source view icon" checkbox in Settings > Appearance, like the other toolbar icons.

Builds clean (0/0). Test build provided so the icon/placement can be checked.

### DIFF
--- a/src/ui/Features/Main/Layout/InitToolbar.cs
+++ b/src/ui/Features/Main/Layout/InitToolbar.cs
@@ -278,8 +278,8 @@ public static class InitToolbar
                 Content = MakeImage("SourceView"),
                 Command = vm.ShowSourceViewCommand,
                 Background = Brushes.Transparent,
-                [AutomationProperties.NameProperty] = languageHints.SourceViewHint,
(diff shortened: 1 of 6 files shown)


niksedk:
No new language tags for now, so this must wait

niksedk:
Also, is the default value false?

muaz978:
Thanks - addressed both:

- Default is now `false` (toolbar unchanged unless the user opts in).
- Removed the new language tags; the button tooltip, accessibility name and the settings checkbox now reuse the existing `Options.Shortcuts.SourceView` ("Source view") string, so no new translation tags are added. Pushed to this branch.

PR #11610, opened 2026-06-14
Insert new line from the waveform selection; honor configurable duration

### LINKED ISSUE #11607: Insert new line forces a fixed ~2s duration instead of using the waveform selection
When inserting a new line it always creates a fixed ~2-second line, even if I've selected a range on the waveform with the mouse. The new line should span the mouse selection, and line creation should be decoupled from the playback cursor that moves while the video plays (the current coupling is confusing).

**Expected:** if a waveform selection exists, insert a line spanning exactly that selection; otherwise use a configurable default duration.

I have a PR for this.

muaz978:
Fixes #11607

Inserting a line always produced a fixed ~2s line even when a waveform range was selected, and a few `InsertService` branches hardcoded the duration.

- `InsertLineBefore`/`InsertLineAfter` now prefer the waveform's new-selection range when present, inserting a line spanning exactly that selection (decoupled from the moving playback cursor); with no selection they fall back to the normal grid insert.
- The remaining hardcoded `2000` ms in `InsertService` now uses the existing `General.NewEmptyDefaultMs` setting (Settings > "Default new subtitle duration (ms)").

Builds clean (0/0). Test build provided to confirm the feel.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -8535,6 +8535,14 @@ public partial class MainViewModel :
     [RelayCommand]
     private void InsertLineBefore()
     {
+        // Prefer the exact range the user dragged on the waveform, so the new line follows the
+        // mouse selection instead of a fixed default duration (and stays decoupled from the
+        // moving playback cursor). Falls back to the normal grid insert when there's no selection.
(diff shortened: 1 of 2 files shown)


niksedk:
Insert before/after should not use selection.

The second fix is ok.

PR #11625, opened 2026-06-14
Add go-to / copy current video position to the player slider

muaz978:
Adds go-to / copy of the current video position, addressing the follow-up on #11598 (@OmrSi: "a way to go to a video position and copy the current video position", which SE4 had on the Video position control).

Right-click the **video position slider**:
- **Go to video position** - opens the existing go-to dialog (previously only reachable via shortcut).
- **Copy** - copies the current position as a timecode, applying the configured video offset.

New `CopyVideoPosition` command; `VideoPlayerControl` raises `GoToPositionRequested`/`CopyPositionRequested` from a context menu on the slider, wired to the view-model in `InitVideoPlayer`. Reuses existing language strings (no new tags). Builds clean (0/0).

Couldn't GUI-test here; a Windows/macOS test build can be provided to confirm the menu and copy if useful.

### DIFF
--- a/src/ui/Controls/VideoPlayer/VideoPlayerControl.cs
+++ b/src/ui/Controls/VideoPlayer/VideoPlayerControl.cs
@@ -366,6 +366,15 @@ namespace Nikse.SubtitleEdit.Controls.VideoPlayer
                 Margin = new Thickness(2, 0, 0, 0),
                 [AutomationProperties.NameProperty] = Se.Language.General.VideoPosition,
             };
+
+            // Right-click the position slider to jump to / copy the current video position
+            // (SE4 parity for the old "Video position" control).
+            var goToPositionItem = new Avalonia.Controls.MenuItem { Header = Se.Language.Options.Shortcuts.GeneralGoToVideoPosition };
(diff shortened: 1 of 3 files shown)


OmrSi:
Hey, Muaz.
Maybe it would be better to add the `Go to video position` to the `Go to line number` window, allowing the user to have both in one location?

The `Go to line number` would have to be changed to just `Go to…` with two items in it, `Line Number` and `Video position`, and the name of the shortcut would be changed to `Go to line number/video position` maybe?

This is how I see it on most platforms, but it's up to @nikolajolsson to decide.

Another option is to add a NumericUpDown with the video position at the bottom in the toolbar, but maybe it wouldn't fit or would look too crowded.

muaz978:
Hey dear,

I can handle both, but let's see what @nikolajolsson say, and then I will act accordingly.

niksedk:
We can take a look after v5.0.0 :)

muaz978:
On your mark.

muaz978:

niksedk:
We already have a "Goto video position" shortcut

PR #11627, opened 2026-06-14
Smooth out waveform/video playback cursor (lag, stutter, seek delay)

shanytc:
## Problem

The waveform/video playback cursor (the vertical position line) had several issues during playback:

- Lagged ~100–200 ms behind playback.
- Froze and then jumped forward when resuming after a seek.
- Stuttered during steady playback (worse over loud audio).
- Clicking the waveform to seek took ~100–200 ms before the line moved to the new spot.
- Pressing play after placing the line caused the cursor to "rush" forward, then settle.
- In "center waveform on current position" mode the waveform scrolled in coarse 20 fps steps.

## Changes

**Interpolated playhead.** The cursor is advanced on a high-resolution wall clock at the current playback speed and reconciled with mpv's real position (snap on a large discontinuity, gently correct small drift, cap a single tick's advance). This hides mpv's chunky `time-pos` updates and its stall-then-jump when resuming after a paused seek.

**Dedicated cursor timer at the right priority.** The cursor is driven by its own ~60 fps `DispatcherTimer`. The mpv video control posts frame requests at `DispatcherPriority.Render`; a default (Background) timer was starved by them for 100–800 ms during decode bursts (the stutter), and even matching `Render` left a play-start rush, so the timer runs at `Normal` (above Render). The heavy per-tick work (subtitle buffer, centering, selection) stays on the existing 50 ms timer.

**Rendering caches.** The waveform geometry and per-paragraph shaped text are cached and replayed, so a cursor-only move no longer re-runs the per-pixel waveform loop or re-shapes `FormattedText` every frame (the latter caused GC pauses). Caches are keyed by view state and cleared in `ResetCache()`.

**Instant waveform seek.** A waveform click/drag pins the cursor to the clicked position immediately and holds it until mpv's reported position arrives, instead of lagging on the old spot while mpv applies the async pause+seek.

**Smooth center-mode scrolling.** In center mode the waveform scroll is driven by the 60 fps cursor timer in lockstep with the cursor instead of the 20 fps heavy timer.

## Notes

- All changes are in `MainViewModel` (timers/interpolation) and `AudioVisualizer` (render caches); no public API/behavior changes outside the waveform.
- The diagnosis was data-driven: temporary instrumentation confirmed the stutter was a `DispatcherPriority` inversion (mpv render posts starving the cursor timer), not GC or our rendering; that instrumentation has been removed.

### DIFF
--- a/src/ui/Controls/AudioVisualizerControl/AudioVisualizer.cs
+++ b/src/ui/Controls/AudioVisualizerControl/AudioVisualizer.cs
@@ -1958,13 +1958,38 @@ public class AudioVisualizer : Control
             waveformHeight = renderCtx.WaveformHeight;
         }
 
-        if (WaveformDrawStyle == WaveformDrawStyle.Classic)
+        // Rebuild the cached geometry only when the view actually changed; otherwise (e.g. the
+        // cursor moved) just replay the existing draw ops. See _waveformCacheDraws.
+        var key = BuildWaveformCacheKey(waveformHeight, ref renderCtx);
+        if (!_waveformCacheValid || !ReferenceEquals(_waveformCachePeaks, WavePeaks) || !key.Equals(_waveformCacheKey))
         {
-            DrawWaveFormClassic(context, waveformHeight, ref renderCtx);
+            _waveformCacheDraws.Clear();
+            if (WaveformDrawStyle == WaveformDrawStyle.Classic)
+            {
+                BuildWaveFormClassic(waveformHeight, ref renderCtx);
+            }
+            else
+            {
+                BuildWaveFormFancy(waveformHeight, ref renderCtx);
+            }
+
+            _waveformCacheKey = key;
+            _waveformCachePeaks = WavePeaks;
+            _waveformCacheValid = true;
         }
-        else
+
+        // The fancy style draws a center line; it depends only on width/height (already in the
+        // key) so it's cheap to draw live each render rather than caching it.
+        if (WaveformDrawStyle != WaveformDrawStyle.Classic)
         {
-            DrawWaveFormFancy(context, waveformHeight, ref renderCtx);
+            var halfWaveformHeight = waveformHeight / 2;
+            context.DrawLine(_centerLinePen, new Point(0, halfWaveformHeight), new Point(renderCtx.Width, halfWaveformHeight));
+        }
+
+        for (var i = 0; i < _waveformCacheDraws.Count; i++)
+        {
+            var draw = _waveformCacheDraws[i];
+            context.DrawGeometry(null, draw.Pen, draw.Geometry);
         }
     }
 
@@ -2025,7 +2050,7 @@ public class AudioVisualizer : Control
     private readonly Dictionary<int, FancyBatch> _fancyBatches = new(64);
     private readonly List<int> _fancyBatchKeysInUse = new(64);
 
-    private void DrawWaveFormFancy(DrawingContext context, double waveformHeight, ref RenderContext renderCtx)
+    private void BuildWaveFormFancy(double waveformHeight, ref RenderContext renderCtx)
     {
         _isSelectedHelper.Reset(AllSelectedParagraphs, renderCtx.SampleRate);
         var isSelectedHelper = _isSelectedHelper;
@@ -2042,9 +2067,6 @@ public class AudioVisualizer : Control
         var peaks = WavePeaks.AsSpan();
         var peaksCount = peaks.Length;
         var highestPeak = renderCtx.HighestPeak;
-
-        // Draw center line first
-        context.DrawLine(_centerLinePen, new Point(0, halfWaveformHeight), new Point(renderCtx.Width, halfWaveformHeight));
 
         // Calculate the threshold for color transitions (as a fraction of the highest peak)
         var lowThreshold = highestPeak * 0.3;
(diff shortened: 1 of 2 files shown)


niksedk:
Two things from reviewing this branch:

### 1. Playhead estimate isn't maintained in layouts without a waveform (functional regression)

In the cursor timer, `UpdatePlayheadEstimate` is only called inside `if (av != null)`:

```csharp
var av = AudioVisualizer;
if (av != null)
{
    var est = UpdatePlayheadEstimate(vp);   // only runs when the waveform exists
    av.CurrentVideoPositionSeconds = est;
    ...
}
```

`_playheadEstimateSeconds` is now the source of truth for the 50 ms `_positionTimer`:

```csharp
var mediaPlayerSeconds = _playheadEstimateSeconds;
```

But the position timer runs whenever `vp != null && _videoFileName` is set — it does **not** require `av`. Several layouts have a video player and no waveform (e.g. `MakeLayout8` in `InitLayout.cs` — video + list). In those, `AudioVisualizer` is `null`, so the estimate is never advanced (and `PinPlayheadTo` only fires from waveform events), leaving `_playheadEstimateSeconds` stuck at `0`. Everything downstream that reads `mediaPlayerSeconds` then breaks during playback: `SelectCurrentSubtitleWhilePlaying`, play-selection end detection, and the auto-scroll branches. The old code read `vp.Position` directly here, so it worked regardless of the waveform being present.

Suggested fix — always maintain the estimate; only the visual updates need `av`:

```csharp
var est = UpdatePlayheadEstimate(vp);
var av = AudioVisualizer;
if (av != null)
{
    av.CurrentVideoPositionSeconds = est;
    if (WaveformCenter && vp.IsPlaying && av.WavePeaks != null)
    {
        var halfSeconds = (av.EndPositionSeconds - av.StartPositionSeconds) / 2.0;
        av.StartPositionSeconds = Math.Max(0, est - halfSeconds);
    }
}
```

### 2. Stale comment

`AudioVisualizer.cs:2217` still says `// See DrawWaveFormFancy:` — that method was renamed to `BuildWaveFormFancy` in this PR.

Otherwise this looks great — nice diagnosis-driven comments, and the waveform geometry/text caching plus the cache-key invalidation all check out. SMPTE handling stays consistent too (the manual `1000/1001` in `UpdatePlayheadEstimate` reproduces what `vp.Position` previously carried).

shanytc:
> Two things from reviewing this branch:
> 
> ### 1. Playhead estimate isn't maintained in layouts without a waveform (functional regression)
> In the cursor timer, `UpdatePlayheadEstimate` is only called inside `if (av != null)`:
> 
> ```cs
> var av = AudioVisualizer;
> if (av != null)
> {
>     var est = UpdatePlayheadEstimate(vp);   // only runs when the waveform exists
>     av.CurrentVideoPositionSeconds = est;
>     ...
> }
> ```
> 
> `_playheadEstimateSeconds` is now the source of truth for the 50 ms `_positionTimer`:
> 
> ```cs
> var mediaPlayerSeconds = _playheadEstimateSeconds;
> ```
> 
> But the position timer runs whenever `vp != null && _videoFileName` is set — it does **not** require `av`. Several layouts have a video player and no waveform (e.g. `MakeLayout8` in `InitLayout.cs` — video + list). In those, `AudioVisualizer` is `null`, so the estimate is never advanced (and `PinPlayheadTo` only fires from waveform events), leaving `_playheadEstimateSeconds` stuck at `0`. Everything downstream that reads `mediaPlayerSeconds` then breaks during playback: `SelectCurrentSubtitleWhilePlaying`, play-selection end detection, and the auto-scroll branches. The old code read `vp.Position` directly here, so it worked regardless of the waveform being present.
> 
> Suggested fix — always maintain the estimate; only the visual updates need `av`:
> 
> ```cs
> var est = UpdatePlayheadEstimate(vp);
> var av = AudioVisualizer;
> if (av != null)
> {
>     av.CurrentVideoPositionSeconds = est;
>     if (WaveformCenter && vp.IsPlaying && av.WavePeaks != null)
>     {
>         var halfSeconds = (av.EndPositionSeconds - av.StartPositionSeconds) / 2.0;
>         av.StartPositionSeconds = Math.Max(0, est - halfSeconds);
>     }
> }
> ```
> 
> ### 2. Stale comment
> `AudioVisualizer.cs:2217` still says `// See DrawWaveFormFancy:` — that method was renamed to `BuildWaveFormFancy` in this PR.
> 
> Otherwise this looks great — nice diagnosis-driven comments, and the waveform geometry/text caching plus the cache-key invalidation all check out. SMPTE handling stays consistent too (the manual `1000/1001` in `UpdatePlayheadEstimate` reproduces what `vp.Position` previously carried).

Thanks for the review! Both addressed in latest commit:
1. `UpdatePlayheadEstimate()` is now always called in the cursor timer, only the visual updates are gated on the waveform existing, so video-only layouts keep _playheadEstimateSeconds current.
2. Fixed the `DrawWaveFormFancy` comment.
3. Also latest changes holds the cursor in place while mpv's clock is frozen after a seek (instead of racing ahead at 1×) and makes the drift correction forward-only, eliminating the "runs fast then slow" play-start behavior.

niksedk:
Also seems to work for vlc :)

shanytc:
> Also seems to work for vlc :)

thanks for merging :) i'd be happy to help here and there! i love subtitleedit! been working with it for years.

PR #11664, opened 2026-06-16
Make SubtitleLineViewModel.StartTime a plain setter; add StartTimeKeepDuration

niksedk:
## Background

`SubtitleLineViewModel.StartTime`'s setter had a hidden side effect: assigning it also moved `EndTime` to preserve the duration ("drag the whole line"). This made the most natural-looking statement — `line.StartTime = x` — the dangerous one:

- Programmatic code almost always wants the end to stay fixed, so the codebase had **~45** `SetStartTimeOnly(...)` calls to opt out of the drag.
- Forgetting `SetStartTimeOnly` and writing `StartTime = x` silently shifted the end too, corrupting timing.

This is the design follow-up to #11661 (which routed `Adjust` through `SetTimes`). It inverts the default so the safe behavior is the obvious one.

## Change

- **`StartTime` setter is now plain**: it moves the start, keeps the end fixed, and recomputes the duration — identical to the existing `SetStartTimeOnly`, which is kept as an alias so the ~45 existing callers are unchanged.
- **New `StartTimeKeepDuration` property** provides the old drag-the-line (preserve-duration) behavior, explicitly.
- **Grid binding** (`InitListViewAndEditBox`): the start-time editor binds to `StartTimeKeepDuration` when no separate end-time editor is shown (preserving that UX), and to `StartTimeOnly` when one is.

## Caller audit

I audited every direct `SubtitleLineViewModel.StartTime = …` assignment in `src/ui` (excluding `Paragraph` and other types). Only **five** relied on the drag and were converted to `StartTimeKeepDuration`:

- `AudioVisualizer.cs` — waveform Alt-drag (move whole line)
- `MainViewModel.cs` — insert subtitle at video position
- `MainViewModel.cs` — transcribe split into new lines
- `MainViewModel.cs` — ripple-delete trailing-line shift
- `PointSyncer.cs` — single-point sync

All other direct assignments either set `EndTime` immediately after or wanted keep-end semantics (one, snap-start-to-shot-change, is actually *fixed* by the new plain setter).

### Bonus fix

Converting single-point sync (`AdjustViaShowEarlierLater`) to `StartTimeKeepDuration` also fixes a real bug: it previously shifted only the **start** time of every line, leaving end times in place, which shrank/inverted durations. It now shifts the whole line by the offset.

## Tests

- `SubtitleLineViewModelTimeTests` — `StartTime` (plain, keep-end), `SetStartTimeOnly` (keep-end), and `StartTimeKeepDuration` (positive/negative shift, duration preserved).
- `PointSyncerTests` — single-point sync shifts both start and end, preserving durations.

Full UI suite: **354/354 passing**.

### DIFF
--- a/src/ui/Controls/AudioVisualizerControl/AudioVisualizer.cs
+++ b/src/ui/Controls/AudioVisualizerControl/AudioVisualizer.cs
@@ -573,7 +573,7 @@ public class AudioVisualizer : Control
             var seconds = RelativeXPositionToSeconds(point.X);
             if (firstSelected != null)
             {
-                firstSelected.StartTime = TimeSpan.FromSeconds(seconds);
+                firstSelected.StartTimeKeepDuration = TimeSpan.FromSeconds(seconds);
                 e.Handled = true;
                 InvalidateVisual();
                 return;
--- a/src/ui/Controls/AudioVisualizerControl/AudioVisualizer.cs
+++ b/src/ui/Controls/AudioVisualizerControl/AudioVisualizer.cs
@@ -573,7 +573,7 @@ public class AudioVisualizer : Control
             var seconds = RelativeXPositionToSeconds(point.X);
             if (firstSelected != null)
             {
-                firstSelected.StartTimeKeepDuration = TimeSpan.FromSeconds(seconds);
+                firstSelected.SetStartTimeKeepDuration(TimeSpan.FromSeconds(seconds));
                 e.Handled = true;
                 InvalidateVisual();
                 return;
--- a/src/ui/Features/Main/Layout/InitListViewAndEditBox.cs
+++ b/src/ui/Features/Main/Layout/InitListViewAndEditBox.cs
@@ -1098,9 +1098,12 @@ public static partial class InitListViewAndEditBox
             UseVideoOffset = true,
             [AutomationProperties.NameProperty] = Se.Language.General.StartTime,
         };
+        // With a separate end-time editor, moving start should keep the end fixed
+        // (StartTimeOnly). Without one, moving start drags the whole line keeping
+        // its duration (StartTimeKeepDuration).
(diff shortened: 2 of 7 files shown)


niksedk:
## Self-review follow-up

Ran a multi-angle review of this branch. It caught one **significant miss** in the original caller audit, now fixed in bd30cfee:

The audit grepped for `.StartTime = ` and so never matched the **compound-assignment** form `StartTime += …`. Six whole-line-shift sites use exactly that and relied on the old end-follows-start behavior — after the inversion they would have moved only the start and corrupted durations:

- `Adjust(...)` — all three modes (all / selected / selected-and-forward), i.e. the **"Show earlier/later" / "Adjust all times"** sync tool
- `WaveformSetStartAndOffsetTheRest` / `WaveformSetEndAndOffsetTheRest`
- `AudioVisualizerSetStartAndOffsetTheRest`

All converted to `StartTimeKeepDuration`.

Also simplified `StartTimeKeepDuration` to delegate to `SetTimes(value, value + (EndTime - StartTime))` — applies start/end atomically (no transient `start > end`) and computes the span from live times instead of the `Duration` field.

Other review findings reviewed and **not** changed (not regressions from this PR):
- Plain `StartTime` can now produce a negative duration if start is set past end — but this was already true via the pre-existing `StartTimeOnly` binding (end-editor mode) and `SetStartTimeOnly`; not introduced here.
- `_skipUpdate` early-return dropping a two-way write-back — refuted: the suppression windows are synchronous on the UI thread, so user input can't interleave; `StartTimeOnly` already carried the identical guard.
- Gap/neighbor brush staleness after a keep-duration drag — same as the previous drag setter (duration unchanged ⇒ no brush re-notify); gap coloring is refreshed by the slow timer.

Full UI suite: 354/354 green.

PR #11689, opened 2026-06-18
Add HTML index export to BinaryEdit and fix five export bugs

mjuhasz:
## Summary

  Ports the SE4 "Images with HTML index" export to the SE5 BinaryEdit window, then fixes five bugs found during testing.

  ### New feature

  - **File > Export > Images with HTML index** in the BinaryEdit window — saves each subtitle bitmap as a numbered PNG and generates `index.html` showing timecodes, images, and OCR'd text, mirroring the SE4 workflow for reviewing OCR results.

  ### Bug fixes

  - **Timestamp overflow** (`ExportHandlerHtmlIndex`): `TimeSpan.hh` silently drops the Days component, producing wrong timecodes past midnight (e.g. 25:00:00). Replaced with an explicit `TotalHours` computation.
  - **Tag encoding** (`ExportHandlerHtmlIndex`): the HTML encode/decode round-trip only preserved `<i>` tags; extended to also preserve `<b>` and `<u>`, which appear when the bold/underline toggles are used.
  - **Sentinel collision** (`ExportHandlerHtmlIndex`): placeholder strings `@1__`/`@2__` could corrupt subtitle text containing those literals. Replaced with control-character sentinels (`\x01`/`\x02`)
  that cannot appear in subtitle text.
  - **Uppercase tags** (`ExportHandlerHtmlIndex`): uppercase variants like `<I>`/`<B>`/`<U>` leaked through as literal text. Now calls `HtmlUtil.FixUpperTags` before sentinel replacement to normalize casing first. Also removed the private `FormatTime` helper in favour of the canonical `TimeCode.ToString()`.
  - **Newline encoding** (`ExportHandlerHtmlIndex`): `Environment.NewLine` is `\r\n` on Windows, so subtitles with Unix line endings (`\n`) were passed through unencoded. Fixed by replacing `\r\n` first, then `\n`, so both line-ending styles render as `<br />`.
  - **Missing OCR text in export** (`BinaryEditViewModel`, `OcrViewModel`): `EditExport` was passing the raw image subtitle path to BinaryEdit, which re-opened the file from disk and discarded all OCR results. Now passes `OcrSubtitleItems` directly. A separate `_sourceFileName` field tracks the origin path for export filename suggestions without triggering a file reload. `GetSuggestedExportFileName` now returns the stem only so the save dialog's `DefaultExtension` doesn't double-append the file extension.

### DIFF
--- a/src/libuilogic/Export/ExportHandlerHtmlIndex.cs
+++ b/src/libuilogic/Export/ExportHandlerHtmlIndex.cs
@@ -42,22 +42,38 @@ public class ExportHandlerHtmlIndex : IExportHandler
         var filePath = Path.Combine(_folderName, imageFileName);
         File.WriteAllBytes(filePath, param.Bitmap.ToPngArray());
 
-        var start = param.StartTime.ToString(@"hh\:mm\:ss\,fff");
-        var end = param.EndTime.ToString(@"hh\:mm\:ss\,fff");
+        var start = FormatTime(param.StartTime);
+        var end = FormatTime(param.EndTime);
         _html.AppendFormat("#{0}:{1}-->{2}<div style='text-align:center'><img src='{3}' />",
             _count, start, end, imageFileName);
 
         if (!string.IsNullOrEmpty(param.Text))
         {
-            var text = WebUtility.HtmlEncode(param.Text
-                    .Replace("<i>", "@1__").Replace("</i>", "@2__"))
-                .Replace("@1__", "<i>").Replace("@2__", "</i>")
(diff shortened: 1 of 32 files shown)


niksedk:
The new language tags are all English... e.g. for Danish we have:

````
"ImagesWithHtmlIndex": "Images with HTML index",
````

mjuhasz:
I removed the untranslated keys from all non-English language files. I was not aware that the C# property defaults in LanguageGeneral.cs serve as the English fallback when a key is missing from a translation file so the addition of those keys was not necessary.

PR #11690, opened 2026-06-18
fix: Tesseract OCR produces empty results for non-English languages

mjuhasz:
## Problem

  Selecting a non-English Tesseract language (e.g. Hungarian) produced all-empty subtitles with no error message. Three compounding bugs caused this:

  1. **Wrong argument order** — `--tessdata-dir` was placed *after* the `hocr` configfile argument. Tesseract treats everything after a configfile as another configfile name, so `--tessdata-dir` was silently ignored and the bundled tessdata directory was never used.

  2. **`hocr` configfile requires `configs/` subdirectory** — user-downloaded `.traineddata` packs don't include a `configs/hocr` file, so Tesseract failed to initialize the hOCR output mode. Replaced with inline `-c tessedit_create_hocr=1 -c tessedit_create_txt=0` flags which have no such requirement.

  3. **Tessdata path inconsistency** — `Se.TesseractModelFolder` probed system/Homebrew paths on macOS/Linux, resolving to a different directory than where SE5 downloads `.traineddata` files. Simplified to always use SE5's own `DataFolder/Tesseract550/tessdata` on all platforms.

  ## Additional hardening

  - Switched from hand-quoted `Arguments` string to `ArgumentList` to correctly handle paths containing spaces or quote characters
  - Kill the Tesseract child process on cancellation (`Process.Dispose()` does not terminate the process)
  - Drain stderr with `CancellationToken.None` and always await it before checking the exit code, preventing a pipe-buffer deadlock and an unobserved `ObjectDisposedException` on the success path
  - Fixed `ParseHOcr`: `Math.Min(-1, alternateLineStart)` discarded `ocr_header`-only hOCR output; guard now correctly handles a missing `ocr_line` span

  ## Files changed

  - `src/ui/Features/Ocr/Engines/TesseractOcr.cs` — argument order, `ArgumentList`, cancellation kill, stderr drain, `ParseHOcr` fix
  - `src/ui/Features/Ocr/OcrViewModel.cs` — pass `tessDataFolder` to `Ocr()`; simplify dictionary null-coalescing
  - `src/ui/Features/Tools/BatchConvert/BatchConverter.cs` — pass `Se.TesseractModelFolder` to updated `Ocr()` signature
  - `src/ui/Logic/Config/Se.cs` — simplify `ResolveTesseractModelFolder()` to always return SE5's bundled tessdata folder

### DIFF
--- a/src/ui/Features/Ocr/Engines/TesseractOcr.cs
+++ b/src/ui/Features/Ocr/Engines/TesseractOcr.cs
@@ -75,32 +75,42 @@ public class TesseractOcr
         {
             await File.WriteAllBytesAsync(tempImage, oneColorBitmap.ToPngArray(), cancellationToken);
 
-            // --tessdata-dir must come before any configfile argument (Tesseract requirement).
-            // Use -c inline variables instead of the "hocr" configfile so the tessdata directory
-            // is not required to contain a configs/ subdirectory (user-downloaded language packs
-            // typically don't include the bundled config files).
-            using var process = new Process
+            // Use -c inline variables instead of the "hocr" configfile — avoids requiring a
+            // configs/ subdirectory in the tessdata folder (user-downloaded packs don't include it).
+            var psi = new ProcessStartInfo
             {
-                StartInfo = new ProcessStartInfo
-                {
-                    FileName = _executablePath,
-                    Arguments = $"\"{tempImage}\" \"{tempTextFileName}\" --tessdata-dir \"{tessDataFolder}\" -l {language} --psm 6 --oem 3 -c tessedit_create_hocr=1 -c tessedit_create_txt=0",
-                    UseShellExecute = false,
-                    RedirectStandardOutput = true,
-                    RedirectStandardError = true,
-                    CreateNoWindow = true,
-                    WorkingDirectory = Se.TesseractFolder
-                },
+                FileName = _executablePath,
+                UseShellExecute = false,
+                RedirectStandardOutput = false, // output goes to temp .hocr file, not stdout
+                RedirectStandardError = true,
+                CreateNoWindow = true,
             };
+            psi.ArgumentList.Add(tempImage);
+            psi.ArgumentList.Add(tempTextFileName);
+            psi.ArgumentList.Add("--tessdata-dir");
+            psi.ArgumentList.Add(tessDataFolder);
+            psi.ArgumentList.Add("-l");
+            psi.ArgumentList.Add(language);
+            psi.ArgumentList.Add("--psm");
+            psi.ArgumentList.Add("6");
+            psi.ArgumentList.Add("--oem");
+            psi.ArgumentList.Add("3");
+            psi.ArgumentList.Add("-c");
+            psi.ArgumentList.Add("tessedit_create_hocr=1");
+            psi.ArgumentList.Add("-c");
+            psi.ArgumentList.Add("tessedit_create_txt=0");
 
 #pragma warning disable CA1416 // Validate platform compatibility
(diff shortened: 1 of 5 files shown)


mjuhasz:
Regarding issue nr. 3, the path inconsistency, I considered 3 options:

  A — Always use SE5's own folder (that's what this PR implements)
  DataFolder/Tesseract550/tessdata on all platforms, always. Downloads go to it, OCR reads from it. Consistent across launches regardless of what's installed on the system.
  Downside: Languages installed via brew install tesseract-lang or apt install tesseract-ocr-hun are invisible — user must re-download through SE5's dialog.

  ---
  B — System tessdata if present, SE5 folder as fallback; scan both
  Surfaces languages installed via brew/apt automatically without a separate download.
  Downside: The resolved path can change between launches if Tesseract is installed after the first run, causing the exact orphaned-file problem (I hit this bug.)

  ---
  C — Always download to SE5 folder, but also scan system tessdata
  Avoids orphaned downloads (SE5's folder is always the download target) while still showing system-installed languages in the picker.
  Downside: Adds scanning complexity for a marginal benefit — SE5 already has a built-in Download dialog.

In the end I decided to keep Subtitle Edit self-contained (Option A): users manage their language downloads from within the app, and we don't show duplicates when a language file is present on system and app level as well, no confusion, also less code, fewer bugs.

niksedk:
Looks good to me - does Linux use the installed Tesseract?

I hope to release the v5.0.0 in the weekend :)

mjuhasz:
> I hope to release the v5.0.0 in the weekend :)

Sounds good! I've mostly been making contained bug fixes lately, with no changes to language strings and limited UX flow adjustments in anticipation of that. I'll be on the road for a few days, so expect very limited activity. After that, I'll be back to help improve the app.

> does Linux use the installed Tesseract?

Yes, on Linux, SubtitleEdit always uses the system-installed Tesseract binary. Nothing is bundled.

I have not made any change related to that. Tesseract is called as an external process, and the binary is discovered at startup via hardcoded path probing, with a `PATH` fallback.

TesseractOcr.cs does the binary discovery, same as before. It probes these paths in order:

```bash
  /usr/bin/tesseract                               ← standard apt/dnf installs
  /usr/local/bin/tesseract
  /snap/bin/tesseract                              ← Snap package
  /home/linuxbrew/.linuxbrew/bin/tesseract
  ~/.local/bin/tesseract
  /opt/homebrew/bin/tesseract                      ← Linuxbrew
  /opt/tesseract/bin/tesseract
  /app/bin/tesseract                               ← Flatpak
```

  Falls back to `"tesseract"` (PATH lookup) if none of the above exist.

  The SeConv CLI  instead does a direct `PATH` environment variable scan.

  **Training Data** is the interesting part for this PR. While the binary must come from the system,  tessdata (`.traineddata` files) are downloaded separately by the app into `~/.config/Subtitle Edit/Tesseract550/tessdata/` (or a similar data path on macOS and Windows.) The `--tessdata-dir` flag is always passed explicitly, so **the app's language data is used regardless of where the system Tesseract was installed**.

I did not test on Linux but it's pretty much the same as macOS in this respect. Windows is what behaves differently, something like this:

  |                    | Linux                    | macOS                    | Windows                     |
  |--------------------|--------------------------|--------------------------|------------------------------|
  | Binary source      | System (required)        | System (required)        | Bundled OR system            |
  | Auto-download binary | No                     | No                       | Yes (via UI prompt)          |
  | tessdata           | User-downloaded to app data | Same                  | Same                         |
  | Error message      | `sudo apt install tesseract-ocr` | `brew install tesseract` | Offers download     |

PR #11714, opened 2026-06-20
Add auto-save of the open subtitle file (Options > General)

muaz978:
## What
Adds a real **Auto-save** that writes changes back to the currently open subtitle file, so the user no longer has to press Ctrl+S / use the toolbar.

## Why
`GeneralSettings.AutoSave` already existed but was dead code: read nowhere and with no UI. Users who expected auto-save (SE4 parity) found their file was never saved automatically.

## How
- New **Auto-save** checkbox in Options > General (next to Auto-backup).
- Debounced via the existing 400 ms slow timer: once the subtitle (main and/or original column) has been idle ~1.5 s with unsaved changes, it writes back to the already-open file.
- Auto-save never opens a Save-As dialog (skips untitled / converted / format-changed files) and never shows modal error dialogs on a timer - failures go to the status bar.
- The separate Auto-backup feature is untouched.

Note: developed on macOS; not yet runtime-tested by CI at time of opening.

### DIFF
--- a/src/ui/Features/Main/AutoSaveDebounce.cs
+++ b/src/ui/Features/Main/AutoSaveDebounce.cs
@@ -1,0 +1,61 @@
+using System;
+
+namespace Nikse.SubtitleEdit.Features.Main;
+
+/// <summary>
+/// Pure decision logic for the debounced auto-save in <see cref="MainViewModel"/>.
+/// Kept side-effect free so it can be unit tested without spinning up the view model.
+/// </summary>
+public static class AutoSaveDebounce
+{
+    public enum Action
+    {
+        /// <summary>Nothing to do this tick (no changes, or still waiting for the idle window).</summary>
+        Skip,
+
+        /// <summary>Content just changed - (re)start the idle window and remember the new content hash.</summary>
+        Arm,
+
+        /// <summary>Content has been idle long enough - write the file(s) now.</summary>
+        Save,
+    }
+
+    /// <summary>
+    /// Decides what auto-save should do this tick.
+    /// </summary>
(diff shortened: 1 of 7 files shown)


muaz978:
Tested on macOS 26.5 (Apple Silicon) from a local build of this branch: enabled Options > General > Auto-save, edited subtitle lines, and confirmed the open file is written to disk ~1.5 s after editing stops (no Ctrl+S), with the title-bar asterisk clearing and a 'Saved changes to …' status. Verified the edit landed on disk by reopening the file externally. Works as intended; Auto-backup behaviour is unaffected.

GrampaWildWilly:
Here's a random thought I'd like to throw out there.  Provide a saving interval instead of automatically triggering a save every time I stop typing.  This would make it more like the existing backup feature.  You might even be able to share code with that if you do this.

muaz978:
Auto back up and save are two different things, but you recommend adding time option?
I can add this if needed, but actually as I have been using this software for a couple of years now, I don't remember needing more than this time of save; it is almost instant, so you lose nothing no matter what happens.

niksedk:
Will look at this after SE 5.0.0 is out (no new language strings)

muaz978:
Shall I make it automatic wihtout a toggle option in th UI?

niksedk:
> Shall I make it automatic wihtout a toggle option in th UI?

No, I think it is good as is :)

muaz978:
Confirming this is merged and verified. Auto-save of the open subtitle file is available under Options > General, writing back to the current file as you edit. Cleaning up the now-merged working branch.

PR #11716, opened 2026-06-20
Reliably remember a subtitle's media across reopen (SE4 parity)

muaz978:
## What
Makes a subtitle reliably reopen with the video/audio it was last worked on, like SE4.

## Why
The media association lives only in the recent-files entry, and persistence was brittle:
- `VideoOpenFile` never wrote the association - it was saved only at the end of `SubtitleOpen` and on a clean `OnClosing`, so a force-quit (or app quit that doesn't raise the window-closing event) lost it.
- `AddToRecentFiles` overwrote `VideoFileName` with the currently-open media, so re-adding a subtitle with no video loaded (auto-open off, or video manually closed) erased the remembered media.

## How
- Persist the media association immediately when a local video/audio opens (guarded by a known subtitle name so opening media alone doesn't create a blank entry).
- When `AddToRecentFiles` is given no media, inherit the previously remembered video/audio (plus its track / offset / SMPTE flag) instead of clearing it. A real media path still overwrites.

Builds on the existing restore path (#11616).

Note: developed on macOS; not yet runtime-tested by CI at time of opening.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -16078,6 +16078,15 @@ public partial class MainViewModel :
             }
         }
 
+        // Persist the media association as soon as the file opens so reopening this subtitle
+        // reloads the same video/audio even if the app is force-quit before OnClosing runs.
+        // Guarded by a known subtitle name so opening media without a subtitle doesn't create
+        // a blank recent entry. AddToRecentFiles itself no-ops while still loading.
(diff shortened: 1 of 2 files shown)


muaz978:
Tested on macOS 26.5 (Apple Silicon) from a local build of this branch: opened a subtitle, loaded a video, quit Subtitle Edit, then reopened the same subtitle, the previously loaded media auto-loads again (SE4 parity), including after a non-clean shutdown. Confirmed working.

It is an important one, and auto backup is different than auto save.

PR #11724, opened 2026-06-21
Add MLX Whisper Mac speech-to-text engine (Apple GPU / Neural Engine)

muaz978:
## What

Adds a macOS speech-to-text engine, **MLX Whisper Mac**, that runs Whisper on the Apple Silicon **GPU / Neural Engine** via Apple's [`mlx-whisper`](https://github.com/ml-explore/mlx-examples/tree/main/whisper) package.

## Why

On macOS the existing CTranslate2 path is **CPU-only** (no Metal/CUDA on Apple Silicon), so large models run near real-time regardless of how powerful the Mac is. MLX uses the GPU, giving a large speedup at the **same accuracy** (same Whisper weights).

## How

- New engine **MLX Whisper Mac**, offered only on macOS **arm64** (where MLX runs).
- Driven through a bundled helper script (`mlx_whisper_transcribe.py`) run via `python3`, the same library-via-script pattern used elsewhere; it writes an SRT that Subtitle Edit loads.
- Models `tiny`..`large-v3` plus **`large-v3-turbo`**, mapped to `mlx-community` Hugging Face repos and downloaded automatically on first use. A full repo id can be passed as the model for custom MLX models.
- The helper decodes Subtitle Edit's 16 kHz mono WAV with the Python standard library, so it does **not** depend on a system `ffmpeg` binary (mlx-whisper's own loader shells out to `ffmpeg`, which may be missing/non-executable).
- Install detection via `python3 -c "import mlx_whisper"`; a pip-install hint is shown when missing (Subtitle Edit cannot download a pip package). One-time setup: `pip3 install mlx-whisper`.

## Notes

- Apple-Silicon only by design (MLX is arm64). Intel Macs are unaffected (engine not listed).
- Verified end-to-end on an Apple M5 (macOS 26) via a local build — see the test report in the comments.

### DIFF
--- a/src/libse/AudioToText/WhisperChoice.cs
+++ b/src/libse/AudioToText/WhisperChoice.cs
@@ -10,6 +10,7 @@
         public const string WhisperX = "WhisperX";
         public const string ConstMe = "Const-me";
         public const string CTranslate2 = "CTranslate2";
+        public const string MlxWhisperMac = "MLX Whisper Mac";
         public const string StableTs = "stable-ts";
         public const string PurfviewFasterWhisperXxl = "Purfview's Faster-Whisper-XXL";
         public const string ChatLlm = "Chat LLM";
--- a/src/ui/Assets/SpeechToText/MlxWhisperMac.txt
+++ b/src/ui/Assets/SpeechToText/MlxWhisperMac.txt
@@ -1,0 +1,22 @@
+MLX Whisper Mac runs Apple's "mlx-whisper" Python library via python3.
+
+Requirements
+- An Apple Silicon Mac (M1 or newer).
+- Install once:  pip3 install mlx-whisper
+
+Why use it
+MLX runs Whisper on the Apple GPU and Neural Engine via Apple's MLX framework,
+so on M-series Macs it is typically much faster than the CPU-only
+"Faster Whisper Mac" engine, at the same accuracy.
+
+Models
+tiny, base, small, medium, large-v2, large-v3, large-v3-turbo. These are
+MLX-format Whisper weights hosted on Hugging Face (the mlx-community org) and are
+downloaded automatically on first use.
+
+Tip: large-v3-turbo is multilingual and the best speed/quality pick.
+
+Advanced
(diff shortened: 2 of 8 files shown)


muaz978:
## Test report

**Hardware / OS:** Apple **M5**, macOS 26.5 (arm64). Built locally from this branch (ad-hoc signed).

**Method:** same audio transcribed through each engine with identical settings (input language English; default post-processing). Timings are wall-clock from the Speech-to-text window; the MLX time below is a warm run (model already downloaded).

### Speed + quality

| Engine | Backend | Model | Time (≈12 min source) | Accuracy / timecodes |
|---|---|---|---|---|
| **MLX Whisper Mac** | Apple GPU / Neural Engine | large-v3-turbo | **~1:49** | **best** — most accurate, cleanest timecodes |
| Whisper CPP | Metal (GPU) | large-v2 | ~1:49 | good |
| Faster Whisper (CTranslate2) | CPU only | large-v3 | **~12:00** | good |

**Takeaways**
- MLX matches Metal whisper.cpp on speed while giving **better accuracy and timecode segmentation**, and is **~6.5× faster** than the CPU-only CTranslate2 path on the same machine.
- Accuracy across MLX `large-v3` and CTranslate2 `large-v3` is equivalent (same weights) — the difference is purely GPU vs CPU, i.e. speed.

### Functional checks
- Engine appears only on Apple Silicon; missing-package case shows the `pip3 install mlx-whisper` hint correctly.
- Models `large-v3-turbo`, `large-v3`, … all selectable; auto-download from `mlx-community` works.
- **Audio decoding:** mlx-whisper's default loader calls the `ffmpeg` binary, which failed with `PermissionError: [Errno 13] Permission denied: 'ffmpeg'` in the host environment. Fixed by decoding Subtitle Edit's 16 kHz mono WAV directly in the helper (stdlib `wave` + numpy), so no external `ffmpeg` is needed. Confirmed transcribing end-to-end after the fix.

### Mixed-language note
Tested on English and Arabic (RTL) content; `large-v3-turbo` handled both well, with `large-v3` available as a higher-accuracy fallback for difficult audio.

Issue #11790, opened 2026-06-24
Questions about the bulk drag time offset feature in Subtitle Edit 5.0


wuwufei:
In Subtitle Edit 4.0, after selecting multiple subtitle lines, you can hold down the Ctrl key, click and drag subtitles in the waveform panel to shift their timestamps as a whole, which is extremely convenient. Is this feature not available in SE 5.0?

PR #11791, opened 2026-06-24
Waveform: drag a multi-selection to shift all timecodes together (#11790)

niksedk:
Implements the SE4 bulk time-offset gesture in the waveform, requested in #11790.

In SE4 you could select several lines and drag them in the waveform to shift all their timecodes together. SE5 only moved the single grabbed line, and a middle-of-paragraph move bailed out entirely when Ctrl was held (`AudioVisualizer.cs` `OnPointerPressed`) - because Ctrl is now used for toggle-select.

## Gesture

A **plain drag started on a line that is already part of a multi-selection** shifts the whole selection as a block. (Ctrl stays bound to toggle-select, so the trigger is a plain drag rather than Ctrl+drag.) Dragging a non-selected line, or a single-line selection, behaves exactly as before.

## Implementation (`AudioVisualizer.cs`)

- New `InteractionMode.MovingSelection`.
- On press, when the grabbed line is in `AllSelectedParagraphs` and more than one line is selected, snapshot every selected line's start + duration and enter the new mode.
- On move, shift every snapshotted line by one delta. The snap is anchored on the grabbed line so relative spacing is preserved, and the delta is clamped so the earliest selected line never moves before zero. Overlap with unselected neighbours is allowed (the block is being repositioned as a whole).
- On release, the follow-up tap is suppressed after a real drag so the multi-selection is not collapsed, and the snapshot is cleared.

Undo is captured by the existing change-detection mechanism (single-line waveform moves already rely on it), so a group move reverts as one step.

## Testing

- Compile-only build of `UI.csproj` succeeds with 0 warnings / 0 errors.
- Manual waveform interaction not yet exercised in a running build (the dev machine had SubtitleEdit open, locking the build output); worth a quick manual check of: drag a 3-line selection left/right, clamp at zero, single-line move unchanged, Ctrl-click still toggles selection, and undo reverts the whole shift.

### DIFF
--- a/src/ui/Controls/AudioVisualizerControl/AudioVisualizer.cs
+++ b/src/ui/Controls/AudioVisualizerControl/AudioVisualizer.cs
@@ -340,6 +340,11 @@ public class AudioVisualizer : Control
     private double _originalDurationSeconds;
     private double _originalPreviousEndSeconds;
     private double _originalNextStartSeconds;
+
+    // Snapshot of every selected paragraph's start/duration captured when a bulk
+    // (multi-selection) move starts, so the whole block can be shifted by one delta.
+    private readonly List<(SubtitleLineViewModel Paragraph, double StartSeconds, double DurationSeconds)> _selectionMoveSnapshot = new();
+
     private bool _preventNextTap;
     private long _audioVisualizerLastScroll;
     private SubtitleLineViewModel? _cachedHitParagraph;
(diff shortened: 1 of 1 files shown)


wuwufei:
Does this mean this feature is not available in the official 5.0 release and will be added in a future version?

niksedk:
> Does this mean this feature is not available in the official 5.0 release and will be added in a future version?

Yes, will be included in v5.1.0 :)
And some beta versions before that too.

Change log: https://github.com/SubtitleEdit/subtitleedit/blob/main/change-log.txt

Issue #11745, opened 2026-06-22
[Accessibility Request] Main menu is still inaccessible to screen readers after #11573 (Alt/F10 navigation and UIA exposure issue)


kylesskim-sys:
First of all, thank you for the accessibility improvements made in #11573. I tested the official SE 5.0 release after the merge, but unfortunately the main issue reported in #11553 still remains.

Current situation

When using the official SE v5.0.0 release with a screen reader on Windows 11, the main menu is still completely inaccessible.

The following top-level menus cannot be reached or read: File, Edit, Tools, Spell check, Video, Synchronization, Translate, Options, Help, Pressing Alt does not allow access to the menu.
F10 does not activate the menu bar.
Tab navigation cannot reach the menu.
The screen reader does not announce the menu items.
The behavior is unchanged compared to what was reported in #11553.

Expected behavior

Like SE4 and standard Windows applications:

Pressing Alt or F10 should move focus to the menu bar.
Arrow keys should allow navigation between top-level menus.
The screen reader should announce the menu names and their submenu items.
Menu controls should be exposed in the UI Automation tree as MenuBar and MenuItem controls.

Cause of the issue

Although #11573 improved accessibility for some controls, accessibility support for the main menu does not appear to have been added yet.

Could support also be added so that the main menu is exposed to UI Automation as MenuBar and MenuItem elements and can be fully accessed using only a screen reader and keyboard?

I would greatly appreciate it if accessibility support could be implemented so that visually impaired users can fully enjoy the many excellent features of SE. Thank you for all the hard work and dedication you have put into this project.

Related issue: #11553

Related PR: #11573

niksedk:
Thanks again for the detailed report and for pushing on this — it really helped.

A lot has changed since the **5.0.0** release you tested. The main-menu keyboard/screen-reader access has been reworked on the development branch, and it will ship in **v5.1.0-beta1, due out in a few days**. I tested the current build on Windows 11 by inspecting the live UI Automation tree (the same API NVDA/Narrator use) and confirmed the following now works:

**Main menu**
- **F10** moves keyboard focus into the menu bar (File).
- **Alt** focuses the menu, and **Alt+letter** mnemonics open a menu directly (Alt+F = File, Alt+E = Edit, Alt+T = Tools, Alt+S = Spell check, Alt+V = Video, Alt+O = Options, Alt+H = Help, …).
[...]

PR #11822, opened 2026-06-25
Let the focused main menu own its navigation keys (#11745)

niksedk:
Follow-up to the F10 menu-focus work (#11759) for #11745.

## Problem
F10 moves keyboard focus into the main menu, but the window's key handler is registered with `RoutingStrategies.Tunnel` (`MainView.cs`), so it — and the shortcut manager it calls — run **before** the focused menu item. Any arrow/Enter/Escape key that matches a configured shortcut was therefore consumed as a shortcut and never reached the menu, so keyboard menu navigation (notably **Left/Right between top-level menus**, the most-requested improvement in the issue thread) didn't actually work even after F10.

## Fix
Add an `IsMainMenuFocused()` guard in `OnKeyDownHandler`: when focus is on the `Menu` or a `MenuItem`, return early (without marking the event handled) so the key tunnels on to the focused menu and Avalonia's built-in menu navigation takes over. Mirrors the existing `IsTextInputFocused` early-out.

Drop-down items live in a popup (outside the menu's visual subtree), so a focused `MenuItem` is detected directly rather than via a visual-parent walk; the walk remains as a fallback for the top-level bar.

## Scope
Main window only — that's the only window with a menu bar. No behavior change when the menu isn't focused (shortcuts work exactly as before). Build passes.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -18012,6 +18012,50 @@ public partial class MainViewModel :
         return false;
     }
 
+    /// <summary>
+    /// True when keyboard focus is on the main menu itself or one of its items (including items in
+    /// an open drop-down, which live in a popup). Used to let the menu own arrow/Enter/Escape keys
+    /// instead of the shortcut manager while it is being navigated (#11745).
+    /// </summary>
+    private bool IsMainMenuFocused()
+    {
+        if (Menu == null)
+        {
+            return false;
+        }
+
+        var focusedElement = Window?.FocusManager?.GetFocusedElement();
+        if (focusedElement == null)
+        {
+            return false;
+        }
+
+        // A focused MenuItem belongs to a menu/flyout being navigated; let it keep the keys.
+        // (Drop-down items are hosted in a popup, so a visual-parent walk to Menu would miss them.)
+        if (focusedElement is MenuItem)
+        {
+            return true;
+        }
+
+        if (ReferenceEquals(focusedElement, Menu))
+        {
+            return true;
+        }
+
+        var parent = (focusedElement as Avalonia.Visual)?.GetVisualParent();
+        while (parent != null)
+        {
+            if (ReferenceEquals(parent, Menu))
+            {
+                return true;
+            }
+
+            parent = parent.GetVisualParent();
+        }
+
+        return false;
+    }
+
     internal void OnKeyDownHandler(object? sender, KeyEventArgs keyEventArgs)
     {
         lock (_onKeyDownHandlerLock)
@@ -18049,6 +18093,15 @@ public partial class MainViewModel :
             if (k == Key.F10 && keyEventArgs.KeyModifiers == KeyModifiers.None && TryFocusMainMenu())
             {
                 keyEventArgs.Handled = true;
+                return;
+            }
+
+            // When the main menu has keyboard focus (opened via F10 or Alt), let it handle its own
+            // arrow/Enter/Escape navigation instead of consuming those keys as shortcuts. The window
+            // key handler tunnels (runs before the focused menu item), so without this the shortcut
+            // manager would eat Left/Right etc. and menu navigation would never work (#11745).
+            if (IsMainMenuFocused())
+            {
                 return;
             }
 


GrampaWildWilly:
I probably should have mentioned this sooner.  Alt or F10 are supposed to be toggles.  Press once to activate the menu bar.  Press again to deactivate.  In the released SE the toggling doesn't happen.  Second press does nothing.  Even hitting Esc (several times) doesn't remove focus from the menu bar.  You have to Tab out of the menu bar to get focus back to somewhere else.  That is totally non-standard behavior for Windows.  Alt & F10 should be toggles, & Esc should have the effect of leaving the menu bar.  Esc does close cascaded submenus & it looks like it closes dropped down menus.  It even looks like it removes focus from the menu bar.  But try this.

Alt+f
DownArrow
Esc
Esc
DownArrow

Pathological.  Maybe your fixes here have taken care of this.  i can't tell.  Of course.  Duh.  But I am hoping.

GrampaWildWilly:
OK.  I've got SE 5 beta 1 now.  Left/RightArrow at long last navigate menus according to Windows standard.  But this is only on the menu bar itself.  You can now open a cascaded submenu with RightArrow & collapse it with LeftArrow.  But when you have a cascaded submenu open, you're supposed to be able to do RightArrow from there & the next primary menu on the menu bar should be displayed.  This is not happening.  With a menu open, Right/LeftArrow do navigate to adjacent menus, so long as you're not in a cascaded submenu.  So that detail still needs to be cleaned up.

Alt & F10 are still not toggles.  So my hope that you covered this was in vain.

My description of the pathological handling of Esc has also not been addressed.

Here's a bit of weirdness.  Suppose you drop down a menu, hit Esc to allegedly close it, then Alt+Tab to another task (for example, Firefox where you might be composing something on GitHub).  If you Alt+Tab back to SE, the initial letters on the menu bar are still underlined.  This underlining is supposed to toggle on only when you activate the menu bar.  Once you activate a menu selection, the underlining on the menu bar is supposed to toggle off, which it does.  But if you do my slightly pathological task switch, the underlines persist.  This tells me the state of the menu bar is half way between activated & idle.  An Alt+Tab task switch away from SE should completely deactivate the menu bar.  If a menu is dropped down but you haven't activated any selection, Alt+Tab to another task is supposed to clear the menu activation . . .  and it does.  But if you hit Esc before you Alt+Tab away, it looks like the menu bar is in this limbo state.  Also, if you have a menu open & you Alt+Tab away, it looks like the menu bar is deactivated.  But if you immediately do Alt+Tab away & then Alt+Tab back to SE, the underlined letters reappear.  You have to do 2 task switches away & back to SE to see this.  It looks like the Alt of Alt+Tab gets recognized by SE before you type Tab to switch away.  I think this has to do with recognizing keys on press down vs release.  You should probably defer recognition of the key press until key-up time.

Yeah, that's a bit of a long paragraph.  Can't be helped.  The issue of adherence to Windows standard is still not addressed.

GrampaWildWilly:
I have beta 2 now & there is movement.  But things are not yet right.

Look at how things work in Notepad++.  This is an app that adheres perfectly to Windows standard behavior.  In Notepad++ Alt & F10 are almost synonyms.  Both Alt & F10 activate the menu bar.  The only difference is you can combine Alt with keys, like Alt+f or Alt+v to drop down the menu with that letter underlined.  You can't combine any letter with F10.  F10 just activates the menu bar.  Alt gives you the added option of also opening a specific menu.

In Notepad++, pressing Alt by itself, even pressing & holding Alt, causes the letters on the menu bar to get underlined.  Releasing Alt leaves the letters underlined & further causes the first menu entry to get highlighted.  A second press of Alt deactivates the menu bar: the underlines go away & so does the highlight.  All of this happens without a menu actually opening.  That's what I've been meaning by toggle.  One press of Alt activates the menu bar; a second press of Alt deactivates the menu bar.  If you have a menu open, pressing Alt closes the menu AND deactivates the menu bar.  This is not how it works in SE.  In SE, press & hold Alt makes the underlines appear; release Alt makes them disappear.  And after you release Alt, the menu bar is in fact not activated.  This is not toggling.  SE is not yet working right with Alt.

Further.  In SE, if you have a menu open, pressing Alt does close it.  The underlines on the menu bar disappear.  But say you don't have any menus open & you Alt+Tab away from SE, then Alt+Tab back.  Suddenly the underlines on the menu bar reappear.  The menu bar is not activated, but the letters are underlined, which they should not be.

In Notepad++ pressing F10 activates the menu bar.  The letters get underlined & the first menu entry gets highlighted.  In addition, the menu bar is activated.  Arrow keys at this point will navigate the menus.  In SE, there appears to be an extra layer of processing.  F10 causes the underlines to appear, the first menu entry is highlighted, AND (this is the extra part that is unnecessary) you get a box around the first menu entry.  If you Alt+Tab away from SE, then come back, the underlines remain, the first menu entry is highlighted (without the box), but the menu bar is not activated.  The box is unique to F10.  There are no boxes with Alt.  But I say there should be no boxes.  Period.

In SE, if you activate the menu bar with F10 & start to arrow around in menus, another F10 correctly does 3 things: (1) closes the menu, (2) removes the underlines, (3) removes any highlight from the menu bar.  The menu bar does successfully get deactivated.  But Alt+Tab away & back makes the underlines reappear, although the menu bar is not activated.

In SE, if you start with either Alt or F10, then you open a menu, Esc once closes the menu.  Esc a second time makes the underlines go away & no menu bar entry is highlighted.  That means the menu bar is not activated, right?  Wrong!  Try any arrow key after this second Esc.  Oops, here comes menu navigation.  You have to hit Esc a third time to actually deactivate the menu bar properly.  No.  Hitting Esc twice should be all you need to do.

Try all of these experiments in Notepad++ to see how a Windows app is supposed to act.  Notepad++ adheres to Windows standards.  Or try all of this in Firefox.  Firefox also adheres to Windows standards.  SE does not.  I would be curious to hear from our visually impaired friends on the subject of screen reader compatibility in this beta.  My suspicion is that these standards violations that I'm pointing out are also a problem for their screen readers.  But I'd be glad to get proper feedback from them to see if my suspicion is confirmed.

GrampaWildWilly:
I've been looking at some other things & there seems to be a variety of approaches.  In Firefox, the menu bar shows underlined letters all the time, independent of whether you are actually navigating menus.  LibreOffice makes the underlining of letters appear & disappear.  The built in Windows 11 tool Resource Monitor makes the underlines appear & disappear.  The built in Windows 11 tool Computer Management puts a box around its menu entries.  But it doesn't take 3 hits on Esc to deactivate the menu bar; twice is enough.  Mind you, I hesitate to offer built in Windows tools as examples because they are so hit & miss.  For example, most Windows tools don't use your defined global system font.  They seem to be too superior for such plebeian constraints.  Despite these variations, I think my discussion in my preceding post is still valid.

GrampaWildWilly:
In RC 3 bare Alt is essentially non-functional.  If you press & hold Alt, the underlines appear in the menu bar.  When you release Alt, the underlines disappear & the menu bar is NOT activated.  If you just quickly click Alt, the underlines appear for a moment & then disappear.  The expected activation of the menu bar does not occur.

If instead you actually open a menu (Alt+letter), a subsequent bare Alt does toggle the menu closed & the menu bar does properly deactivate.

F10 looks like it properly toggles but the menu bar entry does not lose its highlight.  The menu bar is deactivated but the menu entry is still highlighted.

F10 puts a box around the menu bar entry.  To be consistent, Alt should do the same.  Or, you could go the other way & not bother with the box.  The box seems like an unnecessary frill with no added value.  I would vote for ditching the box.  It's probably extra code in SE to draw the box & I suspect there's quite enough code in there now without adding useless code.

When F10 leaves the erroneous highlight on a menu bar entry, Esc does not clear it.  Rather anomalously, Esc after Alt & after F10 causes the first entry on the menu bar to get highlighted, even if you had a different menu dropped down at the moment you hit Esc.  When that unexpected change of focus to the first entry on the menu bar occurs, the menu bar is still activated.  DownArrow does drop down the menu.  I suppose strictly speaking, that's how it's supposed to work.  A second Esc should be necessary at this point & that does deactivate the menu bar.  But the first Esc should leave the current menu bar entry highlighted, not always jump back to the first entry.  Plus the highlight should go away after the second Esc.

Alt+Tab away from SE & back still causes spurious underlining.

At least arrow navigation seems to be durably implemented.

@kylesskim-sys @bichitoxxx Are you guys still participating here?

PR #11825, opened 2026-06-25
Fix Ollama OCR runaway repetition / "hang" and add tools-log logging

niksedk:
## Problem
OCR via Ollama appeared to hang and produced garbage like:
```
- Henry, we're here. - We're here.
- Henry, we're here. - We're here.
...(x hundreds)
```
Vision models (glm-ocr here) loop on subtitle frames, and the request set **no generation limit**, so each line generated until the model's full context — ~50s of garbage per line. Across a subtitle that's the "hang". Reported on both macOS and Windows (it's in the shared request code).

## Test-run evidence (local `glm-ocr:latest`)
| Request | Time | Output |
|---|---|---|
| old (no options) | **53.8s** | 274 lines, looping |
| with `num_predict`+`repeat_penalty`+`temperature 0` | **6.6s** | bounded (~35 lines) |
| + `CollapseRepetition` post-process | — | **single correct line** |

## Changes
- **Bound generation** in the request: `temperature: 0`, `repeat_penalty: 1.3`, `num_predict: 500`. 8× faster on the looping case and no longer runs away.
- **`CollapseRepetition`** safety net: when the output is a short leading block repeated 3+ times, it's a generation loop → reduce to one block. Conservative (only triggers on ≥6 lines where the leading line recurs as a tight 1–4 line cycle), so genuine short OCR text is untouched. Unit-tested.
- **Tools-log instrumentation**: each call logs endpoint, model, image size, elapsed and status to the tools log (`Se.WriteToolsLog`) — previously this path logged *nothing* there, so a slow run looked hung with no diagnostics.
- **Configurable timeout** `OllamaOcrTimeoutMinutes` (default **5**) instead of the hardcoded **25** minutes, matching the llama.cpp engine.
- **Default URL** fixed to `/api/chat` (no trailing slash) and trimmed before POST — avoids a 307 redirect on every request.

Padding to square is kept (a separate test showed it *helps* avoid loops).

### DIFF
--- a/src/ui/Features/Ocr/Engines/OllamaOcr.cs
+++ b/src/ui/Features/Ocr/Engines/OllamaOcr.cs
@@ -16,11 +16,11 @@ namespace Nikse.SubtitleEdit.Features.Ocr.Engines;
 
 public class OllamaOcr
 {
-    // OCR of a single subtitle frame is short, so cap generation and discourage repetition.
-    // Vision models otherwise tend to loop ("- Henry, we're here." x100), which both produces
-    // garbage and makes each line take a very long time (#ollama-ocr-looping).
-    private const int MaxTokens = 500;
-    private const double RepeatPenalty = 1.3;
+    // glm-ocr and similar are "thinking" models that, left unchecked, emit reasoning, markdown
+    // fences and the same line over and over. We disable thinking, use a strict prompt, cap the
+    // tokens (a subtitle frame is short) and discourage repetition. Tuned against local glm-ocr.
+    private const int MaxTokens = 96;
+    private const double RepeatPenalty = 1.1;
 
     private readonly HttpClient _httpClient;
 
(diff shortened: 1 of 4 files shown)


niksedk:
Updated after testing against a local `glm-ocr` — the first attempt (token cap + repetition collapse) wasn't enough because the real cause isn't a simple loop: **glm-ocr is a "thinking" model** that appends reasoning ("Wait, let me look at the image again."), ```markdown fences, and an echo of the prompt after the answer.

New approach (empirically validated):
- Request `"think": false` (disables chain-of-thought — the main fix), a short strict prompt, `temperature 0`, `repeat_penalty 1.1`, `num_predict 96`.
- `CleanOcrText`: the transcription is always first, so keep leading lines and stop at the first duplicate / ``` fence / prompt-echo (max 4 lines).

Measured on local glm-ocr, repeated 3× each:
| input | before | after |
|---|---|---|
| 1-line crop | 50s+, hundreds of repeats | `- Henry, we're here. - We're here.` (~0.6s) |
| 2-line crop | loop + markdown + prompt echo | `Henry, are you there?` / `We're coming for you.` (~0.7s) |

Both correct and stable. Tests updated.

PR #11827, opened 2026-06-25
Fix Tesseract OCR blanking coloured subtitle text, and surface errors

niksedk:
Tesseract OCR failing produced **blank lines with no message**. `RunTesseractOcr` ignored `TesseractOcr.Error` and ran its work in a fire-and-forget `Task.Run` with no try/catch — so a failure (missing binary, bad tessdata/language, or any exception) silently filled the grid with empty lines, and a thrown exception even skipped `PauseOcr` (leaving it stuck).

## Changes
- **Show the actual Tesseract error** when it reports one, and wrap the run in `try/catch/finally` so exceptions are surfaced and `PauseOcr` always runs.
- **All-empty run hint**: when Tesseract exits cleanly but returns no text for every line, tell the user (likely coloured text or wrong language) instead of leaving blank lines.
- **Friendly "can't start" message** if the executable fails to launch (e.g. `brew install tesseract` on macOS).

## Notes on macOS
There's no Tesseract installer for macOS/Linux (only Windows has the in-app downloader) — they use the system install, which is probed across the usual paths and PATH. Verified locally: `tesseract 5.5.1` is installed and SE's exact OCR command produces correct hOCR, so the binary/data/parser are fine; this PR makes the *reason* for any blank output visible. A likely silent-blank cause for coloured subtitles is the blue-channel-keyed `NikseBitmap.MakeOneColor` preprocessing — to be addressed separately once confirmed.

### DIFF
--- a/src/libse/Common/NikseBitmap.cs
+++ b/src/libse/Common/NikseBitmap.cs
@@ -189,6 +189,26 @@ namespace Nikse.SubtitleEdit.Core.Common
             }
         }
 
+        /// <summary>
+        /// Binarizes the image for OCR: bright, sufficiently opaque pixels (the subtitle text)
+        /// become black, everything else becomes white. Keys on overall brightness (the max of the
+        /// R/G/B channels) rather than a single channel, so coloured text such as yellow or red is
+        /// kept instead of being dropped — unlike <see cref="MakeOneColor"/>, which only looks at
+        /// the blue channel and blanks low-blue text.
+        /// </summary>
+        public void MakeBlackAndWhiteForOcr(int brightnessThreshold = 90, int alphaThreshold = 100)
+        {
+            var black = new byte[] { 0, 0, 0, 255 };
+            var white = new byte[] { 255, 255, 255, 255 };
+            for (int i = 0; i < _bitmapData.Length; i += 4)
+            {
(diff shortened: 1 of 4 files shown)


niksedk:
## Root cause found and fixed

Added diagnostics (tools-log `ink=…%` + a saved copy of the exact image fed to Tesseract) which made the cause obvious: coloured subtitle lines logged **`ink=0.0%`** — the preprocessed image was completely white.

`NikseBitmap.MakeOneColor` keys only on the **blue channel** (`_bitmapData[i] > 20`). White/cyan text survives, but **yellow/red/orange text has blue ≈ 0**, so it was dropped entirely and Tesseract got an all-white image → blank lines. (Long-standing — the blue-only logic predates the Avalonia port.)

**Fix:** new `NikseBitmap.MakeBlackAndWhiteForOcr` binarizes on overall brightness (`max(R,G,B)`) instead of one channel, so coloured text is kept. Verified: SE's exact command + parser already work on white-on-black (ink 15%); a unit test proves white/yellow/red → black ink and dark/transparent bg → white.

This PR now covers both the **root-cause fix** and the earlier **error-surfacing + diagnostics**.

### Related (not in this PR)
The same blue-only `MakeOneColor` is used for line-split/alignment detection in `OcrViewModel` (3 calls) — same coloured-text weakness, lower impact. Can switch those to brightness too in a follow-up.

PR #11833, opened 2026-06-25
AudioVisualizer: two micro-opts in the per-pixel waveform loop

niksedk:
Two small, behaviour-preserving optimizations in the waveform per-pixel loop, which runs `Width` times in both `BuildWaveFormFancy` and `BuildWaveFormClassic` on **every** redraw (i.e. every cursor tick during playback).

### 1. Strength reduction — division → multiply
The sample position was computed per pixel as `(StartPositionSeconds + x / div) * SampleRate`, doing a `double` division each column. Precompute once:
```csharp
var startSample = StartPositionSeconds * SampleRate;
var samplesPerPixel = SampleRate / div;
...
var pos = startSample + x * samplesPerPixel;   // multiply instead of divide
```
Algebraically identical; `pos` is truncated to an `int` index, so the sub-ULP float difference is irrelevant.

### 2. Hoist loop-invariant `RenderContext` fields
`RenderContext` is a `ref struct`, so the JIT can't cache field reads across the loop. `Width` (loop condition) and `HighestPeak` / `VerticalZoomFactor` / `Height` (each passed to `CalculateYOptimized` twice per column) were re-read every iteration. Hoisted into locals. (The fancy loop already hoisted `HighestPeak` for thresholds but still re-read it in the call — now fixed too.)

No behaviour change; build is clean.

### DIFF
--- a/src/ui/Controls/AudioVisualizerControl/AudioVisualizer.cs
+++ b/src/ui/Controls/AudioVisualizerControl/AudioVisualizer.cs
@@ -2161,6 +2161,15 @@ public class AudioVisualizer : Control
         var peaksCount = peaks.Length;
         var highestPeak = renderCtx.HighestPeak;
 
+        // Hoist loop-invariant RenderContext fields into locals: it is a ref struct, so the JIT
+        // cannot cache these reads across the loop. Also turn the per-pixel sample position into a
+        // multiply (startSample + x * samplesPerPixel) instead of a division (x / div).
+        var width = renderCtx.Width;
+        var height = renderCtx.Height;
+        var verticalZoomFactor = renderCtx.VerticalZoomFactor;
+        var startSample = renderCtx.StartPositionSeconds * renderCtx.SampleRate;
+        var samplesPerPixel = renderCtx.SampleRate / div;
+
         // Calculate the threshold for color transitions (as a fraction of the highest peak)
         var lowThreshold = highestPeak * 0.3;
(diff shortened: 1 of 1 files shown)


niksedk:
## Benchmark (BenchmarkDotNet)

Measured the changed inner-loop arithmetic (sample-position calc + the two `CalculateYOptimized` calls + accumulation) — old vs new formulation, faithfully replicated (same `WavePeak` short Max/Min, same `RenderContext`-style ref struct, same `CalculateYOptimized`).

```
BenchmarkDotNet v0.14.0, macOS 26.5.1, Apple M4, .NET 10.0.7 Arm64 RyuJIT

| Method | Width | Mean      | StdDev    | Ratio | Allocated |
|------- |------ |----------:|----------:|------:|----------:|
| Old    | 1920  |  5.955 us | 0.017 us  |  1.00 |         - |
| New    | 1920  |  4.486 us | 0.011 us  |  0.75 |         - |
| Old    | 3840  | 11.976 us | 0.037 us  |  1.00 |         - |
| New    | 3840  |  9.015 us | 0.011 us  |  0.75 |         - |
```

**~25% faster** on the changed math, at both 1080p- and 4K-width waveforms, zero allocations either way (error bars <0.3%).

Scope note: this measures the per-pixel arithmetic the PR touches, not the whole `BuildWaveForm*` (which also does color/batch/geometry work that's unchanged), so the end-to-end render win is smaller — but it's a real, free reduction on a loop that runs on every redraw during playback.

PR #11893, opened 2026-06-26
Add two more custom video move forward/back jumps (custom 3 & 4)

niksedk:
## What

Extends the user-configurable **"move video position custom"** jumps from **2** to **4** forward and **4** back.

SE5 already supports moving the video forward/back by a selectable number of milliseconds (Options → Shortcuts → "Video, custom milliseconds … back/forward"), but only offered two slots. This adds custom **3** and **4** so users can map several different jump sizes (e.g. ±250 ms, ±2 s, ±10 s, ±30 s) to different shortcut keys.

- New settings `MoveVideoPositionCustom3/4 Back/Forward` (defaults: custom 3 = 10 s, custom 4 = 30 s).
- New commands `VideoMoveCustom3/4 Back/Forward` in the main view and the binary-edit window.
- Wired through shortcut registration, labels, the `PickMillisecondsWindow` picker in Options → Shortcuts, and `BinaryEditShortcuts`.

Mirrors the existing custom 1/2 pattern exactly; no default keys assigned (user-configurable like the originals).

## Test
- Build is clean.
- Options → Shortcuts now lists custom 1–4 back/forward; clicking each opens the ms picker and persists the value.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -11402,6 +11402,10 @@ public partial class MainViewModel :
             (nameof(VideoMoveCustom1ForwardCommand),VideoMoveCustom1ForwardCommand),
             (nameof(VideoMoveCustom2BackCommand),   VideoMoveCustom2BackCommand),
             (nameof(VideoMoveCustom2ForwardCommand),VideoMoveCustom2ForwardCommand),
+            (nameof(VideoMoveCustom3BackCommand),   VideoMoveCustom3BackCommand),
+            (nameof(VideoMoveCustom3ForwardCommand),VideoMoveCustom3ForwardCommand),
+            (nameof(VideoMoveCustom4BackCommand),   VideoMoveCustom4BackCommand),
+            (nameof(VideoMoveCustom4ForwardCommand),VideoMoveCustom4ForwardCommand),
             (nameof(PlayCommand),                   PlayCommand),
             (nameof(PauseCommand),                  PauseCommand),
             (nameof(TogglePlayPauseCommand),        TogglePlayPauseCommand),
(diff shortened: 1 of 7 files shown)


niksedk:
image

GrampaWildWilly:
Well . . .  It's not quite what I asked for but it may serve my purposes.  I hope others also benefit.  I still think you should not place a limit on the number of skip intervals a user can define.  Nevertheless, I am looking forward to beta 2.

GrampaWildWilly:
I have experimented with this in beta 4 & it seems to work just fine.  Warning.  If you are using an alternative keyboard (I am, for French), that can get turned on by accident.  Win+Spacebar is supposed to switch to your alternative keyboard.  But there's also a couple of other key combinations that unexpectedly activate the alternative keyboard.  I'm not entirely sure what they are but I think they involve the Alt & Arrow keys.  The shortcuts I defined for video skipping do involve various Alt & Arrow combinations, & I think I accidentally turned on my alternative keyboard while I was fooling around with this.  When the alternative keyboard is active, the Right Alt key goes dead.  I think it has a different meaning when the alternative keyboard is active, but I forget what that is.  In any case, this dead Right Alt interfered with my skip shortcuts . . . until I noticed my French keyboard was active.  Going back to my English keyboard, which reactivates Right Alt, restored correct functioning of my SE video skip shortcuts.

I've just found that Alt+Shift switches to the alternative keyboard.  Of course, that kills Right Alt, so you have to toggle back to your primary keyboard via LeftAlt+Shift.  Very tricky.  A Windows-only phenomenon.  And that explains what happened to me, because I do have some Shift+Alt combinations in my SE shortcuts, so I have to be careful I don't accidentally turn on my other keyboard.  Ah the joys of living in Windowsland.

PR #11900, opened 2026-06-27
Main + binary edit menu: Alt/F10 toggle and Escape deactivation (Windows standard)

niksedk:
Follow-up to #11822 for #11745, addressing the keyboard-menu feedback in [PR #11822](https://github.com/SubtitleEdit/subtitleedit/pull/11822#issuecomment-4813859033).

PR #11822 made arrow keys reach the focused menu (Left/Right between top menus, open/collapse submenus). This adds the activation/deactivation half of the Windows menu keyboard interface, to **both windows that have an in-window menu bar — the main window and the binary edit window** (both previously had the same minimal F10-focus code):

- **F10 toggles** the menu bar — first press activates (focus moves in), second press deactivates and restores the previously focused control.
- **Bare Alt** press+release toggles it the same way. The toggle is armed on Alt-down and cancelled if any other key is pressed while Alt is held (so `Alt+F` etc. still works as an access key); the toggle fires on Alt-up.
- **Escape**, once no drop-down is open, fully deactivates the bar and restores focus to the editing control instead of leaving it half-focused with the access-key cues still showing (the "pathological Esc" the reporter described). In the binary edit window this no longer closes the whole window when Esc is used to leave the menu.
- A window **task switch (Alt+Tab)** clears the pending Alt toggle (`Window.Deactivated`) so returning to the window doesn't spuriously open the menu.

### Scope / platforms
- Windows (and the in-window Linux menu). **No effect on macOS**, which uses the native menu bar — the in-window `Menu` has no items / is hidden there, so activation is a no-op and the Option key is left untouched.
- Still **not** covered here (separate framework-level items from the same comment): right-arrow from a deep submenu wrapping to the next top menu (Avalonia `Menu` navigation), and the mnemonic-underline "limbo" after Esc→Alt+Tab→back (Avalonia `AccessKeyHandler` state). The `Window.Deactivated` hook added here is a step toward the latter.

Build passes; app launches cleanly. (Menu-toggle behavior itself is only observable on Windows/Linux.)

### DIFF
--- a/src/ui/Features/Main/MainView.cs
+++ b/src/ui/Features/Main/MainView.cs
@@ -42,6 +42,7 @@ public class MainView : ViewBase
             }
 
             _vm.Window.Closing += _vm.OnClosing;
+            _vm.Window.Deactivated += _vm.OnWindowDeactivated;
             _vm.Window.Loaded += (_, _) =>
             {
                 _vm.OnLoaded();
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -292,6 +292,8 @@ public partial class MainViewModel :
     AudioVisualizerUndockedViewModel? _audioVisualizerUndockedViewModel;
     FindViewModel? _findViewModel;
     Control? _findPreviousFocus;
+    Control? _focusBeforeMainMenu;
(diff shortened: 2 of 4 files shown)


GrampaWildWilly:
😀

PR #11912, opened 2026-06-27
Fix UITests build: expose libuilogic internals

ivandrofly:
## Summary
- `OcrFixEngine.SplitLine` (an `internal` member) was moved from the `ui` project into the new `libuilogic` project, but `InternalsVisibleTo("UITests")` existed only in `UI.csproj`. This broke the `UITests` build with `CS0117: 'OcrFixEngine' does not contain a definition for 'SplitLine'`.
- Added `src/libuilogic/Properties/AssemblyInfo.cs` with `[assembly: InternalsVisibleTo("UITests")]`. A code file is used rather than the MSBuild `<AssemblyAttribute>` item because `LibUiLogic.csproj` sets `GenerateAssemblyInfo=false`, which would skip a generated attribute.

## Test plan
- [x] `dotnet build tests/UI/UITests.csproj` succeeds (previously failed with CS0117).
- [x] `dotnet test tests/UI/UITests.csproj --filter "ClassName~OcrFixEngineSplitLineTests"` runs the `SplitLine` tests.

### DIFF
--- a/src/libuilogic/Properties/AssemblyInfo.cs
+++ b/src/libuilogic/Properties/AssemblyInfo.cs
@@ -1,0 +1,3 @@
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("UITests")]
--- a/tests/UI/Features/SpellCheck/UserDictionaryNormalizationTests.cs
+++ b/tests/UI/Features/SpellCheck/UserDictionaryNormalizationTests.cs
@@ -17,6 +17,7 @@ public class UserDictionaryNormalizationTests : IDisposable
     private readonly string _nfdWord = HangulWord.Normalize(NormalizationForm.FormD);
 
     private readonly string _originalDictionariesFolder;
+    private readonly Func<string> _originalSpellCheckDictionariesFolder;
     private readonly string _tempDictionariesFolder;
 
     public UserDictionaryNormalizationTests()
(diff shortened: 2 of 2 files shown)


ivandrofly:
### Follow-up: fixed a latent test failure surfaced by this change

The `InternalsVisibleTo` fix made `UITests` build and run for the first time, which exposed a pre-existing failure in `UserDictionaryNormalizationTests.AddToUserDictionary_ShouldTreatVisuallyIdenticalHangulAsSameWordAcrossNormalizationForms` (it had never actually executed before because the project didn't compile).

**Root cause:** The test redirected `Se.DictionariesFolder` to an isolated temp dir, but `SpellCheckWordLists` doesn't read that — it reads its folder from `SpellCheckConfig.DictionariesFolder`, a separate `Func<string>` that `Program.cs` wires to `Se.DictionariesFolder` at app startup. Tests don't run that startup wiring, so `SpellCheckConfig.DictionariesFolder` stayed at its default (empty string). As a result `SpellCheckWordLists` looked in the wrong (empty) folder and `HasUserWord(_nfcWord)` returned `false`.

**Fix:** In the test setup, also point `SpellCheckConfig.DictionariesFolder` at the same temp dir and restore the original value on dispose. No production code changed.

Full `UITests` suite now passes locally (466/466).

Issue #11602, opened 2026-06-14
[Feature Request] Suggestions for additional features after testing version 5.0.0 rc4


lsmxpftz-cmyk:
Thank you very much for creating such a powerful subtitle editing software. It has greatly improved my workflow and made the process of creating and editing subtitles much faster and more efficient.

After trying the software, I feel that it already works very well, but I also think there are a few additional features that could make it even better and more convenient for users. Therefore, I would like to share some suggestions based on my experience. I hope these ideas can be considered in future updates.

### **Feature Suggestions**

> ### **1.Synchronization>Adiust all times can add Extend the end point while keeping the start point unchanged/Shift the start point earlier while keeping the end point unchanged.**

<img width="1774" height="834" alt="Image" src="https://github.com/user-attachments/assets/23263ac0-ce7e-4164-a06a-e3cb6fc6bd79" />
（I think the existing option can remain unchanged, while this function could be added as a separate new option.
）

**⬇️The final effect is shown below.**

<img width="779" height="248" alt="Image" src="https://github.com/user-attachments/assets/42f237f4-fe62-4e5b-bf1d-4107e04dbbee" />

（Extend the end point while keeping the start point unchanged.）

<img width="706" height="245" alt="Image" src="https://github.com/user-attachments/assets/1ef235e1-4e3d-45dc-8d4e-0d5b669ef4ff" />

（Shift the start point earlier while keeping the end point unchanged.）

> ### **2.Settings>General>Subetitle grid single-click action/ Subetitle grid double-click action can add Go to subtitle and play and focus text box**

> ### **3.Subetitle grid Triple-click the subtitle and focus the text box.**

This action will not affect timeline playback. It does not need to jump back to the selected subtitle, allowing users to edit the text box while the video continues playing.

> ### **4.Add batch extraction features for audio and video, and allow users to customize the naming format of the extracted files.**

This feature can batch extract corresponding audio or video clips according to the subtitle timecodes. Users should be able to customize the file naming format, for example by adding sequence numbers or using the subtitle text as the file name. 

**Please refer to the interface below.⬇️**

<img width="812" height="631" alt="Image" src="https://github.com/user-attachments/assets/96060a4b-1795-4f57-b033-d240359acd0c" />
<img width="554" height="555" alt="Image" src="https://github.com/user-attachments/assets/e0981d44-b9cf-44dd-ae5f-5902fc5b6fcd" />

**【After export】**

[...]

PR #11982, opened 2026-06-29
Add "Save as" to Modify selection: export the matched lines (#11602)

muaz978:
Part of #11602 (suggestion 6).

Modify selection could only apply the matched lines to the grid selection. This adds a **Save file as** button that exports just the currently matched/checked lines to a new subtitle file (SubRip), without changing the grid selection or closing the dialog, so a subset of a file can be saved directly.

Injects IFileHelper (the view-model is DI-registered) and reuses the standard save-file picker.


niksedk:
We have "Save as..." for selected lines in main view

muaz978:
Understood, thanks for checking. Since the main view already covers "Save as..." for the selected lines, this one is redundant. Happy to leave it closed.

PR #11984, opened 2026-06-29
Subtitle grid: triple-click a row to focus the text box without seeking (#11602)

muaz978:
## Subtitle grid: triple-click to focus the text box without seeking

Implements suggestion #3 from #11602.

### Problem
There was no quick way to jump from a grid row straight into editing its text without also moving the video position. Double-click runs the configured action (which seeks), and a third click just re-fired the single-click action.

### Change
A triple-click on a grid row now moves keyboard focus to the edit text box and leaves the video position untouched.

### Implementation notes
- The third click is detected in `SubtitleGrid_PointerPressed` via `PointerPressedEventArgs.ClickCount`. When count reaches 3 on a valid row with the left button, it cancels the pending single-click timer and marks the upcoming tap as focus-only, then selects the row.
- Both the single-tap and double-tap handlers honour that marker (the third click can surface as either depending on the platform's gesture counting), focusing the edit text box instead of running the seek action.
- The marker is cleared at the start of every fresh click sequence (click count 1), so a missed third tap cannot leave it stuck.
- No seeking, no playback change, and the existing single/double-click actions are unchanged.

### Files
- `src/ui/Features/Main/MainViewModel.cs`

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -19169,6 +19169,13 @@ public partial class MainViewModel :
     public void SubtitleGrid_PointerPressed(object? sender, PointerPressedEventArgs e)
     {
         StopSubtitleGridDragSelectAutoScroll();
+        if (e.ClickCount <= 1)
+        {
+            // Fresh click sequence: clear any triple-click marker that a previous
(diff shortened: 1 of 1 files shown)


muaz978:

niksedk:
Sorry, I'm not too sure about this... will close this for now.

PR #12009, opened 2026-06-30
Fix umlaut diacritics clipped in OCR window on macOS (#11997)

### LINKED ISSUE #11997: Textboxes too small of some Languages
Noticed with German, the Textbox won't correctly show characters such as ÖÄÜ, as the dots above the character are cut off.

Image
The OCR Dialog: Within the list it is cut off, within the "Text" Textbox it is cut off, but the fixes and if I copy the character show it is actually the correct character. 

This behaviour is shown throughout the Application on any Textbox.

Running v5.1.0-beta2 on Mac OS X arm (really love having a Mac version, thank you!!!)

niksedk:
## Problem
On macOS the dots of tall diacritics (Ä/Ö/Ü) are clipped at the *top* in the OCR window — both the subtitle list (grid) and the "Text" edit box. The text data is fine (copy works); it's a pure render clip.

## Root cause
The macOS default UI font (San Francisco) has an ascent metric that sits right at the cap height, so Avalonia's line box clips the diacritic dots. (A font like Arial doesn't clip — so it's partly font-metric driven, but the robust fix is font-independent.)

Verified by rendering real CoreText to a bitmap on macOS:
- `TextBlock` default / `new FontFamily("Default")` → **clipped**
- `TextBox` default → **clipped**; **padding does NOT fix a TextBox** (only a TextBlock)
- **`LineHeight` (extra leading) fixes both** `TextBox` and `TextBlock`

## Fix
New `UiUtil.FixMacDiacriticClipping(control)` — **macOS-only**; binds `LineHeight = FontSize × 1.4` to the live font size (so it scales with the chosen subtitle font, and leaves Windows/Linux line spacing untouched). Applied to:
- the OCR grid's text `TextBlock` (incl. the OCR-fixed-line path), and
- the OCR "Text" edit box.

Confirmed visually fixed on macOS with the reporter's German subtitle.

## Follow-up
The reporter sees this app-wide; the same helper can be applied to the main grid + edit box in a follow-up.

Fixes #11997 (OCR window; main window to follow)

### DIFF
--- a/src/ui/Features/Ocr/OcrSubtitleItem.cs
+++ b/src/ui/Features/Ocr/OcrSubtitleItem.cs
@@ -117,7 +117,9 @@ public partial class OcrSubtitleItem : ObservableObject
     {
         if (FixResult != null)
         {
-            return FixResult.GetFormattedText();
+            var formatted = FixResult.GetFormattedText();
+            UiUtil.FixMacDiacriticClipping(formatted);
+            return formatted;
         }
 
         var tb = new TextBlock { Text = Text };
@@ -125,6 +127,8 @@ public partial class OcrSubtitleItem : ObservableObject
         {
             tb.FontFamily = new FontFamily(Se.Settings.Appearance.SubtitleTextBoxAndGridFontName);
         }
+
+        UiUtil.FixMacDiacriticClipping(tb);
         return tb;
     }
 
--- a/src/ui/Features/Ocr/OcrWindow.cs
+++ b/src/ui/Features/Ocr/OcrWindow.cs
@@ -643,6 +643,7 @@ public class OcrWindow : Window
         textBoxText.Bind(FontFamilyProperty, new Binding(nameof(vm.TextBoxFontFamily)) { Mode = BindingMode.TwoWay });
         textBoxText.Bind(FontSizeProperty, new Binding(nameof(vm.TextBoxFontSize)) { Mode = BindingMode.TwoWay });
         textBoxText.Bind(TextBox.FontWeightProperty, new Binding(nameof(vm.TextBoxFontWeight)) { Mode = BindingMode.TwoWay });
+        UiUtil.FixMacDiacriticClipping(textBoxText);
 
         // Create a Flyout for the TextBox
         var flyout = new MenuFlyout();
--- a/src/ui/Logic/UiUtil.cs
+++ b/src/ui/Logic/UiUtil.cs
@@ -90,6 +90,31 @@ public static class UiUtil
         };
     }
 
+    // On macOS the default UI font's ascent sits right at the cap height, so Avalonia's line box
+    // clips the dots on tall diacritics (Ä/Ö/Ü) at the top - in text boxes and grid cells alike
+    // (issue #11997). Giving the line a bit of extra leading (LineHeight) lifts the line box so the
+    // diacritics fit; padding does not help a TextBox, but LineHeight fixes both TextBox and TextBlock.
+    // Bound to the live FontSize (rather than a fixed value) so it scales with the chosen font size,
+    // and applied only on macOS so Windows/Linux line spacing is unchanged.
+    private static readonly IValueConverter DiacriticLineHeightConverter =
+        new FuncValueConverter<double, double>(fontSize =>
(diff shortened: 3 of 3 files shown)


mjuhasz:
@niksedk Leaving a related note here, in case someone reports it.

Avalonia has another bug related to macOS San Francisco (the default UI) font handling: the caret positioning is slightly off for some characters. I made attempts to fix this in the past but the bug is in Avalonia, and has been for nearly 3 years, and I'd have to fork Avalonia to fix it (some people did...). I'll add more details below, but the tl;dr version is: it's not that much noticeable and the workaround is easy: pick a different font. For the ocr edit box we have a dedicated font setting (right click -> set font), for subtitle edit boxes we also have a dedicated font setting, and the default macOS font is sans so not a great choice for eyeballing subtitle text anyway, it's for the UI. Earlier I sent you several PRs (which you accepted) to use the edit box font in all places where legibility (e.g. lookalike characters) really matters so having the caret slightly off for an ellipsis (…) in the Find dialog or Replace dialog is hardly a big deal. It looks like this:

Screenshot 2026-06-30 at 11 17 28 PM

Screenshot 2026-06-30 at 11 17 18 PM

What could be improved is that if the ocr edit box font is not set it falls back to the default UI font, that's why this user reported the issue. If we face other issues related to the system font we should consider changing the default to "Helvetica Neue".

  What's causing it

  It's the font — specifically the macOS default UI font, San Francisco, which macOS exposes only as the private family .AppleSystemUIFont / .SF NS. On macOS the app deliberately leaves the UI font as the OS default; Program.cs:89-125 only force-registers Inter on Linux (to dodge the $Default glyph-typeface crash) and intentionally lets macOS/Windows use the system font so CJK fallback keeps working.

  The mechanism behind the misplaced caret/selection edge:

  - Avalonia positions the caret and the end-of-selection at a glyph's advance width — the pen-advance after the glyph — computed by its Skia + HarfBuzz text pipeline (GlyphRun.GetDistanceFromCharacterHit, driven by per-glyph advances).
  - It does not account for glyphs whose ink (the painted outline) overruns the advance box — i.e. glyphs with trailing overhang / negative right side-bearing. San Francisco has several such glyphs, the ellipsis … being a clear one: the dots are painted slightly past the advance width. Native macOS apps place the caret correctly because Core Text carries the full system-font metrics/optical tracking; Avalonia sees only the advance, so the caret lands at the advance boundary — visually inside / before the right edge of the rendered glyph.
  
  This is an Avalonia limitation (advance-based caret positioning vs. ink overflow), not a Subtitle Edit bug, and it's inherently font-specific — which is exactly why other fonts I tried don't show it. The same class of bug is reported against other fonts in Avalonia: Inter at small sizes (#12809 (https://github.com/AvaloniaUI/Avalonia/issues/12809)) and Consolas/UniversNextPro (#15768 (https://github.com/AvaloniaUI/Avalonia/issues/15768)). It affects any text control using that font (TextBox, AutoCompleteBox in the Find dialog, etc.), not the Find box specifically.
  
  Workarounds / fixes
  
  1. No-code, available today: Settings → Appearance → Font, pick an explicit well-behaved font (Helvetica Neue, Arial, or the bundled Inter). UiUtil.SetFontName (UiUtil.cs:2528) adds app-wide FontFamily styles. Note its OfType<TextBox>() style also reaches the inner editable TextBox inside the Find dialog's AutoCompleteBox, so the Find box is covered too. This is the quickest confirmation that the font is the cause.
  
  2. Best app-level fix — give macOS a real default font instead of the hidden system font, while keeping CJK fallback. Mirror the Linux branch in Program.cs for macOS using  FontManagerOptions.DefaultFamilyName + FontFallbacks, e.g. default to "Helvetica Neue" (always present on macOS, no ink overhang) with "PingFang SC", "Hiragino Sans", "Apple SD Gothic Neo" as CJK fallbacks. This avoids .AppleSystemUIFont entirely without losing CJK glyphs. Trade-off: the UI no longer renders in the exact system font — a visual/design decision worth confirming before shipping.

If you get reports about this we should consider "Helvetica Neue" as the default on macOS - which is also a workaround for the issue you fixed in this PR.

niksedk:
Thx for the info :)

Trying "Helvetica Neue" seems like it worth trying. 

SE does seem to get a lot of issues regarding IME/font atm!

PR #12020, opened 2026-06-30
Fix bare Alt not activating the main menu bar (Windows)

niksedk:
## Problem
Bare **Alt** (press + release with no other key) is supposed to activate the menu bar for keyboard navigation (Windows standard, added in #11745), but in RC3 it doesn't:
- Press/hold Alt → access-key underlines appear; **release** → underlines vanish, **menu bar not activated**.
- Quick Alt → underlines flash, then nothing.
- After opening a menu with Alt+letter, a subsequent bare Alt **does** toggle it closed correctly.

## Cause
`ActivateMainMenu` focused the first menu item **synchronously inside the Alt key-up handler** (`TryFocusMainMenu` → `firstItem.Focus()`). Avalonia's own access-key handler processes the **same Alt key-up afterwards** — it hides the underlines and resets the access-key session, which **undoes** the focus we just set, so activation never sticks.

This is why *closing* works but *opening* doesn't: `DeactivateMainMenu` already **defers** its focus change (`Dispatcher.UIThread.Post`) with the comment *"moving focus from inside the key handler is racy"* — `ActivateMainMenu` was the one place that didn't.

## Fix
Defer the activation focus the same way, so the key event (and Avalonia's access-key handling) finishes before we focus the menu bar. No-op on macOS (native menu — `ActivateMainMenu` returns early).

## ⚠️ Verification
This path is Windows/Linux only (inert on macOS, where I'm developing), so I couldn't reproduce it locally. Builds clean and the change is low-risk (mirrors the existing `DeactivateMainMenu` pattern), but it should be **confirmed on Windows**: bare Alt should now activate the menu bar on release, and Alt again / Escape should deactivate it.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -18628,7 +18628,14 @@ public partial class MainViewModel :
         }
 
         _focusBeforeMainMenu = Window?.FocusManager?.GetFocusedElement() as Control;
-        return TryFocusMainMenu();
+
+        // Defer focusing the menu bar: when this runs from a key handler (bare Alt key-up, F10),
+        // Avalonia's own access-key handler also processes the same key and resets focus afterwards,
+        // which undid a synchronous focus here - so bare Alt showed the access-key underlines but
+        // never actually activated the bar. Let the current key event finish first, mirroring the
+        // deferred focus restore in DeactivateMainMenu (#11745).
+        Dispatcher.UIThread.Post(() => TryFocusMainMenu());
+        return true;
     }
 
     /// <summary>


GrampaWildWilly:
As a non-coder on this project this all just sounds to me like Avalonia bugs.  If they would have paid at least a little attention to being adherent to Windows standards, you wouldn't be writing all these silly fixes.  Yes, they are silly.  I mean, come on.  Menu operations?  You can't get much more basic than that.  And they don't work properly because of this or that weird way Avalonia does things.  Every other Windows app manages to get this stuff to work because they rely on the operating system to do this stuff for them.  For SE 5, Avalonia has totally dropped the ball, leaving you to code around their bugs, which they seem reluctant to address.  I appreciate your persistence in fixing all these standards violations.  But I believe none of them should have ever occurred.  None of these things was a problem in SE 4 & earlier versions.

Nevertheless, I am looking forward to the next beta.

GrampaWildWilly:
The bare Alt problem is still in beta 4.

GrampaWildWilly:
I should point out a GitHub bug.  I guess it's a GitHub bug.  I mention it here because the only way I could find this PR was to search on the string "bare alt" instead of relying on the `sort:updated-desc` qualifier in the search.  If I search on just `is:pr sort:updated-desc` this PR does not appear even though I've just updated it.  Searching Issues with `is:issue sort:updated-desc` does show updated discussions immediately.  Only Pull Requests seem to suffer this GitHub bug.  It's not an SE bug.  But you should probably know about this anomaly when you're doing searches for things.

GrampaWildWilly:
Now this is showing up in the regular GitHub search.  There must be some delay on GitHub during which fresh PR updates don't show up.  Whatever.  It's something to keep in mind.

PR #12025, opened 2026-06-30
Fix Get dictionaries status dot via BooleanToBrushConverter

ivandrofly:
## Summary

The **Get dictionaries** window's status dot was broken: it was computed once at build time and bound to non-existent DataContext properties, so it never reflected the selected dictionary — and the window did not compile.

This change renders the dot's colour reactively from `SelectedDictionary.IsInstalled` through a new reusable `BooleanToBrushConverter`.

## Changes

- **New** `BooleanToBrushConverter` (`src/ui/Logic/ValueConverters/`) — `IValueConverter` with settable `TrueBrush`/`FalseBrush`, defaulting to green/gray.
- **Status row dot** now binds `Fill` to `SelectedDictionary.IsInstalled` via the converter, so it updates when the selection or installed state changes.
- **Dropdown dots** now use the same converter against `IsInstalled` (they previously bound to the never-updated `StatusBrush`, so they were always gray).
- Removed the now-dead `StatusBrush` property and the VM-side brush bookkeeping (`SelectedStatusBrush`, static `InstalledBrush`/`NotInstalledBrush`).

## Verification

- `dotnet build src/ui/UI.csproj` compiles (it previously did not).
- Spell check → Get dictionaries: status dot is green for installed languages, gray otherwise, and updates instantly on selection change and after a download.

### DIFF
--- a/src/ui/Features/SpellCheck/GetDictionaries/GetDictionariesViewModel.cs
+++ b/src/ui/Features/SpellCheck/GetDictionaries/GetDictionariesViewModel.cs
@@ -1,6 +1,5 @@
 using Avalonia.Controls;
 using Avalonia.Input;
-using Avalonia.Media;
 using Avalonia.Platform;
 using Avalonia.Threading;
 using CommunityToolkit.Mvvm.ComponentModel;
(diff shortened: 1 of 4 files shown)


ivandrofly:
Viewmodel is not meat to have UI stuff

PR #12026, opened 2026-06-30
[macOS] Fix Option+Arrow word navigation in AvaloniaEdit editors (color tags, source view)

mjuhasz:
## Problem
  
  When color tags are enabled or the source view (F2) is open, the edit box uses AvaloniaEdit's `TextEditor` instead of the standard Avalonia `TextBox`. AvaloniaEdit's built-in macOS word navigation (Option+Arrow) behaves differently from the platform convention and from the standard text box:
  
  - **Wrong direction**: Option+Right moved to the *start* of the next word instead of the *end* of the current word (e.g. cursor after `T` in `This is` → jumped to before `is` instead of after `This`)
  - **No punctuation stops**: commas and periods were absorbed into the adjacent word rather than being their own navigation stop
  - **No newline stop**: the newline between subtitle lines was skipped entirely instead of producing a stop at the start of the next line 
    
  The result was inconsistent behavior between color-tags mode and non-color-tags mode, and between macOS apps and Subtitle Edit in color-tags mode.
    
  ## Fix
  
  Adds a tunneling `KeyDown` handler to `TextEditorWrapper` that intercepts Option+Left/Right on macOS and applies the same category-aware word boundary logic used by the standard `TextBox`:
  
  - Alphanumeric runs (`[a-zA-Z0-9_]`) and punctuation runs are separate stops 
  - Newlines are consumed alone — Option+Right lands at the start of the next line, Option+Left lands at the end of the previous line
  - Spaces and tabs are skipped without stopping
  - Shift+Option+Arrow extends the selection correctly, with a stable anchor that persists across consecutive keypresses

  Also fixes a pre-existing stale-key bug: the `OnKeyDownHandler` early-return guard for macOS Option+Arrow did not call `_shortcutManager.ClearKeys()`, unlike the equivalent Option+Backspace and Cmd+Delete handlers. A stale `Key.Left`/`Key.Right` in the active-key set could match an unintended chord on the next keypress if focus switched away before the key-up event fired.
  
  ## Scope
  
  - macOS only — on Windows/Linux, Alt+Left/Right are the video-seeking shortcuts (500 ms back/forward) and are left untouched
  - Applies to all `TextEditorWrapper` instances: main edit box (color tags on) and source view (F2) for files under 2 MB
    
  ## Testing
    
  With `Lorem ipsum dolor sit amet,\nconsectetur adipiscing elit.` in the color-tags edit box, Option+Right from the start now stops at:
  
  `Lorem|` `ipsum|` `dolor|` `sit|` `amet|` `,|` `|consectetur` `consectetur|` `adipiscing|` `elit|` `.|`
  
  This matches the standard text box (no color tags) exactly.

  ## non-macOS

  I have no immediate access to Windows and Linux, but I suspect that the word navigation is inconsistent there as well, just in the opposite way. On Windows, Ctrl+Right should stop after the whitespace, so the behavior of the Avalonia TextEditor is correct. In contrast, the TextBox behavior, which stops right after the word, is incorrect and inconsistent with native platform editors like Notepad. Most likely, a tweak similar to what I just implemented could be applied to a TextBox wrapper. However, without the ability to test it, I only fixed the macOS inconsistency in this PR.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -18834,6 +18834,7 @@ public partial class MainViewModel :
                             && (keyEventArgs.KeyModifiers == KeyModifiers.Alt
                                 || keyEventArgs.KeyModifiers == (KeyModifiers.Shift | KeyModifiers.Alt)))))
                 {
+                    _shortcutManager.ClearKeys();
                     return;
                 }
 
--- a/src/ui/Features/Shared/TextBoxUtils/TextEditorWrapper.cs
+++ b/src/ui/Features/Shared/TextBoxUtils/TextEditorWrapper.cs
@@ -1,11 +1,15 @@
 using Avalonia.Controls;
 using Avalonia.Input;
+using Avalonia.Interactivity;
 using Avalonia.Media;
 using AvaloniaEdit;
+using AvaloniaEdit.Document;
+using AvaloniaEdit.Editing;
 using Nikse.SubtitleEdit.Features.SpellCheck;
 using Nikse.SubtitleEdit.Logic;
 using System;
 using System.Linq;
+using System.Runtime.InteropServices;
 
 namespace Nikse.SubtitleEdit.Features.Shared.TextBoxUtils;
 
@@ -17,6 +21,7 @@ public class TextEditorWrapper : ITextBoxWrapper
     private readonly SubtitleTextAlignmentTransformer _alignmentTransformer;
     private EventHandler? _alignmentUpdateHandler;
     private TextAlignment _currentAlignment = TextAlignment.Left;
+    private int? _wordNavAnchor;
 
     public bool HasFocus { get; set; }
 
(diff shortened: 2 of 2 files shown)


GrampaWildWilly:
I've done a bit of experimenting with beta 4 on Windows.  Ctrl+Arrow keys seem to work fine.  They stop on all the things you'd expect them to stop on.  That includes stopping on line end, then Ctrl+RightArrow from line end stops on line begin on the next line.  The only thing I find mildly disconcerting is that Ctrl+RightArrow stops after the last letter of the current word, before the space between words.  Elsewhere in Windows, Notepad++ for example, Ctrl+RightArrow stops before the first letter of the next word, after the space between words.  But this is such a minor deviation from standard that I say don't bother fixing it.

mjuhasz:
> Ctrl+RightArrow stops after the last letter of the current word, before the space between words. Elsewhere in Windows, Notepad++ for example, Ctrl+RightArrow stops before the first letter of the next word, after the space between words.

That's exactly what I suspected; thank you for confirming it. It's a minor issue that only meticulous individuals with trained eyes notice. However, once you see it, you can't unsee it so I figured, at least on macOS, I would fix it for my own peace of mind 🙂

PR #12035, opened 2026-07-01
Remove checkbox multi-select wiring from Fix Common Errors grids

ivandrofly:
Drops the `DataGridCheckboxMultiSelect` helper from the rules grid and the fixes grid in `FixCommonErrorsWindow`.

This is just dangling instances

Build passes with 0 warnings.

### DIFF
--- a/src/ui/Features/Tools/FixCommonErrors/FixCommonErrorsWindow.cs
+++ b/src/ui/Features/Tools/FixCommonErrors/FixCommonErrorsWindow.cs
@@ -119,8 +119,6 @@ public class FixCommonErrorsWindow : Window
             },
         };
         rulesGrid.Bind(IsVisibleProperty, new Binding(nameof(vm.Step1IsVisible)));
-        new DataGridCheckboxMultiSelect<FixRuleDisplayItem>(rulesGrid,
-            item => item.IsSelected, (item, v) => item.IsSelected = v);
 
         var step2Grid = MakeStep2Grid();
         step2Grid.Bind(IsVisibleProperty, new Binding(nameof(_vm.Step2IsVisible)));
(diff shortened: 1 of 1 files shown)


niksedk (changes requested):
Thanks for the cleanup pass, but I don't think these are dangling — they're intentional and removing them drops functionality.

`DataGridCheckboxMultiSelect` isn't meant to be stored; the grid's event subscriptions keep it alive. That's documented on the type itself:

```csharp
// The caller does not need to store the returned instance.
// Event subscriptions on dataGrid keep it alive for the lifetime of the grid.
```

Its constructor has side effects on the grid — it sets `SelectionMode = Extended` and attaches `KeyDown`/`PointerPressed`/`PointerReleased`/`SelectionChanged` handlers — so the un-assigned `new` is not dead code. The same not-stored form is used in ~15 other windows (Batch convert, Fix Netflix errors, Fix names, Remove text for HI, Apply duration limits, Merge continuation lines, Convert actors, OCR, Binary edit, …), so removing it only here would also be inconsistent.

Concretely, dropping the two instances from `FixCommonErrorsWindow` removes:

- **Rules grid (step 1):** `Space` to toggle a rule, and `Shift`+Up/Down/PageUp/PageDown/Home/End (and shift-click) to select + bulk-toggle a range of rules.
- **Fixes grid (step 2):** the same Space/Shift multi-select, **plus** `onFocusedItemChanged -> SelectAndScrollTo(item)`, which scrolls/selects the corresponding line in the main grid as you move through fixes. That auto-follow is lost.

"Build passes with 0 warnings" is expected (`new X(...)` without assignment doesn't warn), but a clean build doesn't preserve the behavior here.

If the goal is that it reads as intentional rather than accidental, a `_ = new DataGridCheckboxMultiSelect<...>(...)` discard (or a one-line comment at the call sites) would do that without losing the behavior.

PR #12048, opened 2026-07-01
Prevent UI tests from overwriting user settings

mjuhasz:
## Summary

  Prevent UI tests from reading or writing the user's live `Settings.json`.

  ## Problem

  Some UI tests execute commands that call `Se.SaveSettings()`. On macOS, the default settings path resolves to:

  `~/Library/Application Support/Subtitle Edit/Settings.json`

  Running these tests therefore overwrite the user's actual application settings.

  ## Changes

  - Add an internal settings-file path override to `Se`.
  - Configure the UI test assembly to use a unique temporary settings directory.
  - Reset the override and remove temporary files after the test suite.
  - Add a regression test verifying that the settings path is redirected during UI tests.
  - Keep production settings behavior unchanged.

  ## Verification

  - All 490 UI tests pass.
  - The live `Settings.json` hash and timestamp remain unchanged after the full UI test suite.
  - Temporary test settings are removed after the run.

### DIFF
--- a/src/ui/Logic/Config/Se.cs
+++ b/src/ui/Logic/Config/Se.cs
@@ -66,6 +66,7 @@ public class Se
     public static readonly bool IsPortable;
     public static readonly string ExePath;
     public static readonly string DataFolder;
+    internal static string? SettingsFilePathOverride { get; set; }
 
     static Se()
     {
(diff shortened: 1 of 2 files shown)


niksedk:
Ah, that's why I sometimes lost my settings on mac - thx :)

Issue #12087, opened 2026-07-02
Accessibility observations in Subtitle Edit v5.1.0-beta5: menu navigation, DataGrid accessibility, remaining unlabeled controls, and playback behavior


Shaima-Almarzooqi:
I tested Subtitle Edit v5.1.0-beta5 with NVDA and wanted to share a few observations. First, thank you for the accessibility improvements that have already been implemented. The labeling work in the Settings and Shortcuts windows is noticeable, and most of the controls I mentioned previously (such as combo boxes and check boxes) are now correctly announced by NVDA.
1. Menu navigation
When testing v5.1.0-beta1, I found that pressing Alt correctly activated the menu bar, allowing me to navigate both the main menus and their menu items using the arrow keys. This worked very well with NVDA.
In v5.1.0-beta5, I am no longer able to do this. Pressing Alt does not appear to activate the menu bar, and the arrow keys do not navigate between the top-level menus.
Expected behavior
• Pressing Alt should move keyboard focus to the menu bar.
• Left and Right arrow keys should move between the main menus.
• Up and Down arrow keys should navigate the menu items.
2. DataGrid accessibility
The largest remaining accessibility challenge is the application's DataGrid controls.
Wherever a DataGrid is present (for example, in the Shortcuts window and other parts of the application), NVDA simply announces:
"Data grid"
After that, I am unable to interact with or explore the control using the keyboard. I cannot navigate its contents, determine what is currently selected, or obtain any additional information from the screen reader. 
Ideally, the DataGrid should expose its contents through UI Automation so that screen readers can announce information about the currently focused item as keyboard navigation occurs. 
3. Remaining unlabeled controls
Most of the controls mentioned in my previous feedback are now correctly labeled, which is greatly appreciated.
However, there are still a few controls in the main interface and in parts of the Settings and Shortcuts windows that NVDA announces only as generic Avalonia controls, for example:
Avalonia.Controls...
These controls do not expose a meaningful accessible name, making it difficult to determine their purpose. 
4. Playback controls and keyboard shortcuts
I also noticed an issue related to video playback.
Pressing the expected playback shortcuts, such as Space or Ctrl + Space, does not appear to start video playback.
Similarly, activating the Play button in the main interface also does not seem to start playback.
I'm not sure whether this is related to accessibility, keyboard focus, or a separate application issue, but I wanted to mention it in case it is relevant. 
Thank you again for the continued work on accessibility.

[...]

PR #12089, opened 2026-07-02
Accessibility: fix bare-Alt menu activation under NVDA and name remaining unlabeled controls (#12087)

niksedk:
Addresses points 1 and 3 of #12087 (NVDA feedback on v5.1.0-beta5).

## 1. Bare Alt no longer activated the menu bar

Avalonia's built-in `AccessKeyHandler` already owns the bare-Alt gesture — it opens the main menu on Alt release and closes it (restoring focus) when Alt is pressed while the menu is open. That is what beta1 shipped with and what the reporter confirmed working with NVDA.

The custom bare-Alt toggle added in beta2 ran *in addition* to the built-in one. Its tunnel key-up handler runs right after the built-in handler has opened the menu, sees "menu focused", and immediately deactivates it — the two toggles cancel out and Alt appears dead. The built-in handler only acts when a control **inside** the window has keyboard focus, which is the normal state and the one a screen reader always keeps the window in; when nothing inside the window has focus (e.g. right after startup), the built-in handler stays inert and only the custom toggle ran — which is why manual testing could look fine while NVDA users saw a dead Alt key.

Fix: remove the custom bare-Alt toggle (main window + BinEdit) and let the built-in handler own the gesture. Kept:

- **F10** toggle (Avalonia does not handle F10).
- **Escape** deactivation, adjusted for the built-in state model: the "drop-down open?" check now looks at the top-level items instead of `Menu.IsOpen`, because the built-in Alt activation sets `IsOpen` with no drop-down showing — treating that as "drop-down open" would demand an extra Escape to leave an Alt-activated bar.
- On bare Alt key-down we still remember the focused control so an Escape deactivation can restore it.

## 3. Remaining controls announced as generic "Avalonia.Controls..."

- **GridSplitters**: the Fluent theme makes every splitter a focusable tab stop (keyboard resize); the main window alone has a dozen nameless ones across the layouts. One app-level style names them all ("Resize panels") in every window, current and future.
- **Waveform**: the `AudioVisualizer` is focusable with no text content; it now has the name "Waveform".
- **Settings window**: the search box (placeholder text is not exposed as the UIA Name) and the section navigation buttons (icon + label panel content yields no computed name) now have accessible names.
- **Time-code fields**: the icon-only up/down repeat buttons of `TimeCodeUpDown`/`SecondsUpDown` are taken out of the tab order — the text box is the single (named) tab stop and Up/Down arrows already step the value, so screen-reader users are no longer stopped on two nameless buttons per field.

New language strings: `General.Waveform`, `General.ResizePanels` (English.json updated).

## Testing

- All 524 UI tests pass; app starts and runs normally on macOS (native menu unaffected).
- The Alt/NVDA behavior needs verification on Windows with NVDA running — with focus placed on a control (subtitle grid or text box) before pressing Alt, since that is the state where the old code failed.


GrampaWildWilly:
The bare Alt failure is still in beta 5.

GrampaWildWilly:
While F10 seems to continue to toggle the menu bar properly in beta 5, the first entry on the menu bar remains highlighted even after the menu bar is deactivated.

GrampaWildWilly:
The bare Alt issue is finally fixed in beta 7.  F10 followed by Esc still does not clear the highlight on the menu bar entry.  Alt does clear the leftover F10 highlight.  Esc on  a menu initially opened via either Alt or F10 still bounces focus back to the first menu bar entry.  F10 still puts a box around the first menu bar entry.  But the box disappears when you navigate to other entries on the menu bar.  I say the box is a bug.  You should never generate the box.  Alt does not cause the box to appear.

GrampaWildWilly:
Beta 8: no change.

GrampaWildWilly:
Beta 9: no change.

GrampaWildWilly:
With SE 5.1.0 RC 1 the remaining issues are still present.  See my post upthread here dated July 3 for details.

GrampaWildWilly:
No change in RC2.

PR #12100, opened 2026-07-03
TTS: fix cancellation and lifecycle bugs (bug hunt batch 2)

niksedk:
Second batch from the TTS bug hunt — the cancellation/lifecycle cluster. **Stacked on #12099** (base is that branch; merge #12099 first, then this retargets to main automatically).

## Fixes

- **Cancel now actually stops the work**: `StartAndWaitAsync` only stopped *waiting* on cancellation — the child ffmpeg kept running to completion (holding input files that the merge loop was deleting under it), and the timeout overload explicitly excluded user-cancel from `Kill`. Both overloads now kill the process tree on cancellation.
- **Test voice no longer hijacks a running generation**: it assigned the popup's `CancellationTokenSource` into the shared pipeline fields, so the popup's Cancel aborted the whole run and the main Cancel cancelled the popup instead. Now a local token + no-op while generating.
- **TTS window timer leak**: the 100 ms UI timer was never stopped/disposed on close — every open/close leaked a view model ticking for the rest of the session (the review window already cleaned up, confirming the omission).
- **Review history dialog**: playback permanently broke after the first Stop (one-shot CTS cancelled, never renewed; the 200 ms watchdog then killed every later play). Renewed per play.
- **Regenerate (review window)**, four related fixes:
  - cancel threw an unhandled `OperationCanceledException` (only `HttpRequestException` was caught); OCE is quiet now, other failures show a dialog;
  - the row's live `StepResult` was mutated right after `Speak`, so cancel/failed post-steps left a half-updated row that OK/Export published — rolled back on every failure path;
  - engine failures were never checked (row got an empty file name) — now surfaced with the engine's error message (pairs with the ElevenLabs `ErrorMessage` plumbing from #12098);
  - the non-modal progress popup left the grid interactive (`IsRegenerateEnabled` was bound to nothing), so a play/second regenerate mid-run swapped the shared CTS under the first one — all rows are now disabled during regenerate, and `ReplaceCts` no longer disposes a source a popup may still hold (`Cancel()` on a disposed CTS threw).

## Testing

- Remaining hunt findings queued for batch 3: review-window playback state machine (auto-continue follows selection, mpv lock race, missing-file wedge), voice-list refresh hardening (ElevenLabs/Murf/Mistral), export path portability, engine-switch consistency.

### DIFF
--- a/src/ui/Features/Video/TextToSpeech/ProcessExtensions.cs
+++ b/src/ui/Features/Video/TextToSpeech/ProcessExtensions.cs
@@ -19,15 +19,26 @@ public static class ProcessExtensions
     public static async Task StartAndWaitAsync(this Process process, CancellationToken cancellationToken)
     {
         process.StartProcess();
-        await process.WaitForExitAsync(cancellationToken);
+        try
+        {
+            await process.WaitForExitAsync(cancellationToken);
+        }
+        catch (OperationCanceledException)
+        {
+            // WaitForExitAsync only stops *waiting* on cancellation - the child keeps running.
+            // Cancel means stop the work, so kill it; otherwise a cancelled merge/generation
+            // leaves ffmpeg burning CPU (and holding its input files open) to completion.
+            KillNoError(process);
+            throw;
+        }
     }
 
     /// <summary>
     /// Starts the process and waits for it to exit, but no longer than <paramref name="timeout"/>.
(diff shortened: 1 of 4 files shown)


niksedk:
Landed in main via #12103 (commit unchanged).

Issue #12093, opened 2026-07-03
ElevenLabs and tons of issue with nothing on output


cvrle77:
Where exactly is 'Generate speech from text' is, in Elevenlabs model, and why it's so complicated to set it up? Took me 20 minutes to fiddle with options to unravel them.

EDIT: I had to generate speech using Piper, close it down, to get that button show up.

EDIT 2:

Now it's stuck on this.

<img width="1510" height="837" alt="Image" src="https://github.com/user-attachments/assets/11339e32-717a-48f3-8d60-ae2840f4317c" />

<img width="1512" height="838" alt="Image" src="https://github.com/user-attachments/assets/d2aeb6d5-9d65-4600-a25e-dcb5a38fbfeb" />

EDIT 3: I've managed to do 16 lines sub, and reach editor, that's also buggy.

EDIT 4: Now I can't do those 16 lines again, it gets stuck.

Also, TTS window just go in the background sometimes, and I have no idea it's on.

EDIT 5: after 1000s of credits spent this is furthest I've got with 0 result. And I got here only when I turned off every Advanced option.

<img width="1512" height="838" alt="Image" src="https://github.com/user-attachments/assets/d59eb551-d686-4c11-b6a5-28d156da2755" />

This too: 

<img width="3011" height="1657" alt="Image" src="https://github.com/user-attachments/assets/341321bd-3afc-4999-82d1-570687205557" />

If I rename it to, for example to Kajsije.SubtitleEditTts.json, then it's visible.

If I somehow manage to get to Export option, woohoo.

But then, when I import it, I can't export it. I can't get one single wav file no matter what I do.

Too bad it's completely useless (because I can't get a complete output), I love new options. Though you must integrate spacebar as play- stop. And it's kinda messy with me playing the line, but line isn't selected at the same moment. And other 294 bugs.

niksedk:
Thanks for the detailed report and sorry about the credits burned - a few of these I can already confirm from the code, and for the main one (getting stuck) I need your log files.

**Confirmed: export/import file mismatch.** Export writes a file named exactly `SubtitleEditTts.json`, but the import dialog's file filter only matches `*.SubtitleEditTts.json` (something **before** the name, like your `Kajsije.SubtitleEditTts.json` rename). So SE's own export is invisible in SE's own import dialog. Will be fixed - your rename workaround is fine in the meantime.

[...]

PR #12101, opened 2026-07-03
TTS review: fix playback state machine and export round-trip (bug hunt batch 3)

niksedk:
Third batch from the TTS bug hunt — the review-window playback state machine and the export/import round-trip. **Stacked on #12100** (which stacks on #12099); merge in order and GitHub retargets each.

## Playback state machine

- **Auto-continue followed the selection, not the playing row**: grid selection is two-way, so clicking another row during playback stranded the playing row's stop icon and resumed from the clicked row's successor. The playing row is now tracked explicitly (`_playingRow`), and selection follows playback on auto-continue — which is also the "playing line isn't the selected line" complaint from #12093.
- **mpv race**: the 200 ms watchdog read `_mpvContext.IsPaused` outside `_playLock` while Stop/close dispose the player under that lock — NRE / native use-after-free window. State reads now happen under the lock, and all UI property updates from the timer thread are marshaled to the dispatcher.
- **No more wedges**: a throw inside a tick used to kill the timer permanently (all rows stayed disabled); playing a row whose audio file is missing sat in "playing" forever (mpv's failed loadfile is fire-and-forget). Both now reset the playback UI, the latter with an error naming the missing file.

## Regenerate/history correctness

- Picking a history entry now restores `StepResult.SpeedFactor`, not just the display string — Export no longer writes the previous speed next to the restored audio.
- The regenerate pipeline snapshots `SelectedModel`/`Instruction`/language/region up front — clicking another row mid-run (non-modal popup) used to rewrite them via the panel sync, recording settings that never produced the audio.

## Export/import round-trip

- Export no longer crashes unhandled mid-way when a row's audio file is missing (files copied, no JSON written) — such rows are exported without audio and the count is reported.
- **Exports are now portable**: the JSON stores audio file *names* instead of absolute paths, and Import resolves them against the JSON's own folder. Old exports with absolute paths still work (honored when they exist, then retried next to the JSON) — moving or sharing an export folder finally round-trips.
- Import shows a clear error instead of dying on `Voices.First()` when the selected engine's voice list failed to load.

## Testing

- Remaining hunt findings for a final batch: voice-list refresh hardening (ElevenLabs/Murf/Mistral), engine-switch consistency in the main TTS window, FixSpeed dropping trim-failed segments, TtsPostProcessor silent no-op + intermediate leaks, and the smaller engine nits (Google IsInstalled credential, AllTalk escaping/cancellation, Azure SSML locale).

### DIFF
--- a/src/ui/Features/Video/TextToSpeech/ReviewSpeech/ReviewSpeechViewModel.cs
+++ b/src/ui/Features/Video/TextToSpeech/ReviewSpeech/ReviewSpeechViewModel.cs
@@ -121,6 +121,10 @@ public partial class ReviewSpeechViewModel : ObservableObject
     private long _startPlayTicks;
     private readonly List<string> _tempAudioFiles = new();
 
+    // The row whose audio is currently playing. Auto-continue must advance from this row, not
+    // from SelectedLine - grid selection is two-way and can move during playback.
+    private ReviewRow? _playingRow;
+
     public ReviewSpeechViewModel(IFolderHelper folderHelper, IWindowService windowService)
     {
         _folderHelper = folderHelper;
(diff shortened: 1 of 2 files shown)


niksedk:
Landed in main via #12103 (GitHub closed the stacked PRs when the batch-1 base branch was deleted; the commits are unchanged).

PR #12102, opened 2026-07-03
TTS: harden voice-list refresh, engine-switch state, and engine nits (bug hunt batch 4)

niksedk:
Final batch from the TTS bug hunt. **Stacked on #12101** (→ #12100 → #12099); merge in order.

## Voice-list refresh hardening (ElevenLabs / Murf / Mistral)

The list downloads copied HTTP error bodies (or nothing) into the stream and only logged; the engines wrote that over their cached voice JSON, and since the bundled fallback only restores a *missing* file, **one failed refresh permanently emptied the voice list**. The downloads now throw on non-success, and the view model's refresh wrapper catches, keeps the cached list, and shows the error. (Murf's list request also sent `Accept: audio/mpeg` to a JSON API.) Together with the Azure fix in batch 1, all voice-list refreshes are now failure-safe.

## Engine-switch consistency (main window)

A failed or empty voice load used to leave the window describing the *previous* engine (swallowed exception / early return): stale voice list, hidden API-key field — the user couldn't even enter the key to fix the cause — and Generate could hand the wrong engine's `Voice` object to `Speak`. Capability flags and the engine-specific settings blocks now always run; a failed load shows an empty voice list for the new engine.

## Pipeline consistency

- **FixSpeed** kept-original policy now covers the unreadable-trim-output path too (was: silently dropped the line from the final audio).
- **TtsPostProcessor**: failed stages are force-logged instead of silently disabling the user's enabled option, zero-byte outputs are no longer adopted, and superseded intermediates are deleted (a 1000-line run with all options used to leave ~3000 stale wavs).

## Engine nits

- **Google**: `IsInstalled` checked the API key, but the engine authenticates exclusively via the key *file* — configured engines reported "not installed" and vice versa.
- **AllTalk**: JSON escaping removed from a plain multipart field (quotes/backslashes were spoken), the speak request honors the cancellation token, and a missing/unreachable server output path fails the segment with a message instead of an unhandled `File.Move` throw.
- **Azure**: SSML now uses the voice's own locale instead of hardcoded `en-US`.
- **ElevenLabs**: fixed the `"eleven_turbo_v2)"` / `"eleven_multilingual_v1)"` stray-parenthesis comparisons (unreachable branches).

## Testing

- This completes the actionable findings from the hunt. Not addressed (judgment calls / lower value): the Cancel-flips-UI-before-pipeline-stops ordering race, background-thread property sets in older sibling VMs, and the OmniVoice `--ref-text` parameter question (needs a manual check against the CLI's expected form).

### DIFF
--- a/src/ui/Features/Video/TextToSpeech/Engines/AllTalk.cs
+++ b/src/ui/Features/Video/TextToSpeech/Engines/AllTalk.cs
@@ -163,7 +163,17 @@ public class AllTalk : ITtsEngine
 
         var languageCode = language != null ? language.Code : "en";
         Se.WriteToolsLog($"AllTalk: voice={allTalkVoice.Voice}, language={languageCode}, textLen={text.Length}");
-        var allTalkFileNameOutputFileName = await _ttsDownloadService.AllTalkVoiceSpeak(text, allTalkVoice, languageCode);
+        var allTalkFileNameOutputFileName = await _ttsDownloadService.AllTalkVoiceSpeak(text, allTalkVoice, languageCode, cancellationToken);
+
+        // The server reports its output as a local path. It can be empty (unexpected response
+        // JSON) or unreachable from here (server on another machine / different working dir) -
+        // File.Move then threw unhandled instead of failing the segment like other engines.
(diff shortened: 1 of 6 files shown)


niksedk:
Landed in main via #12103 (commit unchanged).

PR #12114, opened 2026-07-03
Fix TTS import OK doing nothing + review window keyboard flow (#12093)

niksedk:
Follow-ups from the [latest feedback](https://github.com/SubtitleEdit/subtitleedit/issues/12093#issuecomment-4876291972) on #12093.

### The main issue: OK after Import did nothing

`Import()` opened the review window and **discarded its result** — so clicking OK behaved exactly like Cancel: no merged wav, no save dialog, nothing. OK now runs the same merge / save-wav / add-to-video tail as the generate pipeline. Two follow-on bugs in that tail are also fixed:

- The silence base track is now sized to cover the imported segments too — an imported session can outlast the subtitle currently open in the main window, which would have cut off the tail of the merged wav.
- A vestigial engine/voice null-guard in `MergeAudioParagraphs` could still silently abort the merge for imported sessions whose selected engine had no loadable voice list (same "OK does nothing" symptom). The merge doesn't need an engine/voice at all — each step result carries its own audio file.

### Review window keyboard flow

- **Space no longer clicks OK.** Space/R are handled at tunnel stage, before the focused control sees the key — previously the initially focused OK button swallowed bare Space as a click and published the whole session. Space now always means play/pause (and still types normally in text boxes).
- **Bare R regenerates the selected line** — pairs with Space for fast keyboard-only review ("thumb on Space, finger on R"). Ctrl+R still works.
- **Playing a row selects it**, so Space (replay) and R (regenerate) always target the line just heard instead of a previously selected row.

### API key

Masked with `●` by default (no more editing screenshots), with an eye button to toggle visibility. Icon names verified against the bundled Material Design set.

### Not addressed here

- The Windows-only "file dialogs / windows drop to the taskbar" behavior needs a Windows repro — likely an Avalonia window-activation issue.
- The cut-off button row at 200% scaling is already fixed on main (#12112, stale saved window size applied to a non-resizable window).

### DIFF
--- a/src/ui/Features/Video/TextToSpeech/ReviewSpeech/ReviewSpeechViewModel.cs
+++ b/src/ui/Features/Video/TextToSpeech/ReviewSpeech/ReviewSpeechViewModel.cs
@@ -952,6 +952,10 @@ public partial class ReviewSpeechViewModel : ObservableObject
         _skipAutoContinue = false;
         _startPlayTicks = DateTime.UtcNow.Ticks;
 
+        // Playing a row selects it, so the keyboard (Space to replay, R to regenerate) always
+        // targets the line just heard - previously selection stayed on the old row (#12093).
+        SelectedLine = line;
+
         line.IsPlaying = true;
         _playingRow = line;
         foreach (var l in Lines)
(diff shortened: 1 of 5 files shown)


niksedk:
Added: the auto-translate window and batch convert's auto-translate view now mask their API keys with the same eye-toggle control (moved to a shared `UiUtil.MakeApiKeyTextBox` helper). Also cherry-picked the test-host hang fix so the full suite runs on this branch: 527/527 green.

PR #11836, opened 2026-06-25
Prep Flathub submission: git-source manifest + refresh metainfo releases

niksedk:
Groundwork for publishing on Flathub (#11800). The existing CI Flatpak (`installer/flatpak/`) already builds from source with bundled mpv/ffmpeg/Tesseract+eng and offline NuGet, and is green in `build-flatpak.yml` — so this is just the small adaptation needed to submit, not a new packaging effort.

## What's here
- **`installer/flatpak/flathub/dk.nikse.subtitleedit.yaml`** — identical to the CI manifest except the app source is a **pinned git tag + commit** (`v5.0.0` / `a44ffa70…`) instead of the local `type: dir` checkout. Flathub builds from the public repo, so this is the form it needs.
- **`installer/flatpak/flathub/nuget-sources.json`** — copy of the offline NuGet sources (must be regenerated for the pinned commit before submitting; noted in the README).
- **`installer/flatpak/flathub/README.md`** — submission steps (regenerate sources → validate → fork `flathub/flathub` → branch `dk.nikse.subtitleedit` → PR to `new-pr`), plus the likely reviewer questions.
- **metainfo**: added the `5.0.0` stable release (Flathub wants the latest release listed; it previously only had `5.0.0-beta9`).

## Not done here (needs a Linux box / the maintainer)
- Regenerating `nuget-sources.json` at the v5.0.0 commit (`generate-nuget-sources.sh`).
- Running `appstreamcli validate` + `flatpak-builder-lint` (no flatpak tooling on macOS; the metainfo is at least well-formed XML).
- The actual `flathub/flathub` PR and flathub.org verification.

## Heads-up for review
The main discussion point will be the **optional AI engines downloading executables at runtime** (whisper.cpp, CrispASR, TTS/STT). Core editing, OCR (bundled Tesseract) and video (bundled mpv/ffmpeg) all work offline, but Flathub scrutinizes runtime executable downloads — be ready to justify or limit those under the sandbox. (Details in the README.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #11800, opened 2026-06-24
Publish on Flathub?


CaptechOmar:
It would be easier to install and update since it will work on all linux distros very easily this way

Spike1377:
I hate the way Linux is going where everyone wants to install packages the wrong way instead of using the system repos.

CaptechOmar:
Sure, let's build a package for each distro out there and hope there isn't a distor left without a package and a maintainer for each of those packages and let's hope the distro have a graphical app for their repos instead of having to use the terminal instead of having a one easy method that will easily work on all distors and is easy to update it.

tbh, I'm not sure if you prefer having a package in the repo of each distro, what do you have against having a one more method to install the app?

Spike1377:
If people are too scared to use the terminal that's their problem. Its up to the pakage maintainers of each distro to compile SumbtitleEdit themselfs so it wont be all done by niksedk. Anyway I think his gonna add flatpak so good for people that don't have it in the repos and don't wanna compile it. Not for me though on Arch.

I have against flatpack because it takes more space since it bundles up its dependencys and Linux already has a great package manager.

niksedk:
I am not very good at Linux. I only run a few distros in Virtual Box / docker, so this is probably above my skill level, as Flathub seems complicated, but a PR is welcome. 
Also, probably not too hard to use the supplied flatpak in the Releases.

muaz978:
I will try my best working on this for the next week, and will submit the PR then hopefully as soon as I test it on my Linux laptop.

muaz978:
Following up on my earlier comment here.

I dug into this and found you already started the groundwork yourself in #11836 (git-source manifest, regenerated nuget-sources, metainfo release entry, and a README with the actual submission steps), it just got closed without a note on why. I left a question there asking if you want that picked back up or if you had a different plan.

Assuming it is fine to continue from that work rather than starting over, here is what I am planning to do, let me know if you want any of this changed or if there is a part you would rather handle yourself:

1. Refresh the manifest to pin the current commit (main has moved on since the v5.0.0 commit that branch pinned).
2. Regenerate nuget-sources.json for that commit and validate the git-source build end to end through the existing Flatpak CI container, the same one used for the local-dir build already in CI.
3. Open a PR here with the refreshed manifest.
4. Once that is in, do the actual submission to flathub/flathub (fork it, branch named after the app id, PR to new-pr).

[...]

PR #12118, opened 2026-07-03
Prep Flathub submission: refreshed manifest, one-command release automation (#11800)

muaz978:
Prep work for #11800 (Flathub request), picking up from the closed #11836.

## Why a fresh branch instead of reopening #11836
That branch was pinned to a commit from months ago, main has since moved through the AI Review feature and a lot more, so the diff had drifted into hundreds of unrelated files. Ported the actual Flathub-specific work forward onto the current tree instead of trying to rebase it.

## What is here
`installer/flatpak/flathub/` holds the Flathub-ready variant of the existing Flatpak manifest (`installer/flatpak/dk.nikse.subtitleedit.yaml`, already validated in CI by `build-flatpak.yml`). It is identical to that manifest in every module except the app's own source: a git checkout pinned to a release tag and commit, instead of the local directory checkout CI builds from, since that is what Flathub actually builds against.

On the "make it auto-update so it is not a lot of work per release" ask:
- `x-checker-data` on that source lets Flathub's own external-data-checker bot open a PR bumping the tag/commit automatically whenever a new stable tag is pushed (pre-releases like `-beta*` are excluded by the tag pattern).
- That bot has no way to regenerate `nuget-sources.json` though, so `prepare-flathub-release.sh` does that half: given a tag, it resolves the commit, regenerates the NuGet source list for that exact commit in a disposable worktree (does not touch whatever branch is currently checked out), and updates the manifest's pin to match. That script is the entire per-release manual step going forward.

Also answered your other question on the issue: yes, third-party downloads are already hash-checked. `src/ui/Logic/Download/DownloadHashManager.cs` verifies every downloaded AI-engine artifact against a known SHA-256 before use. Documented that in the README as a concrete answer for when Flathub reviewers ask about the runtime downloads.

Added the missing stable 5.0.0 release entry to the AppStream metainfo too, it still only listed 5.0.0-beta9.

## What is intentionally not in this PR
`nuget-sources.json` for the pinned commit. It has to be generated on a machine with the Flatpak SDK (or a system .NET 10 install) via `prepare-flathub-release.sh`, the same requirement `generate-nuget-sources.sh` already has for the CI manifest. I do not have that environment here, so I will generate it and validate the actual build on my own Linux machine before this goes further, per your comment on #11800. The Flathub submission itself (forking flathub/flathub, opening the PR there) is a separate step after that.

## Files
- `installer/flatpak/dk.nikse.subtitleedit.metainfo.xml`
- `installer/flatpak/flathub/dk.nikse.subtitleedit.yaml`
- `installer/flatpak/flathub/prepare-flathub-release.sh`
- `installer/flatpak/flathub/README.md`

### DIFF
--- a/.github/workflows/build-ui.yml
+++ b/.github/workflows/build-ui.yml
@@ -139,6 +139,99 @@ jobs:
           path: SubtitleEdit-linux-x64.flatpak
           overwrite: true
 
+  # Same chain as above, but for installer/flatpak/flathub/dk.nikse.subtitleedit.yaml,
+  # the manifest actually submitted to Flathub, which builds from a pinned git tag +
+  # commit instead of this checkout's working tree. Building it here, in CI, on a real
+  # Linux runner is the practical way to validate that exact variant per #11800; it
+  # does not need a contributor's own Linux machine or Flatpak install.
+  regenerate-flathub-nuget-sources:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          # Full history so the commit pinned in the flathub manifest (which can be
+          # older than this checkout's HEAD) is guaranteed fetchable for the worktree
+          # checkout below.
+          fetch-depth: 0
+
+      - name: Set up .NET SDK
+        uses: actions/setup-dotnet@v5
+        with:
+          dotnet-version: '10.0.x'
+
+      - name: Install flatpak
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y flatpak
+
+      - name: Cache flatpak runtimes
+        id: flatpak-cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.local/share/flatpak
+          key: flatpak-fd24.08-dotnet10-v2
+          restore-keys: |
+            flatpak-fd24.08-dotnet10-
+
+      - name: Install/update Freedesktop SDK + dotnet extension
+        run: |
+          flatpak --user remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
+          flatpak --user install -y --noninteractive flathub \
+            org.freedesktop.Platform//24.08 \
+            org.freedesktop.Sdk//24.08 \
+            org.freedesktop.Sdk.Extension.dotnet10//24.08
+          flatpak --user update -y --noninteractive
+
+      - name: Regenerate nuget-sources.json for the pinned commit
+        run: bash installer/flatpak/flathub/prepare-flathub-release.sh
+
+      - name: Verify nuget-sources.json is consistent with the pinned commit
+        run: |
+          if git status --porcelain -- installer/flatpak/flathub/nuget-sources.json | grep -q .; then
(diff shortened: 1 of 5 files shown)


muaz978:
Addressed the Copilot review:

- python3 requirement: added to the "Requires" header, it was missing even though the script uses it to parse/update the manifest.
- Fragile tag fetch: switched to an explicit refspec (refs/tags/TAG:refs/tags/TAG) instead of the "--tags TAG" form, and stopped swallowing its errors, so a genuinely missing tag now fails loudly at the fetch step instead of surfacing later as a confusing "unknown revision" from rev-parse.

On the nuget-sources.json comment: that one is intentional rather than a bug. That file has to be generated on a machine with the Flatpak SDK or a system .NET 10 install, this environment has neither, so per the discussion on #11800 I added a CI job pair (regenerate-flathub-nuget-sources / build-linux-flathub) that generates and build-tests it on a real Linux runner instead. Already documented in the PR description and the README on this branch. Once that job's artifact is downloaded and the actual bundle is run and confirmed working, I will commit the generated file here.

muaz978:
Fixed, and the diagnosis was right: checking the pinned commit out in place (from the previous fix) replaces the whole working tree, including installer/flatpak/flathub/ itself, since that folder is newer than the v5.0.0 pin. Writing the regenerated file there then failed with exactly the FileNotFoundError described.

Added the mkdir -p right before the generator runs (Copilot's fix 1). Fix 2 does not quite apply as written, flatpak-dotnet-generator.py is not a file in this repo, it is downloaded fresh from a pinned upstream commit inside generate-nuget-sources.sh, so a local patch to it would not persist or do anything on the next run.

niksedk:
@muaz978: Can you look at this: https://github.com/SubtitleEdit/subtitleedit/actions/runs/28705401067/job/85130214249 ?

muaz978:
When I run it on my fork, it functioned pretty well.
Please see this photo attached.
image
I will check this one you mentioned as well.

muaz978:
Checked it, this is not a manifest problem, it is a transient external failure.

The build-linux-flathub (x86_64) job failed downloading uchardet's source tarball from freedesktop.org:

    Failed to download sources: module uchardet: The requested URL returned error: 418

That is the same uchardet source URL every other job in that same run used, and all of them succeeded: build-linux-flathub (aarch64) in the same run, and the pre-existing build-linux-flatpak job (identical uchardet source stanza) both completed fine. A one-off 418 from freedesktop.org's server or CDN while four Linux/flatpak jobs hit it in the same few minutes is exactly the kind of blip a re-run clears up, nothing in the manifest or the pin changed between the jobs that succeeded and the one that did not.

Re-running just that failed job should be enough. If it fails the exact same way again, that would be worth a second look, but a single 418 like this is almost certainly the freedesktop.org side, not us.

muaz978:
Confirmed fixed: https://github.com/SubtitleEdit/subtitleedit/actions/runs/28707087697 completed with every job green, including both build-linux-flathub legs (x86_64 and aarch64). That confirms it was the transient freedesktop.org 418 as suspected, nothing wrong with the manifest.

PR #12127, opened 2026-07-03
Fix download dialogs getting stuck when the download itself fails

muaz978:
Found while testing the Flathub build from #11800 on Linux ARM64: opening Speech to text, picking Purfview Faster Whisper XXL, got stuck on "Unpacking 7-zip archive..." forever.

## Root cause

That engine's download method already correctly throws `PlatformNotSupportedException` for Linux ARM64 (and for macOS), since Purfview's upstream project only ships Windows and Linux x64 builds. Nothing showed that error though.

The download-progress timer callback decides "the download finished, proceed to unpack" using:
```csharp
if (_downloadTask is { IsCompleted: true })
```
`Task.IsCompleted` is true for **Faulted** and **Canceled** tasks too, not just ones that ran to completion. So a task that threw took this exact same branch as a successful download and tried to unpack a file that was never actually downloaded, which is what got stuck. The `IsFaulted` check right below it was unreachable for this case, since `IsCompleted` had already matched first.

## Fix

Checked for the same pattern across every download-progress dialog in the app. Nine of them had it: Faster Whisper XXL, Whisper models, Tesseract, Google Lens OCR, PaddleOCR, ffmpeg, libVLC, libmpv, yt-dlp. All switched from `IsCompleted` to `IsCompletedSuccessfully`, which is only true for tasks that actually ran to completion, so a faulted or canceled download now falls through to the existing `IsFaulted` handling and shows the real error instead of hanging.

Three other download dialogs (TTS, Tesseract model, dictionaries) already used `IsCompletedSuccessfully` correctly, this brings the rest in line with that existing, correct convention.

Not Flatpak-specific and not ARM64-specific: this would hang on any download failure on any platform (network failure, unsupported platform/architecture, whatever), across all nine dialogs.

## Files
- `src/ui/Features/Video/SpeechToText/DownloadSpeechToTextEngineViewModel.cs`
- `src/ui/Features/Video/SpeechToText/DownloadSpeechToTextModelsViewModel.cs`
- `src/ui/Features/Ocr/Download/DownloadTesseractViewModel.cs`
- `src/ui/Features/Ocr/Download/DownloadGoogleLensOcrViewModel.cs`
- `src/ui/Features/Ocr/Download/DownloadPaddleOcrViewModel.cs`
- `src/ui/Features/Shared/DownloadFfmpegViewModel.cs`
- `src/ui/Features/Shared/DownloadLibVlcViewModel.cs`
- `src/ui/Features/Shared/DownloadLibMpvViewModel.cs`
- `src/ui/Features/Shared/DownloadYtDlpViewModel.cs`

### DIFF
--- a/src/ui/Features/Ocr/Download/DownloadGoogleLensOcrViewModel.cs
+++ b/src/ui/Features/Ocr/Download/DownloadGoogleLensOcrViewModel.cs
@@ -60,7 +60,10 @@ public partial class DownloadGoogleLensOcrViewModel : ObservableObject
                 return;
             }
 
-            if (_downloadTask is { IsCompleted: true })
+            // IsCompletedSuccessfully, not the broader IsCompleted (also true for
+            // Faulted/Canceled), so a failed download falls through to the IsFaulted
+            // branch below instead of proceeding as if it had succeeded.
+            if (_downloadTask is { IsCompletedSuccessfully: true })
             {
                 _timer.Stop();
                 _done = true;
--- a/src/ui/Features/Ocr/Download/DownloadPaddleOcrViewModel.cs
+++ b/src/ui/Features/Ocr/Download/DownloadPaddleOcrViewModel.cs
@@ -82,7 +82,10 @@ public partial class DownloadPaddleOcrViewModel : ObservableObject
                 return;
             }
 
-            if (_downloadTask is { IsCompleted: true })
+            // IsCompletedSuccessfully, not the broader IsCompleted (also true for
+            // Faulted/Canceled), so a failed download falls through to the IsFaulted
+            // branch below instead of proceeding as if it had succeeded.
+            if (_downloadTask is { IsCompletedSuccessfully: true })
             {
                 _timer.Stop();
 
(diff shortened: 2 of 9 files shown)


tempodat:
I am getting this issue on the flathub build on Linux x86_64... Should I open an issue or is there a way to debug this?

SE V5.1.0-rc12
Fedora 44, KDE Plasma 6.7.3
RTX 5050

niksedk:
Thanks for the report! The fix from this PR is in rc12, so you're likely hitting a related but different problem: if the download succeeds but the **unpack** step fails, the dialog would still hang silently. On the Flatpak build a likely culprit is the bundled 7-zip binary failing to run inside the sandbox.

I've opened #12764 which makes unpack failures show a proper error (and fall back to a built-in extractor), so this shouldn't silently hang anymore.

To help confirm the cause, could you open an issue with:
- which engine/download you picked and the exact text the dialog is stuck on
- terminal output from running `flatpak run org.nikse.subtitleedit`
- the SE error log from `~/.var/app/org.nikse.subtitleedit/config/Subtitle Edit/` (error_log*.txt)

Thanks!

PR #12139, opened 2026-07-04
STT: use model's real folder name for distil Whisper models (#12133)

### LINKED ISSUE #12133: distil models folder (audio to text, Whisper)
hi I think SubtitleEdit creates a folder name `faster-whisper-distil-large-v3.5`  
and it should be `faster-distil-whisper-large-v3.5`.

- https://github.com/Purfview/whisper-standalone-win/issues/586#issuecomment-4879420180

normally when `Purfview-Faster-Whisper-XXL` can't find the model it downloads it
since SE handles downloading and folder creation, `..Faster..` might not be able to find it.

niksedk:
Fixes #12133

## Problem

The Purfview Faster-Whisper-XXL / CTranslate2 model folder was built as `"faster-whisper-" + model.Name`. That's correct for most models, but the **distil** models are published under a different word order — `faster-distil-whisper-<name>`. So SE downloaded e.g. `distil-large-v3.5` into:

```
_models\faster-whisper-distil-large-v3.5\      ← SE (wrong)
```

`faster-whisper-xxl.exe` looks for `faster-distil-whisper-large-v3.5`, doesn't find it, and **silently re-downloads** the model into its own correctly-named folder — wasting bandwidth and confusing users (SE's UI reports the model as installed while the exe fetches it again). Reported for all `distil-*` models, not just v3.5.

## Root cause

Each `WhisperModel` already carries the correct folder name in its `Folder` property (`faster-distil-whisper-large-v3.5`) — the engine just wasn't using it, reconstructing the path from `Name` with a hard-coded prefix instead.

## Fix

Add `WhisperEnginePurfviewFasterWhisperXxl.GetModelFolderName()` which prefers `model.Folder` (falling back to the legacy `faster-whisper-<name>` form), and use it in the three places that built model paths by hand:

- `GetAndCreateWhisperModelFolder` (download destination; CTranslate2 delegates here)
- `IsModelInstalled`
- `WhisperEngineCTranslate2.GetModelForCmdLine` (the `--model` path passed to the exe)

Also kept `IsModelInstalled` requiring the finished `model.bin` so an in-progress `model.bin.$$$` partial is never mistaken for an installed model.

## Verification

Reproduced and confirmed on disk with a real download:

- **Before:** SE created `_models\faster-whisper-distil-large-v3.5\` (transposed, wrong)
- **After:** SE uses `_models\faster-distil-whisper-large-v3.5\`, matching what the exe expects — no more re-download

Build: `dotnet build src/ui/UI.csproj` — succeeded, 0 errors.

### Note on existing installs
Users who already downloaded a distil model on an older build have an orphaned `faster-whisper-distil-*` folder. After this fix SE looks in the correct folder and downloads once more into the right place (harmless; the old folder can be deleted). No automatic migration/rename is included.

### DIFF
--- a/src/ui/Features/Video/SpeechToText/Engines/WhisperEngineCTranslate2.cs
+++ b/src/ui/Features/Video/SpeechToText/Engines/WhisperEngineCTranslate2.cs
@@ -90,8 +90,13 @@ public class WhisperEngineCTranslate2 : ISpeechToTextEngine
 
     public string GetModelForCmdLine(string modelName)
     {
-        var modelFileName = Path.Combine(GetAndCreateWhisperModelFolder(null), "faster-whisper-" + modelName);
-        return modelFileName;
+        // Resolve via the model's real folder name so distil models (published as
+        // "faster-distil-whisper-...") get the correct --model path, not "faster-whisper-<name>" (#12133).
+        var model = Models.FirstOrDefault(m => m.Name == modelName);
+        var folderName = model != null
+            ? WhisperEnginePurfviewFasterWhisperXxl.GetModelFolderName(model)
+            : "faster-whisper-" + modelName;
+        return Path.Combine(GetAndCreateWhisperModelFolder(null), folderName);
     }
 
     public bool SupportsCustomModels => true;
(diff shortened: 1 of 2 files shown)


eladkarako:
that's one solid explanation!

PR #12197, opened 2026-07-05
Show MKV progress window for multi-track files (#12193)

### LINKED ISSUE #12193: MKV parsing progress bar doesn't appear with multiple subtitles
Hello!

In #11712 MKV reading was moved into its own thread and shows a progress bar.

However this only works with exactly 1 subtitle present in the MKV.

When multiple subtitles are present, and therefore the parsing happens right after drawing the "Pick Matroska track" window, the UI remains blocked and the progress window never shows.

Likely there is a different code path when the "Pick Matroska track" window is used.

5.1.0-beta11 is affected.

Thank you!

niksedk:
Fixes #12193

## Problem

The Matroska "please wait" progress window (added in #11712) only appeared when an MKV had a **single** subtitle track. With multiple tracks, the *Pick Matroska track* dialog opens first, the UI froze during parsing, and the progress window never showed.

## Root cause

`MatroskaFile.GetSubtitle(...)` reads and caches the whole file's cluster data on its **first** call; later calls are instant. That first read is the slow part.

- **Single track** → goes straight to the threaded `ExtractMatroskaSubtitleAsync`, which shows the progress window. Works.
- **Multiple tracks** → the picker dialog opens first. When it draws, `SelectAndScrollToRow(0)` fired `TrackChanged()`, which called `GetSubtitle(...)` **synchronously on the UI thread** to build the preview — that first (heavy) read froze the UI with no progress window.

## Fix

In `PickMatroskaTrackViewModel`:

- Rewrote `TrackChanged()` → `async TrackChangedAsync()`. Parsing now runs off the UI thread in `Task.Run` via a static `BuildPreview(...)` that returns plain data (text + pre-decoded Avalonia `Bitmap`s, no controls); grid rows/`Image` controls are built back on the UI thread.
- Shows the same 25 MB-gated `PleaseWaitWindow` with the determinate progress callback, only until the initial cluster read completes (`_clusterLoaded`), so later track switches don't flash a progress window.
- Serialized preview parsing with a `SemaphoreSlim` (`MatroskaFile` isn't thread-safe — single shared `FileStream`) plus a `_trackChangeToken` to discard stale previews on rapid selection changes.
- Added a try/catch that logs via `Se.LogError` so a parse error in the fire-and-forget task isn't unobserved.
- Dropped a dead `format.ToText(...)` call from the old `AddTextContent` (result was unused).

## Testing

Builds clean (0 warnings / 0 errors). Not yet exercised against a real >25 MB multi-track MKV — reviewer confirmation on real media welcome.

### DIFF
--- a/src/ui/Features/Shared/PickMatroskaTrack/PickMatroskaTrackViewModel.cs
+++ b/src/ui/Features/Shared/PickMatroskaTrack/PickMatroskaTrackViewModel.cs
@@ -1,6 +1,7 @@
 using Nikse.SubtitleEdit.UiLogic.Export;
 using Avalonia.Controls;
 using Avalonia.Input;
+using Avalonia.Media.Imaging;
 using Avalonia.Threading;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
@@ -19,6 +20,7 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.IO;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Nikse.SubtitleEdit.Features.Shared.PickMatroskaTrack;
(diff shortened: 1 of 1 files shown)


lukastribus:
> reviewer confirmation on real media welcome.

Confirmed v5.1.0-beta12 fixes this (tested on real media).

PR #12245, opened 2026-07-07
MLX Whisper: chunked decoding with punctuation persistence, beam search, and subtitle standard cues

muaz978:
## Problem

On macOS the MLX Whisper engine produced noticeably worse subtitles than Purfview's Faster-Whisper-XXL on Windows, especially on long recordings and in languages like Arabic:

- Punctuation disappeared after the first minute. mlx-whisper applies `initial_prompt` only to the first 30 second window, and with conditioning reset the later windows decode with an empty prompt, so the model falls back to unpunctuated output in languages whose training data is mostly unpunctuated. Since punctuation drives sentence segmentation, cue quality degraded with it.
- Music and long silences produced hallucination loops (repeated tokens with zero length timestamps, "music" labels, prompt echoes).
- Raw Whisper segments make poor cues: often far too long or too short, with boundaries that ignore sentence structure.
- mlx-whisper only implements greedy decoding, while faster-whisper decodes with beam search (beam size 5) by default, which is one reason its output reads better on difficult speech.

## What changed

The transcribe helper now runs a proper subtitling pipeline:

- **Voice activity detection**: speech regions are detected with Silero VAD (via the faster-whisper package when available) and decoded chunk by chunk; music and silence are never decoded at all, which also removes the hallucination loops those passages caused. Nearby regions merge (gaps up to 3 s), chunks cap at 240 s.
- **Punctuation persistence**: every chunk is decoded with the punctuation prompt re-applied plus the tail of the previous chunk's text, so punctuation and context survive across the whole recording instead of only the first window.
- **Beam search**: the helper adds a beam search decoder to mlx-whisper at runtime (a port of openai/whisper's reference implementation to MLX, submitted upstream as ml-explore/mlx-examples#1429; the runtime patch can be dropped once that ships). `best_of 5` is also passed for the temperature fallback path, matching openai/whisper defaults.
- **Cue building**: cues are rebuilt from word level timestamps following the Netflix Timed Text Style Guide and BBC subtitle guidelines: break at sentence punctuation and speech gaps, balanced splits for oversized sentences (at soft punctuation or the largest pause near the middle), 42x2 characters and 7 s max per cue, 5/6 s minimum with short cues merging into a neighbor, two frame minimum gap, reading speed capped at 20 characters per second. `--raw-segments` restores plain Whisper segmentation.
- **Sanitizing**: degenerate cues are dropped (zero length, hallucination loops, wrong script text, prompt echoes), and doubled punctuation the decoder sometimes emits ("?." style) is collapsed.
- **Console**: each segment prints as `[MM:SS.mmm --> MM:SS.mmm] text` with absolute timestamps while decoding, plus model download progress lines with percentages.

UI additions (post processing dialog, applying to this engine):

- Cue building section: rebuild on/off, max characters per cue, max seconds per cue, max characters per second.
- Vocabulary prompt: Whisper's documented way to get names and technical terms spelled correctly; combined with the punctuation prompt and re-applied on every chunk, so it covers the whole recording.
- Beam size with a guidance note (0 = fastest, 5 = best accuracy at roughly 20-30% more time).

The engine list also shows the usual install dot for MLX Whisper now (green when the pip package is importable, grey when not).

## Practical results

Measured on an Apple M5, whisper-large-v3, beam size 5, a 45.8 minute Arabic TV episode:

- Total transcription time about 12 minutes. VAD reduced the audio actually decoded to 23.5 of 45.8 minutes across 48 chunks, which roughly pays for beam search (measured at about 1.0x to 1.5x of greedy wall time on large models).
- Punctuation (periods, Arabic commas and question marks) present from the first cue to the last minute of the episode; previously it vanished after the opening scene.
- 285 raw segments rebuilt into 330 cues, all within the configured 84 chars / 9 s / 20 cps limits.
- Four minutes of mid episode music produced zero hallucinated cues; the next region resumed with correct absolute timestamps.
- Side by side with Faster-Whisper-XXL (same episode, same model, Windows/CUDA): matching text and timestamps on the same scenes, and the Mac output avoided XXL's music hallucination at 01:54 because VAD skipped that region.

Screenshots of the new dialog section and sample output follow in a comment.

## Default behavior change

With the new defaults (cue rebuild on, beam size 5), existing MLX users' next run switches to beam search, word timestamp resegmentation, and VAD chunking. VAD engages only when the faster-whisper package is importable; otherwise the script prints a note and decodes in one pass. Both behaviors can be turned off in the post processing dialog (cue rebuild checkbox, beam size 0), and `--raw-segments` restores plain Whisper segmentation. On older mlx-whisper versions the script feature-detects what the installed version supports and degrades gracefully with a printed note instead of failing.

## How this was tested end to end

- Synthetic fixtures on Apple Silicon (M5): Arabic and German speech generated with macOS `say` plus generated music, arranged as music / speech / music / speech / music, verifying VAD chunk boundaries and absolute timestamps (first cues at 12.0 s and 75.3 s), punctuation present in both regions, no prompt echo cues over the music, no overlapping cues, and deterministic beam output at temperature 0.
- Real content: a 45.8 minute Arabic TV episode with whisper-large-v3 and beam size 5, about 12 minutes total; console log excerpt in the comments, compared scene by scene against Purfview's Faster-Whisper-XXL output for the same episode on Windows.
- Unit checks for the helper functions (balanced cue splitting, the standards pass including the overlap clamp, sanitizing, punctuation tidy) run directly with python3, plus py_compile on every change.


muaz978:
Console Log:
Example:
Calling speech-to-text (MLX Whisper Mac) with : /Library/Developer/CommandLineTools/usr/bin/python3 "/Users/muazsabbagh/Library/Application Support/Subtitle Edit/SpeechToText/MlxWhisperMac/mlx_whisper_transcribe.py" --audio "/var/folders/df/95g8c4493q162g42ycw955q00000gn/T/se-stt-6b85d100-1815-4dfd-ac7f-a4123e21a7d7/e4e16bdb-5a1c-4675-9de7-66ea7a296b23.wav" --model mlx-community/whisper-large-v3-mlx --output-format srt --output-dir "/var/folders/df/95g8c4493q162g42ycw955q00000gn/T/se-stt-6b85d100-1815-4dfd-ac7f-a4123e21a7d7" --language ar --beam-size 5 --max-cue-chars 84 --max-cue-duration 9 --max-cps 20
Downloading model 'mlx-community/whisper-large-v3-mlx' from Hugging Face (first use; this can take a while)...
Fetching 4 files: 100%
Fetching 4 files:   0%|          | 0/4 [00:00<?, ?it/s]
Fetching 4 files: 100%|██████████| 4/4 [00:00<00:00, 25458.60it/s]
Model download complete.
Loading MLX Whisper model 'mlx-community/whisper-large-v3-mlx' (Apple GPU/Neural Engine)...
Beam search enabled (beam size 5).
VAD: 48 speech chunk(s), 23.5 of 45.8 minutes to decode.
Fetching 4 files:   0%|          | 0/4 [00:00<?, ?it/s]
Fetching 4 files: 100%|██████████| 4/4 [00:00<00:00, 16529.28it/s]
[03:41.200 --> 03:44.960] أعتقد أننا سنصل إلى أطراف مدينة أفراح عند الصباح.
[03:47.200 --> 03:49.800] هل ما زلت مستعداً للإفاء بوعدك يا سعدون؟
[03:50.860 --> 03:52.860] بكل تأكيد أيها المغوار.
[03:55.200 --> 03:59.240] هيا، تقدمني لتثبت قدرتك على حماية المدينة،
[03:59.240 --> 04:04.260] كي يتسنى لك أن تتزوج من شامة، ابنة الملك أفراح، وبعدها
[04:05.200 --> 04:08.840] اتركني لألقن ذلك اللعين درساً لا ينساه أبداً.

niksedk:
Thanks for this — reviewed against current `main`. Good news first: it **test-merges cleanly** into current `main` with no conflict, and it does **not** regress the recently merged pipx/venv/conda MLX detection from #12242 (the base was stale, but the changed regions are disjoint, so `GetShebangInterpreters`/`GetCliShimCandidates` and the improved "not found" message all survive the merge).

A few things worth addressing before merge:

**1. (blocking-ish) Version-guard the new `transcribe` kwargs.** `decode_piece` now passes `hallucination_silence_threshold`, `best_of`, `word_timestamps` (and `beam_size`) to `mlx_whisper.transcribe`. Only `beam_size` is wrapped defensively (the `enable_beam_search` try/except); the others are not. On an older pinned `mlx-whisper` that predates any of these kwargs, `transcribe` raises `TypeError` and the **entire run aborts with no fallback** — a regression for those users. Please degrade gracefully (feature-detect / try-except and drop unsupported kwargs), the way `beam_size` already does.

**2. `best_of=5` on the greedy path.** Passed unconditionally at temperature 0; relies on `mlx-whisper` silently ignoring it when `beam_size` is set / temp is 0. Couple this with #1 so it can't error or silently change cost on versions that don't.

**3. `apply_standards` end-extension can overlap the next cue.** `end = max(latest, start + 0.2)` with `latest = next_start - min_gap`: if two cues start within ~0.16 s, the `start+0.2` floor pushes `end` past the next cue's start (negative gap / overlap), producing exactly the malformed timing this pass is meant to prevent. Clamp so the extension never exceeds `next_start - min_gap`.

**4. Cross-chunk prompt echo not fully sanitized.** `previous_tail` (last 150 chars) is fed into the next chunk's `initial_prompt`, but `sanitize_cues` only strips echoes of the punctuation `prompt`, not the per-chunk tail — so a chunk that opens by echoing the injected tail can emit a phantom duplicate cue at the boundary.

**5. (low) `_has_expected_script` can drop legitimate cues.** For `en`/`tr` it requires a Latin letter/digit, so a valid punctuation/symbol-only cue (e.g. `—`) is discarded.

**6. (low) Single VAD region > 240 s isn't capped.** The `<= 240` guard is only on the extension branch, not the initial seed from the first region, so one long speech region decodes whole and the "prompt refreshed every few minutes" guarantee silently doesn't hold there. Not a crash.

Empty-audio / last-chunk / all-silence paths are handled safely, and arg quoting is fine (`UseShellExecute=false`, quotes stripped from the vocabulary prompt).

**Heads-up on default behavior:** with the new defaults (`WhisperCueRebuild=true`, `WhisperBeamSize=5`, …), existing MLX users' next run silently switches to beam-search + word-timestamp resegmentation + VAD (VAD only engages if `faster-whisper` is also installed; otherwise it prints a note and falls back to single-shot). That appears intentional, but it's a default-behavior shift worth calling out in the PR description.

Net: sound design and the merge is clean — please address **#1** (and ideally #2–#4) and I think it's good to go. The +560 lines of Python are hard to fully validate without an Apple-Silicon runtime, so a note on how you tested end-to-end would help.

_(Review assisted by Claude Code.)_

muaz978:
Thanks for the thorough review. All six points are addressed in 918ac71:

1. **Version guards**: the script now feature-detects the installed mlx-whisper (the `transcribe` signature plus `DecodingOptions` fields) and drops unsupported options with a printed note instead of aborting: `word_timestamps` falls back to raw segments, `hallucination_silence_threshold` relies on the VAD clipping alone, `best_of`/`beam_size` decode greedily, and `initial_prompt` is guarded the same way. If introspection itself fails, previous behavior is kept.
2. **best_of**: covered by the same detection, so it is only passed when the installed version has the field; at temperature 0 mlx-whisper drops it, so the greedy path is unchanged in behavior and cost.
3. **Overlap**: reproduced with two cues starting 0.09 s apart; the old formula produced end = 0.2, overlapping the next cue. The extension is now capped at the gap before the next cue, with a final clamp so a degenerate near-simultaneous pair can touch but never overlap. Verified zero overlapping cues in the regression run.
4. **Cross-chunk echo**: `decode_piece` now drops leading segments of a chunk whose text is contained in that chunk's injected prompt, which covers the previous chunk's 150 character tail as well as the punctuation prompt.
5. **Script check**: cues containing no letters or digits at all are kept; there is no script signal to judge them by.
6. **Region cap**: single VAD regions longer than 240 s are split after merging, so the prompt refresh interval also holds for uninterrupted speech.

The PR description now has a "Default behavior change" section and a note on how the end-to-end testing was done. Regression run after the changes (two speech regions separated by music, beam size 5): 2 VAD chunks, correct absolute timestamps, punctuation in both regions, 16 cues, zero overlaps.

PR #12246, opened 2026-07-07
Add multi window support on macOS with Window menu and Dock integration

muaz978:
## Problem

On Windows you can open several Subtitle Edit windows; on macOS only one window per app was possible, and launching the app again just focused the existing instance. Mac users working on two subtitles at once had no good option.

## What changed

- **File > New window** opens an additional, fully independent editor window in the same process: own MainViewModel/MainView (both are transient in DI), own subtitle, own video player. Windows stack under a single Dock icon like native Mac apps, instead of spawning extra app instances with their own icons.
- **Per window menu bar**: each window gets its own native NSMenuBar wired to its own view model, so the menu always operates on the focused window. The existing menu state sync (gestures, visibilities, toggles, dynamic submenus) became per window state, keyed by the active window.
- **Window menu**: a standard macOS Window menu is added to the menu bar and registered with AppKit (`setWindowsMenu:`), so AppKit itself maintains the open window list with checkmarks, and mirrors it into the Dock icon's context menu, exactly like Finder or Safari.
- **Dock "New window" command**: the Dock icon's context menu also offers New window, via Avalonia's `NativeDock.Menu` attached property (whose native side implements `applicationDockMenu:`); AppKit shows it above the standard Dock entries.
- Closing the last window still exits the process (ShutdownMode.OnLastWindowClose plus the existing safety net), and the batch convert window path keeps its explicit exit.

## Testing

On macOS 15 (Apple Silicon):

- File > New window and Dock > New window both open an extra window; all windows share one Dock icon.
- The Window menu and the Dock menu list every open window and focus the chosen one; the active window carries the checkmark.
- Menus (including shortcuts, dynamic submenus, and toggles) follow the focused window.
- Closing windows one by one exits the app with the last one; no stray background process remains.
- Speech to text, video playback, and file open/save behave independently per window.

Screenshots of the Window menu and the Dock menu follow in a comment.

### DIFF
--- a/src/ui/Features/Main/Layout/InitMenu.cs
+++ b/src/ui/Features/Main/Layout/InitMenu.cs
@@ -69,6 +69,11 @@ public static class InitMenu
                     Command = vm.CommandFileNewKeepVideoCommand,
                     [!MenuItem.IsVisibleProperty] = new Binding(nameof(vm.IsVideoLoaded)),
                 },
+                new MenuItem
+                {
+                    Header = l.NewWindow,
+                    Command = vm.CommandFileNewWindowCommand,
+                },
                 new Separator(),
                 new MenuItem
                 {
--- a/src/ui/Features/Main/Layout/InitNativeMacMenu.cs
+++ b/src/ui/Features/Main/Layout/InitNativeMacMenu.cs
@@ -78,6 +78,15 @@ public static class InitNativeMacMenu
     // Conditional/Toggle helpers register into it (UI thread only, build is synchronous).
     private static MenuState? _building;
 
+    // Windows-menu ownership: when MacWindowsMenuInterop manages to register our
+    // "Window" submenu as NSApplication.windowsMenu, AppKit populates it (and mirrors
+    // it into the Dock icon's menu), and we must never mutate it again, since changing
+    // the NativeMenu would rebuild the NSMenu and wipe AppKit's entries. Until the
+    // first registration attempt resolves, the menu stays empty; on failure the manual
+    // list below takes over.
+    private static bool _appKitOwnsWindowsMenu;
+    private static bool _windowsMenuInteropFailed;
+
     // ── App menu ─────────────────────────────────────────────────────────────
     // Called from Program.cs. Sets the "Subtitle Edit" app dropdown contents.
     // Avalonia also auto-appends the standard macOS items (Hide, Quit, etc.).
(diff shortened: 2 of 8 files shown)


muaz978:
image

muaz978:
image
image

niksedk:
Reviewed — nicely done. The refactor is careful: the per-window `MenuState` replacing the former app-global statics is the right model, the native AppKit/Dock interop is properly isolated and defensive, threading is clean (all `NativeMenu`/interop mutation on the UI thread), and the single-window Windows/Linux path is unchanged. No crash, no cross-platform breakage, no meaningful leak. A few follow-ups:

**1. (should fix) `_registeredMenus` is never pruned on window close.** `MacWindowsMenuInterop.cs` (~line 585) accumulates `NSMenu` pointers for the whole process lifetime. Freed NSMenu addresses can be reused by AppKit, so a later window whose menu lands on a recycled pointer gets `Add(...) == false` and skips seeding the already-open windows via `addWindowsItem:` — its Window menu can omit the other open windows on first display. Please remove the entry on window close (and it avoids unbounded growth).

**2. (should document) All windows share the global `Se.Settings`.** Each window has its own `MainViewModel` (subtitle/video/undo isolated — good), but every window reads/writes the one global settings object, and each `OnClosing` saves the full settings file. So recent-files, `LayoutNumber`, window position/size, and video prefs are **last-writer-wins** across windows. The PR body implies full independence, which only holds for the subtitle itself — worth a note.

**3. (nice to have) Symmetric unsubscribe of the `Activated` handler on close.** `InitNativeMacMenu.cs` (~line 116): on `Closed` you correctly unsubscribe the VM `PropertyChanged` and drop the `MenuState`, but the `Activated` lambda stays attached. It's GC-safe here (the window is being destroyed and nothing else holds it), just asymmetric with the deliberate `PropertyChanged` cleanup.

**4. (low) One-way `_appKitOwnsWindowsMenu` latch.** `InitNativeMacMenu.cs` (~line 94 / `UpdateWindowMenus`): once AppKit registration succeeds it's `true` for the process lifetime, permanently disabling the manual fallback. Only reachable in narrow states (last-window-close force-exits and resets statics), so low severity — a short comment on the intent would help future readers.

Also noting (pre-existing, not a regression): the `Environment.Exit(0)` on last-window-close can truncate the `Task.Run` background saves in `OnClosing` — same tradeoff as the existing single-window path, just now reachable from N windows.

Cross-platform safety looks solid: `TryRegisterWindowsMenu`/Dock interop early-return when `!OperatingSystem.IsMacOS()` and are only wired from the macOS-gated paths; on Windows/Linux `File > New window` just opens a second in-process window with no native calls.

Overall: mergeable — #1 is the one I'd fix before/right after merge; the rest are polish.

_(Review assisted by Claude Code.)_

niksedk (approved):
Approved. The macOS multi-window refactor is careful and cross-platform-safe (all AppKit/Dock interop is IsMacOS-guarded; the Windows/Linux single-window path is unchanged), threading is clean, and there are no crash/leak blockers. The follow-ups noted above (prune `_registeredMenus` on window close, document the shared `Se.Settings`, symmetric `Activated` unsubscribe) are polish and can be handled in a follow-up.

_(Review assisted by Claude Code.)_

PR #12256, opened 2026-07-07
Fix text box orientation in right to left translation mode

### LINKED ISSUE #12249: RTL mode orientation when translating
Hi,

This is how RTL mode currently behaves in SE5 when in translation mode:

Image

The text boxes don't correspond to the list. Original should be on the left, under the Original list, and vice versa, the translation should be on the right. This is how SE4 behaved. Same file on SE4:

Image

muaz978:
Fixes #12249.

### Problem

With right to left mode enabled in translation mode, the subtitle grid and the text boxes below it no longer correspond. The right to left helper sets FlowDirection on the DataGrid, which mirrors the whole list, so the Original column renders on the left. The text edit area below is a plain grid (current text in the first column, original text in the second) that the helper never touched, so the two text boxes kept their left to right positions: Original column on the left, original text box on the right. Subtitle Edit 4 mirrored both together.

### Fix

Two parts, developed and verified iteratively on macOS with a Turkish and Arabic subtitle pair:

1. **Box positions follow the mirrored grid.** Setting FlowDirection on a plain grid turned out not to rearrange its children (only the DataGrid honors it for column order), so the text edit grid now carries a name and the right to left helper mirrors its columns for real: the column definitions are reversed (the original column's width binding travels with its definition object, so the collapse when no original is loaded keeps working) and every child is remapped to the mirrored column index. Idempotent through the grid's Tag, so repeated passes from the window open and toggle paths cannot double swap.

2. **Text direction follows the content; the layout follows the mode.** Right to left mode used to force every text control right to left (pushing a Turkish original under an Arabic working language to the wrong side of its box), and left to right mode forced everything left to right (pushing an Arabic original under a Turkish working language to the wrong side). The two edit text boxes (both the plain and the color tags variants) and the Text and Original text grid columns now detect their content through LanguageAutoDetect.ContainsRightToLeftLetter in either mode, so each language reads from its own side whatever the mode is; empty controls follow the active mode so typing starts on the expected side. The layout itself (column order, box positions) still changes only with the mode toggle, exactly like Subtitle Edit 4. Subtitle Edit 4 resolved box direction per file and only in right to left mode (FixRightToLeftDependingOnLanguage); resolving per line and in both modes came out of testing all four language arrangements. The boxes re-evaluate on every text change, the grid cells bind through a converter (producing explicit directions, since a binding that yields UnsetValue falls back to the property default rather than the inherited direction), and toggling the mode rebinds the grid so the cells update immediately.

3. **The mode is applied after every layout build.** The window level right to left pass runs when the window opens, but the main layout is built later (posted to the dispatcher after the view attaches), so starting the app with right to left mode already enabled walked a tree that did not contain the subtitle grid or the text boxes yet: the layout came up left to right and only looked correct after toggling the mode by hand. Switching layouts had the same effect, since the rebuilt controls start left to right. The pass is now re-applied after the deferred startup layout build, after a layout switch, and after the undocked video layout rebuild; it is idempotent (the grid mirror tracks its state, directions are recomputed from content), so repeated application is safe. Toggling the mode also rebinds the subtitle grid's items source so the per line converter picks up the new mode immediately.

### Scope

- Applies on toggle and on window open (both call sites share the helper).
- The right to left UI language path (whole window flow direction) is unchanged.
- Other windows are unaffected by the mirroring; the helper matches the edit grid by name. Text boxes elsewhere pick up content aware direction only through the existing recursive pass, which already touched them.
- In left to right mode the layout never changes; only text direction inside cells and boxes follows the content.

### Testing

Verified on macOS (Apple Silicon) with a Turkish and an Arabic subtitle in all four arrangements (each language as the working and as the original subtitle, in each mode), including starting the app with right to left mode already enabled from the previous session:

- Right to left mode: the original text box sits under the Original column exactly as in Subtitle Edit 4, the Turkish side reads from the left, the Arabic side from the right, in the grid and in the boxes.
- Left to right mode: layout unchanged; a right to left original next to a left to right working language reads from the right in its column and box, and files with only left to right content look exactly as before.
- Toggling back and forth restores each arrangement with no double swap; other windows are unchanged.

Screenshots for both cases are in the comments.

### DIFF
--- a/src/ui/Features/Main/Layout/InitListViewAndEditBox.cs
+++ b/src/ui/Features/Main/Layout/InitListViewAndEditBox.cs
@@ -1250,6 +1250,9 @@ public static partial class InitListViewAndEditBox
         // Right panel for text editing (show/duration is to the left)
         var textEditGrid = new Grid
         {
+            // RightToLeftHelper mirrors this grid by name so the current/original
+            // text boxes keep matching the mirrored subtitle grid columns.
+            Name = "SubtitleTextEditGrid",
             ColumnDefinitions = new ColumnDefinitions("*,*,Auto"),
             RowDefinitions = new RowDefinitions("Auto,*,Auto"),
         };
--- a/src/ui/Features/Main/Layout/InitListViewAndEditBox.cs
+++ b/src/ui/Features/Main/Layout/InitListViewAndEditBox.cs
@@ -118,6 +118,7 @@ public static partial class InitListViewAndEditBox
         var notNullConverter = new NotNullConverter();
         var nullToOpacityConverter = new NullToOpacityConverter();
         var syntaxHighlightingConverter = new TextWithSubtitleSyntaxHighlightingConverter();
+        var textToFlowDirectionConverter = new TextToFlowDirectionConverter();
         vm.SubtitleDataGridSyntaxHighlighting = syntaxHighlightingConverter;
         // How the Text/Original cells fit their text to the window (feature #11590). Read once here;
         // the grid is rebuilt when settings are applied, so a changed mode takes effect then.
@@ -305,6 +306,7 @@ public static partial class InitListViewAndEditBox
                 {
                     VerticalAlignment = VerticalAlignment.Center,
                     [!TextBlock.InlinesProperty] = new Binding(nameof(SubtitleLineViewModel.Text)) { Converter = syntaxHighlightingConverter, Mode = BindingMode.OneWay },
+                    [!TextBlock.FlowDirectionProperty] = new Binding(nameof(SubtitleLineViewModel.Text)) { Converter = textToFlowDirectionConverter, Mode = BindingMode.OneWay },
                 };
                 SubtitleGridTextDisplayModeDisplay.ApplyTo(textBlock, gridTextDisplayMode);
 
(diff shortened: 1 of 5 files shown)


niksedk:
Could you add a screenshot or two?

muaz978:
Please see the photo below:
image
The problem was solved but we need to adjust it to make auto detect languages and accordingly align the text.

muaz978:
I will compile again, test and attach a screenshot for after polish.

muaz978:
image
image
Please check these dear, @niksedk, and tell me if there is any adjustment you want.
The Screenshots are for both cases; working on translating RTL or LTR languages.

muaz978:
Final state after full testing of all four language arrangements. Three follow up commits since the last screenshots:

1. The flow direction converter now always produces an explicit direction. It returned UnsetValue for right to left content expecting inheritance from the mirrored grid, but a binding that yields UnsetValue falls back to the property default (left to right), which misaligned right to left cells in right to left mode.
2. Text direction follows the content in both modes: a right to left original next to a left to right working language now reads from the right also in left to right mode. The layout itself still changes only with the mode toggle.
3. The right to left pass is re-applied after every layout build. It used to run only at window open, which is before the deferred layout build, so starting the app with the mode already enabled came up left to right until the mode was toggled by hand. Layout switches had the same problem.

Verified on macOS in all four arrangements, screenshots follow: Turkish working subtitle in left to right mode, Arabic working subtitle in right to left mode (mirrored layout), Turkish plus Arabic original in left to right mode (Arabic reads from the right in its column and box), and Arabic plus Turkish original in right to left mode (mirrored, Turkish reads from the left in its column and box). Startup with the mode already enabled and layout switching keep the correct arrangement in every case.

muaz978:
image
image
image
image

muaz978:
One note for completeness: the last commit of this work was pushed shortly after the merge, so it did not make it into this PR. It is now open as #12276 and finishes the polishing of this feature and fix: it re-applies the right to left pass after every layout build, so starting the app with right to left mode already enabled (and switching layouts) comes up in the correct arrangement without toggling the mode by hand. The four screenshots above were taken with that commit included, so #12276 is the version they show.

Issue #9996, opened 2025-10-11
[Feature request] Add documents to give AI translation more context


leegethas:
When running local LLM's with Ollama, you can add documents to a chat to provide context. I would very much like this option in Subtitle Edit too. I use Ollama to translate subtitles with Subtitle Edit, which works surprisingly well. However, I have no option to correct or improve the translation. I can edit the command that Subtitle Edit sends to the selected LLM, but I cannot include documents to provide more context. I think this option would be a very easy and convenient way to improve translations.

aalkz:
> I believe that for best results there should be a way to pass `previous` and `next` line in the prompt, for context. I know we can 'merge' sentences which gives better translation results. But splitting them back gives rather bad results, and needs a lot of manual fiddling after to alter start/finish times of each split line.
> 
> Or maybe improve the `split` function so it would be able to revert to the times before merging?

I think this is the way it should work for LLMs specifically. I was actually looking for a setting to adjust this and I stumbled on this thread so I guess it's still on the works.

The name of the show or movie is a good start so that the LLM knows some stuff beforehand but it really just needs to have a wider view of the scene or dialogue going on. For example I just had a case in which some character asked a question and it was split in two subtitle lines, and the LLM botched the first part of the question and made it into an affirmative statement and it ended up reading very weird. If it had the context of just the line before I'm sure it would have gotten it right.

niksedk:
Mostly covered by #13005: the "llama.cpp advanced" and "Ollama advanced" auto-translate engines have synopsis, glossary, rolling history of previous lines, batch size, custom system prompt and context size (see docs/features/auto-translate-advanced.md), and the SE 5 prompt box has no length limit. Feel free to open a new issue if something is still missing.

PR #12257, opened 2026-07-07
Added context of previous and the following line to automatic translators, improving translation quality.

goodPointP:
Auto-translating text works better when provided with the context of the previous and the next subtitle lines. This feature was initially discussed here: https://github.com/SubtitleEdit/subtitleedit/issues/9996


niksedk:
Thank you for taking the time to work on this, and sorry for the slow reply.

Giving the LLM surrounding context is definitely the right idea, but Subtitle Edit has since gotten there via a different route, and this PR unfortunately collides with it:

**1. `{2}` in a prompt now means something else.** Since #13803, a prompt containing `{2}` marks a *completion-format template* (`{0}` = source language, `{1}` = target language, `{2}` = the text itself), which switches the whole wire format for the local-LLM engines - see `LlmTranslatePrompt.cs`. Using `{2}`/`{3}` for previous/next line would break that.

**2. The "advanced" engines already do this, and more.** "llama.cpp advanced (local LLM)" and "Ollama advanced (local LLM)" translate in numbered batches with a rolling history of the already-*translated* lines (so terminology and pronouns stay consistent, not just grammar), plus an optional synopsis, glossary and formality setting - and a schema-constrained JSON reply, so line alignment is guaranteed rather than hoped for.

**3. Merged sentences already cross line boundaries.** `MergeAndSplitHelper` merges a sentence spanning several subtitles *before* translating it, which covers most of the "a question split over two lines got mistranslated" case.

**4. The new default prompts have no empty-context guard** - the first and last subtitle of every file, and every non-merged single line, would send "the previous subtitle line was , and the following is ." - and they roughly double the prompt tokens per call on the paid engines with no way to turn it off.

The plan instead is to extend the advanced batch/context engine to more backends (LM Studio and a generic OpenAI-compatible one first), which gets the context benefit without the per-line prompt overhead.

Closing this one - but thanks again, and the underlying point was a good one.

goodPointP:
Makes total sense - thanks for reviewing and great to see the idea still got implemented.

PR #12289, opened 2026-07-08
Fix live spell check using the English dictionary for other languages

muaz978:
### Problem

With live spell check enabled and only the English dictionary installed, an Arabic speech to text transcription came back with every single word underlined in red, in the subtitle grid and in the text box.

### Cause

Live spell check picks its dictionary by auto detecting the subtitle's language, but the detection helper used here falls back to English when the subtitle is empty or undetectable. The setup runs at startup before any file is loaded, so the spell checker was armed with the English dictionary against an empty subtitle. Content that arrives without a file open, such as a speech to text transcription or a video OCR result, never re-evaluated that choice, so Arabic text was spell checked against the English dictionary.

### Fix

Two small changes in MainViewModel:

1. `SetupLiveSpellCheck` now uses `AutoDetectGoogleLanguageOrNull`, which returns null instead of the English fallback. An empty or undetectable subtitle disables live spell check instead of arming it with the wrong dictionary. Opening a file re-evaluates as before.
2. The speech to text and video OCR result paths call `SetupLiveSpellCheck()` after inserting their content, the same way opening a file does.

A side effect worth noting: a fresh English transcription now gets live spell check enabled correctly, which previously only happened by accident of the English fallback.

### Testing

Verified on macOS (Apple Silicon), all three directions:

- Arabic transcription (MLX Whisper): no false underlines anywhere.
- Opening an English subtitle file: live spell check works as before, real typos underlined.
- English transcription: live spell check now activates on the result.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -6003,6 +6003,10 @@ public partial class MainViewModel :
             SetSubtitles(_subtitle);
             SelectAndScrollToRow(0);
             ShowStatus(string.Format(Se.Language.Main.TranscriptionCompletedWithXLines, _subtitle.Paragraphs.Count));
+
+            // The subtitle content (and so its language) just changed without a
+            // file open, so re-evaluate the live spell check dictionary.
+            SetupLiveSpellCheck();
         }
         else if (result.OkPressed && result.IsBatchMode && result.BatchItems.Count == 1)
         {
@@ -6088,6 +6092,10 @@ public partial class MainViewModel :
             SetSubtitles(_subtitle);
             SelectAndScrollToRow(0);
             ShowStatus(string.Format(Se.Language.Video.VideoOcr.LinesFoundX, _subtitle.Paragraphs.Count));
+
+            // The subtitle content (and so its language) just changed without a
+            // file open, so re-evaluate the live spell check dictionary.
+            SetupLiveSpellCheck();
         }
     }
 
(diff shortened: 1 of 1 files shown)


muaz978:
Screenshot 2026-07-09 at 1 20 13 AM
Screenshot 2026-07-09 at 1 20 26 AM

muaz978:
@niksedk would you have a look at this one when you get a chance?

It is still clean and mergeable against main, and it is a small change: two lines in `SetupLiveSpellCheck` plus a call after the speech to text and video OCR result paths, 19 added and 2 removed in `MainViewModel.cs`.

The short version: live spell check picks its dictionary by auto detecting the subtitle language, but the helper it used falls back to English when the subtitle is empty. Setup runs at startup before any file is open, so the checker was armed with the English dictionary, and content arriving without a file open (a speech to text transcription, a video OCR result) never re-evaluated it. An Arabic transcription therefore came back with every word underlined.

The fix uses the variant of the helper that returns null instead of falling back to English, so an empty subtitle leaves live spell check off rather than arming it with the wrong dictionary, and the two result paths re-run the setup the same way opening a file does.

Tested on macOS in all three directions: Arabic transcription has no false underlines, opening an English subtitle behaves as before, and an English transcription now enables live spell check correctly (which previously only worked by accident of the English fallback).

No rush, and happy to adjust the approach if you would rather handle the empty subtitle case differently.

muaz978:
Thanks for merging the RTL scroll bar fix just now, much appreciated.

Whenever you have a moment, @niksedk, this one (#12289) is the last from the recent batch still open. It is still clean and mergeable against main, and small (+19/-2): two lines in `SetupLiveSpellCheck` plus a call after the speech to text and video OCR result paths, so an empty subtitle no longer arms live spell check with the English dictionary.

No rush at all, and I am happy to adjust the empty subtitle handling if you would prefer a different approach.

niksedk:
I will try to look at it soon. This just require some testing....

muaz978:
Thanks. On the testing side, I ran it on six different YouTube videos across Arabic, English and Turkish, and all of them passed: Arabic transcriptions no longer come back with every word underlined, English still behaves as before, and an English transcription now enables live spell check correctly. I only attached two of the six as examples above to keep it short.

Happy to share more captures or a specific case if that would help.

Issue #12285, opened 2026-07-08
[Feature request] AI assistant for the subtitle text box (quick actions with context, user's own provider)


muaz978:
### The idea

An optional AI assistant for the subtitle text box: a small icon next to the edit box that opens a set of quick actions for the current line, using the surrounding lines as context (a few lines before and after, so the model understands who is speaking and what is going on).

Quick actions a subtitler needs constantly while working:

- Explain or research a word (the selected text) or the whole line: meaning, idiom, cultural reference
- Rephrase the line to fit the reading speed and line length limits Subtitle Edit already computes, preserving the meaning; nothing else does this context aware
- Fix grammar and punctuation
- Adjust tone (formal or colloquial, or match the style of the earlier lines)
- A free question box for anything else about the line

Suggestions are never applied automatically: the user sees the result and applies it with a button, through the normal undo path.

### How it would connect, without new infrastructure

Two complementary paths, both opt in and both without any bundled key or service:

1. **The existing auto translate providers.** Subtitle Edit already stores API keys and endpoints for OpenAI, Anthropic compatible servers, OpenRouter, and local OpenAI compatible servers (Ollama, LM Studio). The assistant would reuse exactly those settings, so anyone with auto translate configured has the assistant working, including fully offline with a local model.

2. **An MCP server (optional, separate follow up).** MCP is the open standard that lets AI applications control other programs. If Subtitle Edit exposes its editing surface as MCP tools (read lines with context, set text of line N, style operations, reading speed checks), then users of subscription based AI apps (Claude Desktop or Claude Code with a Claude subscription, ChatGPT desktop, or any local MCP client) can work on the open subtitle conversationally with the account they already pay for, no API key needed. Privacy stays in the user's hands: only what the user asks about leaves the machine, and nothing at all with a local model.

### Why it fits Subtitle Edit

The project already embraces AI where it helps (Whisper transcription, LLM auto translate), and this applies the same philosophy to the second half of a subtitler's day: not producing the raw text, but polishing it line by line. The UI cost is one small icon; everything else lives in a flyout.

I would be happy to build this and submit it in reviewable stages (quick actions first, MCP server as a separate follow up), with screenshots and real world testing on long form Arabic and Turkish material, if there is interest.

niksedk:
SE has a "Tools" -> "AI review" which sounds about the same.
We could make an optional button for the text box (like auto-brak/unbreak/italic) with the robot icon.
Should use the same engines and models as AI review.

[...]

PR #12303, opened 2026-07-09
Add an AI assistant button to the subtitle text box (#12285)

muaz978:
Implements the text box AI assistant from #12285, following your suggestion to make it an optional text box button (like auto-break / unbreak / italic) with a robot icon that reuses the same engines and models as Tools -> AI review.

### What it does

A robot icon button next to the subtitle text box opens an assistant for the current line:

- Quick actions: Fix errors (spelling, grammar, punctuation); Fit reading speed (rephrase shorter to the line length and CPS limits); More formal; More casual.
- A free question box for anything else about the line.
- The lines before and after the current one are sent as read-only context, so the model can resolve pronouns, speaker and tone without changing them.
- Nothing is applied automatically: the suggestion appears in an editable box and is written to the line only when the user presses Apply, through the normal edit and undo path.

### How it connects

It reuses `AiReviewClient` and the `Tools -> AI review` settings (Ollama, llama.cpp, or an OpenAI compatible server), so there is no new configuration and no new dependency. The AI review client gained one optional argument so the assistant can request a plain text answer instead of a JSON object; AI review's behavior is unchanged.

The button is on by default and can be hidden in Options -> Settings (Text box: show AI assistant button).

### Notes

New UI strings live in a small `LanguageAiAssistant` section; the English defaults are in the class so untranslated languages fall back cleanly. Tested end to end on macOS against a local OpenAI compatible server, including Arabic lines with surrounding context. Screenshots below.

### DIFF
--- a/src/ui/DependencyInjectionExtensions.cs
+++ b/src/ui/DependencyInjectionExtensions.cs
@@ -379,6 +379,7 @@ public static class DependencyInjectionExtensions
         collection.AddTransient<FindViewModel>();
         collection.AddTransient<FixCommonErrorsProfileViewModel>();
         collection.AddTransient<Features.Tools.AiReview.AiReviewViewModel>();
+        collection.AddTransient<Features.Main.AiAssistant.AiAssistantViewModel>();
         collection.AddTransient<Features.Tools.AiReview.AiReviewPromptViewModel>();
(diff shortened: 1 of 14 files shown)


muaz978:
Screenshot 2026-07-09 at 2 43 43 PM
Screenshot 2026-07-09 at 2 44 08 PM

PR #12306, opened 2026-07-09
Re-apply the layout direction live when the interface language changes

muaz978:
Following the suggestion in this PR's discussion, this now resets the layout when the interface language switches between a right to left and a left to right language, instead of asking the user to restart. The restart message the branch originally added is removed.

### What changes

Changing the interface language already reloads the language and rebuilds the layout. What it did not do was change the layout *direction*, so the subtitle list and the video stayed on the side the previous language used until the app was restarted. This applies the new direction to the window and rebuilds the layout under it, so the panels move to the correct side immediately.

Three details were needed to make the live reset behave exactly like a restart:

1. **Mirror condition.** The window mirrors only for a right to left interface language that also has right to left mode enabled, which is what a restart produces. Right to left mode itself stays the user's own choice from the Edit menu. Previously, switching to a left to right interface language while an Arabic subtitle was open silently turned right to left mode off and it had to be re-enabled by hand.

2. **Video and waveform.** Both keep their own left to right flow direction (#12304), so neither is mirrored when the window is.

3. **Video position.** `LoadLanguage` already calls `SetLayout`, which builds a fresh video player positioned at the beginning. The position is therefore captured *before* the language is loaded and restored once the new layout is in place, so the video stays where the user left it. The restore waits for the player to report a duration before seeking, because the player initializes lazily on its first render pass and drops a seek issued before then.

### Testing

Tested on macOS with an Arabic subtitle and a video loaded, switching Arabic to English and back:

- The subtitle list, video and waveform move to the correct side immediately, with no restart.
- The video and waveform are not mirrored in either direction.
- Right to left mode survives the switch and does not need re-enabling.
- The video keeps its position and continues to play and seek normally after the switch.
- Playback speed and the selected audio track are preserved, as they already were for a layout change.

### DIFF
--- a/src/ui/Assets/Languages/Arabic.json
+++ b/src/ui/Assets/Languages/Arabic.json
@@ -459,7 +459,6 @@
     "replace": "استبدال",
     "replaceWith": "استبدال بـ",
     "requiresRestart": "يتطلب إعادة التشغيل",
-    "languageChangedRestartInfo": "تم تغيير لغة الواجهة. يُرجى إعادة تشغيل Subtitle Edit حتى تُطبَّق جميع التغييرات بالكامل.",
     "reset": "إعادة تعيين",
     "resizePanels": "تغيير حجم اللوحات",
     "resolution": "الدقة",
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -9219,10 +9219,28 @@ public partial class MainViewModel :
     [RelayCommand]
     private async Task CommandShowSettingsLanguage()
     {
+        // The layout direction (which side the subtitle list and video sit on)
+        // follows the interface language, so note it before the change.
+        var wasRightToLeft = Se.Settings.General.IsLanguageRightToLeft();
         var viewModel = await ShowDialogAsync<LanguageWindow, LanguageViewModel>();
         if (viewModel.OkPressed && viewModel.SelectedLanguage != null)
         {
+            // Note where the video is before the language is loaded: loading it
+            // rebuilds the layout, which builds a fresh video player starting at the
+            // beginning. Read afterwards the position is always zero, so it has to be
+            // taken here to be restored once the new layout is in place.
+            var videoPosition = GetVideoPlayerControl()?.VideoPlayer?.Position ?? 0;
+
             await LoadLanguage(viewModel.SelectedLanguage.FileName);
+
+            // Strings update live already. When the new language flips the layout
+            // direction (between a right to left and a left to right language),
+            // re-apply the direction live so the subtitle list and video move to the
+            // correct side at once, instead of asking the user to restart.
+            if (Se.Settings.General.IsLanguageRightToLeft() != wasRightToLeft)
+            {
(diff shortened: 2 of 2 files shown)


niksedk:
What UI elements are left untranslated?

muaz978:
Good question, and it made me realize the message was mislabeled. Nothing is left untranslated: every interface string updates live on a language change.

What does not fully apply live is the overall layout direction. When you switch between a left to right and a right to left interface language, the strings flip immediately, but the side that the subtitle list and the video panel sit on only fully updates after a restart. So it is the layout direction, not any missing translation.

I have reworded the message to say that explicitly, and gated it so it only appears when the direction actually changes (switching between a right to left and a left to right language), not on every language change. If you would prefer, the alternative is to make the direction re-apply live on a language change instead of prompting; I kept the prompt for now because a live direction switch of the docked video and waveform is the part that has been unreliable, but I am happy to go that route if you want.

niksedk:
Perhaps resetting the layout is a better option when switching between RTL/LTR?

muaz978:
Sounds good, I will go with resetting the layout.

On a language change that flips the direction (right to left to left to right, or the other way), I will re-apply the layout direction live using the same reset the window already uses for the Edit menu right to left toggle, so the subtitle list and the video panel move to the correct side immediately and no restart is needed. The restart note then goes away entirely.

I will make it idempotent so it only re-lays-out when the direction actually changes, and I will verify a right to left interface language (Arabic) still flips cleanly both ways with no double flip. Pushing the update shortly.

muaz978:
Thanks, resetting the layout was the better option. I have pushed that instead, and dropped the restart message and its Arabic translation. The PR title and description are updated to match.

Rather than showing a note, the layout direction is now applied to the window and the layout is rebuilt under it, so the subtitle list and the video move to the correct side at once. Getting it to behave exactly like a restart needed three things, which may be worth knowing about:

**Right to left mode was being turned off.** The first version keyed the mirroring off the interface language alone. That meant switching to a left to right interface language while an Arabic subtitle was open silently cleared the user's right to left mode, which then had to be re-enabled from the Edit menu. The window now mirrors only when the interface language is right to left *and* right to left mode is on, which is the state a restart produces, and the mode itself is left alone as the user's own choice.

**The video and waveform must not mirror.** They keep the left to right flow direction they were given in #12304, so the picture, the overlay and the waveform labels render correctly whichever way the window is mirrored.

**The video jumped back to the start.** This was the interesting one. `LoadLanguage` already calls `SetLayout`, so by the time the direction is applied the video player has already been rebuilt and is sitting at position zero. Reading the position at that point returns zero, so restoring it was a no-op. The position is now captured before the language is loaded and restored after the new layout is in place. The restore also waits for the player to report a duration before seeking, since the player initializes lazily on its first render pass and quietly drops a seek issued before then.

Tested on macOS with an Arabic subtitle and a video loaded, switching Arabic to English and back: the panels move immediately, nothing is mirrored that should not be, right to left mode survives, and the video keeps its position and still plays and seeks normally. Playback speed and the selected audio track are preserved as before.

Happy to adjust anything if you would rather it be structured differently.

PR #12307, opened 2026-07-09
MLX Whisper: forward advanced options and window long files without VAD (#12287)

### LINKED ISSUE #12287: MLX Whisper: engine notes say parameters are "passed through", but the helper script silently drops them (e.g. --condition-on-previous-text)
**Environment:** Subtitle Edit 5.1.0-beta12 [replace with your exact beta from Help/About], macOS Tahoe 26.5.1, MacBook Air 15" (M4, 2025, 32 GB RAM), mlx-whisper 0.4.3 via pipx (detected correctly since the shebang fix — thanks for that)

**Issue:**
The MLX Whisper engine notes state: "The command line parameters box is passed through to the helper script." In practice the helper script (mlx_whisper_transcribe.py) defines a narrow argument list, so anything else typed into the box is dropped with a console warning:

    warning: ignoring unrecognized argument(s): --condition-on-previous-text False

The underlying mlx_whisper CLI does support --condition-on-previous-text, so the limitation is only in the wrapper.

**Why it matters:**
--condition-on-previous-text False is the standard mitigation for Whisper's two well-known long-file failure modes: repetition loops and hallucinated subtitle-credit lines. Both are especially frequent in Turkish (training-corpus artifacts like "Altyazı M.K."). On the same machine, same kind of material (Turkish movies, ~90-130 min):

- whisper.cpp engine with the equivalent option (-mc 0): no loops, clean output.
- MLX Whisper engine (option silently dropped): hallucinated credit lines started appearing after ~60 minutes in low-speech regions, and there is currently no way to counter this on the MLX path.

**Suggestion (any of these would resolve it):**
1. Forward unrecognized box parameters to mlx_whisper's transcribe options, matching the documented "passed through" behavior; or
2. Add --condition-on-previous-text to the wrapper's argument list as a supported option; or
3. If the restriction is intentional (I notice that, unlike other engines, the advanced parameters dialog shows no parameter reference for MLX — which suggests it may be), update the engine notes to list what the box actually supports instead of "passed through", and ideally have the wrapper fail loudly rather than warn silently.

**Side note (say if you want this as a separate issue):** in a run without VAD available, the wrapper produced no progress output at all for over an hour on an 88-minute file. A periodic progress line (or forwarding mlx_whisper's own verbose output) would help users distinguish "working" from "hung".

muaz978:
Fixes #12287.

### Problem

The MLX Whisper engine notes said the command line parameters box was passed through to the helper, but the helper only accepted its own narrow argument list and dropped anything else with a warning. So options such as `--condition-on-previous-text False`, the standard mitigation for Whisper's repetition loops and hallucinated subtitle credit lines on long low-speech files, had no effect on the MLX path.

### Fix

The helper now forwards any recognized `mlx_whisper.transcribe` option from the box. Option names and value types are read from the installed mlx-whisper version's own `transcribe` signature and `DecodingOptions`, so nothing is hard coded: values are coerced to the correct type (bool, int, float, tuple), unsupported options on older versions are reported rather than dropped, and options Subtitle Edit already controls (language, task, word timestamps, initial prompt) are refused to avoid a duplicate keyword crash. User options override the helper's own defaults.

A quick check with `--condition-on-previous-text False --temperature 0 --no-speech-threshold 0.5 --totally-bogus-flag xyz` prints:

```
Applying advanced options: condition_on_previous_text=False, temperature=(0.0,), no_speech_threshold=0.5
warning: ignoring unrecognized argument(s): --totally-bogus-flag
```

### No VAD progress

The reporter also noted that without the optional VAD (faster-whisper), a long file was decoded in a single call that printed nothing until it finished, which looks hung for an hour on a 90 minute file, and let repetition loops spread across the whole recording. Without VAD the audio is now decoded in fixed windows with the previous window's tail carried for context, so the console shows steady progress and a failure stays local to its window.

The engine notes are updated to describe the forwarding accurately and list the most useful options.

### Confirmed by the reporter

On an 88 minute Turkish file with `--condition-on-previous-text False`: credit line hallucinations significantly reduced, cleaner output, and runtime dropped from about 40 to about 27 minutes (same 93 VAD chunks, directly comparable).

### DIFF
--- a/src/ui/Assets/SpeechToText/MLXWhisper.txt
+++ b/src/ui/Assets/SpeechToText/MLXWhisper.txt
@@ -22,18 +22,33 @@ option to the decoder. The option name is matched against the installed
 mlx-whisper version, so unsupported options are reported rather than silently
 dropped, and the value is coerced to the right type automatically.
 
-Useful options:
-  --condition-on-previous-text False   turn off conditioning; the standard
-                                       mitigation for repetition loops and
-                                       hallucinated credit lines on long files
-  --temperature 0                      disable the temperature fallback
-  --no-speech-threshold 0.6            silence sensitivity
-  --logprob-threshold -1.0             confidence threshold for the fallback
-  --compression-ratio-threshold 2.4    repetition detector threshold
+Parameters (mlx-whisper transcribe options; type them into the box):
 
-The helper also has its own options, for example --beam-size 5 for beam search.
-Options that Subtitle Edit already controls (language, task, word timestamps, the
-initial prompt) are ignored if typed here.
+  --condition-on-previous-text N       [true   ] feed the previous window's text as
+                                                  context; set False to stop repetition
+                                                  loops and hallucinated credit lines
+  --temperature N                      [0       ] sampling temperature; a single 0
+                                                  disables the temperature fallback
+  --compression-ratio-threshold N      [2.40   ] treat a window as repetitive above this
+  --logprob-threshold N                [-1.00  ] retry a window below this confidence
+  --no-speech-threshold N              [0.60   ] silence sensitivity
+  --best-of N                          [5      ] candidates to sample on a retry
+  --beam-size N                        [0       ] beam search width (0 = greedy)
+  --patience N                         [       ] beam search patience
+  --length-penalty N                   [       ] beam search length penalty
+  --suppress-tokens LIST               [-1     ] token ids to suppress (-1 = default set)
+  --suppress-blank N                   [true   ] suppress blank outputs at the start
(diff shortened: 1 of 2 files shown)


niksedk:
Nice. Could new add the cmd line parameters to the mlx whisper txt file? (like the other Whisper engines have)

muaz978:
Done. Added a parameter reference to MLXWhisper.txt in the same style as the other Whisper engines: a table of the forwardable mlx-whisper transcribe options with their defaults and a short description (condition-on-previous-text, temperature, the thresholds, best-of, beam-size, suppress-tokens, and so on), plus a note that the options are validated against the installed mlx-whisper version.

PR #12325, opened 2026-07-10
Add generic OpenAI-compatible engine to auto-translate

### LINKED ISSUE #12324: Feature request: Support Xiaomi MIMO model or custom OpenAI-compatible API endpoint
Hi, thanks for this great tool! I have a feature request...I would like to suggest supporting the MIMO model in Auto Translate. MIMO is a large language model from Xiaomi – it offers good performance at a low cost, which makes it particularly suitable for subtitle translation tasks.

If integrating a specific model is too complicated, I would also appreciate a more flexible solution: adding a custom service option where users can manually configure any model that exposes an OpenAI‑compatible API endpoint. This would allow us to bring in various models by ourselves, including MIMO.

Thank you for considering this!

niksedk:
Fixes #12324.

Adds an **"OpenAI Compatible API"** engine to the auto-translate window, so any service exposing an OpenAI-style `chat/completions` endpoint can be used without SE needing a dedicated engine — e.g. Xiaomi MIMO, vLLM, LiteLLM/gateways, or any self-hosted server. This mirrors the "OpenAI Compatible Server" option already available in the speech-to-text window.

### What the user configures
- **URL** — full endpoint URL (defaults to `http://localhost:8000/v1/chat/completions`)
- **API key** — optional; sent as `Authorization: Bearer …` when set
- **Model** — free text, optional: single-model servers (llama.cpp, vLLM with one model) ignore it; hosted providers require it, and their error body is surfaced in the existing translation error dialog
- **Prompt** — editable via the engine settings dialog like the other LLM engines

### Implementation
- New `OpenAiCompatibleTranslate` in libse, modeled on `ChatGptTranslate`/`LmStudioTranslate` and reusing the shared ChatGPT helpers (`ListLanguages`, `RemovePreamble`, `FixNewLines`, `DecodeUnicodeEscapes`)
- New settings in `ToolsSettings` + `SeAutoTranslate` (persisted like the other engines)
- Wired into `AutoTranslateViewModel` (engine list, load/save, engine panel) and `TranslateSettingsViewModel` (prompt edit + validation)

### Testing
- Verified end-to-end against a live OpenAI-compatible server (Ollama's `/v1/chat/completions`): en→da translation returns correct multi-line output, and the empty-model case surfaces the server's "model is required" error via `Error` as expected
- `dotnet build src/ui` clean

### DIFF
--- a/src/libse/AutoTranslate/OpenAiCompatibleTranslate.cs
+++ b/src/libse/AutoTranslate/OpenAiCompatibleTranslate.cs
@@ -1,0 +1,110 @@
+using Nikse.SubtitleEdit.Core.Common;
+using Nikse.SubtitleEdit.Core.Settings;
+using Nikse.SubtitleEdit.Core.SubtitleFormats;
+using Nikse.SubtitleEdit.Core.Translate;
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Nikse.SubtitleEdit.Core.AutoTranslate
+{
+    /// <summary>
+    /// Generic engine for any service exposing an OpenAI-compatible "chat/completions"
+    /// endpoint (vLLM, Xiaomi MIMO, Together, etc.) - URL, API key, and model are all
+    /// user-configured, so providers SE has no dedicated engine for still work (#12324).
+    /// </summary>
+    public class OpenAiCompatibleTranslate : IAutoTranslator, IDisposable
+    {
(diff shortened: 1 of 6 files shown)


niksedk:
Second commit extends this to harden all engines: a malformed URL (e.g. missing `http://`) previously crashed the auto-translate window with an uncaught `UriFormatException` from the engine's `Initialize()`. Now the URL is validated as absolute http/https up front (Papago excluded — its URL field is really a client ID), the empty-URL message no longer says "requires an API key", and `translator.Initialize()` has a catch-all that restores the window state and shows the error instead of crashing.

Issue #10209, opened 2026-02-15
.vtt error issues with line% and position% force locking themself to grid and others weirdness


Foxaryse:
if i open with notepad++ the .vtt file it shows in the correct possition BUT when i select the line in Subtitle Editor with F2 inside [Source View] it doesn't show the line and position while showing in the correct place inside the Video Player
<img width="2436" height="717" alt="Image" src="https://github.com/user-attachments/assets/897bc10b-f17b-49ba-b486-7c1f0417f432" />

BUT, if i were to add line and position inside [Source View] everything goes apes shit and it's locking itself to the alignment grid for no apparent reasons.
<img width="1624" height="724" alt="Image" src="https://github.com/user-attachments/assets/376b72b5-b152-4e9c-9a47-ca0725905f23" />

since it locked itself to the alignement grid i've removed all default values in the alignement but it just doesn't care at all.
<img width="554" height="409" alt="Image" src="https://github.com/user-attachments/assets/bda03a03-902c-4871-8c80-54e81ac83b7c" />

Foxaryse:
i've just realised i was running on a old version and updated to 4.0.15 and the problem still presist

PR #12340, opened 2026-07-10
Fix WebVTT cue setting preservation

pdjdev:
## Problem

WebVTT cue settings could be silently lost or changed after loading and saving a file in Subtitle Edit.

In particular, commonly used settings such as `align:left` were not preserved. Cue positions that did not exactly correspond to Subtitle Edit's configured ASS alignment presets were inferred from broad coordinate ranges and converted to a nearby `{\an1}`–`{\an9}` tag.

For example, saving this file again via `Save As...` without any changes:

```vtt
WEBVTT

00:00:20.943 --> 00:00:21.730
♪ If this is it ♪

00:00:21.731 --> 00:00:23.356 position:3% line:5% align:left
🎵 If This Is It
By Huey Lewis And The News

00:00:23.357 --> 00:00:25.900
♪ Please let me know ♪
```

could produce:

```vtt
WEBVTT

00:00:20.943 --> 00:00:21.730
♪ If this is it ♪

00:00:21.731 --> 00:00:23.356 position:20% line:20%
🎵 If This Is It
By Huey Lewis And The News

00:00:23.357 --> 00:00:25.900
♪ Please let me know ♪
```

This both removes `align:left` and changes the author's intended position.

## Cause & Fix

The previous `GetPositionInfo()` logic converted arbitrary `position` and `line` values to ASS alignment tags based on coordinate ranges. This is a lossy conversion because WebVTT cue positioning is more expressive than the nine discrete ASS alignment positions.

This PR replaces that inference with `GetAssAlignmentTagFromCueSettings()`.

* Cue settings are converted to `{\an1}`–`{\an9}` only when the extracted cue settings exactly match one of the configured `SubtitleSettings.WebVttCueAn1`–`WebVttCueAn9` values.
* Non-matching cue settings are preserved instead of being approximated to an ASS alignment tag.
* When an ASS alignment tag is added in Subtitle Edit, its configured `WebVttCueAn*` value overrides existing cue settings during WebVTT export.
* Cue settings that were converted to an ASS alignment tag are no longer retained as stale metadata. Removing the tag therefore also removes the corresponding cue settings.
* Applied the same behavior to regular WebVTT, numbered WebVTT, and MP4 `vttc` parsing.
* Removed the range-based `WebVTT.GetPositionInfo()` method.

For example, when an editor adds `{\an7}` to a cue:

```text
{\an7}🎵 If This Is It
By Huey Lewis And The News
```

and `SubtitleSettings.WebVttCueAn7` is configured as:

```text
position:20% line:20%
```

the exported WebVTT cue uses that configured placement:

```vtt
00:00:21.731 --> 00:00:23.356 position:20% line:20%
🎵 If This Is It
By Huey Lewis And The News
```

## Test

Added regression coverage for:

* preserving non-preset cue settings, including `align:left`;
* converting exact preset matches to ASS alignment tags;
* removing a loaded ASS alignment tag without retaining stale cue settings;
* ASS alignment tags overriding existing cue settings on save;
* preserving non-preset cue settings in numbered WebVTT files.

Tested on Windows 11 x64 with a local Release build.

`WebVttTest`: 14/14 passed.

## Note

WebVTT and ASS/SSA have different styling and positioning models, so mapping between them has inherent limitations. A dedicated styling system and editor UI for WebVTT could be a useful future improvement.

### DIFF
--- a/src/libse/ContainerFormats/Mp4/Mp4Parser.cs
+++ b/src/libse/ContainerFormats/Mp4/Mp4Parser.cs
@@ -706,11 +706,17 @@ namespace Nikse.SubtitleEdit.Core.ContainerFormats.Mp4
         private void AddVttParagraph(double timeTotalMs, string payload, double before, string style)
         {
             var p = new Paragraph(payload, before, timeTotalMs);
-            var positionInfo = WebVTT.GetPositionInfo(style);
+            var positionInfo = WebVTT.GetAssAlignmentTagFromCueSettings(style);
             if (!string.IsNullOrEmpty(positionInfo))
             {
                 p.Text = positionInfo + p.Text;
-                p.Extra = style;
+            }
+
+            p.Style = string.IsNullOrEmpty(positionInfo)
+                ? WebVTT.GetPositionInfoRaw(style)
+                : string.Empty;
+            if (!string.IsNullOrEmpty(p.Style) || !string.IsNullOrEmpty(positionInfo))
+            {
                 VttcSubtitle.Header = "WEBVTT";
             }
             VttcSubtitle.Paragraphs.Add(p);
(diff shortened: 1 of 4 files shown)


pdjdev:
Since #13260 has now been merged and addresses #10209 with a less invasive approach, I'll close this PR to avoid conflicting implementations.

PR #12401, opened 2026-07-12
Perf: remove steady-state allocations in GetFastHash and CalcFactory lookup

niksedk:
Two micro-optimizations on paths that run constantly, follow-up to #12400.

## 1. `GetFastHash` / `GetFastHashOriginal`: no more per-call header copy

These run every 250 ms (undo change detection via `UndoRedoManager.CheckForChanges`) plus every 400 ms (title dirty-star / auto-save; twice with the original column visible) — roughly 6–7 full passes per second, forever, even when the app is idle. Each call allocated:

- `_subtitleFileName + SelectedEncoding.DisplayName` — a throwaway concat string
- `Header?.Trim()` / `Footer?.Trim()` — and since the SE5 editor normalises subtitles to ASSA internally, `Header` is a multi-KB `[V4+ Styles]` block that ends in a newline, so `Trim()` **copied the entire header on every call**.

Now the components are hashed separately and Header/Footer are trimmed as spans: `string.GetHashCode(header.AsSpan().Trim())` — zero allocation. `string.GetHashCode(ReadOnlySpan<char>)` returns the same value as the string's own `GetHashCode()`, and these hashes are only ever compared against other hashes computed in the same process (undo baseline, dirty-star, auto-save settle), so semantics are unchanged.

## 2. `CalcFactory.MakeCalculator`: memoize the strategy resolution

`MakeCalculator` sits on the path of **every** `CountCharacters` call — grid CPS/line-length columns on repaint, the waveform paragraph footer, per-keystroke text info, Netflix checks — and resolved the strategy by comparing `GetType().Name` against up to 11 entries on each call. The strategy is a settings value that practically never changes at runtime.

Added a one-entry memo: an immutable `(strategy, calculator)` pair swapped by reference (volatile read, no locking), so a hit costs a reference read plus one string compare. First call and strategy changes still fall through to the existing scan.

## Test plan

- [x] `dotnet build` libse (no-incremental) + UI — 0 warnings, 0 errors
- [x] LibSETests: 485/485 pass
- [x] Hash semantics: values still compared only against same-process hashes computed by the same function (undo baseline, dirty-star, auto-save)

### DIFF
--- a/src/libse/Common/TextLengthCalculator/CalcFactory.cs
+++ b/src/libse/Common/TextLengthCalculator/CalcFactory.cs
@@ -20,20 +20,48 @@ namespace Nikse.SubtitleEdit.Core.Common.TextLengthCalculator
             new CalcIgnoreArabicDiacriticsNoSpace(),
         };
 
+        // Memo of the last strategy resolution. A single immutable pair swapped by reference
+        // keeps the lookup thread-safe without locking; the strategy is a settings value that
+        // practically never changes at runtime, so this hits on every call but the first.
+        private sealed class ResolvedCalculator
+        {
+            public readonly string Strategy;
+            public readonly ICalcLength Calculator;
+
+            public ResolvedCalculator(string strategy, ICalcLength calculator)
+            {
+                Strategy = strategy;
+                Calculator = calculator;
+            }
+        }
+
+        private static volatile ResolvedCalculator _lastResolved;
+
         public static ICalcLength MakeCalculator(string strategy)
         {
-            // Called per line on grid repaints (characters-per-second) - avoid the
-            // LINQ closure and the fallback allocation; the list entries are shared anyway.
+            // Called for every CountCharacters (grid CPS/length columns on repaint, waveform
+            // footer, per-keystroke text info), so resolving the strategy must be cheap: the
+            // memo avoids re-comparing reflection type names on every call.
+            var last = _lastResolved;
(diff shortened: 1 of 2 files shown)


niksedk:
BenchmarkDotNet numbers for both changes (.NET 10, Release, `MemoryDiagnoser`, i7-1065G7):

**GetFastHash prefix** (filename+encoding+header/footer hashing, ~6 KB ASSA header with trailing newline — the shape the editor's ASSA normalisation produces):

| Method | Mean | Allocated |
|---|---:|---:|
| Old (concat + `Trim()`) | 3,568 ns | 6,256 B |
| New (span trim hash) | 2,980 ns | **0 B** |

~16 % faster per call, and at ~6–7 calls/second the steady-state garbage (~44 KB/s from an idle app, more with larger headers) drops to zero.

**CalcFactory.MakeCalculator**:

| Strategy | Old (type-name scan) | New (memoized) |
|---|---:|---:|
| `""` (default setting — no match, walks all 11) | 49.3 ns | 2.6 ns (**19×**) |
| `"CalcNoSpaceCpsOnly"` (early match) | 12.3 ns | 3.6 ns (3.4×) |

Small per call, but it multiplies across every `CountCharacters` in the app (grid CPS/length columns per row on repaint, waveform footer, per-keystroke text info) — and the default-settings case was the worst case for the old scan.

PR #12409, opened 2026-07-13
Add find history dropdown, text-length sorts, and sort shortcuts

### LINKED ISSUE #12407: [SE5] Sorting features missing
I wish to have back the most recent find text string like the SE4 version
Image

It seems like the sort by total text length and text length in one line is also missing. And also Shortcuts for these sorting options is missing
Image

Image # #

niksedk:
## Summary

Fixes #12407 — all three asks from the report:

**1. Most-recent-find-text dropdown (Find window).** The history was already persisted (last 10 searches in `FindService`) but only surfaced as autocomplete suggestions while typing a matching prefix, so it looked absent. A history dropdown button next to the search box now lists the recent searches SE4-style; picking one fills the search box and refocuses it. The button is disabled while the history is empty, and menu headers are wrapped in `TextBlock` so underscores in search text aren't eaten as access-key markers. This also wires up the previously dead `ShowHistoryCommand`.

**2. Sort by text total length / single line max length.** Added both to the Sort dialog's property list, computed line-break-aware via a shared helper.

**3. Shortcuts for the sort family.** SE5 only exposed sort by number/show/hide as assignable shortcuts. Added the rest of the SE4 set — duration, gap, text alphabetically, text total length, single line max length, chars/sec, words/min, number of lines, actor, style — registered in the shortcuts list and mapped in the SE4 shortcuts importer (which previously silently dropped them on import). The three existing sort commands are refactored onto a shared `SortSubtitlesBy` helper; the new ones reuse it.

New language strings default to English and fall back gracefully for existing translations.

## Testing

- `dotnet build src/ui/UI.csproj` — clean, 0 warnings/errors.


mrklhsnbrg:
Hello, sorry but I do not understand the comment 3. Shortcuts for the sort family. How can I export the SE4 sort by menus and shortcut and import back to SE5?

PR #12433, opened 2026-07-13
Fix common errors: sortable columns in fix list + Japanese translation update

### LINKED ISSUE #12431: Fix common errors: unable to change back to order by action after ordering by #
To reproduce
- go to fix common errors
- click on #
- try change the ordering by clicking on "action"

Also, should "apply" (as in the first column of the list) also function as changing the order?

niksedk:
## Summary
- Fixes #12431: in the Fix common errors fix list, header clicks on Apply/Action/Before/After were silently ignored - template columns have no Binding for the DataGrid sort machinery to inspect, so once sorted by # there was no way back to action order. Adds explicit `SortMemberPath` to all template columns (including the rules grid''s Enabled column). This also makes the Apply column sortable, answering the issue''s second question.
- Update Japanese translation for v5.1.0-rc1 shortcut UI strings - thx hisui3393 ([discussion #10742](https://github.com/SubtitleEdit/subtitleedit/discussions/10742#discussioncomment-17626853)), JSON validated against the existing file structure.
- Starts the v5.1.0-rc2 change-log section (rc1 was tagged today).

## Test plan
- [x] Build succeeds
- [x] Japanese.json parses as valid JSON with the same top-level structure as the previous file


GrampaWildWilly:
This sounds a whole lot like the suggestion I've been making to allow sorting on more columns in the Shortcuts dialog.  Not to mention allowing changes to the order of the columns.  I suspect you could apply much the same code to my suggestion as what you've done here.

Issue #12377, opened 2026-07-11
"Select all" in "Fix common errors" should be "Select all in current category"


subcor:
Unless I'm missing something there is no quick way to select a singular category in fix common errors.
So what I suggest Is to make "select all" and "invert selection" apply in the current category. Also making select all a toggle (un/select all).
and/or
Changing categories would change the pool of errors too, so "Apply selected fixes" would only be applied in current category.
Like this:

<img width="1028" height="795" alt="Image" src="https://github.com/user-attachments/assets/9927dcf8-6964-4936-b44f-fd4c9a49bfc5" />

muaz978:
Good catch, you are right. Select all and Invert selection were applying across every category, and the summary counted the whole list too, so filtering to one category did not scope them.

I have scoped Select all, Invert selection, and the "N fixes, M selected" summary to the active category chip, and the summary now refreshes when you switch categories. So filtering to Fix short gap and clicking Select all selects only those 694, and the count reads 694 fixes, 694 selected, matching your mockup. The All chip still covers everything, and Invert works as a deselect all within a category when it is fully selected.

Applying fixes is unchanged: whatever you have checked across categories still all applies, so you can still build a selection spanning several categories and apply it in one go.

If you would like to try it before it is merged, here is a portable Windows test build with the change (unzip and run the executable, no install):
https://github.com/muaz978/subtitleedit/releases/tag/win-test-fce-select-all

Thanks for the clear report and the mockup.

subcor:
Thank you! I'm on Linux though, so I'm going to wait for official release then.

muaz978:
Since you are on Linux, here is a portable Linux test build with the fix, so you do not have to wait for the release:

https://github.com/muaz978/subtitleedit/releases/tag/linux-test-fce-select-all

No install needed:

```
tar -xzf SubtitleEdit-Linux-x64.tar.gz
chmod +x SubtitleEdit
./SubtitleEdit
```

It is a self-contained x64 build, so it does not need the .NET runtime. Video playback would need libmpv from your distro, but that is not needed to test this change.

Whenever you have a moment, could you try it and let us know how it behaves: open Fix common errors, go to step 2, pick a category chip (for example Fix short gap), and check that Select all and Invert selection now apply only to that category, and that the count reads that category's total (for example 694 fixes, 694 selected) rather than the whole list. A quick note back on whether it matches your mockup would be a big help.

Thanks a lot, and thanks again for the clear report.

[...]

PR #12435, opened 2026-07-13
Fix common errors: clearer step 2 selection UX (#12377)

muaz978:
Follow-up to @subcor's feedback on #12377 after the summary-count fix landed. He tested the Linux build with the fix and suggested four refinements to the step 2 selection controls:

- The **Apply selected fixes** button now shows the live count of fixes it will apply, e.g. `Apply selected fixes (706)`. Because it applies every ticked fix across all categories, not just the active tab, showing that global number removes the ambiguity he raised.
- **Select all** is now a **select all / none** toggle, scoped to the active category. The caption reads `Select none` once everything in view is ticked.
- Category tabs now show **selected / total**, e.g. `Fix short gap (12/694)`, so each tab reflects how many of its own fixes are ticked.
- **Invert selection** is removed, since the toggle now covers both select and deselect.

The toggle and the removal of Invert touch the buttons added in #12408, so happy to adjust or drop either part if you would rather keep those as they were. Screenshots below.

### DIFF
--- a/src/ui/Features/Tools/FixCommonErrors/FixCommonErrorsViewModel.cs
+++ b/src/ui/Features/Tools/FixCommonErrors/FixCommonErrorsViewModel.cs
@@ -32,7 +32,6 @@ public partial class FixCommonErrorsViewModel : ObservableObject, IFixCallbacks
     [ObservableProperty] private ObservableCollection<FixDisplayItem> _fixes;
     [ObservableProperty] private ObservableCollection<FixDisplayItem> _visibleFixes;
     [ObservableProperty] private ObservableCollection<FixFilterChip> _fixChips;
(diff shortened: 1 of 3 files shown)


niksedk:
Could you add a screenshot or two?

muaz978:
The reporter (@subcor) tested this on a portable Linux build and confirmed it, and after his feedback I also removed the now-redundant "N fixes, M selected" summary label, since the active category tab already shows selected / total and the Apply button shows the global selected count. The relevant exchange is on #12377 if useful.

Happy to add screenshots too if you would still like them, just let me know.
If he'd kindly share screenshots from the same file, that would great for a real file, as I don't have one that need fixes.

subcor:
Screenshot_20260713_225527

Issue #12051, opened 2026-07-01
SE 5.1.0 beta 4: Scroll bar anomalies


GrampaWildWilly:
I have just discovered some really weird things with the scroll bars in SOME (not all) of the various windows of SE 5.1.0 beta 4.

In the main grid, clicking in the channel of the scroll bar away from the scroll button is supposed to do the equivalent of a Page Down (or Up).  Supposed to, meaning Windows standard behavior.  Instead, clicking there does a single line scroll, same as clicking one of the arrows at either end of the scroll bar.

Same behavior in the Shortcuts dialog.

Anomalously, the Options dialog does not exhibit this behavior.  Clicking in the scroll bar does scroll a full page.

An additional standard Windows behavior involves the Shift key.  Shift+Click in the scroll bar channel is supposed to make the scroll button jump to the mouse position.  Modifying the click with Shift has no effect in SE.

I would have also tried to test horizontal scroll bars but a quick check of things didn't turn one up.

In any case, I have mentioned only 3 places where scroll bars appear in SE.  There must be dozens of other windows I have not mentioned where scroll bars appear.  You should check whether the anomalies I'm reporting occur elsewhere in the product.  Don't forget horizontal scroll bars as well as vertical scroll bars.

GrampaWildWilly:
These bugs continue to exist in beta 7.

GrampaWildWilly:
These bugs continue to exist in beta 8.

GrampaWildWilly:
These bugs continue to exist in beta 9.

GrampaWildWilly:
These bugs continue to exist in beta 11.

GrampaWildWilly:
No change up through beta 14.

muaz978:
I believe I have this fixed.

The main subtitle grid and the Shortcuts dialog are both Avalonia DataGrids. Avalonia sets the vertical scroll bar's Maximum and ViewportSize but never its LargeChange, so it stays at the default of 1: clicking the trough raised a large increment that moved the value by a single pixel, which looked like a one line scroll. The horizontal scroll bar does get a LargeChange equal to its viewport, and a plain ScrollViewer sets it too, which is why the Options dialog paged correctly while these two did not. The fix keeps the vertical LargeChange in step with the viewport, so a trough click pages the way it does elsewhere and on Windows generally.

If you would like to confirm before I open the pull request, here is a portable Windows test build with the fix (unzip and run the executable, no install needed):
[...]

PR #12438, opened 2026-07-13
Scroll bar: shift+click on the trough jumps to position (#12051 follow-up)

muaz978:
Follow-up to the trough paging fix (#12396) for the scroll bar behavior in #12051.

On Windows a plain click in the scroll bar trough pages, and shift + click jumps the thumb straight to the click position. This adds the shift + click jump to the DataGrid vertical scroll bar, so it now covers both Windows behaviors. It applies to the subtitle grid and the Shortcuts grid, since both share the same helper.

Details:
- Shift + left click in the trough sets the thumb to the click position, centering the thumb on the cursor and clamping to range. Without shift, the existing page behavior is unchanged.
- The DataGrid refreshes its rows off the scroll bar Scroll event rather than ValueChanged, so after setting the value the grid's internal ProcessVerticalScroll is invoked, the same path already used elsewhere for centering a row.

Requested by @GrampaWildWilly on #12051 as the remaining piece for full Windows scroll bar behavior.

### DIFF
--- a/src/ui/Logic/DataGridScrollBarBehavior.cs
+++ b/src/ui/Logic/DataGridScrollBarBehavior.cs
@@ -1,13 +1,18 @@
-using Avalonia;
+using System;
+using System.Linq;
+using System.Reflection;
 using Avalonia.Controls;
 using Avalonia.Controls.Primitives;
+using Avalonia.Input;
+using Avalonia.Interactivity;
+using Avalonia.VisualTree;
 
 namespace Nikse.SubtitleEdit.Logic;
 
 /// <summary>
-/// Gives a <see cref="DataGrid"/>'s vertical scroll bar a working page size so that
-/// clicking the scroll bar trough (the channel above or below the thumb) scrolls a full
-/// page, as it does everywhere else and on other platforms (issue #12051).
+/// Gives a <see cref="DataGrid"/>'s vertical scroll bar Windows-standard trough behavior:
+/// a plain trough click scrolls a full page, and shift + trough click jumps the thumb
+/// straight to the click position (issue #12051 and its follow-up).
 ///
 /// Avalonia's DataGrid sets its vertical scroll bar's Maximum and ViewportSize but never
 /// its LargeChange, so it keeps RangeBase's default of 1. A trough click raises a large
@@ -17,10 +22,17 @@ namespace Nikse.SubtitleEdit.Logic;
 /// pages correctly while the subtitle grid and the Shortcuts grid do not. This keeps the
 /// vertical LargeChange in step with the viewport so a page always matches the visible
 /// height.
+///
+/// The DataGrid hooks the scroll bar's Scroll event (not ValueChanged) to refresh its
+/// rows, so a programmatic jump also invokes the grid's internal ProcessVerticalScroll.
 /// </summary>
 internal static class DataGridScrollBarBehavior
 {
     private const string VerticalScrollBarPartName = "PART_VerticalScrollbar";
+
+    private static readonly MethodInfo? ProcessVerticalScrollMethod = typeof(DataGrid).GetMethod(
+        "ProcessVerticalScroll",
+        BindingFlags.NonPublic | BindingFlags.Instance);
 
     public static void EnableTroughPageScroll(DataGrid grid)
     {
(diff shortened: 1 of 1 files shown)


GrampaWildWilly:
Excellent!  But aren't there other windows in SE that have scroll bars?  This really should be a global fix, not just a fix for the 2 windows I use most often.  I'm not a good example of a typical user.  I actually hardly ever use SE.  I'm just geeky enough to be interested in making the program better so on those rare occasions when I do use it, I like it.  But other people use it every day, all day long, & they should be kept happy, too.

GrampaWildWilly:
Somehow this isn't in 5.1.0 RC 2.  How is that possible?

muaz978:
Right, it is not in 5.1.0 RC 2 because this PR is still open, not merged. RC 2 was built from main before this change, so it does not have the Shift+click jump yet. The scroll bar fixes that did make it into RC 2 are the earlier ones (the overlap fix and the trough paging), which are separate from this follow-up.

To try the Shift+click jump now, before it merges, please use the same portable Windows test build (it contains both this and the shortcut name-weight change):

https://github.com/muaz978/subtitleedit/releases/tag/win-test-scroll-shortcuts

Unzip and run SubtitleEdit.exe, no install, it does not affect your normal install.

Steps to test the Shift+click jump:
1. Open a subtitle with enough lines that the grid scrolls (a few hundred lines is ideal). The Shortcuts window list works too.
2. Plain click in the scroll bar trough, the empty channel above or below the thumb. It should page by one screen, same as RC 2.
3. Now hold Shift and click in the trough, near the very bottom. The thumb, and the list, should jump straight to that position instead of paging.
4. Try Shift+click at a few spots, top, middle, bottom, and confirm the thumb lands where you clicked.

A short screenshot or a quick screen capture of the Shift+click jump would be great to confirm it behaves the way you expect for full Windows scroll bar behavior.

GrampaWildWilly:
I like Copilot.  I never thought to test that.  Put up another special build & I will.

Anyway, see my resp0nse in #12439.

muaz978:
Good call, and Copilot was right. The handler was attached to the whole scroll bar, so a Shift+click on a line arrow or on the thumb would also jump, and near an arrow it would fling to min or max. I have fixed it so the jump only happens on a trough click: clicks outside the track (the arrows) and on the thumb itself (which starts a normal drag) are now ignored. Pushed to this PR.

I refreshed the same Windows test build with the fix if you want to check that exact case:

https://github.com/muaz978/subtitleedit/releases/tag/win-test-scroll-shortcuts

Unzip and run SubtitleEdit.exe, no install. Worth trying:
- Shift+click in the trough above and below the thumb -> jumps to that spot (as before).
- Shift+click directly on the thumb -> nothing jumps, it just drags normally.
- Shift+click on the up or down arrow -> no jump to min or max anymore.

Your earlier confirmation of the core trough jump still holds, this only tightens the edges. Thanks for catching it.

GrampaWildWilly:
I tried to get OBS to do a screen recording of SE but it just won't do it.  I have done it in the past but for some reason, it refuses to see SE.  It shows up in OBS as a black window.  I tried both Window capture & Display capture & it's just a black window either way.  So I can't give you the visual you want.  I did test the older version with Shift+Click on the arrows at the end of the scroll bar & indeed it did jump to either top or bottom.  I never thought to test that because Shift+Click on the arrows is not a defined behavior in Windows.  Shift+Click on the scroll bar on this page, in Firefox for example, acts the same as if I had not pressed Shift, which is to say it scrolls 1 line.

So I got your new test version & Shift+Click on the arrows now acts the same as no Shift: it scrolls one line.  And the jump scroll still works fine.  And the shortcuts are all now normal stress, not bold.  I do wish I could get OBS to cooperate.  Maybe I'll try the built in screen recorder.  I've never used that.  Let's see if I can figure it out.

GrampaWildWilly:
Typical idiotic Microsoft moronic design.  There is a directory in your User tree on C: named `Screen Recorder`.  There is also a directory named `Screen Recordings`.  Once I figured out that there are both of these & that the `Screen Recordings` directory is the one I want, I discovered that I had actually recorded several tests but didn't realize it because I thought I was supposed to look in `Screen Recorder`.  Silly me.  Anyway, I think this will suit our purposes.  The single line scrolls I did while holding down Shift.  That bit of information just can't be captured.  But for the rest, I think this is what we're looking for.


https://github.com/user-attachments/assets/cfc284ff-fdd8-4dc8-bc10-d8056515d596

GrampaWildWilly:
Also, I did mention before that there must be lots of windows in SE besides main & Shortcuts.  This correction needs to be brought to all windows, not just shortcuts.  You didn't respond to that idea.  I think it's worth considering.

muaz978:
You are right, and sorry I did not pick up on that part earlier. Right now only the main subtitle grid and the Shortcuts grid wire this behavior up, so the other grid windows still page a single line and have no Shift+click jump.

Rather than adding a call to every window one by one, the clean way is to turn this into an attached property applied through a single app level DataGrid style, so every current and future grid picks up the trough paging and the Shift+click jump automatically.

@niksedk would you be happy with that approach, applying it app wide via one DataGrid style? If so I will put up a small follow-up PR on top of what just merged. Did not want to change the behavior of every grid in the app without your say-so first.

niksedk:
> @niksedk would you be happy with that approach, applying it app wide via one DataGrid style? If so I will put up a small follow-up PR on top of what just merged. Did not want to change the behavior of every grid in the app without your say-so first.

Sure :)

muaz978:
Done, and thanks again for the nudge. niksedk was happy with applying it app wide, so I have put it up as #12478. It turns the scroll bar behavior into a single DataGrid style, so every grid window, not just the main grid and Shortcuts, now gets the trough paging and the Shift+click jump, including any grid added later.

Here is a portable Windows test build with it if you want to spot-check a few of the other windows:

https://github.com/muaz978/subtitleedit/releases/tag/win-test-scroll-app-wide

Unzip and run SubtitleEdit.exe, no install. Good windows to try, each has a grid that did not have this before: Batch convert, Compare, Multiple replace, and any of the OCR or pick-track lists. In each, a plain trough click should page and Shift+click should jump the thumb to the click position. The main grid and Shortcuts should behave exactly as before.

PR #12478, opened 2026-07-14
Apply the DataGrid scroll bar behavior app wide via one style (#12438 follow-up)

muaz978:
Follow-up to #12438, requested by @GrampaWildWilly and approved by @niksedk on that thread.

Right now the trough paging and Shift+click jump are wired into only two grids (the main subtitle grid and the Shortcuts grid), so every other grid window still pages a single line and has no Shift+click jump.

This applies the behavior app wide from one place:
- DataGridScrollBarBehavior gains an EnableTroughPaging attached property.
- A single DataGrid style in Styles.axaml sets it to True, so every current and future grid picks up the behavior automatically.
- The two explicit per-window calls are removed since the style now covers them; a wired guard prevents double attachment if the property is ever set twice.

No behavior change for the subtitle and Shortcuts grids, they just get it via the style now instead of a direct call. Every other grid window (Batch convert, Compare, Multiple replace, OCR, and the rest) gains trough paging and the Shift+click jump.

Testing scope: this compiles clean and reuses the exact mechanism already shipped and confirmed on the subtitle and Shortcuts grids in #12438; I have not click-tested all of the other grid windows individually.

### DIFF
--- a/src/ui/Features/Main/Layout/InitListViewAndEditBox.cs
+++ b/src/ui/Features/Main/Layout/InitListViewAndEditBox.cs
@@ -97,7 +97,8 @@ public static partial class InitListViewAndEditBox
             FontSize = Se.Settings.Appearance.SubtitleGridFontSize,
         };
 
-        DataGridScrollBarBehavior.EnableTroughPageScroll(vm.SubtitleGrid);
+        // Trough paging and shift+click jump now come from the app-wide DataGrid style
+        // (DataGridScrollBarBehavior.EnableTroughPaging in Styles.axaml), so no per-grid call here.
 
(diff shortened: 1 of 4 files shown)


muaz978:
Good catch from Copilot, the comment was overclaiming. The wired guard only stops a repeat set of the attached property from subscribing to TemplateApplied a second time, which would stack a second set of scroll bar handlers. It does not guard re-templating, and does not need to: TemplateApplied is subscribed once and simply re-runs on each new template, wiring that template's fresh scroll bar a single time, while the discarded template's scroll bar and its handlers are dropped with it. I have corrected the comment to say exactly that.

GrampaWildWilly:
I had a suspicion something like this needed to happen.  Glad to see it.

PR #12510, opened 2026-07-15
Allow assigning shortcuts to waveform seek video back/forward buttons

### LINKED ISSUE #12350: [SE 5.1.0 RC3] Feature Request: Allow assigning shortcuts to "Seek video backward/forward" toolbar buttons
Please refer to my comment below.

~In SE 4.0.16, there are four frame-seeking shortcuts. In SE5 (at least in 5.1 beta 14), only two of them are available, and the "without play" versions (which are more reliable and quite snappy) are missing. Please refer to the table below.~
| Shortcut | SE4.0.16 | SE5.1.0 beta 14 |
| -- | -- | -- |
| One frame back | Available | N/A |
| One frame forward | Available | N/A |
| One frame back (with play) | Available | Available as `Video, one frame back` |
| One frame forward (with play) | Available | Available as `Video, one frame forward` |

~Please add the "without play" versions of "one frame back" and "one frame forward" to SE5 also.~

P.S. Both `Video, one frame back` and `Video, one frame forward` (with play) in SE5 (and in SE4 even!) are quite janky, but that's another issue. I'll try to report it (with video or GIF explanation) another time.

Thank you for your hard work on this AMAZING software!

niksedk:
Fixes #12350

The `<<` / `>>` "Seek video backward/forward" buttons below the waveform could only be triggered by mouse. This makes them assignable keyboard shortcuts.

Changes:
- `WaveformVideoSeekBackCommand` / `WaveformVideoSeekForwardCommand` are now configurable shortcuts under General > Video, listed as "Seek video backward" / "Seek video forward".
- The toolbar button tooltips now show the assigned key combo (the `{0}` placeholder in the existing hint strings was previously always blank for these two buttons).
- The shortcuts also work in the fullscreen video window, like the other video seek shortcuts.

No new language tags — the shortcut names reuse the existing `SeekBackHint`/`SeekForwardHint` strings with the placeholder blanked out, same as the buttons' automation names.

### DIFF
--- a/src/ui/Features/Main/Layout/InitWaveform.cs
+++ b/src/ui/Features/Main/Layout/InitWaveform.cs
@@ -670,7 +670,7 @@ public class InitWaveform
             VerticalAlignment = VerticalAlignment.Center,
             Command = vm.WaveformVideoSeekBackCommand,
             FontWeight = FontWeight.Bold,
-            [ToolTip.TipProperty] = UiUtil.MakeToolTip(languageHints.SeekBackHint, shortcuts),
+            [ToolTip.TipProperty] = UiUtil.MakeToolTip(languageHints.SeekBackHint, shortcuts, nameof(vm.WaveformVideoSeekBackCommand)),
         };
         Attached.SetIcon(buttonSeekBack, IconNames.ChevronDoubleLeft);
         AutomationProperties.SetName(buttonSeekBack, string.Format(languageHints.SeekBackHint, string.Empty).TrimEnd());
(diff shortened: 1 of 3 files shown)


GrampaWildWilly:
The underlying premise here is not 100% true.  It is possible to Tab to the seek buttons under the waveform & hit Enter to activate the seeks.  It should also be possible to hit Spacebar to activate those buttons.  Spacebar is a Windows standard way of pressing buttons, any buttons in any app.  But our waveform controls -- all of them under the waveform -- don't accept Spacebar, only Enter.

In addition, the original poster said he wanted frame movement.  As far as I can tell, single frame movement is not available in the waveform toolbar.  The available intervals range from 1/10 of a second to 10 seconds.  Single frame movement is not in the list.  As an aside, rather unexpectedly, Spacebar does cause the list of available intervals to open, as does Enter.  Why that control uniquely on the waveform toolbar & none of the other controls?

I have to add that I think this effort was unnecessary.  The existing shortcuts for this stuff work just fine when focus is on the waveform, at least in undocked mode.



If playback is paused, the existing shortcuts advance or reverse the video without resuming playback.  If playback is in progress, the skips do their thing & playback continues normally.  I love your commitment to pleasing the users, Nik, but sometimes I think your enthusiasm goes overboard, as here.

Surlime:
I feel like I'm missing something here, but I have to disagree on "The existing shortcuts for this stuff work just fine". I mean, I couldn't find a good way to navigate the video and waveform by frame via keyboard shortcuts, so I opened the issue.

I am a broken record by now, but the existing frame-stepping keyboard shortcuts (the `Video, one frame back/forward`, the exact one-frame navigation backward/forward) are very lagging when triggered rapidly. The other way I can think of is by using a calculator to find the exact millisecond for one frame and set it in the shortcuts you mentioned, which is not practical.

PR #12519, opened 2026-07-15
Time code spinner: nudge 100 ms, not a whole hour (#12506)

### LINKED ISSUE #12506: Clicking the “Show” button changes the time by 1 hour
Hi. 

In the SE5 version, pressing the Show button changes the time by 1 hour. In the SE4 version, the time changed by 100 ms. 
This can be corrected so that, by default, the time advances by 100 ms.

Image

Regards

muaz978:
Fixes #12506.

The time code up/down (the Show and Hide time fields, and every other time field in the app) stepped whichever segment the text caret sat in. The caret defaults to the hours field, so a plain spinner click, or the mouse wheel, jumped a full hour instead of a small nudge.

This makes mouse spinning and wheeling nudge a fixed small step: 100 ms, or one frame in frame mode, regardless of caret position, matching SE4. Keyboard Up/Down still steps the caret's segment, which is the deliberate way to change hours, minutes or seconds precisely.

Reported by @AdiOpi.

### DIFF
--- a/src/ui/Controls/TimeCodeUpDown.cs
+++ b/src/ui/Controls/TimeCodeUpDown.cs
@@ -19,6 +19,12 @@ namespace Nikse.SubtitleEdit.Controls
 {
     public class TimeCodeUpDown : TemplatedControl
     {
+        // Mouse spinner and wheel nudge by a fixed small step, matching SE4, instead of stepping
+        // the segment the text caret happens to sit in. The caret defaults to the hours field, so
+        // a plain spinner click used to jump a full hour (#12506). Keyboard Up/Down still steps the
+        // caret's segment, which is the deliberate way to change hours/minutes/seconds precisely.
+        private const int MouseStepMilliseconds = 100;
+
         public bool UseVideoOffset { get; set; } = false;
 
         private TextBox? _textBox;
@@ -55,14 +61,7 @@ namespace Nikse.SubtitleEdit.Controls
 
             control.AddHandler(InputElement.PointerWheelChangedEvent, (s, e) =>
             {
-                if (e.Delta.Y > 0)
-                {
-                    ChangeValue(+1);                    
-                }
-                else
-                {
-                    ChangeValue(-1);
-                }
+                MouseNudge(e.Delta.Y > 0 ? +1 : -1);
                 e.Handled = true;
             });
         }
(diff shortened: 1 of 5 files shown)


JDJD777:
It would be smarter to put the default caret in the ms section. Ignoring the caret position for the mouse wheel is a very strange approach and doesn't follow common practice.

(I already commented on the strange default of one hour some weeks ago but it was a side remark in a bug report and ignored.)

muaz978:
Fair point, and you did flag the one hour default earlier, sorry that got lost in the other thread.

On the caret: I agree the mouse and keyboard disagreeing is not ideal. Two things made me hesitate to move the default caret into the ms section though. First, the ms section steps 1 ms per click, so the default would become 1 ms and you would need 100 clicks to move the 100 ms that was asked for here. Second, the caret starts on the hours on purpose, so you can tab into the field and type a full time code left to right; starting in the ms would overwrite the milliseconds first.

So the options seem to be: a small fixed step for the mouse (this PR), or caret driven everywhere with the default caret moved and the ms step raised to 100 ms, which loses 1 ms precision.

@niksedk your call on which you prefer, happy to rework it either way.

niksedk:
I think this works as expected. It depends on the cursor placement in text box (default should be the last part which is ms/frames).

We could add a setting for `ms increment for time up/down controls` in Settings or something like that, and set the default to 100ms.
Perhaps the control could in some subtitle way better highlight the active part.

muaz978:
Reworked as you suggested, thanks, this is better.

- Stepping is caret driven again for everything, spinner, wheel and Up/Down alike. No more special case for the mouse, which also addresses @JDJD777's point.
- The caret now starts on the last part, ms or frames, instead of the hours, so a plain spinner click makes a small adjustment.
- New setting in General: `Time up/down increment (ms)`, default 100, range 1 to 5000. The ms step reads from it. Frame mode still steps one whole frame.

One consequence worth knowing: tabbing into the field and typing a full time code now starts on the ms part rather than the hours. If that turns out to annoy anyone, an easy follow-up is to jump the caret back to the start on the first digit typed.

I left the "highlight the active part" idea out of this one, it feels like a separate change to the control's look. Happy to do it as a follow-up if you want it.

Refreshed Windows test build with all of the above:

https://github.com/muaz978/subtitleedit/releases/tag/win-test-timecode-step

Issue #12434, opened 2026-07-13
Persian text rendering


rmtjokar:
I'm not sure why Persian behaves like this in v5x. Sometimes it displays the words in the wrong order. For example, take a look at the ones I marked with the orange and red lines.

<img width="1596" height="1358" alt="Image" src="https://github.com/user-attachments/assets/15bfa297-f4f5-4a9d-b11a-8ec6e483a087" />

There's another issue I highlighted in blue as well, where the words are sometimes rendered like this.
I've attached a test file for second part:
[breaking-and-exiting-2018_farsi_persian_1847705-\[SubSource\].zip](https://github.com/user-attachments/files/29981853/breaking-and-exiting-2018_farsi_persian_1847705-.SubSource.zip)

Additionally, would it be possible to add an RTL/LTR text direction option in the app? This would ensure that right-to-left languages such as Persian are displayed correctly.
Default(wrong text order):
<img width="885" height="1025" alt="Image" src="https://github.com/user-attachments/assets/b831b38d-1cea-4ef5-8c22-c67286d7c2b9" />
RTL:

<img width="873" height="1023" alt="Image" src="https://github.com/user-attachments/assets/aa4c7977-cdfe-4fa6-bc2b-0a0b7c513622" />

rmtjokar:
I initially thought this was an Avalonia text rendering issue, but it doesn't seem to be:

<img width="400" height="339" alt="Image" src="https://github.com/user-attachments/assets/dd0e7262-bd8a-4ead-824d-5bf3539954da" />

Maybe it's an issue with the SE JSON parser.

PR #12551, opened 2026-07-16
Fix RTL text reversed in the text box (vendor AvaloniaEdit with the RTL fix)

### LINKED ISSUE #12386: There's a problem with Arabic letters being reversed
There's a problem with Arabic letters being reversed in the text box, and the video doesn't play automatically when navigating between lines.

Image

niksedk:
Fixes #12386 and #12434 — Arabic/Hebrew rendered **reversed** in the subtitle text box.

## Root cause

The text box is an AvaloniaEdit `TextEditor`. AvaloniaEdit hardcoded `FlowDirection.LeftToRight` in its line layout (`VisualLineTextParagraphProperties`) while the framework applied its automatic mirror transform for a `RightToLeft` control — so LTR-laid-out glyphs got mirror-flipped instead of flowing right-to-left. This is upstream [AvaloniaEdit#401](https://github.com/AvaloniaUI/AvaloniaEdit/issues/401); nothing in SE's own code was wrong.

## Approach

The fix lives inside AvaloniaEdit and there's no upstream release carrying it, so this **vendors** `AvaloniaEdit` + `AvaloniaEdit.TextMate` as source under `src/AvaloniaEdit/` and project-references them from the UI instead of the NuGet packages. (We tried the fork-as-published-package route first; vendoring avoids the publish + flatpak-manifest-hashing overhead entirely.)

- Assembly names and namespaces are unchanged (`AvaloniaEdit`, `AvaloniaEdit.TextMate`), so **no SE code changes** — only the two `PackageReference`s become `ProjectReference`s.
- The vendored `.csproj`s are trimmed to build inside SE (net10.0, Avalonia 12.1.0, no signing/packaging); a scoped `Directory.Build.props` keeps SE-wide settings off the third-party source.
- `src/AvaloniaEdit/README.md` records provenance, the exact patch, and the upstream fork (`niksedk/AvaloniaEdit`).

## The fix itself

Propagate `TextView.FlowDirection` into the line paragraph properties so Avalonia's `TextFormatter` runs the Unicode bidi algorithm and reorders runs; right-align RTL lines; redraw on direction change. **Rendering only** — RTL caret movement and selection are follow-up work.

## Verification

- Full solution builds (the vendored projects are in `SubtitleEdit.sln` and referenced by the UI).
- The fix is covered by a headless test in the fork ([niksedk/AvaloniaEdit#1](https://github.com/niksedk/AvaloniaEdit/pull/1)): a mixed Latin+Hebrew line keeps Latin first under LTR but reorders Hebrew first under RTL; all 516 AvaloniaEdit tests pass.

CI regenerates `installer/flatpak/nuget-sources.json` itself, and the AvaloniaEdit NuGet packages are simply gone from the dependency set (their transitive deps like TextMateSharp remain, restored normally), so the release build is unaffected.

Worth doing separately: upstreaming the fix to AvaloniaUI/AvaloniaEdit so this vendored copy can eventually be dropped for a package release.


niksedk:
@muaz978 : Do you think we can make RTL work for Avalonia Edit ?

muaz978:
Yes, I think we can, and we are closer than the PR text suggests. I read through the vendored tree, and the useful discovery is that AvaloniaEdit delegates almost all of its coordinate work to Avalonia's `TextFormatter` / `TextLine`, which already implement the Unicode bidi algorithm. So the moment your change makes the formatter actually run bidi, a lot more than rendering comes along for free:

- **Caret X**: `VisualLine.GetTextLineVisualXPosition` goes through `TextLine.GetDistanceFromCharacterHit` (bidi aware).
- **Click hit testing**: `VisualLine` uses `TextLine.GetCharacterHitFromDistance`.
- **Selection rectangles**: `BackgroundGeometryBuilder` already loops over `TextLine.GetTextBounds(...)`, which returns several rectangles when a logical range is split across bidi runs, and it even carries the comment *"left>right is possible in RTL languages"* with `Math.Min` / `Math.Abs` handling. Upstream clearly anticipated RTL there.
- **Up/Down**: resolved through `desiredXPos` and hit testing, so it follows the same path.
- **Home/End**: logical start/end, which with `TextAlignment.Right` lands on the correct visual edge.

## The one real gap: horizontal caret movement

`CaretNavigationCommandHandler` maps `CharLeft`/`CharRight` (and `WordLeft`/`WordRight`) to `VisualLine.GetNextCaretPosition(visualColumn, Backward/Forward, ...)`, which walks **visual columns**. In AvaloniaEdit a visual column is a linear logical index into the line; it has no bidi knowledge. Avalonia's `TextLine` exposes exactly the right primitives for this, `GetNextCaretCharacterHit` and `GetPreviousCaretCharacterHit`, and the vendored tree calls **neither** (zero call sites).

## Worth testing before merge

Your fix decouples visual order from visual column order. Before it, the line was formatted LTR and then mirrored by the framework, so the glyphs and the caret X were consistently wrong but aligned with each other. Now the glyphs reorder while the arrows still step logical columns, so on a pure RTL line I would expect Left to move the caret visually **right**. Worth confirming, because it is a behaviour that the rendering fix can surface rather than cause.

Related: SE already has `TryHandleRightToLeftCaretNavigation` in `MainViewModel` (from #11715). It handles RTL word and line navigation for both edit box implementations, but it deliberately leaves bare single character movement to the editor, on the assumption that it is "already bidi correct". That assumption is worth rechecking on the AvaloniaEdit path after this change.

## Suggested staging, each step shippable

1. **Merge this PR as is.** Rendering plus the coordinate wins above is the big user visible fix, and it is low risk.
2. **Caret movement**: in `CaretNavigationCommandHandler`, for `CharLeft`/`CharRight`/`WordLeft`/`WordRight`, when the line actually has bidi runs, resolve the step through `TextLine.GetNextCaretCharacterHit` / `GetPreviousCaretCharacterHit` and map the resulting `CharacterHit` back to a visual column, instead of stepping the column by one. Keep the existing path for pure LTR lines so nothing regresses.
3. **Caret affinity at direction boundaries**: at an LTR/RTL run boundary a single logical offset has two valid X positions (the strong/weak caret). `TextViewPosition` has no affinity concept; `CharacterHit.TrailingLength` carries it. This is the fiddly part and can be deferred.
4. **Upstream it**, so the vendored copy can eventually be dropped.

## The honest caveat

AvaloniaEdit's model assumes visual column order equals visual order, and bidi breaks that assumption. A complete fix means a bidi aware mapping layer between visual column and caret order. The good news is that it is routing, not reimplementing, since Avalonia already does the bidi. But it touches the most delicate part of the editor, which is why I would stage it rather than do it in one go.

For SE specifically the scope is friendly: the box holds short subtitle lines, not code, so folding, rectangle selection and virtual space RTL correctness barely matter. Step 2 is where nearly all of the remaining user visible value is.

## One future upgrade worth knowing about

`CreateParagraphProperties` is built once per layout pass and shared by every line in it, so the whole editor takes a single direction from `TextView.FlowDirection`. But `firstLineInParagraph` is already mutated per line inside that same loop, so `flowDirection` could be set per visual line the same way, auto detected from that line's content. That would match what SE already does in the grid with `TextToFlowDirectionConverter` (per cell auto direction) and it matters for subtitles that mix an Arabic line and a Latin line in the same block.

Happy to take step 2 if you want it.

niksedk:
Please do give it try :)

muaz978:
Done: #12553, against this branch since the change lives inside the vendored AvaloniaEdit.

**First, a correction to my earlier comment.** I said the fix would be to route caret movement through `TextLine.GetNextCaretCharacterHit` / `GetPreviousCaretCharacterHit`. I checked Avalonia's source before writing anything, and that was wrong: those two walk **logical** order (they resolve clusters, not bidi), so routing through them would not have moved the caret visually at all. Good thing it did not go in as written.

## What the bug actually is

Avalonia's `TextPresenter.MoveCaretHorizontal` flips the logical direction when the control is right-to-left:

```csharp
if (FlowDirection == FlowDirection.RightToLeft)
{
    direction = direction == LogicalDirection.Forward ?
        LogicalDirection.Backward : LogicalDirection.Forward;
}
```

So every ordinary Avalonia text box already moves the caret visually left when you press Left in RTL. AvaloniaEdit's `CaretNavigationCommandHandler` never does this: `CharLeft` always steps the visual column backwards. With the old reversed rendering that was invisible, because glyphs and caret were consistently wrong together. Your fix reorders the glyphs, so the mismatch becomes visible: Left moves the caret visually right.

This matters more now because of your latest commit. Disabling the `SubtitleTextBoxColorTags = false` workaround means RTL users land on AvaloniaEdit for the first time, instead of the plain text box where Avalonia's flip already handled this.

## The fix

One flip in `GetNewCaretPosition`, mirroring `TextPresenter` exactly:

```csharp
if (textView.FlowDirection == Avalonia.Media.FlowDirection.RightToLeft)
{
    direction = direction switch
    {
        CaretMovementType.CharLeft => CaretMovementType.CharRight,
        CaretMovementType.CharRight => CaretMovementType.CharLeft,
        _ => direction
    };
}
```

Only the single character steps flip. Avalonia does not flip whole word movement either, and SE already corrects word and line movement for both edit boxes in `TryHandleRightToLeftCaretNavigation` (it marks the event handled, so the editor never sees those keys). Flipping words here too would fight it.

## How it was verified

Four headless tests over a real `TextEditor`, driving the same routed commands the key bindings invoke. The important part is that I checked they actually catch the bug rather than just pass: with the flip disabled, the two RTL cases fail (RTL Left lands on 4 instead of 6) while the LTR ones still pass; with it enabled all four pass. Full UI suite green, 693 tests.

Two things worth knowing that came out of the testing:

1. A `TextEditor` gets **no template** in a headless test unless AvaloniaEdit's own theme is included, so its `TextArea` never enters the visual tree. That is why the shared test app now adds the `avares://AvaloniaEdit/Themes/Fluent/AvaloniaEdit.xaml` include. Until I found that, the TextArea was unattached, could not take focus, and did not inherit `FlowDirection`, which made the first attempts fail for the wrong reason.
2. Following on from that, I did check the obvious worry: `FlowDirection` **does** reach the `TextView` in a properly laid out tree (`editor=RightToLeft area=RightToLeft view=RightToLeft`), so your rendering fix fires as intended. It only looked otherwise in a half-built test tree.

## Still open, deliberately

Caret affinity at a direction boundary, where one logical offset has two valid X positions, is untouched. `TextViewPosition` has no affinity concept and `CharacterHit.TrailingLength` carries it, so it is a separate piece of work. Word and line movement stay as they are for the reasons above.

muaz978:
I went through this branch in depth, including the vendored tree against upstream. Two things to report: one test on the branch is red for a real reason, and the CI here cannot tell you that.

**The CI is blind right now.** AppVeyor says "unable to build non-mergeable pull request" because of the `change-log.txt` conflict with main, so nothing on this branch has actually been run. That is the only conflict, and it is trivial, but it is currently hiding a genuine failure.

**`TaggedLine_DragFromRightEdge_SelectsFromDocumentStart` fails.** Your test is right and the code is wrong. Fix in #12565, against this branch.

The cause is a nice trap. A line starting with `{\an8}` resolves to a **left-to-right embedding level**, whatever the paragraph direction is. The tag is a drawable run with no text, so `TextFormatterImpl.ShapeTextRuns` feeds the bidi algorithm `new string('a', textRun.Length)` on its behalf, and `ResolveEmbeddingLevel` follows Unicode P2/P3: first strong character wins, `ParagraphEmbeddingLevel` is never consulted. So the substituted `'a'` decides the line.

Reordering still uses the paragraph level, so rendering stays correct and `TaggedLine_TagsKeepLiteralLayout` passes. But for drawable runs `GetDirection` cannot infer a direction and falls back to that resolved level, so `GetDistanceFromCharacterHit` and `GetCharacterHitFromDistance` answer with the left-to-right convention. On the test's own line the tag renders at x 343 to 400 and clicks map like this:

| click x | offset | expected |
|--------:|-------:|---------:|
| 396 | 6 | 0 |
| 390 | 9 | 0 or 6 |
| 370 | 10 | 0 or 6 |
| 350 | 14 | 0 or 6 |

Clicking the tag lands in the middle of the Arabic. The snap you added only runs when `TrailingLength > 0`, and these come back `trail = 0`, so it never fires.

`GetTextBounds` is not affected, so #12565 resolves RTL hits from the rendered geometry. One thing worth knowing if you touch this: `TextBounds.FlowDirection` is **also** unusable here, it reports `LeftToRight` for the Arabic columns, since it comes off the same resolved level. The direction has to come from the rendered order.

Some things I checked that turned out fine, so you do not need to look at them:

- `IsPastLineEnd` is correct. I confirmed `TextLine.Start` against Avalonia 12.1.0's `GetParagraphOffsetX`: it is a non-negative offset from the left edge in every alignment, and `Start` normalizes to `Right` under RTL, so the ink does span `[Start, Start + width]` and the left edge is the logical end, exactly as your comment says.
- I suspected `BackgroundGeometryBuilder` line 255 (`left = line.Width`, a width used as an x position, ignoring `Start`) would paint selection on the wrong side of a wrapped RTL line. A pixel probe over a three-line wrapped Arabic selection shows the highlight tracking the text correctly, so it does not bite. It is still off by `Start` on paper and only harmless because the rectangle overlaps already-selected text, so it may be worth tidying at some point, but nothing is broken.
- The vendored tree is clean. Against AvaloniaEdit `master` (which is what it is vendored from, not the `12.0.0` tag) only the 7 files your README documents differ, plus the two trimmed csproj files. Nothing undisclosed. The README is genuinely good, it saved me a lot of time.

The Avalonia behaviour underneath this looks worth an upstream issue on its own: a run with no text should probably not be able to decide a paragraph's direction via a substituted `'a'`. Happy to file that if you want.

niksedk:
I would be nice to help upstream if possible - do try that if you can/have-time.
It seems to be a rather large change to fix this - the selection still looks weird to me (but you might have fixed that)

Also looked at old issues and found this: https://github.com/AvaloniaUI/Avalonia/issues/2842#issuecomment-1705576591
The "main" branch now has this adapted component - how does that work with RTL?

muaz978:
Answering your question about the adapted component, with measurements rather than opinions. Short version: **the native `SyntaxHighlightingTextBox` handles RTL better than I expected, and the case for this PR is now much narrower than it looked.**

## Method

Built `main` (with #12556 + #12564), rendered the real `SyntaxHighlightingTextBox` through Skia, and captured the actual output. Everything below is what the control draws, not a mockup. I also checked the same lines against a plain `TextBox` to separate "bidi does this" from "the new control does this": the results are byte-identical, so none of it is caused by your component.

## What renders correctly

| stored | renders as | |
|---|---|---|
| `<i>…</i>` | `<i>…</i>` | correct |
| `<font color="red">…</font>` | unchanged | correct |
| `هذا <i>اختبار</i> في الوسط` | unchanged | correct |
| `مرحبا <i>hello</i> بالعالم` | unchanged | correct |

`<i>` rendered right-to-left is **pixel identical** to `<i>` rendered left-to-right, just right-aligned. Bidi mirrors `<` and `>`, so the tag still faces the right way. This means **Arabic SRT editing has no tag problem at all** in the native box, since SRT only carries HTML tags.

## What breaks

| stored | renders as |
|---|---|
| `{\an8}` | `{an8\}` |
| `{\pos(320,50)}` | `{pos(320,50)\}` |
| `{\c&HFF0000&}` | `{&c&HFF0000\}` |

Only ASSA brace tags, and mostly by **one character**: the backslash. The braces mirror and stay correct; `\` is a neutral sitting between `{` and the `an8` run, so it resolves to the paragraph level and lands on the far side of that run. `<i>` survives because it is symmetric (one neutral each side); `{\an8}` does not because it has two neutrals at the front. The stored text is always correct, so this is cosmetic.

## Caret, clicks and selection

All three are correct natively, measured on `main`:

- caret at offset 5 in Arabic: Left moves to 6, Right moves to 4 (visually correct for RTL)
- click near the right edge puts the caret at offset 0

Those are exactly what #12553 and #12565 had to add by hand to AvaloniaEdit. A real `TextBox` gets them for free from `TextPresenter`. **So both of my merged fixes buy nothing on `main`** and only exist because AvaloniaEdit is not a TextBox.

## What that means for this PR

The remaining unique value of vendoring AvaloniaEdit is: **a correct backslash inside ASSA override tags in RTL lines.** That is the whole list. Weighed against ~51k vendored lines to re-sync by hand on every AvaloniaEdit or Avalonia upgrade, plus the geometry workaround in #12565 that exists only to route around an Avalonia bug, my honest read is that this PR is no longer worth its maintenance cost, and I say that as the person who wrote two of the commits in it.

If the displaced backslash is worth addressing, there may be a cheaper route than an editor swap: `{\an8}` and `{\pos(…)}` are line-level metadata and could live as a chip above the text box (the row already used for the bookmark), which takes them out of bidi entirely and leaves clean Arabic plus HTML tags in the box, both of which already render correctly. That would not cover mid-line tags like `{\c&HFF0000&}`. Muaz raised that idea and it is his call as the person who actually edits Arabic subtitles; happy to prototype it if it is wanted.

## Upstream

Filed the Avalonia bug from my last comment: **AvaloniaUI/Avalonia#21792**, with a self-contained repro that uses only the public `TextFormatter` API. A `DrawableTextRun` with no text makes the formatter feed the bidi algorithm a substituted `'a'`, and rule P2/P3 takes that as the paragraph's first strong character, so a line starting with a tag element resolves left-to-right no matter the paragraph direction. That is what corrupted the hit APIs in the vendored copy. If it is fixed upstream, the workaround in #12565 becomes deletable.

## One thing I could not reproduce

You mentioned the selection still looks weird. I could not reproduce a selection problem in either editor: in the native box the highlight tracks the text, and in the vendored one the pixel tests pass. Which of the two were you looking at, and on which line? A screenshot would help me chase it properly rather than guess.

Also worth flagging: this PR's CI has not run at all. AppVeyor reports "unable to build non-mergeable pull request" because of the `change-log.txt` conflict with main, which is why the failing test in #12565 was invisible here.

muaz978:
@niksedk, can you please check this?
[rtl-tag-report.html](https://github.com/user-attachments/files/30122571/rtl-tag-report.html)

niksedk:
> @niksedk, can you please check this? [rtl-tag-report.html](https://github.com/user-attachments/files/30122571/rtl-tag-report.html)

Thx, yes. Will try with Claude :)

muaz978:
Adding the one thing my measurements could not tell you: I looked at `{an8\}` as the person who edits Arabic subtitles all day, and it is fine. It does not change the look or the behaviour of anything, the stored text is correct, and it has rendered that way for a long time, including in SE4. So it is not a regression, and it is not something I need fixed.

That removes the last thing this PR uniquely buys. My honest recommendation is to **close it**, and #12553 and #12565 with it. The native `SyntaxHighlightingTextBox` already gets RTL caret movement, clicks and selection right on its own, Arabic SRT editing has no tag problem at all, and the only difference left is a backslash I have lived with for years. That is not worth vendoring AvaloniaEdit and re-syncing it by hand on every upgrade.

Two things worth keeping from the work, so it is not wasted:

- **AvaloniaUI/Avalonia#21792** is confirmed and assigned. Gillibald agreed that embedded object runs should default to the neutral direction and is fixing it with the Unicode Object Replacement Character instead of the substituted Latin letter. That bug affects any Avalonia app that puts object runs in RTL text, so it is worth having found regardless of what happens here.
- The RTL rendering fix itself (`VisualLineTextParagraphProperties` honouring `FlowDirection`) is still real and still worth upstreaming to AvaloniaEdit, whether or not SE vendors it. AvaloniaEdit is still used for the source view and media info, so it may matter there.

Happy to help with the AvaloniaEdit upstreaming if you want it, and thanks for merging the two fixes even though it now looks like they will not ship.

PR #12556, opened 2026-07-17
Add native syntax highlighting text box for the subtitle edit controls

niksedk:
## What

When "Color tags (HTML/ASSA) in subtitle text box" is enabled, the main window edit boxes now use a new `SyntaxHighlightingTextBox` — a plain Avalonia `TextBox` subclass — instead of the AvaloniaEdit editor. Being a real `TextBox`, it gets line alignment (left/center/right), selection, IME/CJK composition, RTL and the context menu natively, without the AvaloniaEdit wrapper workarounds. The AvaloniaEdit editor code is kept for now but is no longer used by the main window.

## How

- **`SyntaxHighlightingTextBox`** (`src/ui/Controls/`) — thin `TextBox` subclass; a template in `Styles.axaml` (copy of the Fluent 12.1 TextBox template, part names preserved so all Fluent styling still applies) swaps in the custom presenter.
- **`SyntaxHighlightingTextPresenter`** — replicates `TextPresenter.CreateTextLayout()` from Avalonia 12.1.0 and merges in syntax-highlight `TextRunProperties` spans. Private Avalonia members are reached via `UnsafeAccessor` (same approach as Pororoca / Carina Studio's AppSuiteBase, MIT). Selection and IME-preedit highlights are merged by span clipping, so composition text is never tag-colored and tags recolor at shifted offsets while composing. Run properties/brushes are cached per color. **Re-verify against the Avalonia source when upgrading Avalonia.**
- **`SubtitleSyntaxTokenizer`** (`src/ui/Logic/`) — the `SubtitleSyntaxHighlighting.ColorizeLine` scanning logic ported to emit plain colored ranges, so both editors share one scheme: delimiters gray, tag names purple, attribute names blue, values orange, and actual-color rendering for `color="red"` / `{\c&HFF0000&}`.

## Bug fix

`SubtitleSyntaxHighlighting.ColorizeLine` threw `ArgumentOutOfRangeException` when a line ended with an unterminated attribute quote (e.g. while typing `<font color="`): the colon-check `text.IndexOf(':', i + 1, valueEnd - i - 2)` computed a negative count. Fixed in both the AvaloniaEdit colorizer and the tokenizer port; a 20k-input fuzzer over the tag alphabet now passes with no exceptions and no range-invariant violations.

## Verification

- Headless Avalonia harness: template applies, granular colors match the AvaloniaEdit scheme, center alignment works, simulated CJK composition (preedit) renders underlined/uncolored with committed text untouched, selection-foreground clipping over tags is correct.
- Edge cases + 20,000-input fuzz on the tokenizer: no throws, ranges always sorted/non-overlapping/in-bounds.
- BenchmarkDotNet (Apple M4, .NET 10): tokenizing a typical tagged line runs in a few hundred nanoseconds; the micro-optimization pass (span comparisons instead of substrings, cached run properties) cut allocations (e.g. mixed HTML+ASSA line 2,000 B → 1,816 B; ASSA line 1,264 B → 1,024 B). A side-by-side old-vs-new timing comparison will be posted as a comment.

### DIFF
--- a/src/ui/Controls/SyntaxHighlightingTextBox.cs
+++ b/src/ui/Controls/SyntaxHighlightingTextBox.cs
@@ -1,0 +1,21 @@
+using Avalonia.Controls;
+using System;
+
+namespace Nikse.SubtitleEdit.Controls;
+
+/// <summary>
+/// A <see cref="TextBox"/> that colors HTML tags and ASS/SSA override tags while typing, using
+/// the same color scheme as the AvaloniaEdit based editor (SubtitleSyntaxHighlighting).
+/// It behaves exactly like a normal text box (alignment, selection, IME, context menu) -
+/// the coloring is done by <see cref="SyntaxHighlightingTextPresenter"/>, which the
+/// ":syntaxHighlighting" style in Styles.axaml swaps in for the default text presenter.
+/// </summary>
+public class SyntaxHighlightingTextBox : TextBox
+{
+    protected override Type StyleKeyOverride => typeof(TextBox);
+
+    public SyntaxHighlightingTextBox()
+    {
+        PseudoClasses.Add(":syntaxHighlighting");
+    }
+}
--- a/src/ui/Controls/SyntaxHighlightingTextPresenter.cs
+++ b/src/ui/Controls/SyntaxHighlightingTextPresenter.cs
@@ -1,0 +1,263 @@
+using Avalonia;
+using Avalonia.Controls.Presenters;
+using Avalonia.Media;
+using Avalonia.Media.Immutable;
+using Avalonia.Media.TextFormatting;
+using Avalonia.Utilities;
+using Nikse.SubtitleEdit.Logic;
+using System;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
+namespace Nikse.SubtitleEdit.Controls;
+
+/// <summary>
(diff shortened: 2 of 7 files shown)


niksedk:
Side-by-side BenchmarkDotNet comparison of the tokenizer before/after the micro-optimization pass (both variants compiled into one run, Apple M4, .NET 10.0.7, default job, MemoryDiagnoser). An earlier sequential run suggested a slowdown, but that was cross-run machine-load noise — measured head-to-head:

| Input | Old | New | Time | Old alloc | New alloc | Alloc |
|---|---:|---:|---:|---:|---:|---:|
| Plain text (no tags) | 73.4 ns | 74.7 ns | ±0 (noise) | 32 B | 32 B | ±0 |
| HTML tags | 272.5 ns | 243.3 ns | **−11%** | 912 B | 880 B | −4% |
| ASSA tags | 406.8 ns | 336.6 ns | **−17%** | 1,264 B | 1,024 B | **−19%** |
| Mixed HTML+ASSA, 2 lines | 488.6 ns | 445.3 ns | **−9%** | 2,000 B | 1,816 B | **−9%** |

The gains come from span comparisons instead of substring allocations in the scan loop (attribute names, ASSA tag names). Either way the tokenizer is sub-microsecond per keystroke; the separately cached `TextRunProperties`/brush instances in the presenter remove another ~2 allocations per tag per layout pass.

PR #12566, opened 2026-07-17
Update Polish translation

potplayer-fanpack:
@niksedk 

Please merge. Thx


niksedk:
Looks good, but the file `Dictionaries/pl_names_user.xml` should not be included - the names should only be in "pl_names.xml" (the _user file is where the users own names ends up).

PR #12567, opened 2026-07-17
Fix ASSA tag backslash rendering in RTL lines in the syntax highlighting text box

niksedk:
Fixes the displaced backslash identified in the [measurement comment on #12551](https://github.com/SubtitleEdit/subtitleedit/pull/12551#issuecomment-5002528544): in RTL lines, `{\an8}` rendered as `{an8\}`, `{\c&HFF0000&}` as `{&c&HFF0000\}` — the tag's neutral characters (backslash, `&`) resolve to the paragraph bidi level and land on the wrong side of the letter runs.

### Approach

Avalonia has no public API to override bidi levels per run (the machinery is internal), so the fix works *with* the bidi algorithm: `AssaTagRtlLayout` permutes the tag characters **in the layout string only** — the stored `Text` is never touched — storing the tag content segment-reversed so the renderer's bidi reversal puts every character back in its true visual spot. Neutral brackets are stored as their mirror pair since the renderer mirrors them at the RTL level.

Which characters ride along with a letter run is decided by a small UAX#9 subset: W4/W7 (digits attach to letter runs, decimal points join numbers), N1 (neutrals between strong runs join them — this is why `(320,50)` and the inner `\fs30` of `{\t(0,500,\fs30)}` already rendered fine), and braces pinned to the RTL level per the paired-bracket rule N0. Directly adjacent blocks (`{\an8}{\i1}`) reorder as one bidi region and are permuted as one region.

### Verification

Verified against Avalonia's actual renderer (headless `TextLayout` + hit-testing every character's visual position) over 23 cases covering the ASSA tag grammar: `{\an8}`, `{\pos(320,50)}`, `{\move}`, `{\c&H..&}`/`{\1c}`/`{\3c}`/`{\alpha}`, `{\bord2.5}`, `{\t(0,500,\fs30)}`, `{\fad}`, multi-tag blocks, adjacent blocks, tags at line start/end/mid-line, plus guards that plain `{...}` text and HTML-only lines are untouched. All pass.

### Notes

- Syntax colors follow each character through the permutation (per-char remap + run-length encoding), so the tag keeps its delimiter/name/value coloring.
- Spell check underline indices are unaffected — the permutation is identity outside tags, which the word scan skips.
- LTR paragraphs, IME composition passes, and the plain `TextBox` are untouched.

### DIFF
--- a/src/ui/Controls/AssaTagRtlLayout.cs
+++ b/src/ui/Controls/AssaTagRtlLayout.cs
@@ -1,0 +1,275 @@
+using System;
+using System.Collections.Generic;
+
+namespace Nikse.SubtitleEdit.Controls;
+
+/// <summary>
+/// Display-only bidi workaround for ASSA override tags ("{\an8}", "{\pos(320,50)}", ...) in
+/// right-to-left paragraphs.
+///
+/// The backslash and other neutral characters inside a tag have no strong direction, so the
+/// Unicode bidi algorithm resolves them to the paragraph (RTL) level and displays them on the
+/// wrong side of the letter/digit runs: "{\an8}" renders as "{an8\}". There is no public
+/// Avalonia API to override bidi levels per run, so instead the tag content is pre-permuted in
+/// the string given to the text layout such that the bidi reordering puts every character back
+/// in its true visual spot. The stored text is never touched - only what the presenter renders.
+///
+/// How the permutation works: within a tag, bidi keeps letter/digit runs intact (left-to-right)
+/// and reverses the display order of the top-level segments (a segment is a letter/digit run or
+/// a single neutral character); neutral brackets are additionally mirrored. So the layout string
+/// holds the segments in reverse order (with neutral brackets swapped for their mirror pair),
+/// and the renderer's reversal restores the original order. Which characters join a letter/digit
+/// run is decided by a small UAX#9 subset (W4/W7 for numbers, N1 for neutrals between strong
+/// characters) - verified against Avalonia's actual renderer for the ASSA tag grammar.
+/// </summary>
+public static class AssaTagRtlLayout
+{
+    /// <summary>
+    /// Permutes the content of every "{\...}" override block for RTL display. Returns false when
+    /// the text contains no override blocks (the common case - no allocation happens then).
+    /// <paramref name="toSource"/> maps each index of <paramref name="layoutText"/> back to its
+    /// index in <paramref name="text"/>, for remapping style spans; it is the identity outside
+    /// override blocks.
+    /// </summary>
+    public static bool TryPermuteForRtlDisplay(string text, out string layoutText, out int[] toSource)
+    {
+        layoutText = text;
+        toSource = Array.Empty<int>();
+
+        var firstBlock = FindBlockStart(text, 0);
+        if (firstBlock < 0)
+        {
+            return false;
+        }
+
+        var chars = text.ToCharArray();
+        var map = new int[text.Length];
+        for (var i = 0; i < map.Length; i++)
+        {
+            map[i] = i;
+        }
+
+        var changed = false;
+        var blockStart = firstBlock;
+        while (blockStart >= 0)
+        {
+            // A run of directly adjacent override blocks ("{\an8}{\i1}") reorders as one bidi
+            // region, so it must be permuted as one region - braces included.
+            var regionEnd = text.IndexOf('}', blockStart + 1);
+            if (regionEnd < 0)
+            {
+                break;
+            }
+
+            while (regionEnd + 1 < text.Length && text[regionEnd + 1] == '{')
+            {
+                var nextEnd = text.IndexOf('}', regionEnd + 2);
+                var nextBackslash = text.IndexOf('\\', regionEnd + 2);
+                if (nextEnd < 0 || nextBackslash < 0 || nextBackslash > nextEnd)
+                {
+                    break;
+                }
+
+                regionEnd = nextEnd;
+            }
+
+            changed |= PermuteRegion(text, blockStart, regionEnd + 1, chars, map);
+            blockStart = FindBlockStart(text, regionEnd + 1);
+        }
+
+        if (!changed)
+        {
+            return false;
+        }
+
+        layoutText = new string(chars);
+        toSource = map;
+        return true;
+    }
+
+    /// <summary>
+    /// An override block is "{" with a "\" before the next "}" - a plain "{...}" without
+    /// override tags is left alone.
+    /// </summary>
+    private static int FindBlockStart(string text, int from)
+    {
+        var start = text.IndexOf('{', from);
+        while (start >= 0)
+        {
+            var end = text.IndexOf('}', start + 1);
+            if (end < 0)
+            {
+                return -1;
+            }
+
+            var backslash = text.IndexOf('\\', start + 1);
+            if (backslash > start && backslash < end)
+            {
+                return start;
+            }
+
+            start = text.IndexOf('{', end + 1);
+        }
+
+        return -1;
+    }
+
+    private enum CharClass
+    {
+        Strong, // renders as an intact left-to-right run (letters, and digits attached to them)
+        Number, // digit run without a preceding letter - still left-to-right but neutral-hostile
+        Neutral, // takes the paragraph (RTL) level and gets displaced
+    }
+
+    private static bool PermuteRegion(string text, int regionStart, int regionEnd, char[] chars, int[] map)
+    {
+        var length = regionEnd - regionStart;
+        if (length <= 1)
+        {
+            return false;
+        }
+
+        // 1) Initial classes for the region characters (ASCII tag grammar). Braces always
+        // resolve to the RTL level in a right-to-left paragraph (paired-bracket rule N0: they
+        // enclose left-to-right tag content in a right-to-left context), so they are never
+        // upgraded below and act as barriers for the N1 upgrade.
+        var classes = new CharClass[length];
+        var isBrace = new bool[length];
+        for (var i = 0; i < length; i++)
+        {
+            var ch = text[regionStart + i];
+            isBrace[i] = ch is '{' or '}';
+            classes[i] = char.IsLetter(ch) ? CharClass.Strong
+                : char.IsDigit(ch) ? CharClass.Number
+                : CharClass.Neutral;
+        }
+
+        // 2) W4: a single '.' ',' ':' between two digits joins the number.
+        for (var i = 1; i < length - 1; i++)
+        {
+            var ch = text[regionStart + i];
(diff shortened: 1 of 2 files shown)


niksedk:
@muaz978 Should we revert or keep this?

muaz978:
I tested it. **The rendering half genuinely works**, and it is a nicer result than I expected: permuting the layout string so the bidi reversal undoes it is a clever way through a problem with no public API. Verified against the real presenter (SE's own `Styles.axaml` template + Skia):

| stored | renders as | |
|---|---|---|
| `{\an8}` | `{\an8}` | correct, colours intact |
| `{\pos(320,50)}` | `{\pos(320,50)}` | correct |
| `هذا {\c&HFF0000&}اختبار{\c} ملون` | correct, red shown | correct |
| `{مرحبا}` (plain braces) | untouched | correct |
| `<i>…</i>`, LTR paragraphs | untouched | correct |

**But it introduces a crash, so my answer is revert (or fix the hit-test path first).**

Clicking inside an ASSA tag in an RTL line throws:

```
System.InvalidOperationException : Covered length must be greater than zero.
  at Avalonia.Media.TextFormatting.TextLineImpl.GetTextBounds(Int32 firstTextSourceIndex, Int32 textLength)
  at Avalonia.Media.TextFormatting.TextLayout.HitTestTextRange(Int32 start, Int32 length)
  at Avalonia.Controls.Primitives.TextSelectionHandleCanvas.MoveHandlesToSelection()
  at Avalonia.Controls.Primitives.TextSelectionHandleCanvas.TextBox_PropertyChanged(...)
  at Avalonia.Controls.TextBox.OnPointerPressed(...)
```

Isolated by clicking the same x positions across `{\an8}هذا` in four configurations (caret index, or THROW):

| configuration | x = 372, 362, 352, 342, 332, 322, 312 |
|---|---|
| **SyntaxBox RTL + ASSA tag (permuted)** | 0, 1, **THROW**, 2, 6, 5, 4 |
| plain TextBox RTL + ASSA tag | 0, 1, 2, 5, 4, 4, 3 |
| SyntaxBox RTL, plain braces `{مرحبا}` | 0, 1, 1, 2, 3, 4, 5 |
| SyntaxBox LTR + ASSA tag | 9, 9, 9, 9, 9, 9, 9 |

Only the permuted row throws. Note the row is also wrong where it does not throw: clicking inside the tag jumps to 6 (past the tag, into the Arabic), so click-to-caret inside a tag does not follow the glyphs.

**Cause:** the permutation is applied to the layout but never inverted on the way back. `toSource` is threaded into `BuildSyntaxSpans` so the colours follow the characters, which is why the colouring is right, but `SyntaxHighlightingTextPresenter` only overrides `CreateTextLayout()`. Caret positioning and hit-testing keep using the permuted layout while `TextBox` treats the results as indices into `Text`. Outside tags the permutation is the identity, so everything else behaves and the problem hides; inside a tag the two index spaces diverge. `MoveHandlesToSelection` is not gated on touch mode (it runs from `TextBox_PropertyChanged` on every caret change), so a plain desktop click reaches it.

**Why I would revert rather than fix it now:** I looked at `{an8\}` as the person who edits Arabic subtitles all day, and it is fine. It does not change the look or the behaviour of anything, the stored text was always correct, and it has rendered that way for years, including SE4. So this trades a cosmetic wart I do not notice for a crash on a common action, which is a bad trade. If you do want to keep it, the fix is to map hit-test and caret indices back through `toSource` (and its inverse for text to layout), so the two index spaces cannot leak into each other.

One other thing worth flagging: the 23 verification cases in the PR description are not in the repo, so nothing guards this. The suite is 689/689 on main because nothing covers `AssaTagRtlLayout` at all. If it stays, it needs those cases committed, plus click-to-caret coverage inside a tag, which is where it actually breaks.

Happy to send the fix or the tests if you would rather keep it than revert. Either way the rendering approach is sound, it is only the return trip that is missing.

niksedk:
Can you try to fix it? (the way you find best)

muaz978:
I tried to fix it, and I could not, so my suggestion is to revert. Below is everything I found, including a second approach I tested that fails the same way, and what I think we should do in the meantime. None of this is a criticism of the approach: the permutation idea is genuinely clever, and the rendering half of it works. The problem sits underneath us, in Avalonia.

## Why it cannot be fixed in place

The fix needs one more step: translate a position from the permuted layout back to a position in `Text`. You already built that map (`toSource`), which is why the colours are correct. It just has to run at the moment a click becomes a caret index, and that moment is inside Avalonia, out of reach:

| what we would need to hook | 12.1.0 |
|---|---|
| `TextPresenter.MoveCaretToPoint` (what a click calls) | not `virtual` |
| `TextLayout.HitTestPoint` / `HitTestTextRange` / `HitTestTextPosition` | not `virtual` |
| `TextPresenter.GetCaretPoints` (draws the caret) | `internal` |
| overridable on `TextPresenter` | only `CreateTextLayout` and `InvalidateTextLayout` |

A click goes `TextBox` to `MoveCaretToPoint` to `TextLayout.HitTestPoint`. Nothing on that path can be overridden, so the permuted layout's indices arrive at `TextBox` as if they were indices into `Text`. Outside tags the permutation is the identity, which is why everything else behaves and the problem stays hidden until a tag is clicked.

## The crash, and what it costs

`InvalidOperationException: "Covered length must be greater than zero"`, thrown from `TextSelectionHandleCanvas.MoveHandlesToSelection` to `HitTestTextRange` to `GetTextBounds`. It is not touch-only: `MoveHandlesToSelection` runs from `TextBox_PropertyChanged` on every caret change, so an ordinary desktop click reaches it.

Clicking the same x positions across `{\an8}هذا` in four configurations (caret index, or THROW):

| configuration | x = 372, 362, 352, 342, 332, 322, 312 |
|---|---|
| **SyntaxBox RTL + ASSA tag (permuted)** | 0, 1, **THROW**, 2, 6, 5, 4 |
| plain TextBox RTL + ASSA tag | 0, 1, 2, 5, 4, 4, 3 |
| SyntaxBox RTL, plain braces `{مرحبا}` | 0, 1, 1, 2, 3, 4, 5 |
| SyntaxBox LTR + ASSA tag | 9, 9, 9, 9, 9, 9, 9 |

Only the permuted row throws. It is also wrong where it does not throw: clicking inside the tag jumps to 6, past the tag into the Arabic. So even with the crash silenced, clicking a tag puts the caret on the wrong character, which makes editing tags worse than the displaced backslash ever did.

## The other approach, tested, fails identically

Before suggesting revert I tried the obvious alternative: isolate the tag instead of permuting it, so nothing moves and no index can diverge.

- **Isolates (LRI/PDI, U+2066/U+2069):** no effect at all. Avalonia rendered the line exactly as without them.
- **LRM (U+200E):** renders **correctly**. `{LRM\an8LRM}هذا` displays as a perfect `{\an8}هذا`, and because the marks live in the text, layout and text stay identical, so no indices diverge.

That looked like the answer, until I asked for the character rectangles, which is what Avalonia does on every caret move:

| index | char | `GetTextBounds(i, 1)` |
|---|---|---|
| 0 | `{` | `[368.2, 382.0]` |
| 1 | LRM | `[368.2, 368.2]` |
| 2 | `\` | `[260.1, 273.8]` |
| 3 | `a` | `[260.1, 260.1]` (zero width) |
| 4 | `n` | **THROWS** |
| 5 | `8` | **THROWS** |

Same exception, reached by arrowing into the tag instead of clicking it. Which is exactly what your own README on #12551 already warns: *"Do not fix this with Unicode directional marks in the displayed text - their zero-width glyph runs break Avalonia's TextLine geometry APIs (GetTextBounds, GetCharacterHitFromDistance), which corrupts selection rendering and mouse hit-testing."* You were right, and I can now confirm it from the other side.

So both routes hit one wall: **Avalonia's text box cannot describe a line whose display differs from its raw characters.** Permuting makes layout and text disagree; isolating adds runs it cannot measure. Neither is fixable from SE.

## What I suggest

**Revert this for now.** It trades a cosmetic backslash for a crash on a common action, and I say that as the person it was written for: I edit Arabic subtitles daily, `{an8\}` never changed the look or the behaviour of anything, the stored text was always right, and it has looked that way since SE4.

**Then let Avalonia fix it properly.** [AvaloniaUI/Avalonia#21792](https://github.com/AvaloniaUI/Avalonia/issues/21792) is already confirmed and assigned to @Gillibald, who agreed object runs should default to the neutral direction. There are two more concrete asks that would each make this feature possible:

1. `GetTextBounds` handling zero-width runs instead of throwing, which would make the LRM route work
2. or a seam (`MoveCaretToPoint` virtual, or a layout transform with an index map), which would make this PR's approach work

I would rather wait for @Gillibald's solution than keep working around it. I am happy to test every change they make and to contribute where it helps: I have real Arabic content, a headless harness that does pixel and geometry checks, and the reproductions above are already reduced to the public `TextFormatter` API. When the fix lands, this PR can come back with the translation it needs, and it will be correct instead of clever.

## In the meantime, a design that does not need any of this

The tags that break are ASSA override tags, and the common ones (`{\an8}`, `{\pos}`) are not really text at all: they are properties of the line. If they are not in the text box, bidi never touches them, and none of the above matters.

**Position picker.** A small 3 by 3 grid beside the edit box, laid out like the numpad, which is exactly the ASSA alignment model:

```
 ┌───┬───┬───┐
 │ 7 │ 8 │ 9 │   top
 ├───┼───┼───┤
 │ 4 │ 5 │ 6 │   middle
 ├───┼───┼───┤
 │ 1 │ 2 │ 3 │   bottom
 └───┴───┴───┘
   left  ctr  right
```

Click the top middle cell and the line becomes `{\an8}`; the active cell stays highlighted so the current position is visible at a glance. This needs no new engine work: `SetAlignmentToSelected("an1")` through `("an9")` already exist in `MainViewModel`, so the grid is a view over commands the app already has. It suits SRT too, where `{\an8}` is widely honoured, and it means the text box holds only the subtitle text.

**Other ASSA tags as metadata.** Anything else (`{\pos}`, `{\fad}`, `{\t}`) shown as a small chip in the row above the text box, next to the bookmark chip, opening the existing `AssaApplyCustomOverrideTags` window on click. Same pattern the bookmark already uses, so nothing new to learn.

**Why this is worth doing even after Avalonia is fixed.** It is not a workaround that gets thrown away:

- A 3 by 3 grid is easier to read than `{\an8}`. You see the position instead of decoding a number.
- It removes a class of typos. `{\an\8}` or `{an8}` cannot be produced by clicking a cell.
- It is discoverable. Most users never learn the alignment codes.
- It helps left-to-right users equally, so it is not RTL special-casing.
- The text box goes back to holding text, which is what it is for.

The only case it does not cover is mid-line tags such as `{\c&HFF0000&}` colouring part of a line, which genuinely belong inline. Those would keep the cosmetic backslash until Avalonia lands a fix, and that seems like a fair place to be.

Happy to prototype the picker if you like the direction.

PR #12579, opened 2026-07-17
Fix alpha-channel check in IsImageOnlyTransparent and CopyRectangle rect math

ivandrofly:
Verified fixes for latent bitmap bugs, plus tests pinning them:

- **`NikseBitmap.IsImageOnlyTransparent` tested the blue byte instead of alpha** (`_bitmapData[i]` in BGRA data) — an opaque blue-less image (black/yellow/red/green text) was reported as fully transparent, dropping line parts in the OCR image splitter. The identical bug in the OCR duplicate `NikseBitmap2` is fixed in the same commit.
- **`NikseBitmap.CopyRectangle(SKRect)` passed width/height into `SKRectI`'s (left, top, right, bottom) constructor** — correct only when the section starts at the origin. The clamping statements in the `SKRectI` overload had the same `Rectangle(x, y, w, h)` → `SKRectI(l, t, r, b)` porting bug and are fixed as a follow-up commit, matching the SE 4 original's clamp-to-bitmap intent.
- **New unit tests** pin the alpha-vs-blue regression (verified to fail against the old code) and the CopyRectangle offset/clamping math, for both `NikseBitmap` and `NikseBitmap2`.

The two `FixInvalidItalicTags` commits from the original version of this PR were dropped after review — see comment below.

All 508 libse and 14 libuilogic tests pass.

### DIFF
--- a/src/libse/Common/NikseBitmap.cs
+++ b/src/libse/Common/NikseBitmap.cs
@@ -1178,7 +1178,7 @@ namespace Nikse.SubtitleEdit.Core.Common
 
         public NikseBitmap CopyRectangle(SKRect section)
         {
-            var rect = new SKRectI((int)section.Left, (int)section.Top, (int)section.Width, (int)section.Height);
+            var rect = new SKRectI((int)section.Left, (int)section.Top, (int)section.Right, (int)section.Bottom);
             return CopyRectangle(rect);
         }
 
--- a/src/libse/Common/NikseBitmap.cs
+++ b/src/libse/Common/NikseBitmap.cs
@@ -1186,12 +1186,12 @@ namespace Nikse.SubtitleEdit.Core.Common
         {
             if (section.Bottom > Height)
             {
-                section = new SKRectI(section.Left, section.Top, section.Width, Height - section.Top);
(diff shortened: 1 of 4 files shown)


niksedk:

niksedk:
Thanks for the PR! The two bitmap fixes are verified and merged (with the `SKRectI` clamp statements completed, since they had the same `(x, y, w, h)` → `(l, t, r, b)` porting issue, plus regression tests).

The two `FixInvalidItalicTags` commits were dropped after careful review:

1. The "duplicated" `lines[1].EndsWith` condition is identical in SE 4 (`se4-legacy` branch) — it's decade-old shipped behavior, not an SE 5 porting typo.
2. Behavioral testing showed the change is a regression: the loose condition is load-bearing. Inputs like `</i>Hello\n</i>a</i>World</i>` or `<i>Hello<i>x\n<i>World<i>` are repaired to valid italics today (`<i>Hello</i>` + `<i>aWorld</i>`), but with the tightened condition they fall through *unfixed* — the function returns the invalid tags unchanged, as there is no downstream fallback for these shapes.

PR #12580, opened 2026-07-17
Fix six silent text-corruption bugs in libse

ivandrofly:
Fixes six verified bugs that silently corrupt subtitle text, one commit each, plus tests pinning each fix:

- **DvdStudioPro**: the `allBoldItalic` branch emitted `^U^B^I|^I^B^U` (copy-paste of the underline+bold+italic template), injecting stray underline codes the reader does not parse, so they survived round-trip as literal `^U` text; now `^B^I|^I^B`.
- **FixStartWithUppercaseLetterAfterColon (Turkish)**: the fix operated at the colon's index instead of the letter's, so `"Ali: iyi"` became `"Aliİ iyi"`; now correctly `"Ali: İyi"`.
- **FixUppercaseIInsideWords**: off-by-one guard (`Length - 3` instead of `- 2`) skipped the consonant check when the match ended the stripped text, so `"You HIt."` became `"You Hlt."`; now `"You Hit."`.
- **FileUtil.ReadAllLinesShared**: tested the UTF-16 *BE* BOM (`FE FF`) for `Encoding.Unicode` (LE) — `ReadAllTextShared` directly below has it right — leaving U+FEFF on the first line of UTF-16LE files.
- **Utilities.SplitEndTags**: appended stripped closing tags instead of prepending; since the loop walks right-to-left, `post` came back reversed, so `<font color="white"><i>Hello</i></font>` round-tripped through RemoveSymbols/AddSymbols/ToggleSymbols as invalid `…Hello</font></i>`.
- **Cavena890 (Hebrew)**: `</i>` set `skipCount` to 2 instead of 3, leaking a literal `>` byte into the output (the Chinese branch and the `<i>` handling show the intended pattern).

**Dropped from the original version**: the Cavena890 Chinese `index++` commit. Review showed the whole Chinese write path cannot round-trip regardless: it writes one byte per character (`encoding.GetBytes(...)[0]` — the high byte of the UTF-16BE pair) while the reader decodes the buffer as UTF-16BE byte *pairs*, so adding `index++` for the italic markers does not make italics (or any Chinese text) survive, and the "fix" claim was not verifiable. That path needs a real rewrite in a separate PR.

**Test coverage** (all verified to fail against the pre-fix code): `SplitEndTags` suffix order (nested tags + trailing space), `FixUppercaseIInsideWords` end-of-text consonant check plus mid-text control, UTF-16LE BOM strip, DvdStudioPro bold+italic encode (exact separator, no `^U`) and a load/save round-trip, and the Turkish colon regression test.

All 529 libse tests pass.

### DIFF
--- a/src/libse/Common/FileUtil.cs
+++ b/src/libse/Common/FileUtil.cs
@@ -102,7 +102,7 @@ namespace Nikse.SubtitleEdit.Core.Common
                 return encoding.GetString(bytes, 3, bytes.Length - 3).SplitToLines();
             }
 
-            if (bytes.Length > 2 && Equals(encoding, Encoding.Unicode) && bytes[0] == 0xfe && bytes[1] == 0xff)
+            if (bytes.Length > 2 && Equals(encoding, Encoding.Unicode) && bytes[0] == 0xff && bytes[1] == 0xfe)
             {
                 return encoding.GetString(bytes, 2, bytes.Length - 2).SplitToLines();
             }
--- a/src/libse/Common/Utilities.cs
+++ b/src/libse/Common/Utilities.cs
@@ -3583,7 +3583,7 @@ namespace Nikse.SubtitleEdit.Core.Common
                 updated = false;
                 if (s.EndsWith(' '))
                 {
-                    post += ' ';
+                    post = ' ' + post;
                     s = s.Remove(s.Length - 1, 1);
                     updated = true;
                 }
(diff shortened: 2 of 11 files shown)


niksedk:

niksedk:
Thanks — six of the seven fixes verified cleanly and are merged, with regression tests added for each (SplitEndTags order, FixUppercaseIInsideWords end-of-text, UTF-16LE BOM, DvdStudioPro encode + round-trip, Turkish colon).

The Cavena890 **Chinese** `index++` commit was dropped: the Chinese write path writes one byte per character (`encoding.GetBytes(new[] { current })[0]` takes only the high byte of the UTF-16BE pair) while the read side decodes the text field as UTF-16BE byte *pairs* (even-length aligned). So the path can't round-trip any Chinese text with or without the marker fix, and inserting a lone `0x88`/`0x98` byte actually shifts the pair alignment on read. Making Chinese writing work needs a real fix (two bytes per char + markers in a form the UTF-16BE decode maps back) — worth its own PR.

PR #12581, opened 2026-07-17
Harden binary subtitle parsers against malformed files (crashes, OOM, infinite loops)

ivandrofly:
Adds missing bounds/sanity guards across the binary parsers so malformed or truncated files degrade gracefully instead of crashing the import or hanging the app. Thirteen fixes, one commit each:

- **BluRaySup**: RLE image decode clamped to the declared `width*height` pixel span — a malformed ODS with long runs wrote past the buffer, and the exception escaped the per-segment try/catch via the post-parse merge path, aborting the file open.
- **VobSub**: `SetColor`/`SetContrast` (commands 0x03/0x04) read two argument bytes with no bounds check, unlike the guarded sibling commands.
- **CEA-608 SEI parsing**: bounded the payload-type loop, the 8-byte cc-data header probe, and the cc-triple loop against the buffer length (reachable from the `MP4Parser` constructor with no try/catch); `GetSeiData` also underflowed `endPos - startPos` in ulong arithmetic for NAL sizes 0/1, attempting a huge allocation.
- **CEA-608 layout**: `CcRow.Position` now clamps to the 32-column range in its setter (a malformed PAC indent could index `Chars[-1]`/`Chars[32]`), `MoveCursor`/`InsertChar` clamp locally, and `RollUp` guards a possibly-negative `RemoveAt` index.
- **CEA-708**: the six one-byte-argument window/Delay commands (0x88–0x8D) now check remaining length, mirroring the guarded `SetPen*`/`DefineWindow` commands.
- **MP4**: `stco`/`co64`/`stts` loops bound file-declared entry counts against the actual buffer (mirroring `stsz`/`ctts`), and the `stts` run-length expansion is capped so a `sampleCount` of 0xFFFFFFFF can no longer allocate toward 4 billion entries; `Box` size-0 ("extends to EOF") Position math is corrected and forward progress enforced so a malformed size can't loop forever; the CMAF parser bails on a zero time scale instead of dividing by zero.
- **Transport stream**: DVB subtitle handling accessed `PageCompositions[0]` before its own `Count > 0` check; the PES constructor now validates the buffer is long enough for the fixed header and the private-stream-1 id byte (a declared packet length of 0 produced a too-short buffer).

Build succeeds; all 499 libse tests pass.


niksedk:

niksedk:
Reviewed all thirteen commits — the guards check out (I verified each bound's arithmetic against the reads it protects, and that valid-file behavior is unchanged; the Box size-0 `+8` correctly lands the next-box position at EOF, matching the `Size == 1` path's math). Merged with two follow-up commits on top:

**Completions** (same-class gaps the review found):
- The SEI cc-data header check needed `pos + 8 < buffer.Length` (not `+ 7`): on a match the caller also reads the cc-count byte at `pos + 8`, so a buffer ending right after the two magic words still threw.
- The PES constructor still threw in `GetBinaryString` for a truncated pack whose flags claim PTS/DTS — the two 5-byte timestamp reads are now guarded too (found by the new tests).
- `ctts` has the identical unbounded run-length expansion that was capped in `stts` (up to 4 billion entries) — same cap applied, shared constant.

**Cleanup**: the `InsertChar` clamp block was dead code (the new `Position` setter already clamps), and `MoveCursor`'s per-index guard was redundant once `newPos` is clamped — both removed.

**Tests**: 27 new malformed-input tests across CcRow, SEI/cc-data, CEA-708, Stbl, PES, and VobSub (via a full `GetBitmap` call). The quick-failing classes were run against the pre-hardening code: 19 fail there, confirming they pin the crashes.

PR #12583, opened 2026-07-17
Cleanup: instance-style extension calls, redundant generic args, and redundant casts

ivandrofly:
Three small, behavior-preserving cleanup commits:

1. **Drop redundant `<string>` type argument from `MakeComboBox` calls** — the generic type is inferred from the source collection, so the explicit `<string>` was just noise. Touches the DCinema SMPTE, iTunes Timed Text, Timed Text 1.0, IMSC 1.1 property windows, and the Cut Video window.

2. **Use instance-style extension method calls** — convert static extension-method invocations to instance call syntax: LINQ (`Enumerable.FirstOrDefault`/`Count`/`ToList`), `ObservableCollectionExtensions.AddRange`, `ClipboardExtensions.TryGetTextAsync`, `UiUtil.RemoveControlFromParent`, and `SplitToLines`.

3. **Remove redundant and unnecessary casts** — drop casts the compiler already supplies or that are no-ops: redundant numeric casts around `Math.Round`/division/`TimeSpan.FromSeconds` (integer-division semantics preserved), pointless `(uint)`/`(int)`/`(byte)` casts on already-correct types, leftover `(IEnumerable<T>)` upcasts before LINQ instance calls, and superfluous nullable/reference casts on ternary branches.

No behavior change. Full solution builds clean.


niksedk:

niksedk:
Reviewed hunk by hunk with focus on the cast removals, since "redundant" casts around arithmetic are where cleanups silently change semantics. All verified behavior-preserving:

- The numeric-division hunks keep their original semantics: every removed `(double)` sat where a `* 100.0` / `- 10.0` / `/ 2.0` literal already forced double arithmetic, or (in `AdvancedEffectAudioTextPulse`) the cast applied to the *result* of the truncating `long` division, which is unchanged.
- The overload-resolution traps check out: `TimeCode(double)` and `Paragraph(string, double, double)` are the only matching constructors, so the implicit `ulong`→`double` conversions pick the same overloads the casts did.
- `SeWindowPosition.Width/Height` are `int`, so the window-centering `(int)` casts were true no-ops (not double→int truncations).
- `nuint`→`ulong` in the mpv stride write is an implicit, value-preserving conversion.
- `ClipboardExtensions` is Avalonia's own extension class and `IClipboard` has no instance `TryGetTextAsync`, so instance syntax binds to the same method.
- The target-typed `null` ternaries and `?? 0` coalescing hunks are compile-proven (a type mismatch would not build).

One leftover fixed on top: `FontNames.FirstOrDefault<string>()` kept the redundant generic argument the same commit was removing elsewhere.

Full solution builds clean; all 1,496 tests pass (libse, libuilogic, seconv, UI).

PR #12624, opened 2026-07-18
Pass spell check dictionary as object instead of file name string

ivandrofly:
## Summary
- Spell check dictionary identity moves from a raw `.dic` file-name string to the `SpellCheckDictionaryDisplay` object (`Name` + `DictionaryFileName`): MS Word provider dictionaries have a name but no file, while hunspell dictionaries are only reliably matched by file name (display names are localized and don't round-trip).
- `GetDictionariesViewModel` (download window) now returns a `SpellCheckDictionaryDisplay` built from the catalog entry's English name and the largest downloaded `.dic` path.
- `SpellCheckViewModel.SetLanguage` matches the passed dictionary by file name when one is present, falling back to name match — so a dictionary just downloaded from the catalog (English name) is still selected on non-English UIs.
- `MainViewModel.ShowSpellCheck` falls back to the last-used dictionary from settings when no session dictionary exists, so the spell check window opens with the previous session's language.
- Fixes two settings guards that checked `LastLanguageDictionaryFile` but matched on `LastLanguageDictionaryName` (a stored name with an empty file path was never applied).

## Test plan
- [x] Open spell check, pick a dictionary, close and reopen the app — spell check window opens with the same dictionary instead of re-detecting.
- [x] In spell check, use "Get dictionaries" to download a new dictionary on a non-English UI locale — the downloaded dictionary is selected when the window returns.
- [ ] Repeat the download flow from the OCR window and the pick-dictionary window — the new dictionary is selected in both.
- [ ] With the MS Word spell check provider, run spell check and confirm the last-used language is restored on next run.
- [ ] Run `dotnet test tests/UI/UITests.csproj` — all tests pass.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -5977,9 +5977,21 @@ public partial class MainViewModel :
             return;
         }
 
-        var dictionaryFileName = _currentSpellCheckDictionary?.DictionaryFileName ?? null;
-        var result = await ShowDialogAsync<SpellCheckWindow, SpellCheckViewModel>(
-            vm => { vm.Initialize(Subtitles, SelectedSubtitleIndex, this, dictionaryFileName, _spellCheckSessionInProgress); });
+        // Fall back to the last dictionary used, so the spell check window opens with the same
+        // language as the previous session instead of re-detecting it.
+        var spellCheckDictionary = _currentSpellCheckDictionary;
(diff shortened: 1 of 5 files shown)


ivandrofly:
Previously if you have en_GB and en_US then select en_US and close & reopen en_GB will always be select instead of the last used dictionary.... 

Filename were being used instead of the dictionary...

PR #12645, opened 2026-07-19
LanguageAutoDetect: single-lookup score increments in GetEncodingViaLetter

ivandrofly:
## Summary
- The per-character scoring loop in `GetEncodingViaLetter` does up to 20 `languageScores["xx"]++` operations per character, and each one compiles to a dictionary get + set — two string-hash lookups.
- Adds an `IncrementScore` helper that uses `CollectionsMarshal.GetValueRefOrAddDefault` on .NET 8+ so each increment is a single lookup; the netstandard2.1 build keeps the plain indexer path (`#if NET8_0_OR_GREATER`, matching the existing guards in this file).
- No behavior change — scores, bonuses, and tie-breaking are untouched.

## Test plan
- [ ] `dotnet test tests/libse/LibSETests.csproj --filter "FullyQualifiedName~LanguageAutoDetect"` — all 36 tests pass
- [ ] Load subtitles in Arabic/Korean/Russian/etc. and confirm encoding/language detection picks the same language as before

### DIFF
--- a/src/libse/Common/LanguageAutoDetect.cs
+++ b/src/libse/Common/LanguageAutoDetect.cs
@@ -1968,6 +1968,15 @@ namespace Nikse.SubtitleEdit.Core.Common
 #endif
         }
 
+        private static void IncrementScore(Dictionary<string, int> scores, string language)
+        {
+#if NET8_0_OR_GREATER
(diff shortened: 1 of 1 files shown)


niksedk:

PR #12646, opened 2026-07-19
TransportStreamParser: single-lookup dictionary updates in packet loops

ivandrofly:
## Summary
- `AddToTeletextDictionary` and `ProcessPackages` run per PES packet / teletext page and used the `ContainsKey` + indexer pattern, walking each dictionary two or three times per update.
- `TeletextSubtitlesLookup` and `SubtitlesLookup` are `SortedDictionary`, so those sites now use `TryGetValue` (one tree walk on the hit path instead of two).
- `teletextPages` and `teletextPesList` are plain `Dictionary`; they now go through a small `GetOrAddList` helper backed by `CollectionsMarshal.GetValueRefOrAddDefault` on .NET 8+ (single hash lookup for both hit and miss), with a `TryGetValue` fallback for netstandard2.1.
- No behavior change — same entries, same ordering.

## Test plan
- [ ] `dotnet test tests/libse/LibSETests.csproj` — all 574 tests pass (no dedicated TS parser tests exist)
- [ ] Open a .ts/.m2ts file with DVB and/or teletext subtitles and confirm the same tracks, pages, and lines are listed as before

### DIFF
--- a/src/libse/ContainerFormats/TransportStream/TransportStreamParser.cs
+++ b/src/libse/ContainerFormats/TransportStream/TransportStreamParser.cs
@@ -392,12 +392,11 @@ namespace Nikse.SubtitleEdit.Core.ContainerFormats.TransportStream
             {
                 if (!string.IsNullOrEmpty(dic.Value.Text))
                 {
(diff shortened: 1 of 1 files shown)


niksedk:

PR #12675, opened 2026-07-20
Rent the PES concatenation buffer in transport stream parsing

ivandrofly:
## Summary
- `AddPesPacket` allocated a fresh concatenation buffer for every assembled PES packet while walking a transport stream; the buffer is only needed while the `DvbSubPes` constructor runs (it copies the data it keeps), so it is now rented from `ArrayPool<byte>` and returned in a `finally`
- Rented buffers can be larger than requested, so `DvbSubPes` gains a constructor overload taking an explicit valid-data length; the existing `(byte[], int)` constructor delegates with `buffer.Length`, leaving all other callers unchanged
- The identical `else if (IsPrivateStream1(...))` / `else` branches are collapsed - both produced `new DvbSubPes(pesData, 0)`, and the check cannot run correctly on an oversized buffer

## Benchmark

BenchmarkDotNet rebuilding DVB subtitle PES packets from 6,000 synthetic transport packets (200 PES packets x 30 TS packets, ~5.5 KB payload each) via `MakeSubtitlePesPackets`:

| | Mean | Allocated | Gen0 / Gen1 (per 1k ops) |
|---|---:|---:|---:|
| main | 125.4 μs ± 2.6 | 2.27 MB | 151 / 61 |
| this PR | 94.3 μs ± 1.1 | 1.21 MB | 81 / 40 |

~25% faster and ~47% less allocated. As a control, a replica of main's `new byte[]` implementation run in the same benchmark process measured 126.6 μs - matching the main baseline, so the gain is attributable to the rented buffer. (BenchmarkDotNet v0.15.2, .NET 10.0.10, Windows 11, i7-13700 pinned to P-cores on an idle machine.)

## Test plan
- [ ] `dotnet test tests/libse/LibSETests.csproj` passes (574/574 locally)
- [ ] Open a .ts / .m2ts file with DVB subtitles and verify subtitles import with the same count, timing, and images as before
- [ ] Open a transport stream with teletext subtitles and verify text extraction is unchanged

### DIFF
--- a/src/libse/ContainerFormats/TransportStream/DvbSubPes.cs
+++ b/src/libse/ContainerFormats/TransportStream/DvbSubPes.cs
@@ -36,8 +36,18 @@ namespace Nikse.SubtitleEdit.Core.ContainerFormats.TransportStream
         private readonly byte[] _dataBuffer;
 
         public DvbSubPes(byte[] buffer, int index)
+            : this(buffer, index, buffer.Length)
         {
-            if (buffer.Length < index + 9)
+        }
+
+        /// <summary>
+        /// Same as <see cref="DvbSubPes(byte[], int)"/>, but only the first
+        /// <paramref name="length"/> bytes of <paramref name="buffer"/> are valid
+        /// data - allows passing an oversized buffer, e.g. one rented from ArrayPool.
+        /// </summary>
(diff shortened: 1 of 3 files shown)


niksedk:
Please write "manually tested" or something ....

ivandrofly:
@niksedk  benchmark added

PR #12676, opened 2026-07-20
Reuse one rented buffer for Blu-ray sup segment parsing

ivandrofly:
## Summary
- `ParseBluRaySup` allocated a new byte array for every PGS segment while looping over the stream; the segment size field is 16-bit, so a single 64 KB buffer rented from `ArrayPool<byte>` now covers all segments and is reused for the whole parse
- Reads are bounded by `segment.Size` instead of the buffer length, since the rented array is larger and holds stale data from previous segments: `ParsePicture`'s length guard now checks `segment.Size`, and the composition-object and window-display loops get explicit bounds checks (previously malformed counts relied on `IndexOutOfRangeException` being caught)
- `ParseOds`'s `buffer.Length` guards are dropped - the loop guarantees `segment.Size` bytes were read, making them redundant against the rented buffer

## Benchmark

BenchmarkDotNet parsing a synthetic 200-caption sup stream (PCS/WDS/PDS/ODS/END plus display-set clear per caption, ~4 KB ODS payloads):

| | Mean | Allocated | Gen0 / Gen1 (per 1k ops) |
|---|---:|---:|---:|
| main | 309.8 μs ± 4.1 | 2.40 MB | 160 / 122 |
| this PR | 267.1 μs ± 2.4 | 1.59 MB | 105 / 81 |

~14% faster and ~34% less allocated per parse; GC collections drop proportionally. (BenchmarkDotNet v0.15.2, .NET 10.0.10, Windows 11, i7-13700 pinned to P-cores on an idle machine.)

## Test plan
- [ ] `dotnet test tests/libse/LibSETests.csproj` passes (574/574 locally)
- [ ] Open a Blu-ray .sup file and verify all subtitle images import with the same count, timing, and rendering as before
- [ ] Import PGS subtitles from an .mkv and verify the extracted subtitles are unchanged

### DIFF
--- a/src/libse/BluRaySup/BluRaySupParser.cs
+++ b/src/libse/BluRaySup/BluRaySupParser.cs
@@ -21,6 +21,7 @@ using Nikse.SubtitleEdit.Core.Common;
 using Nikse.SubtitleEdit.Core.ContainerFormats.Matroska;
 using System.Text;
 using SkiaSharp;
+using System.Buffers;
 using System.Collections.Generic;
 using System.IO;
 using System;
@@ -612,7 +613,7 @@ namespace Nikse.SubtitleEdit.Core.BluRaySup
 
         private static PcsData ParsePicture(byte[] buffer, SupSegment segment)
         {
-            if (buffer.Length < 11)
+            if (segment.Size < 11)
             {
                 return new PcsData
                 {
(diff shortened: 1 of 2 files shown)


niksedk:
Please add BenchmarkDotNet performance numbers (the AI should be able to do this)

ivandrofly:
@niksedk benchmark added

PR #12706, opened 2026-07-21
Clear edit box text selection when a subtitle row is tapped

### LINKED ISSUE #12688: Selecting a line after search, doesn't select the line, but it still has the quote selected.
If I search for a quote, and after I found the line wich I want to change.
I then press the line expecting the line to be selected.
If I press Ctrl-I, I expect to Italic the whole line.

But I have selected the quote itself and not the whole line.
I did this with SE4 a lot.

Image

Image

-
I here select the line, but it doesn't make a difference.
The selecting seems to select the line, but it doesn't. 
Image

Image

muaz978:
Fixes #12688.

When a search finds a match, Find selects the found line in the grid and also selects the matched text inside the edit box. Selecting a different line clears that edit box selection in `SubtitleGridSelectionChanged`, but that only runs on an actual selection change. After a search the found row is already the selected row, so re-tapping it raises no `SelectionChanged` and the matched text stays selected. Line level actions such as italic (Ctrl+I) then apply only to the selected fragment instead of the whole line, which is not what the user expects and differs from Subtitle Edit 4.

This collapses the edit box selection on any grid tap in `OnSubtitleGridSingleTapped`, so tapping a row always acts as selecting that line and line level actions apply to the whole line. Pressing Ctrl+I right after a search without clicking the row still italicizes just the found word, so both behaviors remain available.

A portable Windows x64 test build is attached to the issue for verification.

### DIFF
--- a/.github/workflows/release-linux-test.yml
+++ b/.github/workflows/release-linux-test.yml
@@ -1,0 +1,83 @@
+name: Release Linux test build
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_tag:
+        description: 'Release tag (blank = auto timestamp, e.g. linux-test-YYYYMMDD-HHMMSS)'
+        required: false
+        default: ''
+        type: string
+
+jobs:
+  build-linux:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v5
+        with:
+          dotnet-version: '10.0.x'
+
+      - name: Extract version from Se.cs
+        shell: pwsh
+        run: |
+          $seContent = Get-Content "src/ui/Logic/Config/Se.cs" -Raw
+          $m = [regex]::Match($seContent, 'public\s+static\s+string\s+Version\s*\{[^}]+\}\s*=\s*"v([^"]+)"')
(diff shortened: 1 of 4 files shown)


muaz978:
@niksedk this fixes #12688. After a search the found row is already selected, so re-tapping it raises no SelectionChanged and the matched text stays selected, which made Ctrl+I italic apply to just the found fragment instead of the whole line (a regression from SE4). The change collapses the edit box selection on any grid tap, so tapping the line acts on the whole line again, while Ctrl+I right after a search still italicizes only the found word.

The reporter tested a build and confirmed it works as expected. Ready for review when you have a chance.

niksedk:
Hi @muaz978, thanks for digging into this one and for getting the reporter to verify a test build - much appreciated.

I'm closing this while tidying up the open PR list before the 5.2 release. The linked issue (#12688) has been closed by the reporter, and the PR also carries changes to the test-build workflows that I'd rather keep separate from a grid/edit-box fix. If the "selection survives a re-tap after search" behaviour comes up again, I'll do a small targeted change for it in `MainViewModel` - your analysis of why `SelectionChanged` doesn't fire is spot on and saves the digging.

Thanks again!

muaz978:
Thanks Nikolaj, that all makes sense. Agreed the workflow diff didn't belong in a grid/edit-box fix. I'll leave the targeted MainViewModel fix to you, happy to help track it down again if it resurfaces.

PR #12716, opened 2026-07-21
Replace manual byte-shifting helpers with BinaryPrimitives

ivandrofly:
## Summary
- Remove hand-rolled endian read helpers in libse that duplicate `System.Buffers.Binary.BinaryPrimitives` (available on both `netstandard2.1` and `net10.0` targets)
- `Box` (MP4): static `GetUInt`/`GetUInt64`/`GetWord` span helpers removed; instance readers now delegate to `BinaryPrimitives.Read*BigEndian`
- `GetCcDataHelper`: private `GetUInt32` removed in favor of `ReadUInt32BigEndian`
- VobSub `Helper.GetLittleEndian32` removed; `SpHeader` uses `ReadInt32LittleEndian`
- `Tx3gTextOnly`: `GetUInt` removed in favor of `ReadInt32BigEndian`

## Notes
- All replacements are byte-exact equivalents of the removed helpers (endianness, signedness, and widening preserved)
- Removed methods include a few `public static` helpers on the NuGet-published libse assembly (`Box.GetUInt`/`GetUInt64`/`GetWord`, `Helper.GetLittleEndian32`, `Tx3gTextOnly.GetUInt`) — technically a breaking API change for external consumers

## Test plan
- `dotnet build src/libse/LibSE.csproj` — clean, both targets
- `dotnet test tests/libse/LibSETests.csproj` — 618/618 passed

### DIFF
--- a/src/libse/Cea608/GetCcDataHelper.cs
+++ b/src/libse/Cea608/GetCcDataHelper.cs
@@ -1,4 +1,6 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Buffers.Binary;
+using System.Collections.Generic;
 using System.IO;
 using Nikse.SubtitleEdit.Core.Common;
 
@@ -14,7 +16,7 @@ namespace Nikse.SubtitleEdit.Core.Cea608
                 var buffer = new byte[4];
                 fs.Seek((long)i, SeekOrigin.Begin);
                 fs.ReadFully(buffer, 0, buffer.Length);
-                var nalSize = GetUInt32(buffer, 0);
+                var nalSize = BinaryPrimitives.ReadUInt32BigEndian(buffer);
                 var flag = fs.ReadByte();
                 if (IsRbspNalUnitType(flag & 0x1F) && nalSize < 10_000)
                 {
@@ -144,13 +146,8 @@ namespace Nikse.SubtitleEdit.Core.Cea608
             // buffer[pos + 8], so a buffer ending right after the two magic words must not match
             return payloadType == 4 &&
                    pos + 8 < buffer.Length &&
-                   GetUInt32(buffer, pos) == 3036688711 &&
-                   GetUInt32(buffer, pos + 4) == 1094267907;
-        }
-
-        private static uint GetUInt32(byte[] buffer, int pos)
-        {
-            return (uint)((buffer[pos + 0] << 24) + (buffer[pos + 1] << 16) + (buffer[pos + 2] << 8) + buffer[pos + 3]);
+                   BinaryPrimitives.ReadUInt32BigEndian(buffer.AsSpan(pos)) == 3036688711 &&
+                   BinaryPrimitives.ReadUInt32BigEndian(buffer.AsSpan(pos + 4)) == 1094267907;
         }
     }
 }
--- a/src/libse/ContainerFormats/Mp4/Boxes/Box.cs
+++ b/src/libse/ContainerFormats/Mp4/Boxes/Box.cs
@@ -1,5 +1,6 @@
 ﻿using Nikse.SubtitleEdit.Core.Common;
 using System;
+using System.Buffers.Binary;
 using System.Runtime.CompilerServices;
 using System.Text;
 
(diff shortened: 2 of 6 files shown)


niksedk:
@ivandrofly 
Did you add the BenchmarkDotNet performance tests?

niksedk:
I put BenchmarkDotNet numbers behind this refactor — benchmark project is in #12720 (kept separate so this PR stays a clean refactor).

Apple M4, .NET 10.0.7, arm64 RyuJIT, Release. Each benchmark sweeps a 64 KB buffer with `OperationsPerInvoke` normalizing back to ns/read:

| Method | Categories | Mean | Ratio |
|--- |--- |---: |---: |
| Legacy_Int32LittleEndian | Int32LE | 0.5989 ns | 1.00 |
| BinaryPrimitives_Int32LittleEndian | Int32LE | 0.3812 ns | 0.64 |
| Legacy_WordBigEndian | UInt16BE | 0.4959 ns | 1.00 |
| BinaryPrimitives_WordBigEndian | UInt16BE | 0.4025 ns | 0.81 |
| Legacy_UInt32BigEndian | UInt32BE | 0.5672 ns | 1.00 |
| BinaryPrimitives_UInt32BigEndian | UInt32BE | 0.3806 ns | 0.67 |
| Legacy_UInt64BigEndian | UInt64BE | 0.8609 ns | 1.00 |
| BinaryPrimitives_UInt64BigEndian | UInt64BE | 0.3862 ns | 0.45 |

So `BinaryPrimitives` isn't just tidier here, it's faster in every category — 1.2x to 2.2x, widening as the read gets wider. That matches the expected codegen: one `rev` instruction replacing N byte loads plus shifts and ors. `GetUInt64` benefits most, which makes sense as it had the most shifting to do.

Caveats: sub-nanosecond and machine-specific, and endianness intrinsics differ a lot between arm64 (`rev`) and x64 (`movbe`/`bswap`), so don't read these as x64 numbers. The ratios are the meaningful part.

One extra thing that may be more useful than the timings: the benchmark's `GlobalSetup` asserts old and new agree on **every** offset of the test buffer before timing anything, so it independently checks the byte-exactness claim in your PR description — endianness, signedness and widening all preserved. That passes.

Note the benchmark keeps verbatim private copies of the helpers this PR deletes, so #12720 builds against `main` too and doesn't depend on merge order either way.

niksedk:
Followed up with real x64 hardware (the earlier numbers were arm64, which I flagged as non-transferable). Ran via a CI workflow added in #12720, so these are genuine x64 runners rather than Rosetta.

**Ratio vs the legacy shift-chain (lower is better, 1.00 = legacy baseline):**

| Category | linux-x64<br>Xeon 6973P | windows-x64<br>EPYC 7763 | macos-arm64<br>M1 (virtual) | local<br>M4 |
|--- |---: |---: |---: |---: |
| Int32LE | **0.48** | **0.48** | 0.66 | 0.64 |
| UInt16BE | 0.89 | 0.91 | 1.07 | 0.81 |
| UInt32BE | **0.73** | **0.47** | 0.76 | 0.67 |
| UInt64BE | **0.47** | **0.33** | 0.51 | 0.45 |

Absolute means, for reference:

| Method | linux-x64 | windows-x64 | macos-arm64 |
|--- |---: |---: |---: |
| Legacy_Int32LittleEndian | 0.9972 ns | 1.3225 ns | 1.3138 ns |
| BinaryPrimitives_Int32LittleEndian | 0.4766 ns | 0.6331 ns | 0.8465 ns |
| Legacy_WordBigEndian | 0.7261 ns | 0.7873 ns | 0.8237 ns |
| BinaryPrimitives_WordBigEndian | 0.6490 ns | 0.7188 ns | 0.8800 ns |
| Legacy_UInt32BigEndian | 0.8649 ns | 1.3408 ns | 1.1150 ns |
| BinaryPrimitives_UInt32BigEndian | 0.6314 ns | 0.6324 ns | 0.8430 ns |
| Legacy_UInt64BigEndian | 1.2246 ns | 1.9460 ns | 1.4535 ns |
| BinaryPrimitives_UInt64BigEndian | 0.5798 ns | 0.6340 ns | 0.7433 ns |

So the x64 win is **bigger** than on arm64, not smaller — up to 3x on `GetUInt64` on the EPYC box. That fits `movbe`/`bswap` collapsing a chain of 8 byte loads, shifts and ors into essentially one instruction. `GetUInt64` had the most shifting to lose, so it gains the most.

Two honest caveats: the 16-bit case is close to a wash everywhere (0.89-1.07) — a 2-byte swap was never going to be the bottleneck — and the macos-arm64 runner shows it at 1.07, i.e. marginally *slower*, which at sub-nanosecond on a shared virtual runner is noise rather than a real regression. The local M4 result for the same case was 0.81.

Regardless of the timings, the equality assertion in `GlobalSetup` passed on all three architectures, which is the part I'd weight most for this PR.

PR #12739, opened 2026-07-22
Fix event handler leak by stopping timers on window close

ivandrofly:
## Summary
- Add `TimerExtensions.StopAndDispose(this System.Timers.Timer, ElapsedEventHandler)` — unsubscribes the `Elapsed` handler, stops, and disposes the timer in one call
- Convert lambda `Elapsed` handlers to named instance methods across all dialog view models that run debounce/preview/progress `System.Timers.Timer`s, so the handler can be unsubscribed
- Call `StopAndDispose` in every window-close path (OK, Cancel, Escape, `OnClosing`/`Close()` funnels) — 51 view models across Tools, Video, Files, Shared, OCR, SpellCheck, SSA/ASSA, Options, and Edit features
- Prevents closed windows from being kept alive and ticking by their timers; extends the pattern started in #12715 (Convert Actors)
- Intentionally untouched: app-lifetime timers (`MainViewModel`, `AutoBackupService`), the locally-scoped `using var` timer in `VideoOcrViewModel`, and `DispatcherTimer`s

## Test plan
- [ ] Open a Tools dialog with a live preview (e.g. Bridge gaps, Merge short lines, Sort by, Change formatting), close via OK, Cancel, and Escape — no preview updates or errors after close, and reopening works normally
- [ ] Open Tools > Batch convert, add files, close via OK/Escape — the file scan timer no longer ticks after close
- [ ] Run a download dialog (e.g. FFmpeg, libmpv, Tesseract, Whisper models) to completion and also cancel one mid-download — no timer ticks after the window closes
- [ ] Video > Speech to text / Text to speech / Cut video / Shot changes: close mid-idle and after a run — no background activity remains
- [ ] File > Export (image based / plain text) and Import plain text: close each dialog and verify no further preview updates fire
- [ ] Spell check: close the window and confirm the status text timer stops

### DIFF
--- a/src/ui/Features/Assa/AssaStylesViewModel.cs
+++ b/src/ui/Features/Assa/AssaStylesViewModel.cs
@@ -24,7 +24,7 @@ using System.Threading.Tasks;
 
 namespace Nikse.SubtitleEdit.Features.Assa;
 
-public partial class AssaStylesViewModel : ObservableObject
+public partial class AssaStylesViewModel : ObservableObject, IClosingCleanup
 {
     [ObservableProperty] private string _title;
     [ObservableProperty] private ObservableCollection<StyleDisplay> _fileStyles;
@@ -63,6 +63,7 @@ public partial class AssaStylesViewModel : ObservableObject
     private IApplyAssaStyles? _applyAssaStyles;
     private Subtitle _subtitle;
     private string _subtitleFileName;
+    private volatile bool _isClosing;
     private readonly System.Timers.Timer _timerUpdatePreview;
     private readonly List<string> _extraCategories = new();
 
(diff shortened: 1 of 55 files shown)


niksedk:
Pushed a follow-up commit that reworks the approach to close two gaps in the original per-command disposal:

**1. Leak on title-bar X / Alt+F4.** Disposing only in `Ok`/`Cancel`/`Escape` missed the window-close paths that run none of them, so the timer kept the closed window alive. Now `IClosingCleanup.OnClosingCleanup()` is invoked once from `window.Closed`, wired centrally in `UiUtil.InitializeWindow` (the single funnel every window goes through). Every close path is covered, and the scattered per-command `StopAndDispose` calls are removed.

**2. Dispose-vs-in-flight-`Start()` crash.** The preview/debounce handlers reschedule themselves with `timer.Start()`. If a close disposed the timer mid-handler, `Start()` on the disposed timer throws `ObjectDisposedException` — which `System.Timers.Timer` no longer swallows on modern .NET, so it crashed the process from a thread-pool thread. Self-restarting handlers now carry a `volatile bool _isClosing` guard: they bail out at the top and never `Start()` once closing.

View models already disposing from a window `Closing` hook (`CutVideo`, `SpeechToText`, `TextToSpeech`, `BeautifyTimeCodes`) keep that path; the self-restarting ones among them (`ReviewSpeech`, `ReviewSpeechHistory`, `DownloadTts`) gained the same guard, and `BeautifyTimeCodes` reuses its existing `_disposed` flag.

`TimerExtensions.StopAndDispose` is unchanged. UI test suite: **764 passed**, build clean (0 warnings).

Issue #12233, opened 2026-07-06
Cursor keeps moving after pausing playback in Waveform


geroyannis:
I still have the issue where the cursor in Waveform keeps moving for a while after I pause playback. It's not moving by the same amount every time. Sometimes it's more, sometimes it's less. This is a problem when trying to insert subtitles fast.

I don't know if it helps, but I'm 87% sure that during playback, the cursor doesn't move at constant speed. I think that sometimes it speeds up and then slows down.

We've talked again about it, and you told me to send you a short screen recording if it keeps happening. In the attached video sample, you can see the moment I press the pause button on the toolbar and the cursor keeps moving for a while. I hope it's uploaded correctly.

https://github.com/user-attachments/assets/25cba7d9-0094-439f-84f8-36731ee130a5

I also want to take a moment and just say thank you for the fantastic job you're doing!

PR #12742, opened 2026-07-23
Fix waveform playhead drift after pausing (#12740)

### LINKED ISSUE #12740: Waveform playhead briefly moves after pause while video time is already stopped
I found a waveform playhead issue in Subtitle Edit 5.

After pausing playback, the video time display stops correctly, but the blue waveform playhead sometimes continues to move briefly on its own.

Important detail:

The video time display remains completely unchanged after pausing, so playback itself is already stopped. Only the waveform playhead continues to move visually for a moment.

Observed behavior:

After pausing, the blue waveform playhead may:

move slightly backward,
move slightly forward,
or remain fixed.

The issue is intermittent. When it occurs, the playhead may move slightly in either direction.

I tested this with:

Play/Pause button
Space

The result is the same in both cases.

I also confirmed the same behavior in RC11, so this is not a new RC12 regression.

Tested versions:

Subtitle Edit 5.1.0 RC11 ZIP
Subtitle Edit 5.1.0 RC12 ZIP

Expected behavior:

When playback is paused, the waveform playhead should stop immediately and remain fixed, with no additional visual movement afterward.

For reference, Subtitle Edit 4 stops the waveform playhead immediately and does not show this extra motion after pausing.

I am attaching a short recording showing three Play/Pause cycles. In each case, the video time display remains fixed while the blue waveform playhead may still move briefly after pausing.

Environment:

Windows 11
Subtitle Edit 5.1.0 RC12 ZIP

Attachment:

https://github.com/user-attachments/assets/f659cdec-fb18-4236-bab0-35e69db17173

niksedk:
Fixes #12740.

### Problem
After pausing playback (spacebar, play/pause button, or clicking the video), the blue waveform playhead drifts a tiny bit forward or backward a moment *after* the video time display has already stopped.

### Root cause
The waveform cursor isn't the video player's real position — it's an interpolated estimate (`_playheadEstimateSeconds`, driven at ~60 fps by `_cursorTimer`). On pause it was frozen at the keypress instant and then, once a 300 ms hold window expired, **eased** toward mpv's settled position over several ticks. mpv keeps decoding ~100–200 ms past the keypress, so its final frame differs from the frozen spot — and that delayed easing to close the gap is exactly the drift. Direction depends on where in mpv's decode cycle the pause landed (ahead → forward; behind an over-extrapolated estimate → backward). The numeric time display, by contrast, reads mpv's raw position directly and settles cleanly, which is why only the cursor appears to move.

### Fix
Replace "hold 300 ms, then ease" with "hold dead-frozen through mpv's wind-down, then **snap once** to the true final frame the instant mpv's clock comes to rest" (reported position unchanged for 100 ms). All cursor motion now happens coincident with the pause completing — no separate movement afterward. A new `_playheadPausedSettled` flag guards the one-time snap:

- cleared by `RequestPausePlayheadFreeze` and on the play→pause edge (covers non-command pauses),
- set by `PinPlayheadTo`, so waveform clicks / stop / next-prev hold exactly where the user aimed instead of twitching to mpv's nearest reported position.

Instant freeze-on-keypress (#12033 / #12233) is preserved; centering and selection paths are untouched. Removed the now-unused `PlayheadPausedEaseFactor`.

### Testing
Built clean (0 warnings). Verified in the running app: pausing via spacebar, the play/pause button, and the video surface all stop the playhead with no post-pause drift; stop snaps to 0; waveform clicks / next-prev while paused land and hold on target; normal playback still glides and in-play seeking still tracks.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -382,7 +382,8 @@ public partial class MainViewModel :
     private bool _pauseRequested; // a pause command fired; freeze the cursor now, before mpv's IsPlaying flips
     private long _playheadPauseSettleTs; // when we last paused; briefly hold the estimate so it doesn't snap back
     private const double PlayheadPauseSettleMs = 300; // mpv's reported position lags ~100-200 ms after a pause
-    private const double PlayheadPausedEaseFactor = 0.25; // per-tick fraction for closing small paused-state gaps (~0.1 s converges in ~10 ticks / ~160 ms)
+    private bool _playheadPausedSettled; // set once mpv's clock has come to rest after a pause and the cursor has landed on it
+    private const double PlayheadPausedSettleStableMs = 100; // mpv's position unchanged this long after a pause = its clock is at rest -> snap the cursor there once
     private CancellationTokenSource _videoOpenTokenSource;
     private readonly HashSet<string> _waveformsBeingGenerated = new(StringComparer.OrdinalIgnoreCase);
     private readonly Lock _waveformsBeingGeneratedLock = new();
@@ -1186,6 +1187,7 @@ public partial class MainViewModel :
     internal void RequestPausePlayheadFreeze()
     {
         _pauseRequested = true;
+        _playheadPausedSettled = false; // reconcile the cursor to mpv's final frame once this pause settles
         _playheadPauseSettleTs = Stopwatch.GetTimestamp();
     }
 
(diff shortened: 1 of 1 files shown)


Surlime:
I believe this fix has caused another issue, now the waveform playhead doesn't always stay center when you do small step seeks (both in millisecond and frame mode), please see the examples below.

---
<details>
<summary>Millisecond mode</summary>
SubtitleEdit_7TLl8t5JHL-millisec</summary>
</details>

---
<details>
<summary>Frame mode</summary>
SubtitleEdit_bDgm8ZJhuo-frame
</details>

---
Please take a look when you have a chance. 🙏 @niksedk

niksedk:
Good catch — the pause-drift fix stopped following small paused seeks too. Fixed in #12751: the cursor now follows millisecond/frame step-seeks (so it stays centered) while still holding through the pause wind-down. Will be in the next RC. 🙏

PR #12754, opened 2026-07-23
Simplify encoding detection to return single best match

ivandrofly:
## Summary
- Replace the LINQ-based `DetectInputCodePages(input, maxEncodings)` with a simple loop that returns only the highest-confidence encoding
- Remove the unused `maxEncodings` parameter and its `ArgumentOutOfRangeException` guard, plus the `System.Linq` using
- `DetectInputCodepage` now falls back to `Encoding.Default` only when detection yields no encoding

## Benchmark

BenchmarkDotNet, 1000 iterations per method, in-process (13th Gen Intel Core i7-13700, x64 RyuJIT):

| Method    | Mean      | Error    | StdDev    | Median    | Ratio | Allocated |
|---------- |----------:|---------:|----------:|----------:|------:|----------:|
| Old_Small |  66.57 us | 0.166 us |  1.542 us |  66.20 us |  1.00 |  96.36 KB |
| New_Small |  63.18 us | 0.174 us |  1.611 us |  62.81 us |  0.95 |  95.91 KB |
| Old_Large | 821.40 us | 1.891 us | 17.467 us | 816.26 us | 12.35 |    481 KB |
| New_Large | 778.60 us | 2.285 us | 21.283 us | 774.91 us | 11.70 | 480.55 KB |

`Small` = ~1 KB windows-1252 subtitle text, `Large` = ~460 KB UTF-8. The loop is ~5% faster than the LINQ version on both inputs (outside the 99.9% confidence intervals); allocations are essentially unchanged since `CharsetDetector.DetectFromBytes` dominates. Mainly a simplification, with a small measured speedup.

## Test plan
- [x] Open subtitle files in various encodings (UTF-8 with/without BOM, Windows-1252, UTF-16) and verify they load with correct characters
- [x] Open an empty subtitle file and verify it loads without error
- [x] Run `dotnet test tests/libse/LibSETests.csproj` and verify all tests pass

### DIFF
--- a/src/libse/DetectEncoding/EncodingTools.cs
+++ b/src/libse/DetectEncoding/EncodingTools.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Linq;
 using System.Text;
 using UtfUnknown;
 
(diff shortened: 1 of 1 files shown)


src/libse/DetectEncoding/EncodingTools.cs
```diff
@@ -39,12 +31,24 @@ private static Encoding[] DetectInputCodePages(byte[] input, int maxEncodings)
             // empty strings can always be encoded as ASCII
             if (input.Length == 0)
             {
-                return new[] { Encoding.ASCII };
+                return Encoding.ASCII;
             }
 
             // use UTF.Unknown to detect from input byte string
             var detectionResult = CharsetDetector.DetectFromBytes(input);
-            return detectionResult.Details.OrderByDescending(p => p.Confidence).Select(p => p.Encoding).Take(maxEncodings).ToArray();
```
ivandrofly, line 47: from `nlogN => n` no need to sort as we are always selected the a[0] and taking the encoding with highest confidence, also no array allocation to return list of matchings

niksedk:
Did you forget the BenchmarkDotNet tests? ;)

ivandrofly:
> Did you forget the BenchmarkDotNet tests? ;)

added but I think it was pretty obvious as the old one was using sort

PR #12806, opened 2026-07-24
Micro-optimize hot helpers in UiUtil

niksedk:
Five micro-optimizations in `UiUtil.cs`, each verified with BenchmarkDotNet (.NET 10 Release, Apple M4):

| Method | Change | Old → New | Alloc |
|---|---|---|---|
| `TryGetAccessKey` | `Enum.TryParse(upper.ToString())` / `"D"+upper` → direct enum arithmetic | 915 ns → 4.8 ns | 104 B → 0 |
| `GetTextColor` / `GetBorderBrush` | `new SolidColorBrush` per call → cached `ImmutableSolidColorBrush` per theme/opacity | 126 ns → 0.2 ns | 616 B → 0 |
| `GetGridLinesVisibility` | three `enum.ToString()` comparisons → `nameof` switch | 9.5 ns → 0.03 ns | 24 B → 0 |
| `DrawCheckerboardBackground` | `SKColor.Parse(...)` per call → cached static `SKColor`s | 20 ns → 0.06 ns | – |
| `GetDefaultFontName` | new 14-item `List<string>` per call → static readonly array | 295 ns → 225 ns | 328 B → 0 |

The brush caching is the most meaningful one in practice: `GetBorderBrush`/`GetTextColor` run from dozens of window-construction paths (borders, separators, DataGrid cell themes, NumericUpDown foregrounds), so each window previously churned many short-lived `SolidColorBrush` AvaloniaObjects. The cache is keyed by theme so theme switches keep working, and `ImmutableSolidColorBrush` has no dispatcher affinity so sharing is safe.

No behavior changes intended; `src/ui` builds with 0 warnings/0 errors.

### DIFF
--- a/src/ui/Logic/UiUtil.cs
+++ b/src/ui/Logic/UiUtil.cs
@@ -280,12 +280,29 @@ public static class UiUtil
         if (Se.Settings.Appearance.UseFocusedButtonBackgroundColor)
         {
             var focusStyle = new Style(x => x.OfType<Button>().Class(":focus"));
-            focusStyle.Setters.Add(new Setter(Button.BackgroundProperty, new SolidColorBrush(Se.Settings.Appearance.FocusedButtonBackgroundColor.FromHexToColor())));
+            focusStyle.Setters.Add(new Setter(Button.BackgroundProperty, GetFocusedButtonBackgroundBrush()));
             //focusStyle.Setters.Add(new Setter(Button.ForegroundProperty, Brushes.White));
             button.Styles.Add(focusStyle);
         }
 
         return button;
+    }
+
+    private static IBrush? _focusedButtonBackgroundBrush;
+    private static string? _focusedButtonBackgroundHex;
+
+    // Every button parses the configured hex color and allocated a brush; cache one
+    // immutable brush and only rebuild it when the setting changes.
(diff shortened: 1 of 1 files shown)


niksedk:
Round 2 — four more BenchmarkDotNet-verified micro-optimizations added (`d8fbbbe55`):

| Method | Change | Old → New | Alloc |
|---|---|---|---|
| `MakeButton` | cache focused-button background brush instead of hex parse + `SolidColorBrush` per button | 190 ns → 72 ns | 1144 B → 528 B |
| `MakeColorPickerButton` | cache `PropertyInfo` per (type, name) for swatch reads/refreshes | 33 ns → 16 ns | – |
| `ShowHelp` | remove `string.Format` around an already-interpolated string (also a latent `FormatException` if a help name contained braces) | 53 ns → 15 ns | 496 B → 176 B |
| `MakeToolTip` | early return when `ShowHints` is off, before computing the shortcut string | 280 ns → 0.2 ns | 760 B → 0 |

A fifth candidate (`MakeShortcutsString` LINQ → manual loop + StringBuilder) benchmarked **slower** (300 ns → 311 ns, +80 B) and was rejected.

PR #12809, opened 2026-07-24
Fix out-of-range parsing of culture names with unbalanced parentheses

ivandrofly:
## Summary

Both spell check view models extract the region part of a `CultureInfo` name (e.g. `Português (Portugal)` → `Portugal`) by searching for `(` and `)` independently. The guard only checked that both characters were present, so a name containing a `)` before (or without) a `(` produced a negative length and `Substring` threw `ArgumentOutOfRangeException`.

- `GetDictionariesViewModel.LoadDictionaries` — search for `)` starting at the position after `(` and require it to come later, so unbalanced names fall through to the unmodified name instead of throwing.
- `SpellCheckViewModel` — same guard for the MS Word dictionary lookup, which previously computed `end - start - 1` with no ordering check. The extracted region is now sliced with `AsMemory` so the two `Contains` lookups do not allocate an intermediate string.

## Test plan

- [ ] `dotnet test tests/UI/UITests.csproj` and `dotnet test tests/libse/LibSETests.csproj` pass
- [ ] Open Spell check → Get dictionaries under a normal locale (e.g. `pt-PT`, `en-US`) and confirm the matching dictionary is still preselected
- [ ] With the MS Word spell check provider on Windows, start a spell check and confirm the installed Word language matching the subtitle language is still auto-selected
- [ ] Confirm neither window throws when the current culture's name has no region part (e.g. `en`, `pt`)

### DIFF
--- a/src/ui/Features/SpellCheck/GetDictionaries/GetDictionariesViewModel.cs
+++ b/src/ui/Features/SpellCheck/GetDictionaries/GetDictionariesViewModel.cs
@@ -295,11 +295,11 @@ public partial class GetDictionariesViewModel : ObservableObject, IClosingCleanu
         }
 
         var englishName = CultureInfo.CurrentCulture.EnglishName;
-        if (englishName.Contains('(') && englishName.Contains(')'))
+        var startIndex = englishName.IndexOf('(') + 1;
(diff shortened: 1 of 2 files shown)


niksedk:
Did you test spell check manually? Please always write that you have tested, thx :)

PR #12810, opened 2026-07-25
Deserialize JSON directly from streams instead of via strings

ivandrofly:
Reading a whole JSON file (or embedded asset) into a `string` only to hand that string to `JsonSerializer` allocates the payload twice — once as UTF-16 text, and again while the serializer transcodes it back to UTF-8 to parse it. The `Stream` overloads read UTF-8 straight from the source, so this drops the intermediate string at every such call site.

| File | Change |
| --- | --- |
| `SpellCheck/GetDictionaries/GetDictionariesViewModel.cs` | drop `StreamReader.ReadToEnd()` on the `HunspellDictionaries.json` asset stream |
| `seconv/Core/SeConvSettings.cs` | `File.ReadAllText` → `File.OpenRead` |
| `Logic/Ocr/GoogleLens/CookieManager.cs` | `File.ReadAllText` → `File.OpenRead` |
| `Options/Settings/WaveformThemes/WaveformThemesViewModel.cs` | → `DeserializeAsync` |
| `Options/Settings/ProfilesViewModel.cs` | → `DeserializeAsync` |
| `Options/Shortcuts/ShortcutsViewModel.cs` | → `DeserializeAsync` |
| `Video/TextToSpeech/TextToSpeechViewModel.cs` | → `DeserializeAsync`, keeps the cancellation token |

The four sites already inside `async` methods use `DeserializeAsync`, so importing a settings / profile / waveform-theme / TTS file no longer blocks the UI thread on file I/O. The explicit `Encoding.UTF8` in the shortcuts import goes away with it — the stream overload handles UTF-8 and its BOM.

`SettingsImportExportViewModel` deliberately keeps its string: `TryReadExportSourceOs` needs the same JSON a second time, so streaming there would just mean reading the file twice.

No behaviour change. Solution builds clean.

### DIFF
--- a/src/seconv/Core/SeConvSettings.cs
+++ b/src/seconv/Core/SeConvSettings.cs
@@ -46,8 +46,8 @@ internal sealed class SeConvSettings
             throw new FileNotFoundException($"Settings file not found: {path}", path);
         }
 
-        var json = File.ReadAllText(path);
-        return JsonSerializer.Deserialize<SeConvSettings>(json, JsonOptions)
+        using var stream = File.OpenRead(path);
+        return JsonSerializer.Deserialize<SeConvSettings>(stream, JsonOptions)
             ?? throw new InvalidDataException($"Settings file is empty or null: {path}");
     }
 
--- a/src/ui/Features/Options/Settings/ProfilesViewModel.cs
+++ b/src/ui/Features/Options/Settings/ProfilesViewModel.cs
@@ -117,8 +117,8 @@ public partial class ProfilesViewModel : ObservableObject
         List<ProfileDisplay>? imported = null;
         try
         {
-            var json = System.IO.File.ReadAllText(fileName);
-            var temp = JsonSerializer.Deserialize<ProfileImportExport>(json, new JsonSerializerOptions { PropertyNameCaseInsensitive = true });
+            await using var stream = System.IO.File.OpenRead(fileName);
+            var temp = await JsonSerializer.DeserializeAsync<ProfileImportExport>(stream, new JsonSerializerOptions { PropertyNameCaseInsensitive = true });
             if (temp == null)
             {
                 imported = new List<ProfileDisplay>();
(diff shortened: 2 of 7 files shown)


niksedk:
Please **always** include BencmarkDotNet numbers, thx :)

niksedk:
Benchmarked the two approaches with BenchmarkDotNet on the real `HunspellDictionaries.json` (29 KB) and a ~1 MB payload built by repeating its entries. Each benchmark does the full call-site work: open/read the file + deserialize to `List<DictEntry>` with the same `JsonSerializerOptions` as the PR sites.

| Method | Mean | Ratio | Gen0 | Gen1 | Gen2 | Allocated | Alloc Ratio |
|---|---:|---:|---:|---:|---:|---:|---:|
| `ReadAllText` + `Deserialize(string)`, 29 KB | 48.73 μs | 1.00 | 18.43 | 3.05 | – | 151.11 KB | 1.00 |
| `OpenRead` + `Deserialize(Stream)`, 29 KB | 38.95 μs | 0.80 | 3.42 | 0.31 | – | 28.31 KB | **0.19** |
| `ReadAllText` + `Deserialize(string)`, 1 MB | 1,314.53 μs | 26.97 | 488.28 | 300.78 | 214.84 | 3,475.72 KB | 23.00 |
| `OpenRead` + `Deserialize(Stream)`, 1 MB | 1,227.99 μs | 25.20 | 125.00 | 62.50 | – | 1,034.39 KB | **6.85** |

Environment: Apple M4, .NET 10.0.7, BenchmarkDotNet v0.15.8, macOS 26.5.

Takeaways:

- **~20% faster and ~5× fewer allocations** on the 29 KB asset; the intermediate UTF-16 string plus the serializer's UTF-8 transcode buffer are the bulk of the string path's 151 KB.
- On the 1 MB payload the string path allocates **3.4 MB per call and triggers Gen2 collections** (the UTF-16 string lands on the LOH); the stream path stays out of Gen2 entirely.
- Wall-clock gains are modest since deserialization dominates — the win is chiefly GC pressure.

PR #12831, opened 2026-07-25
Add grammar check via LanguageTool

ivandrofly:
## Summary

A new **Tools → Grammar check (LanguageTool)…** window that checks the subtitle against a [LanguageTool](https://languagetool.org) server - the public API or a self-hosted one (e.g. the [docker image](https://github.com/Erikvl87/docker-languagetool)) - and applies the fixes the user ticks. It complements AI review: rule based, so it points at an exact span and offers concrete replacements instead of rewriting a line, it is much faster, and with a self-hosted server the text never leaves the machine.

- **`src/libuilogic/Grammar/`** — `LanguageToolClient` (`/v2/check`, `/v2/languages`, picky level, disabled rules, optional premium credentials), `LanguageToolAnnotatedText` (builds the annotated document and maps the answer back), `LanguageToolFix` (applies several fixes to one line). Kept out of the UI project so `seconv` can use it later.
- **`src/ui/Features/Tools/GrammarCheck/`** — the window and its view models. Server box with a test-connection button, language list served by the server (defaulting to the auto-detected subtitle language), picky toggle, category filter chips, a checkbox grid with before/after diffs, a replacement picker for the selected row, and a small settings dialog for user name / API key / disabled rules / lines per request.
- Menu entry (both the normal and the native macOS menu), DI registration, `Se.Settings.Tools.GrammarCheck`, localizable strings, and a docs page.

Formatting tags, ASSA override blocks, music symbols and the line break inside a subtitle are sent as `markup`, so LanguageTool neither checks nor counts them as words, while the offsets it returns still point into the original line - a fix lands exactly where it belongs and the tags stay untouched. An issue spanning a tag or a line break is dropped rather than applied. Lines whose sentence continues into the next subtitle are joined with `interpretAs`, so a sentence is checked the way it reads on screen and agreement errors spread over two lines are found; fixes are still applied per line, so timing and reading speed are unaffected.

Nothing is applied automatically: spelling/grammar/punctuation/casing fixes start ticked, style suggestions start unticked, and rules that offer no replacement cannot be ticked at all.

## Test plan

- [x] Tools → Grammar check (LanguageTool)… opens; the language drop-down fills and the status line reads "Connected - N languages available"
- [ ] Point **Server** at an unreachable address and press the refresh button - the error is shown in the status line, the window stays usable
- [ ] Check an English subtitle containing e.g. `<i>He go to school</i>` / `I has a apple` - issues appear per line with category tags and before/after diffs
- [ ] Select a row with several replacements - the drop-down under the grid appears, picking another option updates the After column
- [ ] Filter with the category chips; counts match the number of rows in each category
- [ ] Untick some rows, press **Apply N fixes** - only the ticked fixes are applied, several fixes on one line all land correctly, and Ctrl+Z reverts the whole run in one step
- [ ] A line wrapped in `<i>…</i>` or starting with `{\an8}` keeps its tags after applying a fix inside it
- [ ] A sentence split over two subtitles gets its agreement error reported, and applying the fix does not move words between the lines
- [ ] Press **Stop** mid-check - what was found so far stays in the grid
- [ ] Settings… - user name/API key/disabled rules/lines per request survive a restart; a rule id in "disabled rules" makes that rule stop firing

## UI

image

### DIFF
--- a/docs/features/grammar-check.md
+++ b/docs/features/grammar-check.md
@@ -1,0 +1,52 @@
+# Grammar Check (LanguageTool)
+
+Check spelling, grammar, punctuation, casing and style with a [LanguageTool](https://languagetool.org) server - the public API, or your own installation (for example the [docker image](https://github.com/Erikvl87/docker-languagetool)). Nothing is changed until you apply the fixes you agree with.
+
+- **Menu:** Tools → Grammar check (LanguageTool)...
+
+Unlike [AI Review](ai-review.md), LanguageTool is rule based: it points at an exact span of text and offers concrete replacements, so it never rewrites a line and never invents changes. It is also considerably faster, and with a self-hosted server the subtitle text never leaves your machine.
+
+## Server
+
+Enter the base address of the server in the **Server** box - `https://api.languagetool.org` for the public API, or something like `http://localhost:8010` for a local docker container. The address may include the `/v2` or `/v2/check` suffix; Subtitle Edit strips it.
+
+The circular arrow button next to the box tests the connection and reloads the language list, which also happens automatically when the window opens.
+
+> The public API is rate limited and rejects large amounts of text from one address. For a whole subtitle, a self-hosted server is the better choice.
+
+## Language
+
+The language list comes from the server. It defaults to the auto-detected language of the subtitle, and falls back to the server's own detection if you leave it on **Auto**. For languages with variants (English, German, Portuguese...) pick the right one - the variants disagree about spelling and punctuation.
+
+**Picky** turns on the stricter rules LanguageTool leaves off by default: redundancy, wordiness and typography suggestions. Useful for a final polish, noisy for a first pass.
+
+## Checking
+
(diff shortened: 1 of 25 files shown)


ivandrofly:
I also have https://languagetool.ivandrofly.org that any one can use or you can run your own https://github.com/Erikvl87/docker-languagetool

niksedk:
Hi @ivandrofly, thanks a lot for the work on this - the tag/line-break handling and the cross-subtitle sentence joining are really nicely thought out, and the UI looks good.

I'm going to close this one for now though. The main sticking point is the dependency on a LanguageTool server: either the text goes to the public API, or the user has to run a Java/Docker setup themselves, and neither fits well with where SE 5 is heading (everything running locally, no external services by default).

Longer term I'm hoping we can do an AI-assisted port of the LanguageTool rule engine to C#, so the grammar check can run fully in-process with no server at all. If that materializes, a lot of what you built here - the annotated-text mapping, the fix application, the review grid - could come straight back, so please don't throw the branch away 🙂

Thanks again!

PR #12884, opened 2026-07-27
Replace remaining object lock fields with System.Threading.Lock

ivandrofly:
## Summary
- Finishes the codebase's migration to the .NET 9+ `System.Threading.Lock` type (54 `Lock` fields already existed) by converting the last four `object` lock fields and two method-local lock objects
- `SubtitleFormat.SubtitleFormatsLock` (libse) is guarded with `#if NET9_0_OR_GREATER` because LibSE also targets `netstandard2.1`, where `Lock` does not exist; the `#else` branch keeps the original `object` declaration
- Converts `FontFaces.Gate`, `PaddleOcr._errorLock` (which previously sat alongside an already-migrated `Lock _lock` in the same class), `CrispAsrTtsProvenance.SupportCacheLock`, and both `lockObj` locals guarding `Parallel.For` progress counters in `BatchConverter`
- All fields were verified to be lock-only (no `Monitor.Wait/Pulse`, never passed or reassigned), so `lock (...)` statements need no changes

## Test plan
- [x] `dotnet build SubtitleEdit.sln` succeeds with no warnings (both LibSE target frameworks compile; no CS9216 "Lock converted to object" warnings)
- [x] `dotnet test tests/libse/LibSETests.csproj` passes (745/745)
- [x] `dotnet test tests/UI/UITests.csproj` passes (906/907, 1 pre-existing skip)
- [ ] Batch convert with OCR still reports progress correctly (exercises the `BatchConverter` locals)

### DIFF
--- a/src/libse/SubtitleFormats/SubtitleFormat.cs
+++ b/src/libse/SubtitleFormats/SubtitleFormat.cs
@@ -23,15 +23,18 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
         /// makes concurrent access real. A lock rather than a lazy/volatile publish because the
         /// build deliberately assigns <see cref="_allSubtitleFormats"/> before ordering it, so
         /// <see cref="GetOrderedFormatsList"/> can re-enter this property through
-        /// <c>Utilities.GetSubtitleFormatByFriendlyName</c>; the lock is re-entrant on the same
+        /// <c>Utilities.GetSubtitleFormatByFriendlyName</c>; Monitor is re-entrant on the same
         /// thread, so that keeps working, while another thread can no longer observe the
         /// intermediate unordered list.
         /// </summary>
-#if NET9_0_OR_GREATER
-        private static readonly System.Threading.Lock SubtitleFormatsLock = new System.Threading.Lock();
-#else
+        /// <remarks>
+        /// Stays a plain <see cref="object"/> - unlike the rest of the code base, which uses
+        /// <c>System.Threading.Lock</c> - because LibSE also targets netstandard2.1, where that
+        /// type does not exist. Guarding one field with <c>#if NET9_0_OR_GREATER</c> is not worth
+        /// the target-framework conditional: the lock is taken a handful of times at start-up, so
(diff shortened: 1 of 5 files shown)


niksedk:
Reviewed and verified this locally — thanks, the migration itself is clean and correct. Details of what I checked:

- Both LibSE target frameworks build, plus libuilogic and UI: **0 warnings, 0 errors**.
- There are no raw `Monitor.Enter/Exit/Wait/Pulse/TryEnter` calls anywhere in `src/`. That was the failure mode worth checking, since `Monitor.Enter(x)` on a now-`Lock` field still compiles (`Lock` is an `object`) and would silently lock a different mechanism. Nothing to break.
- `SubtitleFormat.AllSubtitleFormats` deliberately re-enters its own lock on the same thread via `GetOrderedFormatsList` → `Utilities.GetSubtitleFormatByFriendlyName`. `System.Threading.Lock` is recursive like `Monitor`, so that path was safe — I confirmed it empirically rather than trusting the docs.

I've pushed one commit that drops the LibSE hunk and keeps the other four files.

The reason: `#if NET9_0_OR_GREATER` around `SubtitleFormatsLock` would have been the first non-`#if DEBUG` conditional compilation in LibSE, and it lands in the one project that has to keep supporting two target frameworks. The payoff doesn't justify the precedent — that lock is taken a handful of times at start-up, so `Lock` versus `Monitor` is worth nothing measurable there, and future contributors will copy the pattern.

I added a `<remarks>` block on the field recording why it stays an `object`, so the next sweep doesn't "finish" the migration again. The rest of the PR — `FontFaces.Gate`, `PaddleOcr._errorLock`, `CrispAsrTtsProvenance.SupportCacheLock`, and both `BatchConverter` locals — is unchanged and good to go.

One thing left over if you want a follow-up: `LlamaCppServerManager` locks on `_serverLog` (a `static readonly StringBuilder`) at four sites, i.e. on the mutable data itself rather than a dedicated gate. Not a bug today, but it's the real remaining odd-one-out if uniformity is the goal.

Issue #12888, opened 2026-07-27
RC17 feedback: MPV lag, waveform colors, and right-click menu order


Chouu:
I've been testing RC17 as a long-time Subtitle Edit 4 user, and a few things stand out.

1.  The default MPV player is laggy on my system, (at least when playing large WAV files, I haven't tried anything else). Playback is horrendous. V4 is very smooth compared. 

2.  I’d strongly recommend keeping the default waveform colors from Subtitle Edit 4. The current waveform colors in version 5 are hard on the eyes and feel less readable. Also I can't seem to find a way to add gridlines in V5  (not just vertical lines) on the waveform background. 

3.  The right-click text selection menu now has the two lines reversed compared to v4. That change breaks muscle memory and makes the workflow a PITA

<img width="870" height="447" alt="Image" src="https://github.com/user-attachments/assets/34850f99-a636-4811-a4e5-88d56c01b983" />

niksedk:
I don't think we can optimize the waveform or MPV performance any further, so you're probably better off sticking with SE4—sorry.

SE5 is a bit slower, even though the MPV player code is largely the same.

niksedk:
Regarding performance, are you using the center-waveform setting?

Chouu:
> Regarding performance, are you using the center-waveform setting?

I tried both, but center-waveform is a LOT more choppy. I got an older computer i5 6500, 8gb ram, rx480 8gb, running win10.

Chouu:
I did a fresh windows install tonight (for something else unrelated) and tried RC17 again — everything runs smooth now. Not sure what the original issue was, but please don't spend time chasing it. Might've just been my old potato PC acting up.

niksedk:
I'm glad it's working better :)

The new SE 5.1.0 should also perform better for centered waveform: https://github.com/SubtitleEdit/subtitleedit/releases

PR #12901, opened 2026-07-28
Stop rebuilding waveform geometry every frame during playback (#12888)

niksedk:
Follow-up to the performance feedback in #12888 (laggy playback with large WAV files, i.e. waveform-dominated sessions).

### The problem

During playback the smooth-cursor timer invalidates the `AudioVisualizer` at ~60 fps. The waveform, gridline and timeline caches were keyed on the **exact** `StartPositionSeconds` — so in *center video position* mode, where the view scrolls a couple of pixels every 16 ms tick, **every frame missed the cache** and re-ran the full per-pixel waveform build + `StreamGeometry` allocations (plus glow batches in the fancy style). On wide windows that saturates the UI thread and shows up as the stuttery playback reported for SE5.

### The fix

Anchor the caches to 256 px blocks in absolute (time × pixels-per-second) space:

- build the geometry once for the anchor block **plus one block of padding**,
- replay it each frame with a whole-pixel translation while the view scrolls within the padded range,
- rebuild only when the view crosses into the next block (~once per few seconds of playback instead of ~60×/s).

Applied to the waveform (classic + fancy), the optional gridlines, and the timeline ruler — the ruler previously also rebuilt its tick lists + two `StreamGeometry` and resolved a label per ~5 s of visible audio *per frame*, which was significant when zoomed far out.

### mpv polling reduction on the UI thread

Every mpv property get is a synchronous P/Invoke into the mpv core (it takes the core lock, which can spike during decode/seek):

- `UpdatePlayheadEstimate` + the cursor tick read `IsPlaying` up to 6× per 16 ms tick → now read once per tick and passed down.
- The `Position` getter probed `eof-reached` on **every** read whenever an audio end bound was set — i.e. for the whole session with audio-only files (exactly the large-WAV case from the issue). The probe is now gated to when the last seen position is within 2 s of the bound; position is polled ~60×/s and moves continuously, so the real EOF can't be missed.
- Cached the UTF-8 property-name buffers for the hot polled properties (`time-pos`, `pause`, `eof-reached`, `speed`, `duration`, `volume`) instead of allocating a fresh `byte[]` per call.

### Notes

- Non-center playback already mostly hit the caches; the win there is the timeline ruler and the reduced polling.
- All cache-key inputs (zoom, size, colors, selection, frame mode, video offset) still force a rebuild, and `ResetCache()` invalidates the new timeline cache too.
- Builds clean; all 908 UI tests pass.

### DIFF
--- a/src/ui/Controls/AudioVisualizerControl/AudioVisualizer.cs
+++ b/src/ui/Controls/AudioVisualizerControl/AudioVisualizer.cs
@@ -1886,6 +1886,24 @@ public class AudioVisualizer : Control
         return formatted;
     }
 
+    // Anchored timeline cache (see DrawWaveForm for the anchoring scheme). Rebuilding the tick
+    // lists + two StreamGeometry objects and resolving a label per ~5 s of visible audio on every
+    // ~60 fps render added up, worst when zoomed far out (hundreds of labels per frame); now the
+    // layout is built once per 256 px block and replayed with a translation while scrolling.
+    private readonly List<(FormattedText Text, double X, double Y)> _timeLineLabels = new(128);
+    private Geometry? _timeLineMajorGeometry;
+    private Geometry? _timeLineMinorGeometry;
+    private TimeLineCacheKey _timeLineCacheKey;
+    private bool _timeLineCacheValid;
+
+    private readonly record struct TimeLineCacheKey(
+        double Width,
+        double Height,
+        double AnchorPixel,
+        double ZoomFactor,
+        int SampleRate,
+        double VideoOffsetMs);
+
     private void DrawTimeLine(DrawingContext context, ref RenderContext renderCtx)
     {
         if (renderCtx.SampleRate == 0)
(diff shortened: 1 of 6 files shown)


niksedk:
### Benchmark numbers (BenchmarkDotNet)

Added `AudioVisualizerRenderBenchmarks` (second commit): it runs one real ~60 fps playback frame of `AudioVisualizer.Render` headlessly with the real Skia backend, **record-only** (into a `DrawingGroup` context) — that is the UI-thread cost this PR targets; rasterization happens on the render thread. Waveform: 1 h of speech-like peaks at the default 126 peaks/s, zoom 1.0, 2000 subtitle lines, no spectrogram, height 220 px.

Scenarios per frame:
- **StaticView** (baseline): view stands still, only the cursor moves → caches hit on old and new code. This is the floor: the unavoidable cost of re-recording the scene.
- **CenterMode**: "center video position" playback — the view scrolls a couple of px every frame.
- **ForcedRebuild**: view jumps every frame so both branches rebuild — isolates the geometry-build cost.

Apple M4, .NET 10, `--launchCount 3 --warmupCount 8 --iterationCount 20`.

**Center-mode playback frame, relative to the same run's static-view baseline:**

| WidthPx | Style | main | this PR |
|--------:|-------|-----:|--------:|
| 1600 | Classic | 1.27× | **1.01×** |
| 1600 | Fancy | 1.46× | **1.15×** |
| 3200 | Classic | 2.26× | **1.17×** |
| 3200 | Fancy | 1.65× | **1.24×** |

On main, every center-mode frame pays the full per-pixel geometry rebuild; at 3200 px that's **+190–360 µs per frame on an M4** (static floor ≈ 290 µs → center 486–648 µs), sixty times a second, on the UI thread. On the PR branch center-mode is back near the static floor — the rebuild runs once per 256 px block (~every 2 s of playback) instead of every frame. The `ForcedRebuild` scenario confirms the isolated rebuild cost (~1.4× floor on main at 3200 px; slightly higher on the PR branch since it builds one pad wider — but 60× less often).

Caveats, honestly stated:
- An M4 is fast; on the older Windows laptops these reports come from the rebuild is a several-fold larger slice of the 16 ms frame budget, and it competed with mpv's decode on the same UI thread.
- Run-to-run variance on macOS was substantial (several distributions bimodal, thermal/GC effects) — the within-run **ratios** are the meaningful signal, not absolute µs across runs.
- `Allocated` barely moves because the rebuild's churn is mostly native (SkiaSharp path memory), which `MemoryDiagnoser` doesn't see.
- The mpv polling reduction (≈6 → 1 `IsPlaying` P/Invokes per 16 ms tick, gated `eof-reached` probe) isn't benchmarked here — it needs a loaded libmpv core; it's a straight call-count reduction.

Repro:
```
dotnet run -c Release --project tests/benchmarks/UiBenchmarks.csproj -- --filter '*AudioVisualizerRender*' --launchCount 3 --warmupCount 8 --iterationCount 20
```

<details>
<summary>Full tables</summary>

**main (b08911f76):**

| Method | WidthPx | Style | Mean | StdDev | Median | Ratio | Allocated |
|---|---:|---|---:|---:|---:|---:|---:|
| CenterModePlaybackFrame | 1600 | Classic | 340.4 μs | 58.24 μs | 330.5 μs | 1.27 | 207.56 KB |
| ForcedRebuildFrame | 1600 | Classic | 283.2 μs | 42.97 μs | 275.7 μs | 1.05 | 167.35 KB |
| StaticViewPlaybackFrame | 1600 | Classic | 271.5 μs | 29.68 μs | 267.2 μs | 1.01 | 221.71 KB |
| CenterModePlaybackFrame | 1600 | Fancy | 432.7 μs | 71.79 μs | 418.4 μs | 1.46 | 262.33 KB |
| ForcedRebuildFrame | 1600 | Fancy | 364.7 μs | 81.57 μs | 346.9 μs | 1.23 | 220.89 KB |
| StaticViewPlaybackFrame | 1600 | Fancy | 305.8 μs | 56.91 μs | 305.8 μs | 1.03 | 258.28 KB |
| CenterModePlaybackFrame | 3200 | Classic | 648.3 μs | 90.55 μs | 621.8 μs | 2.26 | 385.17 KB |
| ForcedRebuildFrame | 3200 | Classic | 401.5 μs | 101.20 μs | 339.2 μs | 1.40 | 316.96 KB |
| StaticViewPlaybackFrame | 3200 | Classic | 286.8 μs | 1.28 μs | 286.6 μs | 1.00 | 371.7 KB |
| CenterModePlaybackFrame | 3200 | Fancy | 486.3 μs | 3.05 μs | 485.4 μs | 1.65 | 438.19 KB |
| ForcedRebuildFrame | 3200 | Fancy | 410.2 μs | 3.10 μs | 408.8 μs | 1.39 | 370.08 KB |
| StaticViewPlaybackFrame | 3200 | Fancy | 295.3 μs | 1.00 μs | 295.3 μs | 1.00 | 407.39 KB |

**this PR (92ba94263):**

| Method | WidthPx | Style | Mean | StdDev | Median | Ratio | Allocated |
|---|---:|---|---:|---:|---:|---:|---:|
| CenterModePlaybackFrame | 1600 | Classic | 288.1 μs | 66.87 μs | 268.9 μs | 1.01 | 209.51 KB |
| ForcedRebuildFrame | 1600 | Classic | 316.2 μs | 73.60 μs | 290.6 μs | 1.11 | 178.75 KB |
| StaticViewPlaybackFrame | 1600 | Classic | 300.6 μs | 72.46 μs | 268.6 μs | 1.06 | 229.35 KB |
| CenterModePlaybackFrame | 1600 | Fancy | 330.4 μs | 83.74 μs | 314.4 μs | 1.15 | 255.99 KB |
| ForcedRebuildFrame | 1600 | Fancy | 415.7 μs | 99.32 μs | 395.5 μs | 1.45 | 232.17 KB |
| StaticViewPlaybackFrame | 1600 | Fancy | 298.7 μs | 63.28 μs | 274.1 μs | 1.04 | 265.47 KB |
| CenterModePlaybackFrame | 3200 | Classic | 590.8 μs | 138.44 μs | 569.8 μs | 1.17 | 394.59 KB |
| ForcedRebuildFrame | 3200 | Classic | 589.4 μs | 127.39 μs | 552.3 μs | 1.17 | 326.04 KB |
| StaticViewPlaybackFrame | 3200 | Classic | 530.9 μs | 116.17 μs | 522.1 μs | 1.05 | 378.85 KB |
| CenterModePlaybackFrame | 3200 | Fancy | 599.3 μs | 148.41 μs | 589.9 μs | 1.24 | 431.02 KB |
| ForcedRebuildFrame | 3200 | Fancy | 759.9 μs | 170.43 μs | 764.3 μs | 1.57 | 379.25 KB |
| StaticViewPlaybackFrame | 3200 | Fancy | 497.0 μs | 85.98 μs | 485.9 μs | 1.03 | 415.09 KB |

</details>

PR #12971, opened 2026-07-29
Use a bounded channel for auto-backup writes

ivandrofly:
## Summary
- Replace the per-tick `Task.Run` in `AutoBackupService` with a capacity-1 bounded channel (`DropOldest`, single reader), so backup writes can never run concurrently
- Previously a slow save on a large subtitle could overlap the next timer tick's write; two ticks in the same wall-clock second could also race on the same timestamped filename
- Stale snapshots are dropped while a save is in flight — only the newest pending snapshot is written, skipping wasted `ToText` + disk work under backlog
- Channel writer is completed in `StopAutobackup()` so the reader loop exits cleanly

## Benchmark

Standalone simulation of both patterns: 200 timer ticks at 10 ms with a save taking 50 ms (saves 5× slower than ticks arrive, modeling a large subtitle on a slow disk):

| Pattern | Wall time | Allocations | Saves run | Max concurrent saves |
|---|---|---|---|---|
| `Task.Run` per tick (old) | 3390 ms | 59 KB | 200 | 5 |
| Capacity-1 channel (new) | 3380 ms | 34 KB | 54 | 1 |

- The old pattern ran up to 5 backup writes simultaneously; the channel guarantees exactly one writer.
- Under backlog the channel executed only 54 of 200 saves — `DropOldest` skips stale snapshots (~73% less serialization/disk work) while the final backup is always the newest state.
- No latency cost: wall time is identical. In normal operation (interval ≥ 1 minute, fast saves) both behave identically — the channel only changes behavior in the pathological overlap case.

## Test plan
- [x] Enable auto-backup (Options → Settings) with a 1-minute interval, edit a subtitle, and verify a backup file appears in the auto-backup folder after the interval
- [x] Verify backup files are only created when the subtitle has unsaved changes
- [ ] Load a very large subtitle file and confirm backups still appear and the file contents match the newest state
- [x] Toggle auto-backup off and confirm no further backup files are written

### DIFF
--- a/src/ui/Logic/AutoBackupService.cs
+++ b/src/ui/Logic/AutoBackupService.cs
@@ -63,7 +63,14 @@ public partial class AutoBackupService : IAutoBackupService
         {
             await foreach (var (subtitle, format) in channel.Reader.ReadAllAsync())
             {
-                SaveAutoBackup(subtitle, format);
+                try
+                {
+                    SaveAutoBackup(subtitle, format);
+                }
+                catch
+                {
+                    // never let one bad snapshot kill the backup loop
+                }
             }
         });
 
--- a/src/ui/Logic/AutoBackupService.cs
+++ b/src/ui/Logic/AutoBackupService.cs
@@ -9,6 +9,7 @@ using System.Globalization;
 using System.IO;
 using System.Text.RegularExpressions;
 using System.Threading;
+using System.Threading.Channels;
 using System.Threading.Tasks;
 
 namespace Nikse.SubtitleEdit.Logic;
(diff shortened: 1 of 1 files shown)


src/ui/Logic/AutoBackupService.cs
```diff
@@ -47,6 +49,24 @@ public void StartAutoBackup(MainViewModel mainViewModel)
...
+                FullMode = BoundedChannelFullMode.DropOldest,
+                SingleReader = true,
+            });
+        _backupChannel = channel;
+        Task.Run(async () =>
+        {
+            await foreach (var (subtitle, format) in channel.Reader.ReadAllAsync())
+            {
+                SaveAutoBackup(subtitle, format);
+            }
```
niksedk, line 74: `SaveAutoBackup`'s risky operations are guarded internally, but the loop body itself is not: if it ever throws from an unguarded line (as the method evolves), the `await foreach` terminates and the fault goes unobserved — all future auto-backups silently stop until settings are reopened. The old per-tick `Task.Run` failed independently, so its worst case was one missed backup. A defensive catch keeps the loop alive:

```suggestion
            await foreach (var (subtitle, format) in channel.Reader.ReadAllAsync())
            {
                try
                {
                    SaveAutoBackup(subtitle, format);
                }
                catch
                {
                    // never let one bad snapshot kill the backup loop
                }
            }
```

Issue #13003, opened 2026-07-30
[Feature Request] Separate System Prompt and LLM Parameters (Temperature, Reasoning) for AI Translation


subof:
Hi,
First of all, thank you for implementing LLM translation features. The current user prompt window is already a great addition, but we can make AI translation much more flexible and cost-effective by leveraging how modern LLMs actually handle context.
Right now, Subtitle Edit likely sends its hardcoded technical instructions (e.g., "Translate the text inside <translate> tags and return the exact structure") bundled into the API request. However, to get the highest quality and lowest latency, we need better control over the request structure.

1. Dedicate a Field for the "System Prompt"
In modern AI workflows, separating the System Prompt from the User Prompt is crucial:
Current state: Any custom context (character bios, tone, glossary) placed in the current prompt window is sent repeatedly with every batch of subtitle lines as a User message.
The Proposal: Expose a System Prompt field (perhaps under an "Advanced" dropdown). SE can pre-fill it with its standard technical rules (tag handling, strict format preservation), but allow users to append their own foundational rules (e.g., "You are a professional localizer specializing in 18th-century pirate slang").

2. Why this matters: Prompt Caching and Cost Reduction
Most modern cloud APIs (OpenAI, Anthropic, DeepSeek) and local backends (Mistral, vLLM) support Prompt Caching.
If the System Prompt remains completely static across the entire translation session, providers can cache it. This drastically reduces inference costs and speeds up processing.
For certain models (like Mistral), developers can even insert specific cache markers to optimize this further.

3. LLM Parameter Control
To prevent models from hallucinating or changing the tone midway, it would be highly beneficial to expose at least two key parameters in the UI:
Temperature: A slider from 0.0 to 1.0 (setting it close to 0 or 0.2 is essential for stable, predictable subtitle translation).
Reasoning Effort: For newer reasoning models (like Gemini, Claude or DeepSeek), a setting to control thinking budget (low/medium/high) would be amazing.

Conclusion:
Providing access to the System Prompt and adding basic hyperparameter controls would give power-users unparalleled control over translation context, while significantly cutting down API costs due to efficient prompt caching.
Thank you for considering this!

niksedk:
I will play around with this... we almost have something close in AI review.

subof:
@niksedk , I would like to suggest that the json structure might be better suited to working with the model than any tags like [tag]tex[/tag]. The json format appears to be deeply integrated into the models and does not distract from translation like any tags.
[...]

PR #13005, opened 2026-07-30
Add advanced local-LLM auto-translate engines (llama.cpp + Ollama) with batch context and schema-forced output

niksedk:
Towards #13003.

New **"llama.cpp advanced (local LLM)"** and **"Ollama advanced (local LLM)"** engines in the Auto-translate window, alongside the existing engines (llama.cpp variant shares the model list, downloads, server management and remote-server mode; Ollama variant shares the installed-model list and picker but talks to Ollama's OpenAI-compatible `/v1/chat/completions` endpoint).

### What they do
- Translate in **numbered batches** with a **rolling history** of already-translated lines, so terminology, pronouns and formality stay consistent across the file.
- Replies are **grammar-forced** via `response_format: json_schema` — llama-server compiles the schema to a grammar, so the reply always parses and every line number is present. Servers that reject it (older builds) fall back to `json_object`, then plain text.
- On an invalid reply the batch is retried once, then halved; single-line retries go history-free (translation-tuned models like TranslateGemma can echo history into a lone line).
- New **Advanced…** sub-window shared by both engines: synopsis, glossary (`term = translation` per line), history length, batch size, formality, keep-line-breaks, custom system prompt, temperature/top-p/top-k/repeat-penalty, max tokens, server context size. Settings are global (`Se.Settings.AutoTranslate.LlamaCppAdvanced`).
- The system prompt is ordered stable-parts-first and requests set `cache_prompt: true`, so llama-server reuses the KV prefix where the model architecture allows (issue #13003's prompt-caching point). Ollama ignores the field and manages its own cache.

### llama-server launch fixes (benefit OCR and AI review too)
- `-np 1` — llama-server defaults to **4 parallel slots**, silently splitting `-c` four ways: SE's `-c 8192` gave each request only 2048 usable tokens.
- `--swa-full` — sliding-window-attention models (Gemma, Qwen 3.5) otherwise lose prompt-cache reuse entirely (verified: an identical retry dropped from 318 evaluated tokens to 4).
- Context size is now per-caller; the advanced engine defaults to 16384.

### Defaults validated by live tests
EN→DA sweeps on Qwen 3.5 4B and TranslateGemma 4B (24-line dialogue with tags, line breaks and recurring terms): zero schema failures in all configs; temperature 0.7+ drifted terminology ("the Chief" became three different words), history 0 lost term consistency, batch 5 was slower with no quality gain. Defaults: **batch 10, history 12, temperature floor 0.2** (user/curated-model values take precedence). The Ollama path was verified against gemma3 via the OpenAI-compatible endpoint.

### DIFF
--- a/src/libuilogic/LlamaCpp/LlamaCppServerManager.cs
+++ b/src/libuilogic/LlamaCpp/LlamaCppServerManager.cs
@@ -239,6 +239,7 @@ public static class LlamaCppServerManager
     private static Process? _serverProcess;
     private static int _serverPort;
     private static string? _serverModelPath;
+    private static int _serverContextSize;
     private static bool _processExitHooked;
     private static readonly StringBuilder _serverLog = new();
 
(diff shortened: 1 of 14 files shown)


Ajaja:
How to set `thinking_budget_tokens`/`reasoning_budget_tokens`?
https://www.jasonvsthenoise.com/repowatch/2026-07-13-llama-cpp-reasoning-budget-per-request/

PR #13038, opened 2026-07-31
Fix slow Home / long jumps in the subtitle grid

niksedk:
Home in the subtitle grid was noticeably slow on large files while End stayed instant.

## Cause

TableView's virtualizing panel locates an unrealized row by estimating its offset from the average row height. With variable-height rows (two-line subtitles) that estimate drifts, and once it drifts far enough the panel stops jumping and *walks* — realizing every row between where it is and where it is going.

Measured headless on a 5000-line file, row containers realized per keypress:

| round | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 |
|---|---|---|---|---|---|---|---|---|
| End | 17 | 17 | 17 | 17 | 17 | 17 | 17 | 17 |
| Home | 17 | 17 | 18 | 78 | 139 | 184 | 233 | 289 |

Home went from ~90 ms to ~1 s over eight Home/End round trips, and the panel's extent estimate decayed along with it (140000 → 132373).

It was not only Home. Any long jump could hit the same walk — Find, Go to line number, bookmarks and the video-follow all scroll through the same code:

- jump to line 100: `17, 4935, 17, 4935, 17, 4935` containers — 4935 is the whole file, >10 s
- jump to line 400: full walk in 1 of 6 attempts
- jump to line 700: full walk in 1 of 6 attempts

End never walked; the fallback only bites travelling towards the top.

## Fix

`TableViewExtras.PrePositionScroll` moves the scroll offset next to the target before `ScrollIntoView`, so the panel skips those rows instead of realizing them. It measures where the realized rows actually sit, jumps, and remeasures — at most three passes, each realizing one viewport — leaving ScrollIntoView less than half a viewport to cover. Row 0 is special-cased to offset 0, the one scroll position that needs no estimate.

After: 16–77 containers for any target, flat across rounds, and the extent estimate stops drifting.

Two approaches were measured and rejected on the way:

- snapping to the bottom edge for End — the bottom is only known through the *estimated* extent, so the jump teleports the panel past the real content and inflates the estimate further
- blanket snap-to-top for near-top targets — costs roughly `index` realizations (line 200 → 202 containers), so it is worse than plain ScrollIntoView except very close to the top

`SelectAndScrollToSubtitle` now also nudges the row fully into view when the center-selected-row option is off, the same way `SelectAndScrollToRow` already did — ScrollIntoView could leave the row a few pixels past the bottom edge.

## Tests

`SubtitleGridScrollPerformanceTests` drives the real main window headless with 5000 lines and asserts both real key presses (Home/End, 8 round trips) and jumps to lines 0/20/100/400/700/2500 stay within a few viewports of realizations, and that the target row ends up on screen. Both assertions fail on `main`.

Full UI suite: 1020 passed, 1 skipped.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -15602,6 +15602,11 @@ public partial class MainViewModel :
             if (indexToScroll >= 0 && indexToScroll < Subtitles.Count)
             {
                 var itemToScroll = Subtitles[indexToScroll];
+
+                // Get the offset next to the target first - left to itself the virtualizing panel
+                // walks there row by row on long jumps, which is what made Home (and Find/Go to
+                // line hits near the top) crawl on large files (see PrePositionScroll).
+                TableViewExtras.PrePositionScroll(SubtitleGrid, indexToScroll);
                 SubtitleGrid.SelectedItem = itemToScroll;
                 SubtitleGrid.ScrollIntoView(itemToScroll);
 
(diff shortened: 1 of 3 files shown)


niksedk:
Release-build A/B, same test, 5000 lines, jumping from the last row to the target and back (containers realized / wall clock):

| target | without PrePositionScroll | with |
|---|---|---|
| 0 (Home) | 17, 17, 18, 78 → 246 ms and climbing | 16–77, flat |
| 20 | 17 / ~70 ms | 17 / ~70 ms |
| 100 | 17 / 88 ms, then **4935 / 15 986 ms** | 18–34 / 67–172 ms |
| 400 | 17 / ~140 ms ×3, then **4635 / 18 286 ms** | 18–52 / 211–347 ms |
| 700 | 17 / ~130 ms | 18–44 / 159–848 ms |
| 2500 | 17 / ~90 ms | 18–72 / 72–397 ms |

The typical jump is fine either way — that is why the fix looks like it does nothing. The walk is intermittent (every second attempt for line 100, 1 in 6 for lines 400/700), and when it fires in a Release build it is a 16–18 second freeze, not a stutter.

Honest cost: on long jumps that do not walk, pre-positioning realizes 1–4 viewports instead of 1 (17 → 35–72 containers) because it measures, jumps and remeasures. Raising the convergence threshold from half a viewport to a full one was measured and changed nothing, so the tighter landing is kept.

PR #13055, opened 2026-08-01
Fixed Italian Translation

SiriosDev:
- In Italian, both `brackets`, `curly brackets` and `parentheses` are called `parentesi`, and they are distinguished by the adjective that follows them

- `Leggero` is the "weight" translation of `Light`, `Chiaro` is the "color" translation


niksedk:
@bovirus: OK?

Issue #12504, opened 2026-07-15
[SE 5.1.0-rc3] F10 menu focus shortcut can't be overwritten


Khaztaroth:
The recent inclusion of F10 as a way to highlight and select menu items has made it so F10 can no longer be used as a shortcut, since it defaults to menu selection and the user-configured action is ignored.

niksedk:
Fixed. F10 only grabs the menu bar when no user shortcut is bound to it — MainViewModel now guards the handler with `!_shortcutManager.HasSingleKeyShortcut("F10")`, so assigning F10 in Shortcuts takes precedence and your configured action runs.

Landed in 8d84c0fe7.

PR #13086, opened 2026-08-02
Restore F10/Alt keyboard activation of the menu bar

### LINKED ISSUE #13083: SE 5.2.0 beta 1: Loss of keyboard navigation of the menu bar
I wanted to look at some of the things you changed in the Shortcuts dialog so I did Alt+o then executed the Shortcuts entry.  Scrolled around a bit in the shortcuts.  Closed it.  Alt stopped activating the menu bar.  F10 stopped activating the menu bar.  Closing & reopening SE did not restore the Alt nor F10 functionality.  It's like there is a one-time switch somewhere.  You get the mouse-free access on your first invocation of SE after you install it -- well, I'm actually using the portable version so strictly speaking it's not installed -- as I was saying, on your first invocation of SE, mouse-free access is available.  Once.  After that, no go.  Not in that session.  Not in any subsequent session.  Not in the same session is the sort of thing we've all seen before.  But not in later sessions is very surprising.

niksedk:
Fixes #13083 — after visiting the Shortcuts window, F10 and Alt no longer activated the main menu bar, and the F10 loss survived restarts.

## F10 (the persistent part)

The F10-suppression check from #12504 lets a user-assigned F10 shortcut win over the menu toggle, but it cannot tell a user assignment from the shipped `F10 → set end and go to next` default (shipped since v5.0.0) — and OK in the Shortcuts window persists the whole seeded shortcut list to Settings.json, which made the suppression permanent ("a one-time switch", as the reporter put it).

- Shipped defaults now bind no bare F10, keeping the Windows-standard menu key free.
- A one-time versioned migration (`Se.MigrateShortcuts`, same pattern as the macOS font migration) clears the stale persisted F10 copy from existing Settings.json files. It runs once, so re-assigning F10 afterwards sticks; F10 on other actions and modified bindings are never touched.
- User shortcuts are still checked first: assigning F10 to any action (including "set end and go to next", for the SE4 behavior) wins over the menu toggle.

## Alt (per-session, re-triggered each session)

Avalonia's `AccessKeyHandler` settles bare-Alt state only on the Alt **KeyUp**; a modal stealing focus mid-gesture (Alt, O, … opening the Shortcuts window) eats the release and strands the handler with "ignore the next Alt up" armed — verified against the Avalonia 12.1.1 source and reproduced in a headless test. `OnWindowDeactivated` now raises a synthetic Alt KeyUp to complete the cycle (main window + binary edit window); if the synthetic release opens the menu, the existing deactivation cleanup closes it again.

## Tests

- `ShortcutsF10MenuTests` — defaults never bind bare F10; the migration clears the stale binding, runs only once, and preserves user assignments.
- `AltMenuAccessKeyResetTests` — headless keyboard tests: bare Alt opens the menu; a missed Alt release kills the next Alt press (a canary that flags when Avalonia fixes this upstream, so the workaround can be dropped); the synthetic KeyUp restores it.

Full UI suite: 1191 passed, 0 failed.

### DIFF
--- a/change-log.txt
+++ b/change-log.txt
@@ -3,6 +3,13 @@ Subtitle Edit Changelog
 
 -----------------------------------------------------------------------------------------------------
 
+v5.2.0-beta2 (unreleased)
+
+* Fix F10 no longer activating the menu bar: remove the default F10 "set end and go to next" binding (a one-time migration clears the stale persisted copy; re-assign F10 to get the SE4 behavior back) - thx GrampaWildWilly
+* Fix Alt no longer activating the menu bar after a dialog opened while Alt was held - thx GrampaWildWilly
+
+-----------------------------------------------------------------------------------------------------
+
 v5.2.0-beta1 (2nd of August 2026)
 
 * Auto-translate: advanced llama.cpp and Ollama engines with batch context - thx subof
--- a/docs/reference/keyboard-shortcuts.md
+++ b/docs/reference/keyboard-shortcuts.md
@@ -79,7 +79,11 @@ See also: [Shortcuts Settings](../features/shortcuts.md)
 |----------|--------|
 | F11 | Set start time |
 | F12 | Set end time |
-| F10 | Set end time and go to next line |
+
+"Set end time and go to next line" has no default shortcut: bare F10 activates the
+main menu bar (the Windows standard). Assign F10 to the action in Options →
+Shortcuts if you prefer the Subtitle Edit 4 behavior — a user-assigned shortcut
+wins over the menu activation.
 
 ## Waveform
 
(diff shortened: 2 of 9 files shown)


GrampaWildWilly:
I'm eager for the next beta.  I never thought about a shortcut collision.  Sure enough, I did have the shortcut you identified set to F10.  So I removed that binding.  But even after restarting SE, F10 didn't activate the menu bar.  It mystifies me that Alt did work exactly once the very first time I ran beta 1 & it hasn't worked since.  Similarly, F10 is permanently dead . . . until (we hope) the next beta.  But I didn't think to try F10 until after Alt had already quit working so I don't know if F10 similarly had a first-time-only functioning in it.

I have to admit I don't understand whether your explanation above says why restarting SE does not restore functioning Alt & F10.  I'm glad the next beta is supposed to fix this, but it's just a big surprise to me that cycling SE beta 1 doesn't fix this.

PR #13148, opened 2026-08-03
Localize Video OCR llama.cpp labels and download prompts

Ironship:
## Problem

Follow-up to #13135 (Copilot review comments). VideoOcrViewModel still showed hardcoded English "Start server"/"Stop server" button labels and the three llama.cpp OCR download prompt messages, while AutoTranslate/Ocr view models and LanguageGeneral/LanguageOcr already have the keys.

## Fix

- "Start server"/"Stop server" -> Se.Language.General.StartServer / StopServer
- The three "llama.cpp requires ... Download now?" prompts -> Se.Language.Ocr.LlamaCppDownloadEngineAndModelPrompt / LlamaCppDownloadEnginePrompt / LlamaCppDownloadModelPrompt

## Verification

- dotnet build src/ui/UI.csproj - 0 errors, 0 warnings

### DIFF
--- a/src/ui/Features/Ocr/Download/DownloadCrispEmbedViewModel.cs
+++ b/src/ui/Features/Ocr/Download/DownloadCrispEmbedViewModel.cs
@@ -153,7 +153,7 @@ public partial class DownloadCrispEmbedViewModel : ObservableObject, IClosingCle
         if (_downloadStream.Length == 0)
         {
             ProgressText = Se.Language.General.DownloadFailed;
-            Error = "No data received";
+            Error = Se.Language.General.NoDataReceived;
             return;
         }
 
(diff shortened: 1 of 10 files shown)


Ironship:
CI note: the failing test (`HasErrors_MatchesGetErrors_WithAllRulesOn`) is unrelated to this change — it fails with "[Test Case Cleanup Failure] The calling thread cannot access this object because a different thread owns it" (Avalonia ServerCompositor teardown). The same flaky cleanup failure hits a *different* test on every run (also observed on #13144, #13150 and locally), and the test passes in isolation on main with these changes. Pre-existing UI-test infrastructure flakiness.

Ironship:
Follow-up: the window-leak root cause behind these flaky cleanup failures is fixed in #13154 (tests now close every window they create and settle the dispatcher after Show).

PR #13150, opened 2026-08-03
Localize native Mac export menu items

Ironship:
## Problem

Follow-up to #13131 (Copilot review comment). The native macOS menu (InitNativeMacMenu.cs) still hardcoded the English export captions "DOST/png", "Final Cut Pro + image" and "WebVTT png", while the main menu and the new LanguageExport keys (TitleExportDostPng, TitleExportFcpImage, TitleExportWebVttThumbnails) are localized.

## Fix

- The three native menu items now use the LanguageExport keys (lExport is already in scope).

## Verification

- dotnet build src/ui/UI.csproj - 0 errors, 0 warnings

### DIFF
--- a/src/ui/Assets/Languages/English.json
+++ b/src/ui/Assets/Languages/English.json
@@ -2956,6 +2956,7 @@
     "newNOcrDatabase": "New nOCR database",
     "newBinaryImageCompareDatabase": "New/rename Binary Image Compare database",
     "renameNOcrDatabase": "Rename nOCR database",
+    "renameBinaryImageCompareDatabase": "Rename Binary Image Compare database",
     "nOcrDatabase": "nOCR database",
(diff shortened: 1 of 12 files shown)


Ironship:
CI note: the failing test (`PickTsTrackWindow_HasATitle`) is unrelated to this change — it fails with "[Test Case Cleanup Failure] The calling thread cannot access this object because a different thread owns it" (Avalonia ServerCompositor teardown). The same flaky cleanup failure hits a *different* test on every run (also observed on #13144, #13148 and locally), and the test passes in isolation on main with these changes. Pre-existing UI-test infrastructure flakiness.

Ironship:
Follow-up: the window-leak root cause behind these flaky cleanup failures is fixed in #13154 (tests now close every window they create and settle the dispatcher after Show).

PR #13153, opened 2026-08-03
Address Copilot review comments from the i18n PR batch

niksedk:
Follow-up addressing the valid Copilot review comments left on the merged Ironship i18n batch (#13119–#13152):

**Localization gaps Copilot caught that were still on main**
- BinaryEditWindow + native Mac BinaryEdit export menus still had hardcoded "IMSC 1.1 image profile" / "DOST/png" / "Final Cut Pro + image" / "WebVTT png" (from the #13131 comment) — now reuse `File.Export.TitleExport*`.
- Translate's llama.cpp download prompts in `LlamaCppDownloadHelper` (from the #13135 comment) — three new `LanguageTranslate` keys.
- Remaining `"Download failed"` / `"Download canceled"` / `"No data received"` / `"Unknown error"` literals in the STT engine/model, TTS, OpenFromUrl and OCR download view models — now `General.*`.

**Duplication / value fixes**
- #13141's `Edit.BoxType` / `Edit.BackgroundColor` duplicated `Video.BurnIn.BoxType` and `General.BackgroundColor`; SetTextWindow now reuses those (colon appended in code) and the duplicate keys are removed.
- `NewBinaryImageCompareDatabase` no longer says "New/rename …" now that rename has its own key.

**Code-quality comments**
- `FileSizeConverter` picks the suffix via `switch` — no per-call array allocation, bytes label still read per conversion (also resolves the #13127 static-init comment).
- `GetDictionariesViewModel` timer callback now marshals `StatusText`/`Close()`/`OkPressed` through `Dispatcher.UIThread.Post` (the success branch had the same cross-thread issue Copilot flagged on the cancel branch).
- Removed the accidental UTF-8 BOM #13148 introduced in `VideoOcrViewModel.cs`.
- `AssaStyleWillBeUsed` uses `Environment.NewLine` like the class's other multi-line defaults.

English.json intentionally untouched — regenerate from the language classes.

Not addressed (judgment calls): moving `ForcedAligner`/`PostProcessing` under `video.audioToText` and relocating the color-picker keys in `LanguageAssa` (organizational churn); the #13123 `Key.Return` comment is moot (Avalonia aliases `Enter`/`Return` to the same value); the #13119 `isAssa` comment is a theoretical edge case (literal `\fn` in non-ASSA text) left as-is.

### DIFF
--- a/src/ui/Features/Ocr/Download/DownloadCrispEmbedViewModel.cs
+++ b/src/ui/Features/Ocr/Download/DownloadCrispEmbedViewModel.cs
@@ -142,7 +142,7 @@ public partial class DownloadCrispEmbedViewModel : ObservableObject, IClosingCle
                 else
                 {
                     ProgressText = Se.Language.General.DownloadFailed;
-                    Error = ex?.Message ?? "Unknown error";
+                    Error = ex?.Message ?? Se.Language.General.UnknownError;
                 }
             }
         }
--- a/src/ui/Features/Ocr/Download/DownloadGoogleLensOcrViewModel.cs
+++ b/src/ui/Features/Ocr/Download/DownloadGoogleLensOcrViewModel.cs
@@ -114,7 +114,7 @@ public partial class DownloadGoogleLensOcrViewModel : ObservableObject, IClosing
                 else
                 {
                     ProgressText = Se.Language.General.DownloadFailed;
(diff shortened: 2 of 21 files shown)


Ironship:
@niksedk I did follow ups ;d i think there will be merge conflicts in some cases

PR #13157, opened 2026-08-03
fix: stop burn-in progress bar overlapping target file size box

Ironship:
## Summary

In the "Generate video with burned-in subtitles" dialog, three related layout bugs:

1. While generating, the progress bar is drawn **through the "File size in MB" field** (the bar visibly overlaps the field's lower half).
2. The window can be shrunk so far that the **"Preview" label is covered by the video** and the video's **timestamp + file name overlap the "Audio encoding" section**.
3. The window can be shrunk (or restored) below the content height so the **button row (Generate / Help / Batch mode / OK / Cancel) is clipped in half** while generating.

### Root causes

1. The left settings column (subtitle + video encoding + target file size) is **taller than the middle column's rows** (cut / preview / audio / video info), because the preview row is a star row sized by the preview. The three left boxes were packed into one `StackPanel` spanning rows 0–3; a spanning child's height is not accounted for in the row layout, so on DPI-scaled displays the panel overflowed into row 4 — the progress-bar row — and the bar rendered through the bottom of the target-size box.
2. The saved window size (restored after the content minimum is locked) can be **smaller than the content minimum**; the preview row then shrinks below what its label + player need and the player spills over the audio settings box.
3. `LockMinimumToContentSize` used the **client size** as the window minimum, but `Width/MinHeight` cover the whole window including the title bar and borders (~30 px too small), and the minimum was measured **with the progress row hidden** (pre-generation) — so when generating started, the content grew and the window clipped the button row.

### Fix

- The left panel keeps the **v5.1.0 look** (three boxes packed, no gaps).
- The preview row (row 1) gets **`MinHeight = 400`** — two guarantees:
  - the packed panel always fits inside rows 0–3, so it can never overflow into the progress-bar row (the bar stays below every box, in single and batch mode);
  - the preview box is always taller than its label + player, so the player can never spill over the audio settings box.
- `LockMinimumToContentSize` now adds the **window chrome** (window size − client size) to the locked minimum, so the window can never be shrunk below its content.
- **Starting/stopping a generation re-locks the minimum** (height-only, enlarge-only), so the window always fits the content while the progress bar and status text are shown — the button row is never clipped. The saved window size is still respected as long as it is at least the content minimum.

### Tests

- `BurnInWindowTests.Window_Constructs` — construction smoke test (code-built layout, like `TextToSpeechWindowTests`).
- `BurnInWindowTests.PreviewRow_HasMinimumHeight_KeepingPanelAboveProgressBar` — asserts the star row carries `MinHeight >= 350` plus rendered geometry (left panel bottom above the progress view top); fails on both pre-fix layouts.
- `BurnInWindowTests.VideoPlayer_StaysAboveAudioSettingsBox_WhenWindowIsAtMinimumHeight` — renders the window at the clamped minimum and asserts the player's bottom stays above the audio settings box.
- `BurnInWindowTests.Window_EnlargesToFitProgressBar_WhenGeneratingStarts` — starts a generation and asserts the progress view is visible and the button row stays inside the window; fails without the re-lock.

### DIFF
--- a/src/ui/Features/Video/BurnIn/BurnInWindow.cs
+++ b/src/ui/Features/Video/BurnIn/BurnInWindow.cs
@@ -45,8 +45,13 @@ public class BurnInWindow : Window
         var videoInfoView = MakeVideoInfoView(vm);
         var progressView = MakeProgressView(vm);
 
-        // Keep the left column (subtitle + video + target size) together and top-aligned so the
-        // extra vertical space goes to the preview instead of opening a gap between the boxes.
+        // The left column (subtitle + video settings + target size) is taller than the middle
+        // column's cut/preview/audio/video-info rows. Keeping all three boxes in one packed
+        // panel preserves the v5.1.0 look (no gaps), and the preview row's MinHeight below
+        // guarantees the panel fits in rows 0-3 - so it can never overflow into the
+        // progress-bar row (which used to draw the bar through the "File size in MB" field)
+        // - and the preview box never gets shorter than its label + player, so the player
+        // cannot spill over the audio settings box.
         var leftPanel = new StackPanel
         {
             Orientation = Orientation.Vertical,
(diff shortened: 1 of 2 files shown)


Ironship:
I found one more thing during testing locally. When maximized gaps are to big. I will make fast fix and then confirm.

Ironship:
@niksedk now its 100% ready. Tested it  on Windows 11

before 

{DDBD179F-9E48-4C73-B5FB-FECCF710EFDF}


fixed

{54FAE7DE-E4C1-49CB-A6A6-4CD902FE9BD3}

niksedk:
nice :)

PR #13158, opened 2026-08-03
Apply dark theme background for SyntaxTextEditor

pdjdev:
## Problem

`SyntaxTextEditor` kept its default white background in dark mode, making the dark-theme foreground and syntax colors difficult to read.

## Cause & Fix

The virtualizing editor owns its text surface instead of using Avalonia's `TextBox` template. Its background therefore does not inherit the dark input background applied to `TextBox`.

This change adds a dark-theme style for `SyntaxTextEditor` using the configured dark background color. The same style also covers the source view, format preview, and ASSA batch-convert editor.

## Screenshots
|Before|After|
|-|-|
|image|image|

## Testing

- Confirmed correct display in Light, Dark, Classic and Pastel themes.

### DIFF
--- a/src/ui/Logic/UiTheme.cs
+++ b/src/ui/Logic/UiTheme.cs
@@ -637,6 +637,16 @@ public static class UiTheme
                 }
             },
 
+            // The source editor owns its text surface, so unlike a TextBox it does not inherit
+            // the dark input background from Fluent.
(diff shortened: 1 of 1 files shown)


niksedk:
I normally always use Dark theme, but have been running Light theme the last few days because I often forget something there ;)

Issue #13191, opened 2026-08-04
SE 5.2.0 beta 3: Initial video position is not time index 0


GrampaWildWilly:
Here's another one I've mentioned before but should put it in its own issue.

When you first open a subtitle file, the video is positioned on the time index of the first subtitle.  This is hardly ever time index 0.  Seems to me we dealt with this in SE 4.  Initial position of the video should always be time index 0 no matter where the first caption might fall.

niksedk:
Fixed in PR #13193. A remembered selection on the *first* line is no longer treated as a position worth restoring, so the video now stays at time index 0 after opening. If you were in the middle of the file when you last closed it, reopening still jumps to that line like before (the SE 4 behavior). Will be in the next beta - this also covers #12898.

GrampaWildWilly:
This is solved in beta 4.

Issue #13190, opened 2026-08-04
SE 5.2.0 beta 3: Stuck on first caption


GrampaWildWilly:
I've reported this elsewhere but not as a separate issue.  It is a separate issue.

When you first open a subtitle file, the first caption is highlighted.  If you still have the default binding of Alt+DownArrow to navigate to the next caption, it doesn't work when you have just opened a subtitle.  Even clicking the caption in the grid many times does not get Alt+DownArrow to work.  When you go to another subtitle & come back, things work.

I have just noticed, which I had not in the previous times I've mentioned this, that the first subtitle is not showing in the text box right after a fresh open.  Going to another caption & coming back makes the caption appear in the text box & that activates Alt+DownArrow.

niksedk:
Root cause found: on a fresh file open the grid control can mark row 1 as selected without firing its selection-changed event, so the app's notion of the selected line was still empty - that's why the text box was blank and "go to next line" did nothing (it bails when no line is selected), and why clicking the already-selected row didn't help but going away and back did. Fixed in PR #13193 - the selection is now synced explicitly after open. Will be in the next beta.

GrampaWildWilly:
This is still happening in beta 4.  There's nothing in the text box despite the first caption being the one highlighted in the grid.  I've just noticed that when I first open a subtitle file, the first caption does flash in the text box for a fraction of a second, then it disappears.  I hope that's a clue.  And this is consistent across multiple Ctrl+n & loads of a new subtitle files, as well as across closing & reopening SE.

fraternl:
I suspected it was related somehow, but I thought  I'll wait beta 4.
When it still was in there I saw it wasn't fixed.

It did have some attention.

It happened in 5.2

GrampaWildWilly:
This is fixed in beta 5.

Issue #13187, opened 2026-08-04
SE 5.2.0 beta 3: In undocked mode menus are obscured behind the Audio visualizer


GrampaWildWilly:
I've mentioned this in passing inside discussions on other topics.  It merits dealing with on its own.

In undocked mode, menus on the main SE window open behind the undocked Audio visualizer, obscuring some of the choices on the menus.

niksedk:
Fixed in PR #13193. The undocked tool windows deliberately stay on top of the main window while SE is active (so they don't hide behind it), but that also put them above the drop-down menus. The menus now temporarily drop the tool windows' always-on-top state while open and restore it when they close. Will be in the next beta - this also covers #12899.

GrampaWildWilly:
This is solved in beta 4.

GrampaWildWilly:
<img alt="#01" src="https://github.com/user-attachments/assets/455cc8b6-46af-40bb-981d-648d25539182" />

I think you covered the case of the primary menus, but it looks like you need to revisit this for cascaded submenus.

GrampaWildWilly:
I confirm this is fixed in beta 6.  However, there is something I discovered later.  Please see #13325.

Issue #13185, opened 2026-08-04
SE 5.2.0 beta 3: Settings dialog does not open with focus


GrampaWildWilly:
I can't remember if I've raised this before.  I also can't remember if this was solved before.  Either way, this is an issue now.

When you first open the Settings dialog, the Settings window gets the title bar color that claims the window has focus.  But in fact it does not.  If you Tab a few times, you will see focus bouncing around in the main SE window behind the Settings dialog.  It's easy enough to get focus into the Settings dialog.  Alt+Tab does it as does a mouse click in the window.  But those should not be necessary.  The window should open & have focus.

As a counterexample, the Shortcuts dialog doesn't have this problem.  But I have a vague recollection that once upon a time, the Shortcuts dialog also didn't open with focus.  That's why I'm thinking this is something we've dealt with in the past & this might be a regression.

niksedk:
Your recollection was right: the Shortcuts dialog once had the same problem, and it was fixed by explicitly activating the window and focusing its search box when it opens. The Settings dialog never got that treatment and relied on default focus, which lands wrong when the undocked topmost tool windows are around. Fixed in PR #13193 - Settings now activates and focuses its search box on open. Will be in the next beta.

GrampaWildWilly:
What took you so long?  (kidding, kidding).  Thanks.  Seems I'm always eager for the next beta.

GrampaWildWilly:
This is solved in beta 4.

PR #13193, opened 2026-08-04
Fix beta 3 focus, first-caption selection, initial video position, and undocked menu z-order

niksedk:
Fixes four issues reported against 5.2.0 beta 3:

- **#13185 – Settings dialog does not open with focus**: `SettingsWindow` never focused anything on open, unlike `ShortcutsWindow` (which explicitly activates and focuses its search box — that was the earlier fix the reporter half-remembered). The dialog now activates and focuses the settings search box on `Opened`. Reproducible in undocked mode with the floating topmost tool windows.

- **#13190 – Stuck on first caption**: TableView can initialize row 0 as selected when `ItemsSource` is assigned without raising `SelectionChanged`, so after a fresh file open the grid showed the first row highlighted while `SelectedSubtitle` was still `null` — empty edit box, and go to next/previous line (Alt+Down/Alt+Up) dead because `GoToNextLine` bails on a null `SelectedSubtitleIndex`. `SelectAndScrollToRow` now syncs the view model explicitly when the assignment raises no event.

- **#13191 / #12898 – Initial video position is not time index 0**: `SeekVideoToSelectedLineAsync` (startup restore + Reopen) now treats a remembered *first-line* selection as "no previously established position" and leaves the video at 0:00; only a mid-file line restores the position (the #12222 behavior is unchanged for that case). Note the #13190 fix would otherwise have made the first-caption jump deterministic, since `SelectedSubtitle` is now reliably set after open.

- **#13187 / #12899 – Menus obscured behind the undocked audio visualizer**: the undocked tool windows are topmost while SE is active (#11971), which also covered the main menu's popups. Their topmost is now suppressed while a main-window menu is open and restored (via the same owner-active rule) when it closes.

All 1276 UI tests pass.

### DIFF
--- a/src/ui/Features/Main/Layout/InitMenu.cs
+++ b/src/ui/Features/Main/Layout/InitMenu.cs
@@ -38,6 +38,11 @@ public static class InitMenu
         menu.Items.Clear();
         menu.Opened += (s, e) => DisplayShortcuts(menu, vm);
 
+        // In undocked mode the tool windows are topmost while SE is active (#11971), which
+        // covers the menu popups - drop their topmost while a menu is open (#13187/#12899).
+        menu.Opened += (s, e) => vm.SetUndockedWindowsTopmost(false);
+        menu.Closed += (s, e) => vm.SetUndockedWindowsTopmost(true);
+
         // Drop the menu's font one notch below the theme default and tighten
         // each item's vertical padding — a denser menu reads better when there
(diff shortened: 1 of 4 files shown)


GrampaWildWilly:
How lovely!  Let me say thank you just once here instead of in all the issues.  I will close those issues one by one myself when I get to test the next beta.  Thanks again.

PR #13206, opened 2026-08-04
Fix names: replace 500 ms polling timer with push-based debounce

ivandrofly:
## Summary
- Replace the `System.Timers.Timer` polling loop in `FixNamesViewModel` with a push-based debounce (`CancellationTokenSource` + `Task.Delay`). The old timer fired every 500 ms for the entire lifetime of the window, re-joining the checked names into a string on each tick just to detect whether anything changed — even while the user did nothing. Preview generation is now only scheduled when a checkbox actually changes, and rapid toggles collapse into a single regeneration 500 ms after the last change.
- Checkbox cell template now uses a two-way `IsChecked` binding with `UpdateSourceTrigger.PropertyChanged` and invokes `FixNameItemChangedCommand`; the space-bar multi-toggle helper also requests a preview.
- Select all / invert selection / add extra name regenerate the preview directly, removing the `_loading` / `_oldNames` / lock bookkeeping the polling approach required.
- Minor cleanup in `TableViewExtras.AddSpaceToggle` (parameter rename, lambda simplification).

## Test plan
- [x] Open Tools → Change casing → Fix names with a subtitle containing names; verify the hits preview populates on open
- [x] Toggle a single name checkbox; verify the preview refreshes ~500 ms later
- [x] Rapidly toggle several checkboxes; verify only one refresh happens after the last click
- [x] Use Select all and Invert selection from the context menu; verify the preview updates
- [x] Select multiple rows and press Space; verify all checkboxes toggle and the preview refreshes
- [x] Add an extra name; verify it is picked up and the preview updates
- [x] Toggle a checkbox and immediately close the window; verify no exception from the pending preview


src/ui/Features/Tools/ChangeCasing/FixNamesViewModel.cs
```diff
@@ -122,6 +122,14 @@ private void FindAllNames()
             }
         }
 
+        foreach (var item in names)
+        {
+            // Single notification point: checkbox clicks, the space toggle and
+            // select-all/invert all go through IsChecked, so no call site can
+            // forget to refresh the preview.
+            item.PropertyChanged += (_, _) => RequestPreview();
```
ivandrofly, line 130: I wouldn't go with this as when user choose option to Select All or several useless event/ allocation will happen.
you want preview to be generate once the items are all checked to generate once!

niksedk: thx :)

PR #13209, opened 2026-08-04
Only show video controls when the mouse is over the video window

### LINKED ISSUE #13207: SE 5.2.0 beta 3: Undocked video controls hover up on any mouse movement
This isn't new.  It's been going on for quite a few betas/release candidates going back a fair bit.  I'm only just now thinking to report this.

Any mouse movement causes the video controls to hover up.  They do go back down like they're supposed to.  But I've got the video player on the second monitor & the controls hover up when the mouse moves on the main monitor.  And SE doesn't even need to be in foreground.  For example, I'm here typing in Firefox & the video controls still hover up with any mouse movement in this window.

I would expect those controls to hover up only when the mouse cursor is actually within the boundaries of the video player window.  The video player shouldn't have to be in foreground.  The most common case would be the main SE window is in foreground on monitor 1 when I slide the mouse over to monitor 2 WITHOUT clicking over there.  That's when the controls should hover up.  Now, I've got the video player in fullscreen so anywhere over monitor 2 means anywhere over the video player.  But if it weren't in fullscreen, the hovering should not happen until the mouse cursor actually travels within the borders of the video player window.

niksedk:
Fixes #13207.

The undocked video player window and the fullscreen video window poll the desktop cursor position on a 100 ms timer — the mpv "wid" player renders into a native child window that swallows Avalonia pointer events, so `PointerMoved` alone can't drive the auto-hide.

That poll sees the whole desktop, and any movement over the 20 px threshold counted as user activity. So the auto-hidden controls popped back up whenever the mouse moved in *any* app on *any* monitor, with Subtitle Edit not even in the foreground.

## Changes

- New `CursorPositionHelper.IsCursorOverWindow(window, cursorPosition)`: converts the polled desktop position with `PointToClient` (Avalonia handles per-monitor DPI) and checks it against `ClientSize`; false for a null, hidden, or minimized window.
- Both pollers gate on it before calling `NotifyUserActivity()` — `VideoPlayerUndockedViewModel` (the case in the issue) and `FullScreenVideoWindow`, which had the same copy-pasted block.

`_lastCursorPosition` is still updated on every move, so moving the mouse in from another monitor registers as activity the moment it crosses into the window — no click or focus needed, which is what the issue asks for.

## Tests

7 new headless tests in `tests/UI/Logic/CursorPositionOverWindowTests.cs` (inside, off each of the four edges, minimized, null). Full UI suite green: 1318 passed.

## Notes

- The coordinate spaces of `GetCursorPos` / CoreGraphics / `XQueryPointer` and Avalonia's `PixelPoint` line up by platform convention; Windows (the reported case) is the straightforward one, but a quick check on a Retina Mac and on X11 is worthwhile.
- This is pure geometry — if another window sits on top of the player, movement over that window still counts. Fixing that needs per-platform "window under cursor" calls; left out for now.

### DIFF
--- a/src/ui/Features/Shared/FullScreenVideoPlayer.cs
+++ b/src/ui/Features/Shared/FullScreenVideoPlayer.cs
@@ -103,7 +103,14 @@ public class FullScreenVideoWindow : Window
                         Math.Abs(cursorPos.Value.Y - _lastCursorPosition.Y) > mouseMovementMinPixels)
                     {
                         _lastCursorPosition = cursorPos.Value;
-                        videoPlayer.NotifyUserActivity();
+
+                        // Only movement over this window counts - the poll is desktop-wide,
+                        // so mouse movement in another app or on another monitor must not
+                        // bring the controls back up (issue #13207).
+                        if (CursorPositionHelper.IsCursorOverWindow(this, cursorPos.Value))
+                        {
+                            videoPlayer.NotifyUserActivity();
+                        }
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


GrampaWildWilly:
> . . . if another window sits on top of the player

I say if the video player isn't visible, the controls hovering up is a non-issue.  You can't see them.  It's like if a bear growls in the forest when nobody is there to hear him, is there any sound?  Who cares?  Don't even think about fixing that.

PR #13213, opened 2026-08-04
Make combo/spinner value changes audible to screen readers and let Alt close an open drop-down

niksedk:
Fixes the three remaining NVDA issues from the latest feedback in #12087 (https://github.com/SubtitleEdit/subtitleedit/issues/12087#issuecomment-5181740518) — the ones fixable on our side. The edit-box caret/character review issue is an upstream Avalonia gap (no UIA TextPattern implementation at all) and can't be worked around locally.

### Combo boxes: value changes are now announced (NVDA/Narrator)

Two Avalonia gaps kept them silent on Windows:

1. The Win32 UIA backend forwards a peer's property changes only for properties listed in its private `AutomationNode.s_propertyMap` — and the **Value pattern's properties are missing**, so every Value change was dropped before reaching UIA. `ScreenReaderAnnouncements` patches the two missing entries in via reflection at startup (fail-soft; a unit test acts as the upgrade canary).
2. A collapsed non-editable `ComboBox` never raises a Value change on its own peer when Up/Down moves the selection. A `ComboBox` class handler now raises it the way WPF's combo does, and only when a UIA client (screen reader) has actually materialized a peer and the combo has keyboard focus.

### Spin boxes: stepping with Up/Down is now announced

Screen readers deliberately stay quiet when a plain edit control's value changes (caret events cover real text fields), and the focused element inside `SecondsUpDown`/`TimeCodeUpDown`/`NumericUpDown` is exactly such an inner `TextBox`. They are now exposed as **spinners** (`AutomationProperties.ControlTypeOverride`), so every value change is spoken. (This is also why upstream's own `NumericUpDown` RangeValue events never helped — they fire on the wrong, unfocused element.)

### Alt now closes the menu while navigating a drop-down

Avalonia's `AccessKeyHandler` owns the bare-Alt toggle, but its key-down handler ignores any key whose focused element is not a visual descendant of the window — and drop-down items live in a `PopupRoot`. So Alt opened the menu fine, but a second Alt while inside a drop-down did nothing and only Escape got you out. The window key handler now arms on that exact case and closes the menu on the Alt **release** — closing on the press would immediately be undone by the built-in handler's leftover access-key state calling `MainMenu.Open()` on the release.

### Tests

- Map-patch canary (fails if an Avalonia upgrade renames the internals or ships the entries itself).
- Combo announcement raised on the focused combo's peer / suppressed while unfocused.
- Alt closes the menu with focus inside a drop-down (behavioral, overlay-hosted in headless) plus a focus-topology test that models the desktop `PopupRoot` case with a second top-level.
- Full UI suite: 1360/1360 passing.

Upstream issues worth filing from the analysis: the missing `ValueValue`/`ValueIsReadOnly` map entries (one-liner), Alt-in-popup not closing the main menu, and UIA TextPattern support (tracked loosely under AvaloniaUI/Avalonia#10143).

### DIFF
--- a/src/ui/Controls/SecondsUpDown.cs
+++ b/src/ui/Controls/SecondsUpDown.cs
@@ -153,6 +153,11 @@ public class SecondsUpDown : TemplatedControl
             // screen readers to announce it (e.g. "Duration") instead of just the value.
             _textBox.Bind(AutomationProperties.NameProperty, this.GetObservable(AutomationProperties.NameProperty));
 
+            // Screen readers deliberately stay quiet when a plain edit control's value changes,
+            // so stepping with Up/Down was inaudible; announced as a spinner, every value change
+            // is spoken (#12087).
+            AutomationProperties.SetControlTypeOverride(_textBox, Avalonia.Automation.Peers.AutomationControlType.Spinner);
+
             _textBox.KeyDown += OnTextBoxKeyDown;
             _textBox.LostFocus += (_, _) => ParseAndUpdate();
             _textBox.PointerWheelChanged += (_, args) =>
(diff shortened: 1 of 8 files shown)


niksedk:
Also ported the Alt-closes-drop-down fix to `BinaryEditViewModel`'s parallel copy of the menu logic (second commit), per the keep-in-sync note from #13112.

Issue #13205, opened 2026-08-04
Linux: CrispEmbed server exits with code 127


perkesfurmah-collab:
Hi Nikse,

I tested this on two different Linux distributions:

EndeavourOS
Linux Lite

On both systems:

Downloaded the CrispEmbed CPU engine successfully.
Selected GLM-OCR (glm-ocr-q8_0.gguf).
Started OCR on a Blu-ray PGS subtitle.

Result:

CrispEmbed server exited with code 127

OCR never starts.

Paddle OCR Standalone works correctly with the same subtitle.

Running Subtitle Edit from a terminal only prints:

Cannot load libcuda.so.1

This message appears when Subtitle Edit starts, before using OCR, and nothing else is printed when the CrispEmbed error occurs.

<img width="1280" height="719" alt="Image" src="https://github.com/user-attachments/assets/0df891de-4a46-4c20-b5bf-355ef8a99850" />

niksedk:
Thanks for the detailed report - the exit code 127 is the clue.

Exit code 127 from a Linux binary means the dynamic loader could not resolve a required shared library, so `crispembed-server` dies before it prints anything of its own. I checked the release archive Subtitle Edit downloads (`crispembed-linux-x86_64.tar.gz` from CrispEmbed v0.16.1) and the dependency chain is:

```
crispembed-server  ->  libggml.so.0        (bundled, RUNPATH $ORIGIN - ok)
libggml.so.0       ->  libggml-blas.so.0   (bundled - ok)
libggml-blas.so.0  ->  libopenblas.so.0    <-- NOT bundled
```

`libopenblas.so.0` is a hard link-time dependency and it is not included in the archive, and it is not installed by default on most distros. That is why it fails the same way on both of your systems, while PaddleOCR (a separate binary) works fine.

The `Cannot load libcuda.so.1` message at startup is unrelated - that is just ffmpeg/mpv probing hardware decoders.

Two things:

1. **This needs to be fixed upstream** - could you please report it at https://github.com/CrispStrobe/CrispEmbed/issues? The Linux release tarballs should either be built with `-DGGML_BLAS=OFF` (which is what their own CI workflow already specifies for linux-x86_64) or ship `libopenblas.so.0` alongside the binaries. Subtitle Edit only downloads their prebuilt binaries, so I cannot fix it from this side. The Linux arm64 build has the same issue, and the CrispASR Linux build additionally misses `libgomp.so.1`.

[...]

PR #13218, opened 2026-08-04
Update CrispEmbed to v0.17.2

niksedk:
Bumps the CrispEmbed engine from v0.17.0 to [v0.17.2](https://github.com/CrispStrobe/CrispEmbed/releases/tag/v0.17.2).

Two upstream releases are picked up:

- **v0.17.1** dropped the OpenBLAS link-time dependency that made *every* Linux tarball exit with code 127 in the dynamic loader before reaching `main()` — the cause of #13205. Verified by extracting the v0.17.2 Linux x86_64 and arm64 archives: `libggml-blas` is gone, and no binary has a `libopenblas` `DT_NEEDED` entry any more.
- **v0.17.2** fixes two embedding-quality bugs (Qwen-family tokenizer newline handling, a missing GTE v1.5 per-layer bias) and adds a Linux CUDA archive that bundles the CUDA runtime. The OCR engines are untouched by both fixes.

### Linux CUDA now uses the self-contained archive

The plain `crispembed-linux-x86_64-cuda.tar.gz` requires a CUDA 12.x **toolkit** on the host (`libcudart.so.12` / `libcublas.so.12`), not just a driver, and fails with the same silent exit 127 when it is missing — Subtitle Edit has no way to detect that beforehand. `crispembed-linux-x86_64-cuda-bundled.tar.gz` ships those libraries and needs only the NVIDIA driver, so that is what the CUDA option downloads now. It is ~705 MB instead of ~138 MB, so the Linux download prompt spells out both sizes: "CPU (~10 MB)" / "GPU CUDA (~705 MB)".

### Verification

- All seven pinned URLs cross-checked against the live release digests; every `DownloadHashManager` index 0 matches the archive its URL resolves to.
- macOS smoke test: `crispembed-server --ocr` from v0.17.2 returns byte-identical text and confidences to the installed v0.16.1 for the same image and model (GOT-OCR2 q4_k).
- Build clean, CrispEmbed and download-hash tests pass.

Also documents the PP-OCRv6 backend added in #13210, which never made it into the OCR help page.

### DIFF
--- a/docs/features/ocr.md
+++ b/docs/features/ocr.md
@@ -56,8 +56,9 @@ Local LLM-based OCR using a llama.cpp-compatible server.
 
 ### CrispEmbed
 Local OCR engine with multiple model backends (free/open source).
-- Backends: **GLM-OCR**, **GOT-OCR2**, and **Qwen3-VL-2B**
-- Each backend offers a smaller `q4_k` and a higher-quality `q8_0` model (from about 445 MB for GOT-OCR2 q4_k up to about 2.3 GB for Qwen3-VL-2B q8_0)
+- Backends: **PP-OCRv6**, **GLM-OCR**, **GOT-OCR2**, and **Qwen3-VL-2B**
+- **PP-OCRv6** is a text detector plus recognizer rather than a vision language model, so it needs only two small files (about 79 MB in total) and is the quickest way to get started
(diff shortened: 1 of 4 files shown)


perkesfurmah-collab:
Thanks again for the incredibly fast turnaround.

I really appreciate how responsive you've been to all these reports. In just a few days you've fixed the draggable splitter issue, improved the CrispEmbed error reporting, updated to the upstream release that removes the Linux dependency problem, and even added tests for it.

Subtitle Edit keeps getting better with every release. Thank you for all the hard work!

PR #13238, opened 2026-08-05
Update CrispEmbed to v0.17.5

niksedk:
Bumps the CrispEmbed pin from v0.17.2 to [v0.17.5](https://github.com/CrispStrobe/CrispEmbed/releases/tag/v0.17.5), the first release where both defects behind #13205 are fixed upstream:

- **No unbundled `libopenblas.so.0`** — the Linux legs build `-DGGML_BLAS=OFF`, so the loader no longer aborts with exit 127 before `main()`.
- **glibc floor 2.38 → 2.27** — the CPU archives now build inside `quay.io/pypa/manylinux_2_28`, so Ubuntu 18.04+, Debian 10+ and EL 8+ can run them at all. Previously anything on glibc ≤ 2.37 (including Ubuntu 22.04 LTS and derivatives) could not.

### Verified against the published archives, not the release notes

| archive | hash matches pin | unresolved `DT_NEEDED` | glibc floor |
|---|---|---|---|
| `linux-x86_64` | yes | none | **2.27** |
| `linux-arm64` | yes | none | **2.27** |
| `linux-x86_64-cuda-bundled` | yes | `libcuda.so.1` (the driver) only | 2.38 |
| `macos-arm64` | yes | runs | – |

No `libggml-blas` in any archive, and ggml is 0.17.0 — so this is not the mis-built v0.17.4. All seven pinned SHA-256s were checked against the release API digests.

### OCR output is unchanged

On macOS arm64, v0.17.2 against v0.17.5 for the same image:

- CLI pipeline (PP-OCRv6 detector + recognizer): byte-identical JSON.
- Server path (GLM-OCR q4_k, `POST /ocr/model`): identical text, confidence and every token confidence; only the `ms` timing field differs.

### Notes

- The Linux CUDA pin stays on `crispembed-linux-x86_64-cuda-bundled.tar.gz`, which ships `libcudart` / `libcublas` / `libcublasLt` and so needs only an NVIDIA driver. The plain `-cuda` archive still requires a CUDA 12.x *toolkit* and exit-127s without it. The bundled archive grew from ~705 MB to ~718 MB, so the Linux download prompt states the new size.
- The CUDA archives keep the glibc 2.38 floor — they build outside the manylinux container because they need the toolkit on the host. Only the CPU archives got the lower floor. Noted in a comment next to the constant.
- v0.17.3 (shipped with no Linux CUDA archive) and v0.17.4 (shipped ggml 0.10.2 instead of 0.17.0) are skipped.
- v0.17.5's `DS2_CROP_MODE` default change affects DeepSeek-OCR-2, which is not one of SE's four backends.

Thanks to @CrispStrobe for the fast turnaround and for supplying the hash set.

### DIFF
--- a/src/ui/Features/Ocr/OcrViewModel.cs
+++ b/src/ui/Features/Ocr/OcrViewModel.cs
@@ -1325,7 +1325,7 @@ public partial class OcrViewModel : ObservableObject
                     MessageBoxButtons.Cancel,
                     MessageBoxIcon.Question,
                     "CPU (~10 MB)",
-                    "GPU CUDA (~705 MB)");
+                    "GPU CUDA (~718 MB)");
 
                 if (answer == MessageBoxResult.Cancel)
                 {
--- a/src/ui/Logic/Download/CrispEmbedDownloadService.cs
+++ b/src/ui/Logic/Download/CrispEmbedDownloadService.cs
@@ -21,16 +21,18 @@ public class CrispEmbedDownloadService : ICrispEmbedDownloadService
 {
     private readonly HttpClient _httpClient;
 
-    private const string WindowsCudaUrl = "https://github.com/CrispStrobe/CrispEmbed/releases/download/v0.17.2/crispembed-windows-x86_64-cuda.zip";
-    private const string WindowsVulkanUrl = "https://github.com/CrispStrobe/CrispEmbed/releases/download/v0.17.2/crispembed-windows-x86_64-vulkan.zip";
-    private const string WindowsCpuUrl = "https://github.com/CrispStrobe/CrispEmbed/releases/download/v0.17.2/crispembed-windows-x86_64.zip";
-    private const string MacUrl = "https://github.com/CrispStrobe/CrispEmbed/releases/download/v0.17.2/crispembed-macos-arm64.tar.gz";
-    private const string LinuxUrl = "https://github.com/CrispStrobe/CrispEmbed/releases/download/v0.17.2/crispembed-linux-x86_64.tar.gz";
+    private const string WindowsCudaUrl = "https://github.com/CrispStrobe/CrispEmbed/releases/download/v0.17.5/crispembed-windows-x86_64-cuda.zip";
+    private const string WindowsVulkanUrl = "https://github.com/CrispStrobe/CrispEmbed/releases/download/v0.17.5/crispembed-windows-x86_64-vulkan.zip";
+    private const string WindowsCpuUrl = "https://github.com/CrispStrobe/CrispEmbed/releases/download/v0.17.5/crispembed-windows-x86_64.zip";
+    private const string MacUrl = "https://github.com/CrispStrobe/CrispEmbed/releases/download/v0.17.5/crispembed-macos-arm64.tar.gz";
+    private const string LinuxUrl = "https://github.com/CrispStrobe/CrispEmbed/releases/download/v0.17.5/crispembed-linux-x86_64.tar.gz";
     // The plain "-cuda" archive expects a CUDA 12.x *toolkit* on the host (libcudart/libcublas),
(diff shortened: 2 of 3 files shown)


perkesfurmah-collab:
## Feedback: OCR engine comparison (Japanese PGS subtitles, Linux)

Hi Nikse,

First of all, thank you for all the work you've been doing on the OCR engines recently. The pace of development has been incredible, and it's been really interesting to follow all the improvements over the last few days.

I spent some time comparing most of the OCR engines currently available in Subtitle Edit using exactly the same input.

This is **not intended as a scientific benchmark**, only as practical user feedback from a real subtitle translation workflow.

### Test environment

- Subtitle Edit 5.2.0-beta5
- EndeavourOS (Arch Linux)
- Intel Core i3-6006U
- 4 GB RAM
- Same Japanese PGS subtitle stream
- Same subtitle images for every OCR engine (around 40 subtitle images, except where noted)

### Results

| Engine | Time | Notes |
|--------|------|-------|
| Google Lens Sharp | ~19 s | Fastest overall and excellent OCR quality. |
| Paddle OCR Standalone | ~47 s (about 20 s startup) | Excellent quality and very stable. |
| CrispEmbed PP-OCRv6 | ~4 min 16 s | Surprisingly good quality. Much slower than Paddle, but very accurate. |
| CrispEmbed GLM-OCR q8 | ~16 min | Produced repeated words/characters in several places (hallucination-like behaviour). |
| CrispEmbed DeepSeek-OCR | Not usable | On three different attempts it never completed the first subtitle and froze my desktop before producing any OCR output. |
| Tesseract | ~33 s | Much faster than GLM, but significantly worse recognition quality on Japanese subtitles. |
| PaddleOCR-VL (llama.cpp) | Extremely slow | Needed several minutes for only a few subtitle images and skipped one subtitle completely. |
| glm-ocr (Ollama) | ~6 s for 109 subtitle images | Finished almost instantly but returned empty subtitles. This may be an integration issue rather than a model issue. |

### A few observations

- On this hardware and with this subtitle sample, **Google Lens Sharp** provided the best combination of speed and recognition quality.
- **Paddle OCR Standalone** remains an excellent default local OCR engine.
- **PP-OCRv6** was a pleasant surprise and performed much better than I expected.
- **GLM-OCR** occasionally produced repeated words or characters, suggesting hallucination-like behaviour.
- **DeepSeek-OCR** could not be properly evaluated because it consistently froze my system before producing the first subtitle.
- The VLM-based engines generally appeared much slower than the dedicated OCR engines for subtitle extraction on this hardware.

Again, these results only reflect one hardware configuration and one subtitle source, but I thought they might be useful while you're actively improving the OCR engines.

Finally, I'd like to say that I really appreciate how easy Subtitle Edit makes it possible to compare different OCR engines. Having all these backends available in a single application made this comparison possible, and I think it's a fantastic feature for users.

Thanks again for all your recent work!

perkesfurmah-collab:
Follow-up after some more testing of Google Lens Sharp OCR on Linux (EndeavourOS, Subtitle Edit 5.2.0-beta6).

I believe I found another issue, this time related to large PGS subtitle streams.

### Test results

| Source | Images | Result |
|--------|-------:|--------|
| Japanese PGS | 1579 | OCR stopped prematurely (different runs stopped around image 385, 450 and 603) |
| English PGS | 785 | OCR stopped prematurely at image 667 |
| Spanish VobSub (.idx/.sub) | 876 | Completed successfully |

Observed behavior:

- OCR quality is very good before it stops.
- No error dialog is shown.
- No entries are written to `error-log.txt` or `tools-log.txt`.
- The OCR window returns to its initial state ("Start OCR") as if processing had completed normally.

Since both Japanese and English PGS files show the same behavior while a large VobSub completes successfully, it seems to be related to PGS processing rather than the OCR language itself.

Hopefully this helps narrow it down.

PR #13244, opened 2026-08-05
Fix #10193: harden SoftNI regression tests (review follow-up)

### LINKED ISSUE #10193: Save As to SoftNI sub (*.sub) always exports framerate as 25fps (hardcoded in SoftNiSub.cs)
When exporting subtitles using File → Save As → SoftNI sub (*.sub), the framerate from the current project/video is not saved into the generated .sub file.

Instead, the SoftNI timing metadata is always written as 25fps, which appears to be hardcoded in the exporter implementation.

This creates incorrect timing for projects using other frame rates such as 23.976 / 24 / 29.97 / 30 fps.

In `SoftNiSub.cs` (line 119 and line 120), the framerate is hardcoded:
```
sb.AppendLine(@"*TIMING*");
sb.AppendLine(@"1` 25 0");
```

Because of this, the exported .sub always contains:

```
*TIMING*
1 25 0
```

### Expected

Use the current project/video FPS in the *TIMING* header.

Ironship:
## Problem
Follow-up to merged PR #13241 (issue #10193, SoftNI sub export frame rate). Copilot review on #13241 noted that the regression tests index `lines[timingIndex + 1]` right after asserting the *TIMING* header exists, without checking that a following line actually exists — if the output is truncated or *TIMING* is the last line, the test would throw IndexOutOfRangeException instead of failing with a clear assertion message. The comment applies at two sites (line 31 and line 51 of tests/libse/SubtitleFormats/SoftNiSubTest.cs).

## Fix
Added a bounds assertion before indexing in both SoftNI tests:
`Assert.True(timingIndex + 1 < lines.Count, "Missing timing line after *TIMING* section");`

Test-only change; no production code touched.

## Verification
- [x] `dotnet build tests/libse/LibSETests.csproj` — EXIT:0, 0 errors
- [x] `dotnet test tests/libse/LibSETests.csproj --filter FullyQualifiedName~SoftNiSubTest` — Passed: 4, Failed: 0
- [x] Diff limited to 2 added assertion lines

## Notes
- Copilot review finding from PR #13241 addressed here as a follow-up PR since #13241 was already merged by the maintainer.

### DIFF
--- a/tests/libse/SubtitleFormats/SoftNiSubTest.cs
+++ b/tests/libse/SubtitleFormats/SoftNiSubTest.cs
@@ -28,6 +28,7 @@ public class SoftNiSubTest
             var lines = text.SplitToLines();
             var timingIndex = lines.FindIndex(line => line == "*TIMING*");
             Assert.True(timingIndex >= 0, "Missing *TIMING* section");
+            Assert.True(timingIndex + 1 < lines.Count, "Missing timing line after *TIMING* section");
             Assert.Equal(expectedTimingLine, lines[timingIndex + 1]);
         }
         finally
@@ -50,6 +51,7 @@ public class SoftNiSubTest
             var lines = text.SplitToLines();
             var timingIndex = lines.FindIndex(line => line == "*TIMING*");
             Assert.True(timingIndex >= 0, "Missing *TIMING* section");
+            Assert.True(timingIndex + 1 < lines.Count, "Missing timing line after *TIMING* section");
             Assert.Equal("1 24 0", lines[timingIndex + 1]);
         }
         finally


Ironship:
CI note: the `test` job failure is a pre-existing Avalonia headless flake, not a regression from this PR:

- Failing test: `UITests.Features.Shared.TextBoxTagTogglerTests.ToggleTag_SelectedItalicWord_RemovesItalic` — a **Test Case Cleanup Failure** (`System.InvalidOperationException: The calling thread cannot access this object because a different thread owns it`), i.e. thread-affinity in teardown.
- The test is unrelated to this PR (test-only hardening of `SoftNiSubTest`); LibSE tests all passed (859/859, incl. the 4 SoftNI tests).
- The same flake class hits a *different* test on every full-suite run (locally: `SubtitleGridScrollPerformanceTests.HomeAndEnd_RealizeOnlyAViewportOfRows` failed in two consecutive full runs while passing in isolation; here CI failed `TextBoxTagTogglerTests`, which passes 5/5 in isolation on this exact branch).
- Different test each run + teardown thread-affinity exception + passes in isolation = infrastructure flake, not a code issue.

PR #13246, opened 2026-08-05
Fix #10955: ASSA subtitle grid colors not working with fading colors

### LINKED ISSUE #10955: [beta] ASSA Subtitle grid colors not working when using fading colors
```
everything
{\c&HFFFFFF&\t(20,1000,0.9,\c&H29F2FF&)}became clear.{\c&HFFFFFF&}
```

Image

I don't know really how to normalize this - if there are two colors, how to decide, which one should it show? 

Perhaps the destination color is the most logical to take in these cases (if it's white then colorize it as white)

Ironship:
## Problem
Fixes #10955 — ASSA subtitle grid colors are not shown when the line uses fading colors: `{\c&HFFFFFF&\t(20,1000,0.9,\c&H29F2FF&)}became clear.{\c&HFFFFFF&}` renders uncolored in the grid.

Root cause: `TextWithSubtitleSyntaxHighlightingConverter.ParseAssaColor` (src/ui/Logic/ValueConverters/TextWithSubtitleSyntaxHighlightingConverter.cs:1052). Tag blocks are split on `\`, so the destination color inside a `\t(...)` fade transition becomes the segment `c&H29F2FF&)` — carrying the transition's closing paren. `ParseAssaColor` rejected the trailing `)` as a non-hex character and returned `null`, which reset the color state and left the following text with no foreground color at all.

## Fix
In `ParseAssaColor`, drop a trailing `)` from the color value before trimming the `&H` wrapper. A `)` can only appear there as the closing paren of a `\t(...)` transition, and the color inside the transition is the fade's destination color — the most logical one to display in the grid, as suggested in the issue. Behavior for all other tags is unchanged; malformed colors still fall back to no color.

## Verification
- [x] `dotnet build src/ui/UI.csproj` — 0 errors (EXIT:0)
- [x] New regression test(s): `ShowFormatting_UsesDestinationColorOfAssaFadeTransition` (2 theory rows: `\c` and `\1c` destination inside `\t()`) — fails on pre-fix code (Failed: 2, Passed: 0), passes post-fix (Failed: 0, Passed: 2)
- [x] Existing tests run: `ValueConverterTests` class — 57 passed; full `UITests` suite — all passed
- UI-only change; the failing/passing state is covered headlessly via the converter's rendered inline runs (no manual path needed)

## Notes
- Interpretation taken from the issue: when a `\t()` transition contains two colors, the grid displays the destination color (the one inside the transition). 
- Deliberate non-change: `\t()` timing/acceleration arguments are still ignored by the grid formatter, as before — only the color segment's parsing is fixed.

### DIFF
--- a/src/ui/Logic/ValueConverters/TextWithSubtitleSyntaxHighlightingConverter.cs
+++ b/src/ui/Logic/ValueConverters/TextWithSubtitleSyntaxHighlightingConverter.cs
@@ -1056,7 +1056,11 @@ public class TextWithSubtitleSyntaxHighlightingConverter : IValueConverter
             return null; // Skip empty or excessively long color strings
         }
 
-        // ASSA colors are in format &HAABBGGRR& or &HBBGGRR& (BGR order, not RGB)
+        // ASSA colors are in format &HAABBGGRR& or &HBBGGRR& (BGR order, not RGB).
+        // A trailing ')' is left over when the color is the last tag inside a \t(...)
+        // fade transition ({\c&HFFFFFF&\t(20,1000,0.9,\c&H29F2FF&)}); for fading colors
+        // the grid shows the destination color, so drop the transition's closing paren.
+        colorStr = colorStr.TrimEnd(')');
         colorStr = colorStr.Trim("&Hh".AsSpan());
 
         if (colorStr.Length < 6 || colorStr.Length > 8)
--- a/tests/UI/Logic/ValueConverters/ValueConverterTests.cs
+++ b/tests/UI/Logic/ValueConverters/ValueConverterTests.cs
@@ -209,6 +209,21 @@ public class ValueConverterTests
         Assert.False(run.IsSet(TextElement.ForegroundProperty));
     }
 
+    [AvaloniaTheory]
+    [InlineData("{\\c&HFFFFFF&\\t(20,1000,0.9,\\c&H29F2FF&)}became clear.{\\c&HFFFFFF&}")]
+    [InlineData("{\\c&HFFFFFF&\\t(20,1000,0.9,\\1c&H29F2FF&)}x")]
+    public void ShowFormatting_UsesDestinationColorOfAssaFadeTransition(string text)
+    {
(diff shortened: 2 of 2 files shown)


Ironship:
**Independent audit: PASS** (A1)\n- Issue #10955: ASSA grid colors with fading colors \u2014 root cause confirmed, surgical fix\n- ValueConverter tests 57/57 PASS; CI green; Copilot: no comments\n- Mergeable.

niksedk (commented):
Thanks, but I don't think this is the right fix.

The whole change is `colorStr.TrimEnd(')')`, and it works by accident. The tag splitter breaks `{\c&HFFFFFF&\t(20,1000,0.9,\c&H29F2FF&)}` on `\` into `c&HFFFFFF&`, `t(20,1000,0.9,` and `c&H29F2FF&)`. The last fragment wins simply because `state.Color` gets overwritten — the code has no concept of `\t(...)` at all. So "the grid shows the destination color" isn't a decision the code makes; it's whichever color happens to be textually last.

That ordering isn't stable. `{\t(0,500,\c&HFF0000&)\c&HFFFFFF&}` (transition first, static color after) gives white, while `{\c&HFFFFFF&\t(...,\c&H29F2FF&)}` gives the transition target. Same two colors, opposite result, purely from tag order.

The other half is visibility. There's no contrast handling anywhere in the converter — `CreateBrush` is just `new ImmutableSolidColorBrush(color)` and the converter has no idea what the row background is. `&H29F2FF&` is `#FFF229`, a pale yellow: fine on the dark theme (which is what the screenshot in #10955 shows), close to invisible on light. When there are two candidate colors I'd rather pick the most visible one against the current background, and leave the foreground unset when neither is usable.

I'll do that picking in a separate PR — collect the candidate colors instead of overwriting, then choose by contrast, with the paren trim included as part of it. I'll link it here when it's up.

niksedk:
The contrast-based picking is up as #13269 — it includes the paren trim from this PR, collects all primary colors in the block, and picks by WCAG contrast against the grid background (or leaves the foreground unset when nothing is visible). Closing this one in favor of that.

PR #13250, opened 2026-08-05
Fix #12717: Settings icons poor contrast with custom colors

### LINKED ISSUE #12717: Settings icons - poor contrast in certain color settings
With my custom set colors looks like this:

I guess either need to set some default color (white?), or add separate color option for these

Image

With white:

Image

Ironship:
## Problem
Fixes #12717 — with custom colors set in Settings (dark theme + custom foreground), the icons in the Settings pages have poor contrast (dark-on-dark), while with the default white background they look fine. Root cause: `UiTheme.ApplyLighterDark()` registers a global style forcing every Optris `Icon`'s `Foreground` to the user's custom dark-theme foreground color. Styles beat property inheritance, so the Settings icons — designed as white glyphs on colored square tiles — render in the custom dark color instead of white.

## Fix
Pin the glyph `Foreground` to White as a **local value** on the affected icon containers (a local value beats the theme style) in the settings-related windows:
- `SettingsPage.cs` — left-nav menu tiles (the reported bug)
- `SettingsSection.cs` — section headers in the settings scroll view
- `WordListsWindow.cs` — panel-header glyph
- `ForcedAlignerSetupWindow.cs` — intro glyph
- `ShortcutsWindow.cs` — group tiles (via a PropertyChanged hook, since the icon is created by a data-template binding)

Deliberate non-change: toolbar icons intentionally follow the custom foreground (`MatchIconColorToDarkTheme` is the setting's purpose) and icon-only buttons elsewhere use theme-colored text icons — both left untouched.

## Verification
- [x] `dotnet build src/ui/UI.csproj` — EXIT:0, 0 warnings, 0 errors
- [x] `git diff --stat` — 5 files, +39/−0, CRLF preserved
- [x] Manual verification path (UI-only fix, no automated test per campaign/AI-guideline policy): Settings → Appearance → dark theme with a custom dark foreground color → left-nav tiles and section headers must render white glyphs on the colored squares; regression cases: Word lists window header, Forced-aligner setup icon, Shortcuts group tiles.

## Notes
- The fix relies on Avalonia's deterministic precedence (local value > style).

### DIFF
--- a/src/ui/Features/Files/ImportPlainText/ForcedAlignerSetup/ForcedAlignerSetupWindow.cs
+++ b/src/ui/Features/Files/ImportPlainText/ForcedAlignerSetup/ForcedAlignerSetupWindow.cs
@@ -38,13 +38,8 @@ public class ForcedAlignerSetupWindow : Window
             VerticalAlignment = VerticalAlignment.Center,
         };
         Attached.SetIcon(introIcon, IconNames.Waveform);
-        if (introIcon.Content is Optris.Icons.Avalonia.Icon glyph)
-        {
-            // The dark-theme icon style (UiTheme.ApplyLighterDark) overrides the inherited
-            // white foreground with the custom dark-theme foreground color, which can make
-            // the glyph hard to see on the colored square (#12717) - pin it to white locally.
-            glyph.Foreground = Brushes.White;
-        }
+        // Keep the glyph white on the colored square in the dark theme too (#12717).
+        introIcon.Classes.Add(UiTheme.IconOnAccentClassName);
 
         var introGlyph = new Border
         {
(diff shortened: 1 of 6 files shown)


Ironship:
**Independent audit: PASS** (A2)\n- Issue #12717: settings icons poor contrast with custom colors \u2014 local Brush override beats the global style; PropertyChanged hook for the shortcuts window\n- Build 0 errors; CI green; manual verification path documented\n- Mergeable.

niksedk (commented):
The diagnosis is right but this fixes it at the wrong layer.

The root cause is a single blanket style in `UiTheme.cs:664`:

```csharp
new Style(x => x.OfType<Optris.Icons.Avalonia.Icon>())
{
    Setters = { new Setter(Optris.Icons.Avalonia.Icon.ForegroundProperty, new SolidColorBrush(foreColor)) }
},
```

Rather than narrowing that selector, the PR pastes the same seven-line workaround into five call sites. The sixth colored-square icon someone adds later will be wrong again.

Two more things:

- The four `if (x.Content is Optris.Icons.Avalonia.Icon)` sites assume `SetIcon` materialises the child synchronously. The fifth site (`ShortcutsWindow`) needed a `PropertyChanged` handler precisely because it doesn't — so that assumption is already known to be unreliable, and having both patterns in one PR is a smell.
- Pinning `Brushes.White` doesn't really address "custom colors" in the title. If the accent brush behind the glyph is light, white-on-light is still poor contrast.

Better: either add `.Not(x => x.Class("on-accent"))` to the theme style and one class at each site, or add a shared `UiUtil` helper that builds the square + glyph together — that would also remove the five duplicated `Border { Width = 22, Height = 22, CornerRadius = 6, ... }` blocks.

PR #13257, opened 2026-08-05
Fix #10476: copy ASSA lines to clipboard without file headers

### LINKED ISSUE #10476: Please fix copy content of Advanced Sub Station Alpha format
In Subtitle Edit, when I paste .ass code into Aegisub, it mistakenly treats the file headers (like [Script Info]) as actual subtitle lines.

Image

This is the bad result.

Image

Ironship:
## Problem
Fixes #10476 — copying subtitle lines from SE as ASSA and pasting them into Aegisub makes Aegisub treat the file headers (`[Script Info]`, `Title:`, `[V4+ Styles]`, `Format:`, `Style:` lines, `[Events]`) as actual subtitle lines. Root cause: the grid's Copy/Cut put the **full** `AdvancedSubStationAlpha.ToText(...)` output on the clipboard, file headers included. Aegisub's paste interprets every non-`Dialogue:`/`Comment:` clipboard line as a plain-text subtitle line, so the headers became fake lines. Aegisub's own copy puts only the entry data on the clipboard.

## Fix
`src/ui/Logic/SubtitleGridCopyPasteHelper.cs` — new `GetClipboardText(...)` used by both `Copy` and `Cut`: for ASSA, everything before the first `Dialogue:`/`Comment:` line (the file headers) is dropped, trailing empty lines are trimmed. SE's own paste parses the bare event lines back into paragraphs, so the SE-to-SE round-trip keeps working.

## Verification
- [x] Probe (throwaway, deleted before commit): payload contains only Dialogue/Comment lines; `Subtitle.Parse(payload, "ass")` round-trips both lines with correct text/times — 2/2 passed
- [x] Manual verification path: SE grid copy 2 ASSA lines → paste into Aegisub → only the 2 dialogue lines appear (no header garbage); paste back into SE → identical lines.

## Notes
- Deliberate non-change: other formats (SRT etc.) keep the plain `ToText` payload; only ASSA needs the event-only payload.

### DIFF
--- a/src/ui/Logic/SubtitleGridCopyPasteHelper.cs
+++ b/src/ui/Logic/SubtitleGridCopyPasteHelper.cs
@@ -22,7 +22,7 @@ internal static class SubtitleGridCopyPasteHelper
             subtitle.Paragraphs.Add(item.ToParagraph(subtitleFormat));
         }
 
-        var text = subtitleFormat.ToText(subtitle, string.Empty);
+        var text = GetClipboardText(subtitleFormat, subtitle);
         await ClipboardHelper.SetTextAsync(window, text);
     }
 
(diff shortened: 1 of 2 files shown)


Ironship:
**Independent audit: PASS-WITH-WARNINGS** (A3, regression test added)\n- Issue #10476: ASSA copy payload included file headers, which Aegisub pastes as fake subtitle lines \u2014 GetClipboardText now keeps only event lines; probe: payload = exactly the Dialogue lines, round-trip preserves texts+times; CI green\n- Warning (edge): footer segments ([Fonts]/[Graphics]/\u2026) still survive for embedded-font files \u2014 non-blocking\n- Audit follow-up (c90fd0c2): one committed regression test (event-only payload) per minimal-test policy\n- Mergeable.

niksedk (commented):
Correct, and the round-trip test is exactly the right shape.

One gap: the check is `subtitleFormat is AdvancedSubStationAlpha`, so plain SubStationAlpha (.ssa) still puts its `[Script Info]`/`[V4 Styles]` headers on the clipboard with the exact same Aegisub paste problem. `SubStationAlpha` events use the same `Dialogue:`/`Comment:` prefixes, so extending the filter should be trivial.

PR #13259, opened 2026-08-05
Fix #9631: keep multi-word 'do not break' phrases together

### LINKED ISSUE #9631: Prioritize Multi-Word Entries Over Single Words in "Do Not Break" List
I'm using the "do not break" list to control subtitle line breaks, and I’ve encountered an issue when both multi-word phrases and their individual components are present.

Example:
In the "do not break" list, I have:

    SORT OF

    OF

When the line contains “sort of,” the line break logic still prefers the individual word OF, so the break occurs before “of,” which splits the phrase unnaturally. This happens even though “sort of” is in the list and should ideally be kept together.

Expected Behavior:
Multi-word entries like SORT OF, KIND OF, AT LEAST, etc., should take precedence over single-word matches in the "do not break" list. If a longer phrase is matched, it should block line breaks within it, even if one of its words also appears as a standalone entry.

Suggested Fix:
Prioritize longer (multi-word) matches before checking individual words. This way, phrases that commonly go together are preserved as intended.

Ironship:
## Problem

Fixes #9631 — when the "do not break" list contains a multi-word phrase (e.g. `SORT OF`) and also one of its single words (`OF`), the auto line break still splits the phrase right before the single word (`HE WAS SORT` / `OF WEIRD`).

Root cause: `Utilities.CanBreak` (src/libse/Common/Utilities.cs:241) only checks the text *before* the candidate break point against the list (`head.EndsWith(entry)`). A multi-word entry like `SORT OF` therefore only ever matches at the break *after* the complete phrase — the break point *between* the words (`SORT|OF`) is never protected, so the balance-based splitter (TextSplit → `GetBestLengthSplit`) lands on it. Reproduced on current main with the list `[SORT OF, OF]`, and also with `[SORT OF]` or `[AT LEAST]` alone. The identical logic exists in SE4 (4.0.8 tag) — a long-standing flaw. The plain-text importer path (`PlainTextImporter.NextWordInDoNotBreakList`) already looks ahead and is not affected.

## Fix

In `Utilities.CanBreak`, after the existing head check, add a look-ahead: when the text up to the **next** space ends with a literal multi-word list entry (an entry containing a space), the break is also forbidden. This keeps the whole phrase together — including the issue's exact case where both `SORT OF` and `OF` are in the list.

- Single-word entries are completely untouched: an `OF`-only list still forbids the break after `of` exactly as before (verified).
- The change only *adds* prohibitions, never relaxes any: existing multi-word entries in the shipped dictionaries (e.g. `bg`/`mk` `на пр.`) keep their "no break after" behavior and additionally can no longer be split internally.

## Verification

- [x] New regression test: `AutoBreakLine_NoBreakAfterMultiWordEntry_KeepsPhraseTogether` (Theory, 3 rows) — **fails on pre-fix code** (3/3 failed with the fix stashed), passes post-fix (3/3)
- [x] Existing tests run: full `tests/libse` suite — 862 passed, 0 failed

## Notes

- Interpretation (per the issue's expected behavior): a multi-word entry means "keep the phrase together" — the line must not be split between the phrase's words. When both `SORT OF` and `OF` are present, the multi-word entry now protects the between-words break point, where the single word alone never applied.
- Known limitation (pre-existing, shared with the importer path): a phrase followed by punctuation (`SORT OF,`) is not protected by the look-ahead, since matching is word-boundary based. Deliberately not addressed to keep the change minimal.
- Deliberate non-change: `PlainTextImporter` already protects between-word breaks via `NextWordInDoNotBreakList` — no change needed there.

### DIFF
--- a/src/libse/Common/Utilities.cs
+++ b/src/libse/Common/Utilities.cs
@@ -272,6 +272,27 @@ namespace Nikse.SubtitleEdit.Core.Common
                         return false;
                     }
                 }
+
+                // A multi-word entry (e.g. "SORT OF") means the whole phrase must stay together,
+                // so also forbid a break between its words - otherwise the line could still be
+                // split right before the last word of the phrase (issue #9631).
+                var nextSpaceIndex = s.IndexOf(' ', index + 1);
+                if (nextSpaceIndex < 0)
+                {
+                    nextSpaceIndex = s.Length;
+                }
+
+                if (nextSpaceIndex > index + 1)
+                {
+                    var s3 = s.Substring(0, nextSpaceIndex);
+                    foreach (NoBreakAfterItem ending in NoBreakAfterList(language))
+                    {
+                        if (ending.Regex == null && ending.Text != null && ending.Text.IndexOf(' ') >= 0 && ending.IsMatch(s3))
+                        {
+                            return false;
+                        }
+                    }
+                }
             }
             else
             {
--- a/tests/libse/Common/UtilitiesTest.cs
+++ b/tests/libse/Common/UtilitiesTest.cs
@@ -86,6 +86,54 @@ public class UtilitiesTest
         Assert.Equal(input, result);
     }
 
+    // A multi-word "do not break after" entry (e.g. "SORT OF") must keep the whole phrase
+    // together: the line may not be split between its words, even when a single word of the
+    // phrase ("OF") is in the list too (issue #9631).
+    [Theory]
+    [InlineData("HE WAS SORT OF WEIRD", 10, new[] { "SORT OF", "OF" })]
+    [InlineData("HE WAS SORT OF WEIRD", 10, new[] { "SORT OF" })]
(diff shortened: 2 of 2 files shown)


Ironship:
**Independent audit: PASS** (A3)\n- Issue #9631: multi-word do-not-break entries \u2014 CanBreak look-ahead to the next word; RED/GREEN proven by the auditor (3/3 fail on parent, 3/3 pass with fix); libse 862/862; CI green\n- Documented limitation (pre-existing): phrase followed by punctuation unprotected\n- Mergeable.

niksedk (commented):
Works for the two-word entries, and since no shipped `*_NoBreakAfterList.xml` currently has an entry with three or more words the remaining gap is theoretical — but worth noting: the look-ahead only extends to the *next* space, so a 3+-word phrase ("AS WELL AS") can still be split at its first internal gap. A user who adds such an entry gets half a fix.

Minor perf note: this adds a second full list scan plus a `Substring` per candidate break position. It only runs when `UseNoLineBreakAfter` is on (off by default), so it's acceptable, but it could be folded into the existing loop over the list.

PR #13260, opened 2026-08-05
Fix #10209: keep exact WebVTT cue position settings on save

### LINKED ISSUE #10209: .vtt error issues with line% and position% force locking themself to grid and others weirdness

Ironship:
## Problem
Fixes #10209 — WebVTT cue settings with exact percentages (`line:25% position:25%`) are lost on save: the export re-maps them to the coarse `WebVttCueAnX` grid defaults (`line:50%` etc.) and drops `position:`/`size:`/`align:` entirely. Reproduced on current main with the issue's exact repro (probe: `line:25% position:25%` → exported as `line:50%`, `position` gone).

## Fix
`src/libse/SubtitleFormats/WebVTT.cs` — `GetPositionInfoFromAssTag` now remembers which `{\anX}` tag was actually present and returns the user's stored cue settings (`p.Style` / `p.Extra`) verbatim when they match that tag's alignment, instead of always overwriting with the coarse grid default. Exact percentages survive the load/save round trip.

## Verification
- [x] New regression test `WebVttTest.ToTextKeepsExactCuePositionSettings` (3 Theory rows: `line:25% position:25%`, `line:72.69% align:left position:44.90% size:10.21%`, `line:90% position:50%`) — fails pre-fix (proven by probe on unmodified main), **3/3 PASS post-fix**
- [x] Manual verification path: open a VTT with `line:25% position:25%` → Source View / save → the cue settings are unchanged.

## Notes
- Deliberate non-change: when no `\an` tag is present or no custom settings are stored, the existing `WebVttCueAnX` defaults still apply.

### DIFF
--- a/src/libse/SubtitleFormats/WebVTT.cs
+++ b/src/libse/SubtitleFormats/WebVTT.cs
@@ -143,41 +143,61 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
         internal static string GetPositionInfoFromAssTag(Paragraph p)
         {
             string positionInfo;
+            string currentTag = null;
             if (p.Text.StartsWith("{\\an1", StringComparison.Ordinal))
             {
                 positionInfo = Configuration.Settings.SubtitleSettings.WebVttCueAn1;
+                currentTag = "{\\an1}";
             }
             else if (p.Text.StartsWith("{\\an3", StringComparison.Ordinal))
             {
                 positionInfo = Configuration.Settings.SubtitleSettings.WebVttCueAn3;
+                currentTag = "{\\an3}";
             }
             else if (p.Text.StartsWith("{\\an4", StringComparison.Ordinal))
             {
                 positionInfo = Configuration.Settings.SubtitleSettings.WebVttCueAn4;
+                currentTag = "{\\an4}";
             }
             else if (p.Text.StartsWith("{\\an5", StringComparison.Ordinal))
             {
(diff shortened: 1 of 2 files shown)


Ironship:
**Independent audit: PASS-WITH-WARNINGS** (A3)\n- Issue #10209: exact WebVTT cue position settings (line%/position%) lost on save \u2014 GetPositionInfoFromAssTag returns stored settings when they match the anX tag; discriminating test 3/3 (pre-fix 3/3 FAIL verified); 16/16 WebVtt tests; libse 862/862; CI green\n- Warnings non-blocking\n- Mergeable.

niksedk (commented):
Logic looks sound. I traced the three test cases through `GetPositionInfo`/`GetPositionInfoRaw`: the tag-equality guard correctly falls back to the coarse grid settings when the user changes the alignment tag after load, and `GetPositionInfoRaw` on an ASSA style name returns empty, so cross-format conversions are safe.

One process note: #12340 is an open PR for this same issue (#10209) with a different approach — we should pick one and close the other rather than land both.

PR #13263, opened 2026-08-05
Fix #12144: Auto-trim white space must not merge Dutch 's (ze 's avonds)

### LINKED ISSUE #12144: Correction is immediatly reverted to a faulty sentence.
If I correct the wrongly spelled sentence 

"Ik hoorde `ze's avonds` ruzie maken.
Mama wilde het land verlaten."

to the correct sentence:

"Ik hoorde `ze 's avonds` ruzie maken.
Mama wilde het land verlaten."

it reverts **immediately** to the wrong sentence.
No spelling check, no "fix common errors"

[nl.srt.txt](https://github.com/user-attachments/files/29655469/nl.srt.txt)

Ironship:
## Problem
Fixes #12144 — with "Auto-trim white space" enabled (and in Fix common errors), the Dutch line `Ik hoorde ze 's avonds ruzie maken.` is silently rewritten to `Ik hoorde ze's avonds ruzie maken.` because the `" 's "` → `"'s "` merge in `Utilities.RemoveUnneededSpaces` treated every language except Dutch as English-possessive territory.

Root cause (verified on current main with the reporter's sample `simple.srt`):
- The merge is gated on `language != "nl"` (src/libse/Common/Utilities.cs:2607), but both consumer paths feed it a language code that is `"en"` for Dutch text whenever detection fails:
  - Auto-trim white space (`MainViewModel.AutoTrimWhiteSpaces()`, src/ui/Features/Main/MainViewModel.cs:16858, and the selection-changed trim at :22351) calls `Subtitles.AutoDetectGoogleLanguage() ?? string.Empty`. The grid detector (src/ui/Logic/ObservableCollectionExtensions.cs:56) delegates to `LanguageAutoDetect.AutoDetectGoogleLanguage(Subtitle)`, which falls back to `?? "en"` when detection fails (src/libse/Common/LanguageAutoDetect.cs:922) — so the `?? string.Empty` added in PR #12926 never fired and the `nl` gate was skipped.
  - Fix common errors ("Remove unneeded spaces", the mechanism behind the reported "fix common OCR") passes the dialog language (`callbacks.Language`), which defaults to the same detection with the same `"en"` fallback (src/ui/Features/Tools/FixCommonErrors/FixCommonErrorsViewModel.cs:129).
- Probe on unmodified code: the reporter's 2-line file detects as `"en"` (`AutoDetectGoogleLanguageOrNull` = null) → `RemoveUnneededSpaces(text, "en")` → `ze's avonds`.

## Fix
Two small changes:
1. **`Utilities.RemoveUnneededSpaces`** — the `" 's "` possessive merge (including the `" 's"` + newline variant) now runs only for `language == "en"` instead of `language != "nl"`. The merge is an English possessive OCR fix; Dutch `'s avonds` is a separate genitive word and must never be merged regardless of which language code reaches the method.
2. **`ObservableCollectionExtensions.AutoDetectGoogleLanguage`** (the grid-based detector used by the auto-trim path) now returns `LanguageAutoDetect.AutoDetectGoogleLanguageOrNull(...)`, honoring its already-nullable `string?` signature: detection failure yields `null`, so the existing `?? string.Empty` at the auto-trim call sites actually produces `""`, which the English-only gate blocks. The other call sites already null-guard (`?? "en"` at MainViewModel:10811; `MergeContinuationLinesViewModel.Initialize`/`MergeContinuationLinesHelper` accept and fall back from null/empty).

## Verification
- [x] New regression test `UtilitiesTest.RemoveUnneededSpaces_ApostropheSMerge_EnglishOnly` (3-row Theory):
  - `("Ik hoorde ze 's avonds ruzie maken.", "nl")` keeps the space
  - `("Ik hoorde ze 's avonds ruzie maken.", "")` keeps the space — this row **fails on pre-fix code** (`"" != "nl"` → merged to `ze's avonds`), passes post-fix
  - `("John 's car is red.", "en")` still merges to `John's car is red.`
  - Pre-fix run: `Failed: 1, Passed: 2`; post-fix run: `Passed! - Failed: 0, Passed: 3`
- [x] Full `tests/libse` suite: `Passed! - Failed: 0, Passed: 862, Total: 862`
- [x] Throwaway console probe (referencing libse, deleted before commit): reporter's `simple.srt` → detection `null` → trim language `""` → line **unchanged** (pre-fix it became `ze's avonds`); language matrix confirms only `"en"` merges, while `"nl"`, `""`, `"de"`, `"fr"` all keep the space.

## Notes
- Interpretation: the design intent (per the issue) is "Dutch must never merge" — a single-language-argument API cannot distinguish English text from Dutch text whose detection failed, so the merge is restricted to positively-English input; this is the trade-off documented in the campaign brief. A genuinely English subtitle whose language detection fails no longer gets the cosmetic `" 's "` merge in the auto-trim path — cosmetic only.
- Deliberate non-change: the FixCommonErrors dialog language combo is user-visible and defaults to detection; for a short Dutch file the default can still show "English" and "Remove unneeded spaces" would then merge. Changing the dialog's language default is a broader behavior change (it gates every language-aware rule in the dialog) and was left out — flagging for the owner.
- Verified there is no separate `" 's "` rule in FixCommonErrors or the OCR fix engine (grep across `src/`) — both reported triggers funnel into `Utilities.RemoveUnneededSpaces`, so this single change covers both.
- Related history: PR #12926 fixed the empty-language-code variant of this bug ("Fixes the auto-trim part of #12144"); this PR closes the remaining detection-fallback hole and restricts the rule to English.

### DIFF
--- a/src/libse/Common/Utilities.cs
+++ b/src/libse/Common/Utilities.cs
@@ -2602,7 +2602,8 @@ namespace Nikse.SubtitleEdit.Core.Common
                 text = text.Replace(" ,", ",");
             }
 
-            if (language != "nl")
+            // " 's " is an English possessive fix - Dutch keeps it as a separate word ("'s avonds", issue #12144)
+            if (language == "en")
             {
                 while (text.Contains(" 's "))
                 {
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -10808,7 +10808,7 @@ public partial class MainViewModel :
             ? text.Substring(0, selectionStart)
             : string.Empty;
 
-        var language = Subtitles.AutoDetectGoogleLanguage() ?? "en";
+        var language = Subtitles.AutoDetectGoogleLanguage();
         ReplaceSelectedText(tb, SentenceCaser.SentenceCase(textBefore, tb.SelectedText, language));
     }
 
@@ -16858,7 +16858,9 @@ public partial class MainViewModel :
     {
         if (Se.Settings.General.AutoTrimWhiteSpace)
         {
-            _autoTrimLanguageCode = Subtitles.AutoDetectGoogleLanguage() ?? string.Empty;
+            // OrNull + empty fallback: undetected text must not get the English-only rules
+            // (e.g. the " 's " merge corrupting Dutch, #12144).
+            _autoTrimLanguageCode = Subtitles.AutoDetectGoogleLanguageOrNull() ?? string.Empty;
             foreach (var item in Subtitles)
             {
                 item.Text = Utilities.RemoveUnneededSpaces(item.Text, _autoTrimLanguageCode).Trim();
@@ -22351,7 +22353,7 @@ public partial class MainViewModel :
 
         if (Se.Settings.General.AutoTrimWhiteSpace && e.RemovedItems.Count < 10)
         {
-            var languageCode = _autoTrimLanguageCode ??= Subtitles.AutoDetectGoogleLanguage() ?? string.Empty;
+            var languageCode = _autoTrimLanguageCode ??= Subtitles.AutoDetectGoogleLanguageOrNull() ?? string.Empty;
             foreach (SubtitleLineViewModel? item in e.RemovedItems)
             {
                 if (item != null)
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -15213,7 +15213,10 @@ public partial class MainViewModel :
         var countOfTrimmedLines = 0;
 
         var selectedItems = SubtitleGridSelectedItems.Cast<SubtitleLineViewModel>().ToList();
-        var languageCode = LanguageAutoDetect.AutoDetectGoogleLanguage(GetUpdateSubtitle());
+        // Same nullable-detection pattern as the auto-trim path (#12144 audit follow-up): on
+        // detection failure the empty language code keeps the English-only rules (e.g. the
+        // " 's " merge) from corrupting e.g. Dutch text.
+        var languageCode = LanguageAutoDetect.AutoDetectGoogleLanguageOrNull(GetUpdateSubtitle()) ?? string.Empty;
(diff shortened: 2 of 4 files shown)


Ironship:
**Independent audit: PASS-WITH-WARNINGS** (A3, audit follow-up applied)\n- Issue #12144: auto-trim merged Dutch 's (ze 's avonds \u2192 ze's avonds) \u2014 's merge now English-only in RemoveUnneededSpaces + detection returns nullable so auto-trim gets '' on failure instead of 'en'; discriminating test 3/3 (pre-fix fail verified); libse 862/862; CI green\n- Residuals disclosed in PR body (FixCommonErrors default-English path); audit follow-up (e534a62): manual trim-selected-lines uses the same nullable detection\n- Mergeable.

niksedk (commented):
The libse part is correct and the Dutch case is a real bug, but the UI part has a wider blast radius than the title suggests.

Changing `ObservableCollectionExtensions.AutoDetectGoogleLanguage` from `LanguageAutoDetect.AutoDetectGoogleLanguage` to `...OrNull` changes the return value for **every** caller of that extension, not just auto-trim. `LanguageAutoDetect.AutoDetectGoogleLanguage(Subtitle)` is literally:

```csharp
return AutoDetectGoogleLanguageOrNull(subtitle) ?? "en";
```

so four unrelated call sites that used to get `"en"` on detection failure now get `null`:

- `MainViewModel.cs:5990` — merge continuation lines
- `MergeManager.cs:186` — `ContinuationUtilities.MergeHelper`
- `MergeManager.cs:232` and `:254` — `Utilities.AutoBreakLine`

I traced all four sinks and none of them throw, so this isn't a crash — but continuation-style handling and auto-break silently lose their English default there. Either add a separate null-returning helper for the two auto-trim sites, or put `?? "en"` back at those four.

Also worth calling out explicitly in the PR description: `language != "nl"` → `language == "en"` disables the `" 's "` merge for *every* language, not just Dutch, and for the empty code that detection returns on short files. That's defensible but it's a deliberate SE 4 behaviour break, not just a Dutch fix.

PR #13279, opened 2026-08-05
Fix #12626: TTS OK button bug

### LINKED ISSUE #12626: TTS OK button bug
After I generate the TTS, wait for it to finish, and obtain the WAV file I need, clicking the OK button in the Text-to-Speech window causes the entire application to close instead of simply closing that window and returning to the main interface.

Image

Ironship:
## Problem
Fixes #12626 — after TTS generation finishes, clicking **OK** in the Text-to-Speech window closes the whole application instead of just the window.

Issue report (verbatim): "After I generate the TTS, wait for it to finish, and obtain the WAV file I need, clicking the OK button in the Text-to-Speech window causes the entire application to close instead of simply closing that window and returning to the main interface."

Reporter's Event Viewer (verbatim): "The process was terminated due to an unhandled exception. Stack: at Avalonia.Controls.Window.CloseInternal() at Avalonia.Controls.Window.Close() at Nikse.SubtitleEdit.Features.Video.TextToSpeech.TextToSpeechViewModel.<Close>b__147_0() ..." (+ AppCrash event, ExceptionCode c0000005, ModuleName coreclr.dll).

Root cause: `TextToSpeechViewModel.Close()` (the `Close` command used by both OK and Cancel) posts `Window?.Close()` to the UI dispatcher without any guard. The reporter's crash stack shows exactly this frame — `<Close>b__147_0` is the generated lambda `() => Window?.Close()` — throwing out of `Avalonia.Controls.Window.CloseInternal()` on the UI thread. An exception escaping a dispatcher operation is unhandled and terminates the process. PR #12628 guarded every step *inside* the close sequence (`Ok()`, `Cancel()`, `OnClosing()` teardown, post-close `ApplyTtsChanges`) but left the `Window.Close()` invocation itself able to throw its way out of the dispatcher lambda — the one frame the crash report actually points at.

## Fix
`src/ui/Features/Video/TextToSpeech/TextToSpeechViewModel.cs` — `Close()`: wrap `Window?.Close()` in try/catch, log via `SeLogger.Error` and write a forced tools-log breadcrumb, matching the guard pattern #12628 established for the rest of the close path ("nothing in the close path may throw its way out of the command"). If Avalonia's close machinery throws, the window stays open, the failure is logged (`error-log.txt` + `tools-log.txt`), and SE keeps running instead of dying.

## Verification
- [x] `dotnet build src/ui/UI.csproj -v q -nologo` — 0 errors, 0 warnings (exit code 0)
- [ ] No new automated tests: UI-only close-path guard, headless tests impractical (per campaign policy for UI fixes)
- Manual verification path:
  1. Open Tools → Generate speech (Text-to-Speech window), generate a clip to completion.
  2. Click **OK** — window closes, app returns to main interface, merged lines/review edits are applied.
  3. Repeat with **Cancel**, **Escape**, and title-bar **X** — same result.
  4. Optionally use **Test voice** before OK (mpv preview path) and repeat.
  5. With "Write tools log" enabled, the tools log should end with `TTS window: closing ...` / `TTS window: close cleanup done`; any failure would appear as `TTS window: Window.Close() failed` and the app would remain alive.

## Notes
- Interpretation: the bug is the OK button taking down the whole app via an unhandled exception in the close path; the fix makes that exception impossible to escape the dispatcher.
- The sibling unguarded `Window?.Close()` in `OnKeyDown` (Escape, `TextToSpeechViewModel.cs:3545`) is left unchanged: same mechanism, but not the reported path, and the window is not reported to die there; it can be hardened in a follow-up if the underlying Avalonia close crash ever shows up again.
- Deliberate non-change: no reformatting, no new dependencies; only the `Close()` method is touched.

### DIFF
--- a/src/ui/Features/Video/TextToSpeech/TextToSpeechViewModel.cs
+++ b/src/ui/Features/Video/TextToSpeech/TextToSpeechViewModel.cs
@@ -2129,22 +2129,32 @@ public partial class TextToSpeechViewModel : ObservableObject
 
     private void Close()
     {
-        Dispatcher.UIThread.Post(() =>
+        Dispatcher.UIThread.Post(CloseWindowSafely);
+    }
+
+    /// <summary>
+    /// Closes the window without letting a failure escape, for every path that closes the TTS
+    /// window (#12626). An exception thrown out of a dispatcher callback is unhandled and takes
+    /// the whole process down, which is what the crash report shows for the OK button:
+    /// <c>Window.CloseInternal()</c> under the <c>Close</c> command's dispatcher lambda. #12628
+    /// guarded the steps inside the close sequence; this guards the close call itself.
+    /// <para>
+    /// Caveat: this can only contain a <em>managed</em> exception. The reporter's Event Viewer
+    /// entry is an access violation (<c>c0000005</c> in coreclr.dll), which no catch block can
+    /// intercept - if that is the real fault, the crash needs a different root cause.
+    /// </para>
+    /// </summary>
+    private void CloseWindowSafely()
+    {
+        try
         {
-            // Guarded (#12626): the crash report showed the app dying on an unhandled exception
-            // thrown out of Window.Close()/CloseInternal() on the UI thread. The close sequence
-            // itself is guarded in OnClosing; this keeps a failure in Avalonia's close machinery
-            // from escaping the dispatcher lambda and terminating the whole application.
-            try
-            {
-                Window?.Close();
-            }
-            catch (Exception ex)
-            {
-                SeLogger.Error(ex, "TTS window: Window.Close() failed");
-                Se.WriteToolsLog("TTS window: Window.Close() failed: " + ex, true);
-            }
-        });
+            Window?.Close();
+        }
+        catch (Exception ex)
+        {
+            SeLogger.Error(ex, "TTS window: Window.Close() failed");
+            Se.WriteToolsLog("TTS window: Window.Close() failed: " + ex, true);
+        }
     }
 
     private async Task PlayAudio(string fileName)
@@ -3554,7 +3564,9 @@ public partial class TextToSpeechViewModel : ObservableObject
                 return;
             }
 
-            Window?.Close();
+            // Same guard as the Close command: Escape reaches Avalonia's close machinery through
+            // exactly the same path, so it must not be able to take the app down either (#12626).
+            CloseWindowSafely();
         }
         else if (UiUtil.IsHelp(e))
         {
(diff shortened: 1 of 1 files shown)


Ironship:
**AUDIT #13279 (issue #12626): PASS — mergeable**

- A. Fixes issue: PASS — the issue's crash stack shows the app dying on an unhandled exception escaping the UI-thread dispatcher lambda during Window.Close(); the guard catches exactly that site (with #12628's existing guards covering the rest of the close sequence). Catch path is safe: SeLogger.Error + tools-log breadcrumb, no rethrow.
- B/C: PASS — 1 file, +13/−1, matches the owner's #12628 guard style.
- D: PASS — UI-only, no new tests per policy; manual path documented.
- E: Compilation: PASS — UI build exit 0.
- F: Regression risk: low — happy path unchanged; failures are now logged and diagnosable instead of killing the app.
- 1 commit (Ironship), DRAFT.

niksedk (commented):
## Approve, with a caveat worth recording

The guard is harmless and the log breadcrumbs are useful, so I am not blocking. But please be clear in the description about what this can and cannot fix.

**`ExceptionCode c0000005` in `coreclr.dll` is an access violation, not a managed exception.** `try`/`catch` does not catch an AV. If the reporter's crash is a native fault in Avalonia's close/render path, this change does nothing for them and the issue will come back. The PR reads as if the crash is definitely fixed; it isn't established that it is.

If it *is* a managed exception, the catch works — but it leaves the window open after `OnClosing` teardown has already run, i.e. a half-torn-down window the user then clicks OK on again. Worth a follow-up thought about what state the window should be left in.

**Do both call sites or neither.** The PR deliberately leaves the identical unguarded `Window?.Close()` in `OnKeyDown` (Escape) at line 3545. If the guard earns its place, Escape deserves it too; a shared `CloseSafely()` used by both would be cleaner than the asymmetry.

Also: #12628 already guarded the rest of this path and the crash still happened. That is a signal the root cause is still unknown, and a fourth layer of try/catch may be papering over it.

niksedk (approved):
## Approve

Follow-up pushed as a8431e5: the guard is extracted into `CloseWindowSafely()` and the Escape path in `OnKeyDown` now uses it too. Escape reaches Avalonia's close machinery through the same path as the OK/Cancel command, so guarding one and not the other left the same crash reachable by a different key.

The doc comment now also records the limit of the guard: it contains a **managed** exception only. The reporter's Event Viewer entry is an access violation (`c0000005` in coreclr.dll), which no catch block can intercept — if that is the real fault, #12626 needs a different root cause and this change will not close it. Worth keeping the issue open until a user confirms.

Build clean, CI green.

Issue #13082, opened 2026-08-02
In Dutch, like Spanish there are also abbreviations that cause a capital letter fot the next word


fraternl:
<img width="1257" height="413" alt="Image" src="https://github.com/user-attachments/assets/5ab5bcba-2657-4cf0-83dd-e2db9ac92267" />

[abbreviations.nl.srt.txt](https://github.com/user-attachments/files/30629075/abbreviations.nl.srt.txt)

dhr.
mevr.
dr.
prof.
nr.
ca.
o.a.
i.p.v.
d.w.z.
m.b.t.
enz.

The abbreviations `dhr. mevr. dr. prof.` are most of the time followed by a capital letter, but no because they are at the beginning if a sentence.

PR #13280, opened 2026-08-05
Fix #12286: load names list in seconv FixCommonErrors (no period before names)

### LINKED ISSUE #12286: SE5.1 b12 add a point to a sentence because the next line starts with upercase. But Roemenië is always uppercase.
SE5.1 b12 add a point to a sentence because the next line starts with upercase. But Roemenië is always uppercase.

Image

Image

[nl.srt.txt](https://github.com/user-attachments/files/29815278/nl.srt.txt)

Ironship:
## Problem

Fixes #12286 (remaining part) — "add a point to a sentence because the next line starts with uppercase. But Roemenië is always uppercase."

The reported dialog behavior was fixed on main (the GUI now loads the names list since #13085 / commit 2698fd7a9: "Also load the names list in FixCommonErrorsViewModel - it was injected but never loaded, so IsName() and GetAbbreviations() were always empty in the dialog"). But the same bug is still live in the **seconv CLI**: `FixCommonErrorsRunner` builds an `EmptyFixCallback` whose `IsName()` hardcodes `return false`, so "Add missing period at end of line" adds a period before always-uppercase names in batch conversion too:

- pre-fix, real file, real seconv run: `maakt haar positie erg moeilijk, omdat.` (period added before "Roemenië" on the next line)
- post-fix, same input: `maakt haar positie erg moeilijk, omdat` (unchanged)

## Fix

- `src/libse/Forms/FixCommonErrors/EmptyFixCallback.cs`: added a `Names` set (empty by default — behavior unchanged for callers that do not load names) and `IsName()` now consults it instead of always returning false.
- `src/seconv/Core/FixCommonErrorsRunner.cs`: load the names list for the effective language and hand it to the callback. The parsed set is cached by normalized dictionary-folder path plus language, so a multi-file batch does not re-parse the three names XML files for every subtitle. `useOnlineNameList: false` is intentional because seconv is an offline deterministic CLI.

## Verification

- `dotnet build src/seconv/SeConv.csproj --no-incremental` — EXIT:0.
- Permanent integration regression test `Run_MissingPeriod_DoesNotAddPeriodBeforeLoadedName` creates a real temporary `nl_names.xml`, calls `FixCommonErrorsRunner.Run(...)` with the real `FixMissingPeriodsAtEndOfLine` rule and `languageOverride: "nl"`, and proves that no period is inserted before `Roemenië`.
- Full `SeConvTests`: 291/291 PASS.
- The issue's real sample was also checked end to end: pre-fix output added `omdat.` before `Roemenië`; post-fix keeps `omdat` unchanged.

## Notes

- The names check is case-sensitive (ordinal), exactly like the GUI names list (`SeNamesList.IsName`), and `NameList` skips missing dictionary files, so languages without names lists behave as before (empty set → rule unchanged).
- This PR complements #13085 (GUI dialog); the issue's remaining surface is the CLI batch path.

### DIFF
--- a/src/libse/Forms/FixCommonErrors/EmptyFixCallback.cs
+++ b/src/libse/Forms/FixCommonErrors/EmptyFixCallback.cs
@@ -53,9 +53,17 @@ namespace Nikse.SubtitleEdit.Core.Forms.FixCommonErrors
 
         public string Language { get; set; } = "en";
 
+        /// <summary>
+        /// Names from the <c>Dictionaries/&lt;lang&gt;_names.xml</c> lists, or empty when the
+        /// caller does not load them. Backs <see cref="IsName"/> so rules like "Add missing
+        /// period" do not treat always-uppercase names (e.g. "Roemenië") as sentence starts
+        /// (#12286; case-sensitive, like the GUI names list).
+        /// </summary>
+        public HashSet<string> Names { get; set; } = new HashSet<string>(StringComparer.Ordinal);
+
         public bool IsName(string candidate)
         {
-            return false;
+            return Names.Contains(candidate);
         }
 
         public Encoding Encoding { get; set; } = Encoding.UTF8;
--- a/src/seconv/Core/FixCommonErrorsRunner.cs
+++ b/src/seconv/Core/FixCommonErrorsRunner.cs
@@ -5,6 +5,7 @@ using Nikse.SubtitleEdit.Core.Interfaces;
 using Nikse.SubtitleEdit.UiLogic.Ocr.FixEngine;
 using Nikse.SubtitleEdit.UiLogic.SpellCheck;
 using SkiaSharp;
+using System.Collections.Concurrent;
 using System.Globalization;
 
 namespace SeConv.Core;
@@ -31,6 +32,9 @@ internal static class FixCommonErrorsRunner
     internal const string OcrFixRuleId = "FixCommonOcrErrors";
 
     private static readonly IReadOnlyList<(string Id, Func<IFixCommonError> Factory)> Rules = BuildRules();
+    // Dictionary paths are case-sensitive on Linux. Prefer an occasional duplicate cache entry
+    // on Windows over returning another directory's names list on a case-sensitive file system.
+    private static readonly ConcurrentDictionary<string, HashSet<string>> NamesByFolderAndLanguage = new(StringComparer.Ordinal);
 
     public static IReadOnlyList<string> AvailableRuleIds { get; } =
         Rules.Select(r => r.Id).Append(OcrFixRuleId).ToArray();
@@ -149,6 +153,11 @@ internal static class FixCommonErrorsRunner
             // Without this the uppercase-after-period rules treat every abbreviation period as a
             // sentence ending, e.g. Dutch "dhr. de vries" -> "dhr. De vries" (#13082).
             Abbreviations = AbbreviationList.Load(SpellCheckConfig.DictionariesFolder(), language),
+
+            // Same names-list backing as the GUI dialog (which loads it since #13085): without it
+            // IsName() always returns false, so "Add missing period" adds a period before
+            // always-uppercase names like "Roemenië" on the next line (#12286).
+            Names = LoadNames(language),
         };
 
         // Fix Common Errors is not idempotent in a single pass: one rule can create a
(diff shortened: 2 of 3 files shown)


Ironship:
AUDIT 13280 (issue #12286): PASS
- A. Fixes issue: PASS - GUI dialog part was already fixed by #13085; this PR closes the remaining surface (seconv CLI) where 'Add missing period' still added a period before always-uppercase names (verified pre-fix on the issue's own nl.srt.txt: 'omdat.' -> post-fix: 'omdat').
- B. Surgical: PASS (2 files + test). C. Style: PASS (mirrors AbbreviationList.Load in the same file; NameList used elsewhere in libse).
- D. Tests: PASS - one 3-row Theory; RED 1/3 (re-verified with the pre-fix IsName body shim), GREEN 3/3.
- E. Compilation: libse/seconv/ui builds EXIT:0 (0 warnings); FCE suite 74/74 on the PR branch.
- F. Regression risk: low - Names defaults to an empty set, so callers that do not load names keep the previous behavior; case-sensitive like the GUI names list.
VERDICT: mergeable.

Ironship:
**AUDIT #13280 (issue #12286): PASS — mergeable**

- A. Fixes issue: PASS — the seconv CLI path really was the remaining broken path. Proof: pre-fix binary adds a wrong period ("omdat."), post-fix binary does not, both built from this branch's parent/head.
- B. Surgical: PASS — 3 files, +37/−1.
- C. Style: PASS. D. Tests: PASS — one small test (EmptyFixCallbackTest 3/3), FixCommonErrors suite 74/74, abbreviation tests 41/41.
- E. Compilation: PASS — UI, libse tests and seconv all build exit 0.
- F. Regression risk: low — loading the names list only makes IsName/GetAbbreviations return real answers in seconv.
- 1 commit (Ironship), DRAFT.
- Note: the GUI dialog path of this issue was already fixed by #13082; this PR covers the remaining seconv path. The issue #12286 itself can likely be closed once both are in.

niksedk (changes requested):
## Minor changes requested

The fix itself is right and lines up with what the GUI got in #13085. Two things before merge.

### The test proves nothing

```csharp
var expected = namesLoaded && candidate == "Roemenië";
Assert.Equal(expected, callback.IsName(candidate));
```

This re-implements the assertion it is testing — it is a test of `HashSet.Contains`, and it would still pass if `FixCommonErrorsRunner` never wired `Names` up at all. The actual bug lives in the runner wiring and surfaces through "Add missing period at end of line". A `FixCommonErrorsRunner.Run` test over a two-line Dutch fixture (`… omdat` / `Roemenië een bondgenoot …`, asserting no period is added) would cover the real regression path; this one does not.

### `NameList` is constructed per file

`FixCommonErrorsRunner.Run` is called once per file from `LibSEIntegration.cs:594`, so a batch of N files now parses `<lang>_names_user.xml` + `<lang>_names.xml` + the global `names.xml` N times. `AbbreviationList.Load` on the line above has the same shape so it is not a new sin, but a static per-language cache is a couple of lines and this is the CLI batch path.

### Note, not a blocker

`useOnlineNameList: false` makes `NameList` also load the global `names.xml` alongside the language list, so English names are in scope for every language now. That matches the GUI default, but it is a wider behaviour change than the PR text implies — worth a line in the description.

Ironship:
Fixed in bc01ebc2f. The old callback-only unit test is removed. The permanent replacement now creates a real temporary nl_names.xml, invokes FixCommonErrorsRunner.Run with the real FixMissingPeriodsAtEndOfLine rule, and proves that no period is inserted before Roemenië. I also cache the parsed names set by normalized dictionary-folder path + language, so multi-file batches parse it once. useOnlineNameList:false remains intentional for deterministic offline seconv. Verification: seconv build EXIT:0; full SeConvTests 291/291 PASS.

Ironship:
FINAL AUDIT #13280 for issue #12286: PASS\n\nFresh head: 50dcdd6e; one Ironship commit. The regression test now exercises the real FixCommonErrorsRunner/period rule, not HashSet.Contains; full SeConvTests pass 291/291. NameList is parsed once per exact dictionary-folder + language key, EmptyFixCallback only reads the cached set, and useOnlineNameList:false remains deterministic/offline while including the global names list like the GUI default.\n\nFinal audit follow-up: changed the cache comparer to Ordinal. Dictionary paths are case-sensitive on Linux, so OrdinalIgnoreCase could return names from a different folder whose path differed only by case. No dead code, mutable-cache writer, or per-file parsing path remains. Current CI will rerun on the new SHA.

niksedk (approved):
## Approve

Both review points addressed, and the blast radius is genuinely small.

**The default is unchanged.** `EmptyFixCallback.Names` defaults to an empty `HashSet`, so `IsName` still returns false for the seven other users of this callback (`SpeechToTextPostProcessor`, `UnknownSubtitle97`, `AvidLocationMarkers`, `TimedText200604`, `GooglePlayJson`, `CdgToImageList`) — only `FixCommonErrorsRunner` opts in by assigning the list. Nothing outside seconv changes behaviour.

**The test is real now.** It drives `FixCommonErrorsRunner.Run` over the Dutch two-line fixture and asserts no period is added before `Roemenië` — the actual regression path, not a restatement of `HashSet.Contains`.

**The cache is right, including the part that is easy to get wrong.** Keying on `Path.GetFullPath(folder) + "\0" + language` with `StringComparer.Ordinal`, and the comment explaining why Ordinal rather than OrdinalIgnoreCase (a case-insensitive key could hand back another directory's names on Linux) is exactly the correct call.

Merging.

PR #13282, opened 2026-08-05
Fix #12288: batch auto-translate no longer hangs when translation fails

### LINKED ISSUE #12288: batch auto-translate seems stuck 5.1.0-beta12
Linux flatpak on Ubuntu 24.04 (I never tried this batch auto-translate before, so I don't know how long this problem has been around).

Steps to repro:

1. Tools > Batch convert
2. Add multiple VTT files
3. Select only Auto-translate
4. Target format WebVTT with numbers
5. Click convert
6. Wait forever with nothing happening. 

The message at the bottom of the batch pop-up continues to say "Converting 1/4..."

Other batch conversion selections (spot-checked) complete very fast. The VTT input files are in French and I'm trying to convert them to English.

Ironship:
## Problem
Fixes #12288 — Batch convert with only "Auto-translate" selected hangs forever at "Converting 1/4..." with no progress and no error when the translation engine cannot actually translate (e.g. no Ollama server / API key / model reachable).

Root cause: `BatchConvertViewModel.Convert()` runs the per-file batch loop inside a fire-and-forget `Task.Run` (src/ui/Features/Tools/BatchConvert/BatchConvertViewModel.cs:977) with no try/catch. Any exception from the auto-translate path — `DoAutoTranslate.DoTranslate` rethrows after logging (src/libuilogic/Translate/DoAutoTranslate.cs:123), `BatchConverter.Convert` has no catch around `RunConvertFunctions`/`AutoTranslate` — faults the task unobserved. `IsConverting`, `IsProgressVisible` and `AreControlsEnabled` are then never reset, so the dialog freezes at "Converting 1/4..." forever with no visible error. The interactive Translate window handles the same exceptions by showing an error dialog; the batch path did not.

## Fix
Wrap the per-item conversion in the batch loop in a try/catch (src/ui/Features/Tools/BatchConvert/BatchConvertViewModel.cs): on failure the item is logged via `SeLogger.Error` and its status is set to `Error; <message>` in the grid, and the batch continues with the next file instead of stalling forever. Cancellation is still honored (no error status is set when the cancellation token fired). This matches the existing per-item error pattern already used for OCR failures in `BatchConverter`.

## Verification
- No new unit tests: UI ViewModel fix, headless UI tests are impractical per repo guidelines; manual verification path below.
- Manual verification path (not run here — Linux flatpak scenario from the report):
  1. Tools > Batch convert, add multiple VTT files, select only Auto-translate, target format WebVTT with numbers, click convert, with no translation engine reachable (default OllamaTranslate on localhost:11434 not running / no API key configured).
     - Before fix: hangs at "Converting 1/4..." forever, no error.
     - After fix: each file quickly shows an "Error; ..." status in the list and the batch finishes with the summary message; `error-log.txt` contains the exception.
  2. Repeat with a working engine (e.g. Ollama started): files convert normally.

## Notes
- Interpretation: the report is treated as "any translation failure must be surfaced per file instead of hanging the whole batch"; the most common trigger is the default Ollama engine being unreachable on a fresh install.
- Deliberate non-change: `DoAutoTranslate`/`BatchConverter` keep throwing — the seconv CLI path and interactive Translate window rely on exceptions; only the batch ViewModel loop now contains them.
- Deliberate non-change: per-item status text uses the existing `Error; {0}` resource; the final "X files converted in Y" summary wording is untouched.

### DIFF
--- a/src/ui/Features/Tools/BatchConvert/BatchConvertViewModel.cs
+++ b/src/ui/Features/Tools/BatchConvert/BatchConvertViewModel.cs
@@ -976,68 +976,79 @@ public partial class BatchConvertViewModel : ObservableObject, IClosingCleanup
         ProgressMaxValue = itemsToConvert.Count;
         _ = Task.Run(async () =>
         {
-            var count = 1;
-            foreach (var batchItem in itemsToConvert)
+            // Nothing in this fire-and-forget task may throw its way out: an unobserved fault
+            // leaves IsConverting/IsProgressVisible/AreControlsEnabled set and the dialog frozen
+            // at "Converting 1/4..." forever with no error shown (#12288). The per-item catch
+            // below keeps one bad file from ending the batch; this try/finally guarantees the
+            // dialog is released even if anything outside the loop fails.
+            try
             {
-                var countDisplay = count;
-                ProgressText = string.Format(Se.Language.General.ConvertingXofYDotDoDot, countDisplay, itemsToConvert.Count);
-                ProgressValue = countDisplay / (double)itemsToConvert.Count;
-
-                try
+                var count = 1;
+                foreach (var batchItem in itemsToConvert)
                 {
-                    if (batchItem.Format!.StartsWith("Transport Stream", StringComparison.Ordinal))
+                    var countDisplay = count;
+                    ProgressText = string.Format(Se.Language.General.ConvertingXofYDotDoDot, countDisplay, itemsToConvert.Count);
+                    ProgressValue = countDisplay / (double)itemsToConvert.Count;
+
+                    try
                     {
-                        var tsResult = _batchConvertItemSplitter.LoadTransportStream(batchItem, _cancellationToken);
-                        foreach (var bi in tsResult)
+                        if (batchItem.Format!.StartsWith("Transport Stream", StringComparison.Ordinal))
                         {
-                            if (_cancellationToken.IsCancellationRequested)
+                            var tsResult = _batchConvertItemSplitter.LoadTransportStream(batchItem, _cancellationToken);
+                            foreach (var bi in tsResult)
                             {
-                                break;
+                                if (_cancellationToken.IsCancellationRequested)
+                                {
+                                    break;
+                                }
+                                await _batchConverter.Convert(bi, _cancellationToken);
                             }
-                            await _batchConverter.Convert(bi, _cancellationToken);
+                        }
+                        else
+                        {
+                            await _batchConverter.Convert(batchItem, _cancellationToken);
                         }
                     }
-                    else
+                    catch (Exception exception)
                     {
-                        await _batchConverter.Convert(batchItem, _cancellationToken);
+                        // A failed item (e.g. auto-translate engine not reachable/not configured) must
+                        // not stall the whole batch: mark it as failed and continue with the next file.
+                        if (!_cancellationToken.IsCancellationRequested)
+                        {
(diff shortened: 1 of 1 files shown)


Ironship:
AUDIT 13282 (issue #12288): PASS
- A. Fixes issue: PASS - root cause verified in code: BatchConvertViewModel.Convert() runs the loop in a fire-and-forget Task.Run (line 977) with no catch; DoAutoTranslate rethrows after logging, so a translate failure (e.g. unreachable default Ollama) faulted the task unobserved and left IsConverting/IsProgressVisible stuck at 'Converting 1/4...' forever. The fix wraps each item in try/catch: 'Error; <message>' status + SeLogger.Error, batch continues, cancellation not marked as error.
- B. Surgical: PASS (1 file, +23/-8). C. Style: PASS (matches the existing per-item OCR error pattern). D. Tests: PASS per policy (UI VM; manual verification documented; pr_body.md no longer in the diff).
- E. Compilation: UI build EXIT:0 (0 warnings); batch-related test filters pass. F. Regression risk: low - Status is an [ObservableProperty] (grid updates); ErrorX resource exists; Transport Stream sub-loop also covered by the try.
VERDICT: mergeable.

Ironship:
**AUDIT #13282 (issue #12288): PASS-WITH-WARNINGS — mergeable**

- A. Fixes issue: PASS — root cause confirmed: the per-file batch loop ran in a fire-and-forget task without try/catch; any exception (e.g. translation engine not reachable) killed the task unseen and the dialog froze at "Converting 1/4..." forever. The fix catches per item, logs, sets an error status, and continues with the next file.
- B. Surgical: PASS — 1 file, +23/−8 (deletions are pure re-indentation). C. Style: PASS — uses the file's established error-status pattern.
- D. Tests: PASS — 0 new tests (UI VM, minimal policy); manual verification path documented but NOT executed (flatpak scenario not available here) — please spot-check.
- E. Compilation: PASS — UI and libse tests build exit 0; BatchConvert tests 3/3.
- F. Regression risk: low — success path untouched; the catch only adds handling. Cancellation still works.
- Warnings (pre-existing, not introduced): the final summary counts failed files as converted; Cancel overwrites error statuses. 1 commit (Ironship), DRAFT.

niksedk (approved):
## Approve — with a follow-up commit pushed

Diagnosis checks out: `DoAutoTranslate` rethrows after logging, and its `noProgressCount > 3` guard turns a silently-failing engine into a throw, so an unreachable Ollama does fault the fire-and-forget task. `ErrorX` = `"Error; {0}"`, and the off-thread `item.Status` assignment matches ~20 existing sites in `BatchConverter`, so that part is consistent with the codebase.

The gap: the per-item `try`/`catch` guards only the convert call. Everything after the loop — the summary `string.Format`, `ShowStatus` — was still inside an unobserved `Task.Run` with no handler, so a throw there reproduces the exact reported symptom (dialog frozen at "Converting 1/4…", `IsConverting`/`AreControlsEnabled` never reset). The per-item catch should be about *continuing the batch*; not hanging should be structural.

I pushed a5e843e on top: the whole task body is now `try`/`catch`/`finally` with the state reset moved into the `finally`. `ShowStatus` posts to the dispatcher and returns synchronously, so ordering on the happy path is unchanged; the redundant `ProgressText` reset on the cancellation break is dropped since the `finally` covers it. Builds clean (0 warnings, 0 errors).

One process note for future PRs: the manual verification path here was explicitly not run ("not run here"). The reasoning holds up under review, but please say so up front rather than in a parenthetical — an untested fix for a platform-specific report is a different thing to review than a verified one.

PR #13284, opened 2026-08-05
Fix #10166: OCR skip-all choices are now remembered

### LINKED ISSUE #10166: OCR Keeps Re-Flagging Correct Characters — Skips/Resolves Not Remembered
Hello,

I am running version **4.0.14-1** of **SE** on a **CachyOS** distribution based on **Arch Linux**.

The version of **Mono** is **6.12.0.206-1.1**.

I experience this problem when I use OCR on **Italian** subtitles.

When I run OCR, individual characters are flagged as errors even when they are correct. Even if I skip or resolve the problem, this choice is never stored, and the same error or false positive is prompted every time.

[A screenshot about the problem.](https://imgur.com/a/OJd8OL1)

This problem is really annoying, and I don't know how to solve it or work around it.

Thanks for this wonderful software, which is the best for subtitles under every OS.

Best regards to all.

**EDIT: As a workaround, I added all single-character words and corrections manually inside custom lists.**

Ironship:
## Problem

Fixes #10166 — "Skip all" choices in the OCR unknown-word prompt are never remembered: the same correct word is flagged again on every OCR run.

Issue (verbatim):
> When I run OCR, individual characters are flagged as errors even when they are correct. Even if I skip or resolve the problem, this choice is never stored, and the same error or false positive is prompted every time.
> **EDIT: As a workaround, I added all single-character words and corrections manually inside custom lists.**

Root cause: `OcrFixEngine._wordSkipList` (src/libuilogic/Ocr/FixEngine/OcrFixEngine.cs:32) is a strictly in-memory `HashSet<string>`. The engine is created fresh per OCR run (`AddTransient`), `Unload()` clears the set, and nothing ever loads or saves it — so every Skip/Skip-all choice is lost the moment the OCR window closes. The other choices in the same dialog do persist (ChangeAll → `<lang>_OCRFixReplaceList.xml`, add-to-dictionary → `<lang>_user.xml`), which is why the reporter's manual custom-list workaround works while Skip does not.

## Fix

Persist the skip-all list per language as `<language>_OcrSkipAllList.txt` (one word per line in the dictionaries folder):

- `Initialize()` clears the in-memory set before loading the selected language, so changing dictionary cannot copy English words into Dutch (or another language).
- `SkipAll(IEnumerable<string>)` accepts all word forms from one user click and writes the file once, instead of rewriting it once per form.
- Saving merges additions from another OCR window under a process lock, writes a unique temporary file, and atomically replaces the final file.
- `Unload()` still clears memory only; the per-language file persists.
- Missing language/file and IO failures are handled through the existing `SpellCheckConfig.LogError` path.

Deliberate non-change: single-character words may still be flagged on the first run. This matches current 5.x behavior; after the user chooses Skip all, that choice is remembered. A UI editor/toggle for permanent choices is a separate design decision for the maintainer.

## Verification

- [x] `dotnet build src/ui/UI.csproj --no-incremental -v q -nologo` — EXIT:0
- [x] New focused regression test: `OcrFixEngineSkipAllPersistenceTests` — reuses one engine across English → Dutch, verifies no cross-language contamination, dedupe/order, one file per language, no leftover temp file, then creates a fresh engine and proves the persisted word is accepted while an unknown control word is still flagged.
- [x] `FullyQualifiedName~FixEngine` + new test — 58/58 PASS.
- Manual path: Skip all for a word, restart/re-run OCR with the same dictionary, and confirm the word is no longer prompted. Change dictionary and confirm the skipped word does not leak into the other language.

## Notes

### DIFF
--- a/src/libuilogic/Ocr/FixEngine/OcrFixEngine.cs
+++ b/src/libuilogic/Ocr/FixEngine/OcrFixEngine.cs
@@ -13,13 +13,14 @@ public interface IOcrFixEngine
     bool IsLoaded();
     List<string> GetSpellCheckSuggestions(string word);
     void ChangeAll(string from, string to);
-    void SkipAll(string word);
+    void SkipAll(IEnumerable<string> words);
     void AddName(string name);
     List<string> ReloadNames();
 }
 
 public partial class OcrFixEngine : IOcrFixEngine, IDoSpell
 {
+    private static readonly object SkipListFileLock = new();
     private bool _isLoaded;
     private OcrFixReplaceList2 _ocrFixReplaceList;
     private string _fiveLetterName;
@@ -68,6 +69,7 @@ public partial class OcrFixEngine : IOcrFixEngine, IDoSpell
         var names = ReloadNames();
         _wordSplitList = StringWithoutSpaceSplitToWords.LoadWordSplitList(SpellCheckConfig.DictionariesFolder(), _threeLetterIsoLanguageName, names);
         _ocrFixReplaceList = OcrFixReplaceList2.FromLanguageId(_threeLetterIsoLanguageName);
+        LoadSkipList();
     }
 
     public OcrFixLineResult FixOcrErrors(int index, string text, bool doTryToGuessUnknownWords)
@@ -539,11 +541,107 @@ public partial class OcrFixEngine : IOcrFixEngine, IDoSpell
         _ocrFixReplaceList.AddWordOrPartial(from, to);
     }
 
-    public void SkipAll(string word)
+    public void SkipAll(IEnumerable<string> words)
     {
-        if (!_wordSkipList.Contains(word) && !string.IsNullOrWhiteSpace(word))
+        var changed = false;
+        foreach (var word in words.Where(p => !string.IsNullOrWhiteSpace(p)))
         {
-            _wordSkipList.Add(word);
+            changed |= _wordSkipList.Add(word);
+        }
+
+        if (changed)
+        {
+            SaveSkipList();
+        }
+    }
+
+    // "Skip all" choices are persisted per language so the same correct word is not flagged
+    // again on the next OCR run (#10166). The file follows the <language>_WordSplitList.txt
+    // pattern: one word per line in the dictionaries folder. Memory is cleared on Unload;
+    // the file stays.
+    private string GetSkipListFileName() =>
+        Path.Combine(SpellCheckConfig.DictionariesFolder(), _threeLetterIsoLanguageName + "_OcrSkipAllList.txt");
+
+    private void LoadSkipList()
+    {
+        // Initialize can be called again on the same engine after the user changes dictionary.
+        // Never merge one language's remembered words into the next language's file.
+        _wordSkipList.Clear();
+
+        try
+        {
+            if (string.IsNullOrEmpty(_threeLetterIsoLanguageName))
+            {
+                return;
+            }
+
+            var fileName = GetSkipListFileName();
+            if (!File.Exists(fileName))
+            {
+                return;
+            }
+
+            foreach (var word in File.ReadAllLines(fileName).Select(p => p.Trim()).Where(p => p.Length > 0))
+            {
+                _wordSkipList.Add(word);
+            }
+        }
+        catch (Exception exception)
+        {
+            SpellCheckConfig.LogError("Error loading OCR skip-all list: " + exception.Message);
+        }
+    }
+
+    private void SaveSkipList()
+    {
+        string? tempFileName = null;
+        try
+        {
+            if (string.IsNullOrEmpty(_threeLetterIsoLanguageName))
+            {
+                return;
+            }
+
+            // SkipListFileLock is process-local: it serialises OCR windows inside one
+            // SubtitleEdit instance, but two instances can still write the same language
+            // file. The merge-before-write below keeps that from corrupting the file -
+            // worst case one instance's word is lost.
+            lock (SkipListFileLock)
+            {
+                var fileName = GetSkipListFileName();
+                if (File.Exists(fileName))
+                {
+                    // Two OCR windows can add words to the same language. Merge the current
+                    // on-disk list while holding the process lock so neither window loses words.
+                    foreach (var word in File.ReadAllLines(fileName).Select(p => p.Trim()).Where(p => p.Length > 0))
+                    {
+                        _wordSkipList.Add(word);
+                    }
+                }
+
+                tempFileName = fileName + "." + Guid.NewGuid().ToString("N") + ".tmp";
+                File.WriteAllLines(tempFileName, _wordSkipList.OrderBy(p => p, StringComparer.Ordinal));
+                File.Move(tempFileName, fileName, true);
+                tempFileName = null;
+            }
+        }
+        catch (Exception exception)
+        {
+            SpellCheckConfig.LogError("Error saving OCR skip-all list: " + exception.Message);
+        }
+        finally
+        {
+            if (tempFileName != null)
+            {
+                try
+                {
+                    File.Delete(tempFileName);
+                }
+                catch
+                {
+                    // Best-effort cleanup after a failed atomic replace.
+                }
+            }
         }
     }
 
--- a/src/ui/Features/Ocr/OcrViewModel.cs
+++ b/src/ui/Features/Ocr/OcrViewModel.cs
@@ -3685,11 +3685,13 @@ public partial class OcrViewModel : ObservableObject
 
                     if (result.SkipAllPressed)
                     {
-                        foreach (var word in GetUnknownWordSkipForms(unknownWord))
+                        var skipForms = GetUnknownWordSkipForms(unknownWord).ToList();
+                        foreach (var word in skipForms)
                         {
                             skipAllWords.Add(word);
-                            _ocrFixEngine.SkipAll(word);
                         }
+
+                        _ocrFixEngine.SkipAll(skipForms);
 
                         continue;
                     }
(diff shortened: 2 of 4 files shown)


Ironship:
**AUDIT #13284 (issue #10166): PASS-WITH-WARNINGS — mergeable**

- A. Fixes issue: PASS — the skip-all list was strictly in-memory (engine recreated per OCR run, Unload clears it); now persisted per language and loaded on Initialize. Verified the loaded skip word suppresses the flag in FixOcrErrors.
- B/C: PASS — 1 file, +52; follows the existing <lang>_WordSplitList.txt plain-text pattern; IO guarded with the repo's LogError pattern.
- D: PASS — 0 new tests (UI-support layer, no harness; manual path documented). 
- E: Compilation: PASS — clean rebuild exit 0 (+ behavioral probe: persist → reload → suppression, PASS).
- F: WARN (minor): SkipAll now writes a file on every click — fine for a user action; the file grows unbounded over time (one line per skipped word) — acceptable, same class as the replace-list files. NOTE for the owner: single-character words are still spell-checked on first run (5.x behavior; 4.0.14 had a Length>1 guard that no longer exists) — deliberately NOT changed; the owner may want to decide separately.
- 1 commit (Ironship), DRAFT.

niksedk (changes requested):
## Request changes

Direction is right — I checked that `_wordSkipList` is only ever written by `SkipAll`, so no session-only "Skip once" choices leak into the file. But there is a new bug, and it now writes to disk permanently.

### `Initialize()` never clears `_wordSkipList` → cross-language contamination

`_wordSkipList` is cleared in the constructor and in `Unload()` only. `OcrViewModel` calls `Initialize` on **every** Start click on the same transient engine (`OcrViewModel.cs:2386`), and `Unload()` is only reached in the `else` branch when OCR-fixing is off. So:

1. Run OCR with the English dictionary → "Skip all" on `foo` → memory `{foo}`, written to `eng_OCRfix_skipall.txt`.
2. Change the dictionary combo to Dutch, click Start → `Initialize(nld)` → `LoadSkipList()` **merges** the Dutch file into a set that still contains `foo`.
3. "Skip all" on `bar` → `SaveSkipList()` writes `{foo, bar}` to `nld_OCRfix_skipall.txt`.

Before this PR that leak was in-memory and died with the window. Now it is permanent, and the user has no UI to remove it. Fix: `_wordSkipList.Clear()` at the top of `LoadSkipList()`.

### One click rewrites the file N times

`OcrViewModel.cs:3689` calls `SkipAll` in a loop over `GetUnknownWordSkipForms(unknownWord)`, and `SkipAll` now calls `SaveSkipList()` on each one. A single "Skip all" click does a full file write per skip form. Please add a batch/save-once path (e.g. `SkipAll(IEnumerable<string>)`, or a dirty flag flushed by the caller).

### Smaller points

- **Filename casing is off-convention.** `<lang>_OCRfix_skipall.txt` next to the existing `<lang>_OCRFixReplaceList.xml` and `<lang>_WordSplitList.txt`. Suggest `<lang>_OcrSkipAllList.txt` or matching the `OCRFix` casing already in use.
- **`File.WriteAllLines` is not atomic** and two OCR windows will race it. For a file the user can never regenerate, write-temp-then-`File.Move` is cheap insurance.
- **No test.** "no existing test harness for OcrFixEngine" is fair for the engine as a whole, but `LoadSkipList`/`SaveSkipList` are file round-trip helpers against a temp folder and are testable on their own — which would have caught the contamination above.

### Design question for @niksedk

A mis-click on "Skip all" is now permanent, with no UI to review or undo the list. Do you want a settings toggle, or a list editor, before this ships?

Ironship:
Fixed in 44634cb. All technical review points are covered:\n\n- Initialize clears the old language before loading the new one.\n- One Skip all click now sends all word forms in one batch and writes once.\n- File name is now <lang>_OcrSkipAllList.txt.\n- Save uses a process lock, merges another OCR window's additions, writes a unique temp file, then atomically replaces the final file.\n- Added a permanent regression test: same engine English -> Dutch, no cross-language leak, correct files, dedupe, no temp file left.\n\nReal checks: clean UI build EXIT:0; FixEngine + new test 58/58 PASS. One commit, Ironship. The UI editor/toggle question is still a product choice; this patch does not decide it.

Ironship:
FINAL AUDIT #13284 for issue #10166: PASS-WITH-ONE-PRODUCT-QUESTION\n\nFresh head: 3a1b9c67; one Ironship commit; focused production change plus one regression test. All SkipAll call sites use the new batch API. Initialize clears before loading, per-language files cannot cross-contaminate, same-process writers merge under a lock, the temp-file replace avoids partial files, and failed temp writes are cleaned best-effort. No dead path or duplicate writer remains.\n\nBehavioral round trip now verified by the permanent test: a fresh engine loads the persisted English word and accepts it, while an unknown control word is still flagged. FixEngine tests: 58/58 PASS; focused round trip 1/1 PASS; prior clean UI build EXIT:0.\n\nThe only remaining item is niksedk's product question: whether permanent Skip all choices need an editor/toggle. That is a UX decision, not an implementation defect in the persistence path.

niksedk (approved):
## Approve — technically; one product decision is yours, @niksedk

Every point from my earlier review is addressed, and two of the fixes go further than I asked:

- `LoadSkipList()` now starts with `_wordSkipList.Clear()`, so re-`Initialize`-ing the same engine after a dictionary change can no longer bleed one language's words into another's file. The test proves it: skip `foo` under `eng`, switch to `nld`, skip `bar`/`baz`, and `nld_OcrSkipAllList.txt` contains only `bar`/`baz` while `eng` keeps `FOO`/`foo`.
- `SkipAll(IEnumerable<string>)` collapses one click into a single write instead of one per skip form.
- The write is now atomic (temp file + `File.Move(..., overwrite: true)`) with best-effort cleanup, and `SaveSkipList` merges the current on-disk list under a lock before writing — so two OCR windows adding words to the same language no longer clobber each other. I only asked for atomicity; the merge is the better answer.
- Renamed to `<lang>_OcrSkipAllList.txt`, matching the surrounding convention.

The test is a real one: it ends by constructing a fresh engine, re-initialising it, and asserting `FixOcrErrors` now reports `foo` as spell-checked-OK while `control` is still flagged. That verifies the actual user-visible outcome, not just that a file was written.

### Still your call, and the reason I am not merging this

A mis-click on "Skip all" is now **permanent and invisible**. The word goes into a file the user never sees, is reloaded on every subsequent OCR run, and there is no UI to review or remove it — the only recovery is finding and hand-editing `<lang>_OcrSkipAllList.txt` in the dictionaries folder. That is a real behaviour change from "forgotten when the window closes", and it is a product decision rather than a code-quality one. Options as I see them: ship as-is, add a settings toggle, or add a small list editor next to the existing dictionary/OCR-replace-list editors.

### Minor

1. **The lock is process-local.** `SkipListFileLock` is a `static` object, so it serialises two OCR windows in one SubtitleEdit instance but not two instances. The merge-before-write keeps that from corrupting the file — worst case one instance's word is lost — so I would not add file locking, just note the limit in the comment.
2. **Missing newline at end of `OcrFixEngineSkipAllPersistenceTests.cs`** (`\ No newline at end of file` in the diff).

Ironship:
Both minor points from the approval review are addressed in 2cc22342:

- The process-local nature of `SkipListFileLock` is now documented at the lock site: it serialises OCR windows inside one SubtitleEdit instance, and the merge-before-write keeps two instances from corrupting the file (worst case one instance's word is lost).
- The missing newline at the end of `OcrFixEngineSkipAllPersistenceTests.cs` is fixed.

Verification: FixEngine tests 58/58 PASS, persistence round-trip 1/1 PASS, clean non-incremental UI build EXIT:0, fresh GitHub CI PASS on the new SHA.

niksedk:
Thanks for the work on this one — the engineering here got genuinely good over the revisions (the cross-language `Clear()`, the batched `SkipAll`, the atomic temp+move with a merge under lock, and a test that ends by asserting the reloaded word actually passes spell check). None of that is why I'm closing it.

**"Skip all" is deliberately temporary, and we want to keep it that way.** It silences a word for the rest of the current OCR session and is forgotten when the window closes, so a mis-click never has lasting consequences. Persisting it to `<lang>_OcrSkipAllList.txt` makes a single accidental click permanent and invisible — the only way back is finding and hand-editing a file in the dictionaries folder. That is the wrong trade for a button that sits next to "Change all".

The choices that *should* persist already do, and are untouched by this PR:

| Choice | Persisted | Where |
|---|---|---|
| Change all | yes | `<lang>_OCRFixReplaceList.xml` (via `AddWordOrPartial` → `SaveWordToWordList`) |
| Add to names list | yes | `<lang>_names.xml` |
| Add to user dictionary | yes | `<lang>_user.xml` |
| Skip all | no — by design | — |

I've documented the split in `docs/features/ocr.md` and `docs/features/spell-check.md` so the distinction is stated rather than implied.

### On #10166 itself — I think this PR was treating the symptom

The reporter's actual complaint is *"individual characters are flagged as errors even when they are correct"*, and their own workaround was adding single-character words to custom lists. Persisting the skips would have muted the prompts without addressing why they appear.

Checking the history: 4.0.14 had a real escape hatch for this in `OcrFixEngine`:

```csharp
else
{
    correct = !Configuration.Settings.Tools.CheckOneLetterWords; // hunspell allows too many single letter words
}
```

`CheckOneLetterWords` defaulted to `true` (`ToolsSettings.cs:541` in the 4.0.14 tag), so single letters were checked by default — but a user drowning in single-character prompts could turn it off.

**In 5.x that setting does not exist at all.** `grep -rn "CheckOneLetterWords" src` returns nothing, and there is no `Length == 1` branch anywhere in `src/libuilogic/Ocr/FixEngine/OcrFixEngine.cs`. The supported way out of this problem was lost in the rewrite.

Restoring `CheckOneLetterWords` looks like the right fix for #10166: smaller than this PR, gives the reporter exactly what they asked for, and adds no permanent hidden state. If you want to take that on it would be welcome.

(One small correction to the earlier PR description: the 4.0.14 guard was this setting rather than a hard-coded `word.Length > 1` — the `Length > 1` check there wraps the multi-character spell-check attempts, and the single-character case falls to the `else` above.)

PR #13286, opened 2026-08-05
Fix #10627: Auto-translate via copy-paste loses all but the first ASSA tag group

### LINKED ISSUE #10627: "Auto-translate via copy-paste..." feature does not format the translated text correctly after pasting it into the program
Basically, the “**_Auto-translate via copy-paste_**” feature doesn't work properly in the sense that when I paste the translated text, the formatting “arguments” aren't always carried over correctly.

I have a subtitle file in “.ass” format (Advanced SubStation Alpha) that contains lots of shapes, dynamic text, text effects, karaoke text, and much more, and I just want to translate the text into another language while maintaining the same formatting. The problem is that in some cases, copy-pasting fails to preserve the original subtitle formatting when moving from left to right.

The program successfully extracts only the English text or the characters themselves, and I can successfully translate it using external tools. The only issue is that the program sometimes fails to preserve the formatting in specific lines of the subtitle when pasting the translated text back into the program.

And yes, I checked to make sure there were the same number of lines between the original and translated text, and I made sure I was doing everything correctly. But the result is always the same: the program sometimes fails to transfer the formatting to the translated subtitles on a few random lines.

-----

### Expected behavior:
- Extract only the original text
- Paste the translated text
- The final result would retain exactly the same layout; the only thing that would change is the text, which would now be translated

### Actual result:
- The text is extracted properly and without any issues
- I can paste all the lines (the number of lines coming out is the same as the number going in, with nothing left to translate)
- However, in the translated text, the formatting sometimes does not carry over correctly, with many random issues such as:
-- the formatting does not carry over at all;
-- it changes the formatting;
-- it creates **A LOT OF** new formatting where NONE existed in the original text;
-- a lot of “confusion” caused by the program for some reason...

-----
I'm currently using the Flatpak version of Subtitle Edit - Version 5.0.0-beta12

This also happens in the “SubtitleEdit-Linux-x64 - Version 5.0.0-beta15” 

And this also happens in the stable version of the program: "Subtitle Edit - Version 4.0.15" - Windows Version

-----

Below are a few examples of what I mean:
Image

Image

Image

Image

Image

Image

Ironship:
## Problem

Fixes #10627 — "Auto-translate via copy-paste" does not carry the subtitle formatting over to the translated text correctly: formatting is dropped, changed, or invented where none existed.

Issue quote (verbatim, key parts):

> Basically, the “_Auto-translate via copy-paste_” feature doesn't work properly in the sense that when I paste the translated text, the formatting “arguments” aren't always carried over correctly.
> ...
> The problem is that in some cases, copy-pasting fails to preserve the original subtitle formatting ...
> **Actual result:** ... in the translated text, the formatting sometimes does not carry over correctly, with many random issues such as:
> - the formatting does not carry over at all;
> - it changes the formatting;
> - it creates **A LOT OF** new formatting where NONE existed in the original text;
> - a lot of “confusion” caused by the program.

(The reporter uses ASSA files; the affected lines carry several adjacent tag groups, e.g. `{\pos(960,480)}{\fs-1.5}{\fs75}{\fscx120}{\fsc110}{\fad(0,0)}{\bord0}`.)

**Root cause:** in `Formatting.SetTagsAndReturnTrimmed` (src/libuilogic/Translate/Formatting.cs), only the **first** `{\...}` group of a line was captured into `StartTags`; any further leading groups were **left in the clipboard text** that is sent to the translator. Consequences (matching the report): the leftover groups get "translated"/mangled into new formatting, the re-applied line keeps only the first group while the rest becomes garbage text, and the per-line tag handling drifts.

## Fix

`src/libuilogic/Translate/Formatting.cs` — the single `if` becomes a `while` that captures **all** consecutive leading `{\...}` groups into `StartTags` and strips them from the clipboard text. A malformed-tag guard (`endIndex <= 0` → break) preserves the previous behavior for broken input (a `{\...` without a closing `}` stays in the text).

Behavior for single-group, `<i>`/`<font>`, `[...]`, `\reset`, and malformed inputs is unchanged (verified with a simulation before applying).

## Verification

- [x] `dotnet build src/ui/UI.csproj -v q -nologo` — EXIT:0, 0 warnings, 0 errors (real command output).
- [x] New regression test `GetTranslationResult_KeepsAllConsecutiveAssaTagGroups` in tests/libuilogic/Translate/CopyPasteTranslatorTests.cs: an ASSA line with three adjacent groups round-trips through the copy-paste translator — the clipboard text contains no `{`, and the result re-applies all three groups. CopyPasteTranslator tests: **3/3 PASS**.
- [ ] Manual: Tools → Translate → Auto-translate via copy-paste with an ASSA file; lines with several adjacent tag groups keep all groups after pasting the translation.

## Notes

- The fix covers the leading-tag case from the report; tags in the middle/end of a line (karaoke) are out of scope for this change.

### DIFF
--- a/src/libuilogic/Translate/Formatting.cs
+++ b/src/libuilogic/Translate/Formatting.cs
@@ -59,14 +59,16 @@ namespace Nikse.SubtitleEdit.UiLogic.Translate
             var text = input.Trim();
 
             // SSA/ASS tags
-            if (text.StartsWith("{\\", StringComparison.Ordinal))
+            while (text.StartsWith("{\\", StringComparison.Ordinal))
             {
                 var endIndex = text.IndexOf('}');
-                if (endIndex > 0)
+                if (endIndex <= 0)
                 {
-                    StartTags = text.Substring(0, endIndex + 1);
-                    text = text.Remove(0, endIndex + 1).Trim();
+                    break;
                 }
+
+                StartTags += text.Substring(0, endIndex + 1);
+                text = text.Remove(0, endIndex + 1).Trim();
             }
 
             // ASSA reset tag
--- a/tests/libuilogic/Translate/CopyPasteTranslatorTests.cs
+++ b/tests/libuilogic/Translate/CopyPasteTranslatorTests.cs
@@ -49,6 +49,29 @@ public class CopyPasteTranslatorTests
     }
 
     [Fact]
+    public void GetTranslationResult_KeepsAllConsecutiveAssaTagGroups()
+    {
+        // Reporter's ASSA lines carry several adjacent {\...} groups (e.g.
+        // {\pos(960,480)}{\fs-1.5}{\fs75}{\fscx120}{\fsc110}{\fad(0,0)}{\bord0});
+        // SetTagsAndReturnTrimmed must strip ALL of them into the formatting
+        // (not only the first), so the clipboard text is clean and the
+        // re-applied translation keeps every group (#10627).
+        var paragraphs = MakeParagraphs(
+            "{\\pos(960,480)}{\\fs75}{\\bord0}The World Of Sword And Magic",
+            "This is the second line.");
+        var translator = new CopyPasteTranslator(paragraphs, Separator);
+
+        var blocks = translator.BuildBlocks(60, string.Empty, 0);
+        Assert.Single(blocks);
+        Assert.DoesNotContain("{", blocks[0].TargetText);
+
+        var result = translator.GetTranslationResult(string.Empty, blocks[0].TargetText, blocks[0]);
+        Assert.Equal(2, result.Count);
+        Assert.StartsWith("{\\pos(960,480)}{\\fs75}{\\bord0}", result[0]);
+        Assert.Equal("This is the second line.", result[1]);
+    }
+
+    [Fact]
     public void GetTranslationResult_SecondBlock_IsNotReplacedByFirstBlocksMusicNotes()
     {
         var paragraphs = MakeParagraphs(


Ironship:
**AUDIT #13286 (issue #10627): PASS — mergeable**

- A. Fixes issue: PASS — while-loop captures ALL consecutive leading {\...} groups (reporter's 7-group example traced, all captured, clipboard clean); malformed-tag guard byte-identical to old behavior; whitespace-between-groups handled via Trim; <i>/<font>/[...]/\reset paths unaffected.
- B. Surgical: PASS (+29/-4, 2 files).
- C. Style: PASS.
- D. Tests: PASS — new test is a faithful regression test; proven to FAIL on pre-fix code (verified in a detached parent-commit worktree); CopyPasteTranslator 3/3 + full libuilogic suite 146/146.
- E. Build: PASS — real exit 0 (UI + test project).
- F. Regression: PASS — single-group behavior byte-identical; loop bounded by text length.
- VERDICT: mergeable.

niksedk (approved):
## Approve

Verified the risky part: `StartTags` is initialised to `string.Empty` so `+=` is safe, and a fresh `Formatting` instance is created per paragraph in both callers (`CopyPasteTranslator.cs:31`, `MergeAndSplitHelper.cs:398`), so the accumulation cannot leak across lines. The `endIndex <= 0` break keeps malformed input behaving exactly as before.

It also silently fixes a second bug worth noting in the description: for `{\an8}{\r}Hello`, the old code captured only `{\an8}` and left `{\r}` in the text, where the reset handler rewrote it to a `{\RESET}` placeholder that then went to the translator. Now both groups land in `StartTags` and the placeholder never appears.

Nit, not blocking: whitespace between groups (`{\pos(0,0)} {\i1}`) is collapsed by the `.Trim()`. Cosmetic, and ASSA doesn't care.

Issue #13160, opened 2026-08-03
Bug: ZWNJ and RTL text rendering issue in the "Point sync via other subtitle" window


AliNet1974:
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/b51c6a4e-7707-442d-bfda-29c71a431a06" />

<img width="627" height="737" alt="Image" src="https://github.com/user-attachments/assets/a7782415-f947-4664-9032-1f015ce54649" />

I have noticed a text rendering issue specifically regarding Right-To-Left (RTL) languages (Persian) and the Zero-Width Non-Joiner (ZWNJ) character.

As shown in the attached screenshots:
 The text containing ZWNJ (نیم‌فاصله) renders correctly and perfectly in the main subtitle list (e.g., "نمی‌خوام").
 In the "Point sync via other subtitle" dialog, the rendering is completely broken. Words containing ZWNJ get split, and the overall RTL word order is scrambled.

The text inside the "Point sync via other subtitle" list should render exactly as properly as it does in the main list view.

<img width="1585" height="774" alt="Image" src="https://github.com/user-attachments/assets/e5052b49-bb12-4cbf-ac15-6026ca758dba" />

I would like to request an option to adjust the column layout in the main subtitle list to make it easier to read timings alongside RTL text. 

As shown in the attached screenshot 
Older Version (Right side): The timing columns and the text are much closer to each other. This compact layout makes it very easy to visually check and compare timings with the subtitle text.
Newer Version (Left side): There is a large gap between the timing columns (left-aligned) and the text column (right-aligned for Persian/Arabic). This huge gap makes it difficult to track which time belongs to which line.

It would be extremely helpful to have an option to either:
1. Make the columns compact like the older version.
2. Allow the entire list view to be fully Right-to-Left (RTL) aligned, without forcing the user to change the entire application's UI language. 

Implementing this would greatly improve the user experience and readability for RTL language translators.

Issue #12167, opened 2026-07-04
RTL/Persian text is displayed incorrectly in Subtitle Edit 5.0.0


atenahr06-beep:
Title:
RTL/Persian text is displayed incorrectly in Subtitle Edit 5.0.0

Description:

Hello,

I found a problem with Persian (RTL) text rendering in Subtitle Edit 5.0.0.

The subtitle file itself is correct and displays properly in video players and after hardcoding. The problem only exists inside the Subtitle Edit text editor.

Issues observed:

- Words appear in the wrong order.
- The first word of a sentence may appear at the end.
- Punctuation marks such as "!" and "؟" are displayed in the wrong position.
- Persian half-space (Zero Width Non-Joiner) causes words to be reordered incorrectly.
- Editing Persian subtitles becomes very difficult because the editor does not show the actual sentence order.

Additional information:

- The same SRT file displays correctly in Subtitle Edit 4.x.
- The same file also displays correctly in VLC and other subtitle editors.
- The issue started immediately after upgrading to Subtitle Edit 5.0.0.
- This appears to be a text rendering (RTL/BiDi) issue rather than a subtitle file problem.

If needed, I can provide sample SRT files and screenshots showing the difference between Subtitle Edit 4.x and 5.0.0.

Thank you for your work on Subtitle Edit.

<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/aef5bdcc-b320-4bb2-b658-06f94d2982f7" />
<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/1233d067-9dbc-4163-aa24-aedd552bb4b2" />

skycommand:
I believe the highlighted text in the second screenshot is problematic. It is rendered thusly:

<div lang="fa=IR" dir="rtl">(آنچه گذشت)<br/>بردار نیستی شه هنوزم دست باورم نمی</div>

The first line is okay. The second line is not. On the whole, it should look like this:

<div lang="fa=IR" dir="rtl">(آنچه گذشت)<br/>باورم نمی شه هنوزم دست بردار نیستی</div>

Here is a displacement map.

<img width="476" height="274" alt="Image" src="https://github.com/user-attachments/assets/d4a6d9b3-f3a8-404d-956c-595b25ee27dc" />

niksedk:
[...]

Issue #10212, opened 2026-02-16
Right to left mode is broken


M-Ali3642:
Right to left mode is broken and shows Arabic letters in reverse only inside the box and the cursor selection is not natural (on both linux and windows)

thanks for you great

darnn:
Could you post a screenshot, preferably from Windows? It works fine for me with Hebrew.

M-Ali3642:
It worked after disabling Colors tags (HTML/ASSA) in subtitle box
Now my only issue is setting actor name

niksedk:
Solved above by turning off "Color tags (HTML/ASSA) in subtitle box"; the RTL edit box has also been improved since (#11715, #13197, #14266, and the native text box in #12556). Please open a new issue with a screenshot if RTL still misbehaves on the latest version.

PR #13287, opened 2026-08-05
Fix #11663: Persian (Farsi) texts are rendered in reverse order after translation

### LINKED ISSUE #11663: Persian (Farsi) texts are rendered in reverse order after translation
Image

We have identified a minor issue within the AI translation feature where Persian (Farsi) texts are rendered in reverse order after translation. This effectively makes it impossible to assess the translation quality in real time. Consequently, for large files, we are forced to wait several hours for the entire process to complete before we can review the output—a workflow that I find far from ideal. If possible, we kindly request that this issue be addressed prior to the final release of version 5. Thank you for your consideration.

Ironship:
## Problem

Fixes #11663 — Persian (Farsi) texts are rendered in reverse order in the Auto-translate window's live table after translation, making it impossible to assess translation quality in real time.

> We have identified a minor issue within the AI translation feature where Persian (Farsi) texts are rendered in reverse order after translation. This effectively makes it impossible to assess the translation quality in real time. Consequently, for large files, we are forced to wait several hours for the entire process to complete before we can review the output—a workflow that I find far from ideal. If possible, we kindly request that this issue be addressed prior to the final release of version 5. Thank you for your consideration.

### Root cause

The stored translation is correct — `MainViewModel.ShowAutoTranslate` copies `TranslateRow.TranslatedText` verbatim into the subtitle (src/ui/Features/Main/MainViewModel.cs:8630), and no code path in the translate pipeline reverses the string. The bug is display-only: the Auto-translate window's table (`BuildRowGridCard` in src/ui/Features/Translate/AutoTranslateWindow.cs:428-443) defines its "Text" and "Translation" columns with a plain `Binding`. A plain binding leaves the cell with the window's left-to-right flow direction; Avalonia then lays a right-to-left line out under an LTR base direction, cuts it at zero-width non-joiners (U+200C, bidi class BN — common in Persian), and places the resulting runs left to right, so the word order renders reversed. This is the same failure mode fixed for other TableView windows in #13160 via `TableViewExtras.MakeTextCellTemplate` (src/ui/Logic/TableViewExtras.cs:334-356), which binds the cell's `FlowDirection` to its own text content through `TextToFlowDirectionConverter`. The Auto-translate window was the last TableView window of the #13160-converted set still using a plain binding for subtitle text columns.

## Fix

Use the established `TableViewExtras.MakeTextCellTemplate` for the Auto-translate window's "Text" and "Translation" columns (src/ui/Features/Translate/AutoTranslateWindow.cs), exactly as the sibling Copy-paste translate window (src/ui/Features/Translate/CopyPasteTranslateWindow.cs:141-156) and 9 other TableView windows already do. The cell template binds `TextBlock.FlowDirection` to the cell's content, so Persian/Arabic text flows right-to-left and renders in correct word order.

## Verification

- [x] New regression test(s): none — UI-only change; per repo policy headless UI tests are reserved for keyboard-dispatch regressions
- Manual verification path: open Auto-translate (Tools > Auto-translate), pick any engine, translate an English subtitle to Persian (fa); the Translation column now shows the Persian text in correct right-to-left word order while translation runs, instead of reversed. Compare against the Copy-paste translate window which already renders correctly.
- `git diff --stat`: 1 file, 2 insertions, 2 deletions (no reformatting, no other files touched)

## Notes

- Interpretation: issue reports the real-time preview in the Auto-translate window; the final applied subtitle is verified correct by the reporter after the run completes, consistent with a display-only cause.
- The main window's edit box RTL issues (#12167, #10212) are separate — they concern the main subtitle text box/RTL mode and are NOT changed here.
- Deliberate non-change: other windows' plain-binding columns for non-text data (numbers, times) intentionally keep `Binding`.

## Notes

- Follow-up opportunity (out of scope here): an audit found ~11 other TableView windows that still use plain bindings for subtitle text and share this RTL display bug — a future pass could convert them to `MakeTextCellTemplate` the same way.

### DIFF
--- a/src/ui/Features/Translate/AutoTranslateWindow.cs
+++ b/src/ui/Features/Translate/AutoTranslateWindow.cs
@@ -428,7 +428,7 @@ public class AutoTranslateWindow : Window
         tableView.Columns.Add(new SeTableViewColumn
         {
             Header = Se.Language.General.Text,
-            Binding = new Binding(nameof(TranslateRow.Text)),
+            CellTemplate = TableViewExtras.MakeTextCellTemplate(nameof(TranslateRow.Text)),
             Width = new GridLength(1, GridUnitType.Star),
             CellTheme = UiUtil.TableViewCellTheme,
             HeaderTheme = UiUtil.TableViewColumnHeaderTheme,
@@ -436,7 +436,7 @@ public class AutoTranslateWindow : Window
         tableView.Columns.Add(new SeTableViewColumn
         {
             Header = Se.Language.General.Translation,
-            Binding = new Binding(nameof(TranslateRow.TranslatedText)),
+            CellTemplate = TableViewExtras.MakeTextCellTemplate(nameof(TranslateRow.TranslatedText)),
             Width = new GridLength(1, GridUnitType.Star),
             CellTheme = UiUtil.TableViewCellTheme,
             HeaderTheme = UiUtil.TableViewColumnHeaderTheme,


Ironship:
**AUDIT #13287 (issue #11663): PASS-WITH-WARNINGS — mergeable**

- A. Fixes issue: PASS — TextToFlowDirectionConverter detection re-verified with a real probe against the built libse: Persian → RTL, Persian+ZWNJ (U+200C) → RTL, Arabic → RTL, Mixed EN+FA → RTL, English → LTR (5/5); the Auto-translate window's Text/Translation columns now use MakeTextCellTemplate (same as the Copy-paste translate window and the other #13160-converted windows).
- B. Surgical: PASS (+2/-2, 1 file).
- C. Style: PASS.
- D. Tests: PASS per policy — UI-only, 0 new tests, manual path in PR body.
- E. Build: PASS — real exit 0.
- F. Regression: PASS — English/LTR unchanged; converter null/empty fallback verified by source; one-way bindings never reach ConvertBack.
- WARN (addressed): the body previously claimed Auto-translate was 'the remaining' plain-binding TableView window — an audit found ~11 other TableView windows that still share this RTL display bug. Body updated with an accurate claim + follow-up note.
- VERDICT: mergeable.

niksedk (approved):
## Approve

Byte-for-byte the pattern the sibling `CopyPasteTranslateWindow` already uses (`CellTemplate` + `Width` + `CellTheme` + `HeaderTheme`), applied to exactly the two subtitle-text columns and correctly leaving Number/Show/Duration on plain `Binding`. Established across 10 windows by #13160. Nothing to add.

PR #13288, opened 2026-08-06
Fix #13278: avoid repeated paths in managed 7z extraction

### LINKED ISSUE #13278: Purfview Faster Whisper XXL installation and execution issues
Using Ubuntu 26.04 and SE 5.1.0 (and 5.2.0-beta4), a series of problems occur when trying to install this speech-to-text engine.

[1] The binary faster-whisper-xxl is installed in an unexpected location:

Image

It is in fact installed in a further subdirectory name "Faster-Whisper-XXL"; moving the binary up one directory level allowed the binary to be located.

[2] Execute permissions were not applied to the binary; doing so allowed the process to proceed further.

[3] Numerous dependencies were not properly located; the specific error message regarded Python.  Similar to 1), this (and presumably other) libraries are expected by the engine to reside directly under the `_xxl_data` directory, but these are instead appearing under _`xxl_data/Faster-Whisper-XXL/_xxl_data`.  Moving these dependencies up the two directory levels allowed the process to advance further.

[4] At this stage, a Python initialization error occurs:
ModuleNotFoundError: No module named 'numpy.core._multiarray_umath'

This one may be outside of SE's purview and instead an issue with the Purfview package instead, but thought I would include it for completeness sake.

Thank you.

Ironship:
## Problem

Fixes #13278 — on Linux, Purfview Faster Whisper XXL can be extracted into repeatedly nested paths, so the executable is not at `<install dir>/faster-whisper-xxl`.

Issue quote (verbatim):

> Using Ubuntu 26.04 and SE 5.1.0 (and 5.2.0-beta4), a series of problems occur when trying to install this speech-to-text engine.
> [1] The binary faster-whisper-xxl is installed in an unexpected location: 
> It is in fact installed in a further subdirectory name "Faster-Whisper-XXL"; moving it up one directory level fixed the execution issue.

**Root cause:** when the external 7-Zip paths are unavailable/fail, `Extract7ZipSlow` uses SharpCompress. It first computes the correct destination after stripping `Faster-Whisper-XXL/`, but then calls `reader.WriteEntryToDirectory(fullPath)`. That API appends the entry's original full archive key again. For an entry such as `Faster-Whisper-XXL/_xxl_data/numpy/core/file`, the resulting path becomes:

`<install>/_xxl_data/numpy/core/Faster-Whisper-XXL/_xxl_data/numpy/core/file`

This is the repeated path visible in the issue screenshot.

The current real Linux archive was downloaded and inspected byte-for-byte. Its wrapper is exactly `Faster-Whisper-XXL` (same case and Unicode code points), plus a root `license.txt`; therefore the previous theory about a differently named wrapper was incorrect and that fallback has been removed.

## Fix

- Normalize the archive entry and strip only an exact wrapper directory boundary.
- Canonicalize the extraction root and each destination, reject rooted/absolute entries and any normalized `..` path that escapes the root, then call `WriteEntryToFile(fullFileName)`. This restores the path-containment property that `WriteEntryToDirectory` previously provided while avoiding its duplicate-key behavior.
- Preserve the old fallback when the requested wrapper genuinely does not exist; no speculative single-directory unwrap remains.

## Verification

- [x] `dotnet build src/ui/UI.csproj --no-incremental -v q -nologo` — EXIT:0.
- [x] Permanent `UnpackerTests.Extract7ZipSlow_StripsWrapperWithoutRepeatingOriginalEntryPath` uses a real embedded 7z fixture with the same wrapper, nested `_xxl_data/numpy/core` path and loose root `license.txt`.
- [x] RED/GREEN proven: the layout test fails 0/1 on the old code and passes on the fix. A second real `7z -spf` fixture proves an absolute archive entry is rejected instead of being written outside the target; `UnpackerTests` 2/2 PASS.
- [x] Full real asset `Faster-Whisper-XXL_r245.4_linux.7z` (1,657,634,355 bytes compressed; 4,393 files and 634 folders) was extracted through `Extract7ZipSlow`: PASS in 4m19s. `faster-whisper-xxl`, `license.txt` and `_xxl_data` were at the install root, with zero nested `Faster-Whisper-XXL` directories.

## Notes

- The issue's later execution-environment points are outside this extraction-layout fix and are not claimed as resolved here.

### DIFF
--- a/src/ui/Logic/SevenZipExtractor/Unpacker.cs
+++ b/src/ui/Logic/SevenZipExtractor/Unpacker.cs
@@ -320,7 +320,6 @@ public static class Unpacker
 
         if (!Directory.Exists(skipPath))
         {
-            // If the skip path doesn't exist, just move everything
             skipPath = sourceDir;
         }
 
@@ -373,7 +372,7 @@ public static class Unpacker
         }
     }
 
-    public static void Extract7ZipSlow(string tempFileName, string dir, string skipFolderLevel, CancellationTokenSource cancellationTokenSource, Action<string> updateProgressText)
+    public static void Extract7ZipSlow(string tempFileName, string dir, string? skipFolderLevel, CancellationTokenSource cancellationTokenSource, Action<string> updateProgressText)
     {
         using Stream stream = File.OpenRead(tempFileName);
         using var archive = SevenZipArchive.OpenArchive(stream);
@@ -381,6 +380,7 @@ public static class Unpacker
         double unpackedSize = 0;
 
         var reader = archive.ExtractAllEntries();
+        var targetRoot = Path.GetFullPath(dir);
         while (reader.MoveToNextEntry())
         {
             if (cancellationTokenSource.IsCancellationRequested)
@@ -390,16 +390,46 @@ public static class Unpacker
 
             if (!string.IsNullOrEmpty(reader.Entry.Key))
             {
-                var entryFullName = reader.Entry.Key;
-                if (!string.IsNullOrEmpty(skipFolderLevel) && entryFullName.StartsWith(skipFolderLevel))
+                var entryFullName = reader.Entry.Key.Replace('\\', '/');
+                var normalizedSkipFolder = skipFolderLevel?.Replace('\\', '/').Trim('/') ?? string.Empty;
+                if (!string.IsNullOrEmpty(normalizedSkipFolder))
                 {
-                    entryFullName = entryFullName[skipFolderLevel.Length..];
+                    if (entryFullName.Equals(normalizedSkipFolder, StringComparison.Ordinal))
+                    {
+                        entryFullName = string.Empty;
+                    }
+                    else if (entryFullName.StartsWith(normalizedSkipFolder + "/", StringComparison.Ordinal))
+                    {
+                        entryFullName = entryFullName[(normalizedSkipFolder.Length + 1)..];
+                    }
                 }
 
                 entryFullName = entryFullName.Replace('/', Path.DirectorySeparatorChar);
-                entryFullName = entryFullName.TrimStart(Path.DirectorySeparatorChar);
+                if (Path.IsPathRooted(entryFullName))
+                {
+                    throw new InvalidDataException($"Archive entry is rooted outside the extraction folder: {reader.Entry.Key}");
+                }
 
-                var fullFileName = Path.Combine(dir, entryFullName);
+                entryFullName = entryFullName.TrimStart(Path.DirectorySeparatorChar);
+                if (string.IsNullOrEmpty(entryFullName))
+                {
+                    if (reader.Entry.IsDirectory)
+                    {
+                        Directory.CreateDirectory(dir);
+                        continue;
+                    }
+
+                    throw new InvalidDataException("Archive contains an empty file entry name.");
+                }
+
+                var fullFileName = Path.GetFullPath(Path.Combine(targetRoot, entryFullName));
+                var relativePath = Path.GetRelativePath(targetRoot, fullFileName);
+                if (relativePath.Equals("..", StringComparison.Ordinal) ||
+                    relativePath.StartsWith(".." + Path.DirectorySeparatorChar, StringComparison.Ordinal) ||
+                    Path.IsPathRooted(relativePath))
+                {
+                    throw new InvalidDataException($"Archive entry is outside the extraction folder: {reader.Entry.Key}");
+                }
 
                 if (reader.Entry.IsDirectory)
                 {
@@ -417,6 +447,8 @@ public static class Unpacker
                     continue;
                 }
 
+                Directory.CreateDirectory(fullPath);
+
                 var displayName = entryFullName;
                 if (displayName.Length > 30)
                 {
@@ -427,7 +459,10 @@ public static class Unpacker
                     updateProgressText(string.Format(Se.Language.General.UnpackingX, displayName));
                 });
 
-                reader.WriteEntryToDirectory(fullPath);
+                // WriteEntryToDirectory appends the entry's original full key again. Since
+                // fullPath already contains the stripped relative path, that produced the
+                // repeated .../numpy/core/Faster-Whisper-XXL/... nesting reported in #13278.
+                reader.WriteEntryToFile(fullFileName);
                 unpackedSize += reader.Entry.Size;
             }
         }
--- a/tests/UI/Logic/SevenZipExtractor/UnpackerTests.cs
+++ b/tests/UI/Logic/SevenZipExtractor/UnpackerTests.cs
@@ -1,0 +1,61 @@
+using Nikse.SubtitleEdit.Logic.SevenZipExtractor;
+
+namespace UITests.Logic.SevenZipExtractor;
+
+public class UnpackerTests : IDisposable
+{
+    private const string ArchiveBase64 = "N3q8ryccAASRme6J7wAAAAAAAAAWAAAAAAAAANtGIGABABZlbmdpbmVleGVjdXRhYmxlbGljZW5zZQDgAlUAzF0AAIEzB64Pz7XvEA/Ual595dfdm+avUngQmcP0+WCm1/B0sii/q/CHmny4m6aKmeaETC1j6oYCbmCvutV7/FCFvtanCgHlX/zjPbLzpoLXTRu5YlZGWdBR+YNjTU+jaIr+BgzCBmucvUyzM93+Zf2XK7H3EeosnQ88EhVRM096o8AbJeFkMSXMddJen437oUr4YA7ZtjnpAZxnmMuFuDTcR0hzuZ2JwNeddfoONNW9EvkwlskRI/+0AJxZRToJCGFpXlUTmVVoSecnAAAAABcGGwEJgNQABwsBAAEhIQEYDIJWAAA=";
+    private const string AbsolutePathArchiveBase64 = "N3q8ryccAASLHia9CwAAAAAAAACCAAAAAAAAACgymrgBAAZvdXRzaWRlAAEEBgABCQsABwsBAAEhIQEADAcACAoBpv3qtQAABQEZDAAAAAAAAAAAAAAAABE7AC8AdABtAHAALwBzAGUAMQAzADIAOAA4AGUAdgBpAGwALwBvAHUAdABzAGkAZABlAC4AdAB4AHQAAAAZABQKAQB8salAgSXdARUGAQAggKSBAAA=";
+
+    private readonly string _tempFolder;
+
+    public UnpackerTests()
+    {
+        _tempFolder = Path.Combine(Path.GetTempPath(), "SeUnpackerTest_" + Guid.NewGuid().ToString("N"));
+        Directory.CreateDirectory(_tempFolder);
+    }
+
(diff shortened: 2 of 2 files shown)


niksedk (changes requested):
## Request changes

### Top-level files are silently deleted

```csharp
var topLevelDirs = Directory.GetDirectories(sourceDir);
if (topLevelDirs.Length == 1)
{
    skipPath = topLevelDirs[0];
}
```

`GetDirectories` counts only directories. An archive laid out as

```
Wrapper/…            <- one top-level directory
LICENSE.txt          <- top-level file
run.sh               <- top-level file
```

takes the new branch, moves only `Wrapper/`'s contents, and leaves the two files in the temp folder — which `Extract7Zip`'s `finally` then removes with `Directory.Delete(extractPath, true)`. They are gone, with no error and no log line. The old fallback moved them.

`Directory.GetFileSystemEntries(sourceDir)` is the check you want: unwrap only when the extraction produced exactly one entry and that entry is a directory. This matters because the fallback now fires for every engine using `skipFolderLevel` (PaddleOCR, Chrome-Lens-OCR, VLC, …), not just Faster-Whisper-XXL.

### The root cause is not established

The verification says the current archive contains a single top-level `Faster-Whisper-XXL/` matching the hard-coded skip name, and `DownloadSpeechToTextEngineViewModel.cs:173` passes exactly `"Faster-Whisper-XXL"`. If both are true, `Directory.Exists(skipPath)` is **true** on the reporter's machine and the fallback this PR changes is never reached — yet the reporter's layout (`<dir>/Faster-Whisper-XXL/faster-whisper-xxl`) is precisely the old fallback's output.

Something makes that `Directory.Exists` fail on Ubuntu and the PR does not say what. The obvious candidate is case sensitivity — `Directory.Exists` is case-sensitive on Linux and case-insensitive on Windows, which would explain a Linux-only report exactly. If that is the cause, the correct fix is a case-insensitive match on the wrapper name, and the single-directory heuristic is treating a symptom.

Please pin this down before merging. As written the PR may not fix #13278 at all, and its own evidence argues that it doesn't.

### No test

`MoveFilesSkippingFolderLevel` is static filesystem logic — build a temp tree, call it, assert the layout. Three cases (matching wrapper / differently-named single wrapper / wrapper plus sibling files) would take a few minutes and would have caught the deletion above. The manual Linux path is checked off as not run, so right now nothing verifies this change.

Ironship:
Fixed in cc28a8c2. The review was correct: the previous wrapper-name diagnosis contradicted the real archive. I removed that fallback and traced the actual root cause to the managed SharpCompress path: after computing the stripped destination, WriteEntryToDirectory appended the original full entry key again, reproducing the issue's repeated numpy/core/Faster-Whisper-XXL/... path. The fix uses WriteEntryToFile at the computed path and deletes emptied recursive source directories after successful moves/cancellation checks. Added a permanent real-7z regression test: RED 0/1 on old code, GREEN 1/1. I also ran the complete 1.65 GB Linux asset through Extract7ZipSlow: PASS in 4m19s; executable, license and _xxl_data at root; no nested wrapper. UI build EXIT:0.

Ironship:
FINAL AUDIT #13288 for issue #13278: PASS\n\nFresh head: cc28a8c; one Ironship commit; mergeable; CI success. Root cause is now established in the managed SharpCompress fallback: WriteEntryToDirectory appended reader.Entry.Key below an already-computed destination path, producing the exact repeated numpy/core/Faster-Whisper-XXL/... layout from the report. The incorrect wrapper-name heuristic was removed.\n\nRegression coverage: the committed real-7z fixture test is RED on the old code and GREEN on the fix; it checks the executable, _xxl_data, root license, absence of a nested wrapper, and exact file count. The complete current 1,657,634,355-byte Linux asset also passed through Extract7ZipSlow in 4m19s with the executable/data/license at root. UI build EXIT:0.\n\nMove cleanup deletes a source directory only after its recursive move returns, after cancellation is checked, and never after an exception. No dead fallback, lost top-level files, or unrelated extractor behavior remains in the diff.

niksedk (changes requested):
## Request changes — root cause confirmed, but the rewrite drops path containment

I verified this one by running it rather than reading it. The diagnosis is **correct** and a clear improvement over the previous revision; there is one regression I'd want fixed before merge.

### Verified: the root cause is real

`ExtractionOptions.ExtractFullPath` defaults to **`true`** in SharpCompress 0.50.2 (decompiled from the pinned package), so `WriteEntryToDirectory(fullPath)` re-appends the entry's whole key under a path that already contains the stripped relative path. Exactly as the PR says.

I built a realistic archive with the real `7z` binary (wrapper folder, nested `_xxl_data/numpy/core`, several files) and ran the old and new `Extract7ZipSlow` logic against it through the same SharpCompress API:

```
===== OLD (main) =====
  _models/Faster-Whisper-XXL/_models/model.bin
  _xxl_data/numpy/core/Faster-Whisper-XXL/_xxl_data/numpy/core/engine.bin
  Faster-Whisper-XXL/faster-whisper-xxl        <-- the reporter's exact symptom
  Faster-Whisper-XXL/license.txt
  Faster-Whisper-XXL/whisper-faster

===== NEW (this PR) =====
  _models/model.bin
  _xxl_data/numpy/core/engine.bin
  faster-whisper-xxl                            <-- where GetExecutable() looks
  license.txt
  whisper-faster
```

That reproduces #13278 precisely and shows the fix resolves it. It also retires my earlier question about `Directory.Exists` — `Extract7ZipSlow` never calls `MoveFilesSkippingFolderLevel`, so the managed fallback (used when no system 7z is installed, which fits an Ubuntu reporter) was always the wrong place to look. Good find.

### Verified: the new test is genuine

I reverted only `Unpacker.cs` to main and kept the new test:

- against the fix: **1 passed**
- against unpatched `Unpacker.cs`: **1 failed**

It fails for the right reason and is not vacuous. Thank you for the base64 fixture — that is the right way to do this.

### Blocking: `WriteEntryToFile` has no containment check, and `TrimStart` was removed

Two changes compound. `WriteEntryToDirectory` routes through `IEntryExtensions.WriteEntryToDirectoryCore`, which calls:

```csharp
DirectoryManagement.EnsurePathInDestinationDirectory(entryDestinationFileName, fullDestinationDirectoryPath, "Entry is trying to write a file outside of the destination directory.");
```

`WriteEntryToFile` has **no equivalent** — it only handles `LinkTarget` and `Overwrite`. At the same time this PR deletes `entryFullName.TrimStart(Path.DirectorySeparatorChar)`, so an absolute entry key survives into `Path.Combine(dir, entryFullName)`, where the absolute operand wins and `dir` is discarded.

Demonstrated with an archive whose entry key is absolute (produced by plain `7z a -spf`, a documented option, no crafting tricks):

```
===== OLD (main) =====  ExtractionException: Directory does not exist to extract to: <target>/private/tmp/.../evilsrc/
===== NEW (this PR) =====  ok
>>> file written to its ORIGINAL absolute path, outside the extraction target
```

Old code refused, loudly. New code reports success and writes outside the destination, silently. This is not only a security question — a publisher who archives with `-spf` would have engine files scattered across the filesystem instead of a clean failure.

Suggested fix, restoring both properties:

```csharp
entryFullName = entryFullName.Replace('/', Path.DirectorySeparatorChar);
entryFullName = entryFullName.TrimStart(Path.DirectorySeparatorChar);   // keep this

var fullFileName = Path.GetFullPath(Path.Combine(dir, entryFullName));
var targetRoot = Path.GetFullPath(dir) + Path.DirectorySeparatorChar;
if (!fullFileName.StartsWith(targetRoot, StringComparison.Ordinal))
{
    continue;   // or throw - but do not write outside dir
}
```

A `..`-containing key deserves the same treatment; `TrimStart` alone does not cover it, and neither old nor new code does today.

### Minor

1. **`skipFolderLevel.Replace(...)` before the null check.** `var normalizedSkipFolder = skipFolderLevel.Replace('\\', '/').Trim('/');` runs unconditionally, then `IsNullOrEmpty` is tested. Null throws. All nine call sites pass literals so it is unreachable today, but `Extract7ZipSlow` is `public static` and the old code guarded with `!string.IsNullOrEmpty(skipFolderLevel)` first.

2. **`entryFullName == string.Empty` edge.** When the key exactly equals the skip folder and the entry is *not* flagged as a directory, `fullFileName` becomes `dir` itself and `WriteEntryToFile` would try to open a directory as a file. Normally the wrapper is a directory entry and hits the `continue`, so this needs a slightly odd archive — worth a guard rather than an exception.

3. **`Directory.Delete(subDir)` in `MoveDirectoryContents` is unexplained.** It is in a different method from the bug, on the managed *move* path rather than the extract path, with no comment and not covered by the new test. It looks safe (files are moved with overwrite, the cancellation check precedes it, so the directory should be empty) but it is an unrelated behaviour change riding along in a bug fix. Either give it a comment saying what it fixes, or split it out.

Fix the containment issue and I'm happy to approve — the core of this is right and well evidenced.

Ironship:
Fixed the containment regression in b16d8f2e. Thanks for running the SharpCompress comparison and the -spf case.\n\nChanges:\n- reject rooted/absolute entry names before trimming;\n- canonicalize target root + destination and reject any relative result that is .., starts with ../, or remains rooted;\n- restore TrimStart for safe relative names;\n- guard the empty non-directory entry case;\n- accept nullable skipFolderLevel safely;\n- remove the unrelated recursive Directory.Delete change from the diff.\n\nAdded a second embedded fixture produced by real 7z a -spf with an absolute /tmp/... entry. It must throw InvalidDataException and cannot write outside the target. Current checks: UnpackerTests 2/2 PASS; canonical UI build EXIT:0; one Ironship commit. The PR body is updated and CI is rerunning on b16d8f2e.

niksedk (approved):
## Approve

The containment issue from my last review is fixed in `b16d8f2e9`, and your version is better than the one I had staged — `Path.GetRelativePath` beats the string-prefix check I was going to push (mine would have been fragile on case-insensitive filesystems), throwing beats my skip-and-log for a guard like this, and removing `Directory.Delete(subDir)` is the right call over the comment I'd written for it. I dropped my patch rather than pushing it.

Re-verified both behaviours against real archives built with the `7z` binary, driving the same SharpCompress API this code path uses:

- **Escape fixture** (absolute entry key, `7z a -spf`): `InvalidDataException: rooted outside: /tmp/sezt/outside.txt` — nothing written outside the target. The pre-fix revision wrote the file to its original absolute path and still reported success.
- **Realistic Faster-Whisper-XXL fixture** (wrapper folder, nested `_xxl_data/numpy/core`): flat correct layout, `faster-whisper-xxl` at the root where `GetExecutable()` looks — identical to the previous revision, so the #13278 fix is intact.

Also confirmed earlier that the first test is genuine: reverting only `Unpacker.cs` to main makes it fail, and it passes with the fix.

### One note, not blocking

In `Extract7ZipSlow_RejectsAbsoluteEntryPath`:

```csharp
Assert.Empty(Directory.GetFiles(_tempFolder, "outside.txt", SearchOption.AllDirectories));
```

The escape target is `/tmp/se13288evil/outside.txt`, which is not under `_tempFolder` — so this line cannot fail whether or not the entry escaped. `Assert.Throws` is what actually guarantees containment here (nothing is written before the throw). Worth a comment saying so, otherwise someone later softening the throw to a skip will see this assertion still passing and assume they are covered.

Merging.

Issue #13230, opened 2026-08-05
5,2 release ~ 5.2 beta 4  -- When you open a subtitle the 1st entry doesn't seem to be selected.


fraternl:
When you open a subtitle the 1st entry doesn't seem to be selected.

It works with any subititle.
You have to open the application and select the 1st row.
Then you have to close the application with saving anything and then re-open the application.

I had to try a lot of versions to see where it happened.
Easiest is to wait for the 1st row to disappear.

5.1 is the last version that doesn't do it.
5.2 is the first version with the bug.

You see it displeys the first row 
<img width="831" height="722" alt="Image" src="https://github.com/user-attachments/assets/e5a577bc-0b6d-4600-a3e0-23ef7eca678e" />

You don't do anything, just wait 5 seconds and the text disappears.

<img width="735" height="797" alt="Image" src="https://github.com/user-attachments/assets/fdf52641-8c6b-4014-a20d-c35368ea654a" />

You think you made a selection with row 1, but it's omitted. 

<img width="790" height="312" alt="Image" src="https://github.com/user-attachments/assets/a638fac5-604c-431b-a317-702729e671f9" />

You apply italic

<img width="791" height="307" alt="Image" src="https://github.com/user-attachments/assets/89e1236d-cb4f-49d5-8a8c-35eaf545bdd8" />

<img width="742" height="580" alt="Image" src="https://github.com/user-attachments/assets/cb9f32d2-6cd5-4548-a900-1e1862494e88" />

GrampaWildWilly:
#13190

PR #13306, opened 2026-08-06
Fix #13303: delete dead on first row of the first file opened after startup

### LINKED ISSUE #13303: Delete does not work on the first subtitle row in the initial file after startup
Delete does not work initially on the first subtitle row in the first subtitle file opened after starting Subtitle Edit.

The issue affects both the Delete key and Delete from the context menu.

Steps to reproduce:

1. Start Subtitle Edit.
2. Open a subtitle file.
3. Select the first subtitle row.
4. Press Delete or choose Delete from the context menu.
5. Nothing happens and no confirmation dialog appears.
6. Select the second subtitle row.
7. Press Delete or choose Delete from the context menu.
8. The confirmation dialog appears.
9. Cancel the deletion.
10. Select the first subtitle row again.
11. Delete now works correctly.

Additional observation:

If another subtitle file is opened during the same Subtitle Edit session, Delete works immediately on its first subtitle row.

The problem appears only with the initial subtitle file after a fresh application launch.

Actual behavior:

Delete does not react on the first subtitle row in the initial file until the command has been used on another row.

Expected behavior:

Delete should work immediately on the first selected subtitle row.

The issue is reproducible after several fresh application launches.

W11
SE 5.2.0-beta5 (ZIP)

niksedk:
Fixes #13303.

### Symptom
After a fresh application launch, Delete (key or context menu) silently does nothing on the first row of the first subtitle file opened - no confirmation dialog. Moving the selection to another row once fixes it for the rest of the session, and a second file opened in the same session is unaffected.

### Root cause
This is a residual case of #13230. On the first file open:

1. `SetSubtitles` reattaches ItemsSource, and the grid picks row 0 on its own (`SelectionMode.AlwaysSelected`) **without raising SelectionChanged**.
2. The TwoWay `SelectedItem` binding writes `SelectedSubtitle` directly, bypassing `SubtitleGridSelectionChanged`.
3. The repair in `SelectAndScrollToRow` (added for #13190) used `!ReferenceEquals(SelectedSubtitle, itemToScroll)` as proof that the selection pipeline had run - so with `SelectedSubtitle` already correct via the binding, it skipped the repair.
4. `_selectedSubtitles` - the cache every selection command reads (delete, copy, italic, ...) - stayed `null`, so those commands hit their `selectedItems.Count == 0` early return.

The edit box and time codes looked fine the whole time (they follow `SelectedSubtitle`), which is what made only the commands appear dead.

### Fix
In `SelectAndScrollToRow`, run the repair whenever the `_selectedSubtitles` cache does not match the single row being selected, not only when `SelectedSubtitle` diverges. Before re-reading the selection, sync the grid's `SelectedItems` from its selection model (`SyncSelectedItemsWithSelection`, idempotent) so the re-read can never blank the selection the way #13230 did.

### Test
New headless test drives the real user scenario - fresh main window, `SubtitleOpen` on a temp .srt, then the delete command - and failed before the fix (3 rows remained) / passes after. Full UI suite: 1486/1486 green.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -15973,9 +15973,17 @@ public partial class MainViewModel :
                 // shows the first row highlighted while SelectedSubtitle is still null -
                 // empty edit box and go to next/previous line dead until the selection is
                 // moved away and back (#13190). Assigning the same item above raises no
-                // event either, so sync the view model explicitly.
-                if (!ReferenceEquals(SelectedSubtitle, itemToScroll))
+                // event either, so sync the view model explicitly. SelectedSubtitle alone
+                // is no proof the pipeline ran: the TwoWay SelectedItem binding writes it
+                // directly when the grid picks row 0 on its own, while _selectedSubtitles -
+                // what delete/copy/italic read - stays empty, leaving those commands dead
+                // on the first row of the first file opened after startup (#13303).
+                var selectionCacheStale = _selectedSubtitles == null
+                                          || _selectedSubtitles.Count != 1
+                                          || !ReferenceEquals(_selectedSubtitles[0], itemToScroll);
+                if (!ReferenceEquals(SelectedSubtitle, itemToScroll) || selectionCacheStale)
                 {
+                    TableViewExtras.SyncSelectedItemsWithSelection(SubtitleGrid);
                     SubtitleGridSelectionChanged();
                 }
 
--- a/tests/UI/Features/Main/SubtitleOpenFirstRowDeleteTests.cs
+++ b/tests/UI/Features/Main/SubtitleOpenFirstRowDeleteTests.cs
@@ -1,0 +1,115 @@
+using Avalonia.Controls;
+using Avalonia.Headless.XUnit;
+using Avalonia.Threading;
+using Microsoft.Extensions.DependencyInjection;
+using Nikse.SubtitleEdit;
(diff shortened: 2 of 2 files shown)


GrampaWildWilly:
I'm still baffled by this.  I'm not seeing the problem.  But then, my test case is a VTT.  The reporter's test case is an ASSA subtitle.  Is that the key?

In any case, while trying again to replicate the reporter's results (and still failing), I did notice some other things.  First, I hit the Delete key & got this:



Bringing something to foreground is easy enough with Alt+Tab.  I did Alt+Ctrl+Tab so I could capture this screenshot:




The Audio visualizer going into hiding may not be a bug or even an issue.  I just noticed it.  It's probably not a problem.  It does properly come back into view after dispensing with the delete confirmation window.  The problem is the delete confirmation dialog doing a fake job on us.  It isn't really in foreground from the outset & it should be.

Now, going back to the original state, instead of typing the Delete key, let's pop up the context menu on that famous first line.



This again.  I feel like we're playing whack-a-mole.  How many things are going to get obscured in undocked mode?  I don't know how to hunt for all the places in the product where this sort of thing can happen.  Can you do it?  I'm sure it's quite annoying to you to have me keep reporting this problem.  No doubt you'd really like to shut me up on this subject.  I would certainly like to stop reporting it.

GrampaWildWilly:
Come to think of it, a window looking like it's in foreground but isn't really is another whack-a-mole game we've been playing.  Aren't there central service routines that everybody calls that could deal with these things once & for all?

PR #13331, opened 2026-08-07
Fix: stop play selection 0.05s early to prevent auto-selecting next subtitle

ghostminhtoan:
fixed an issue where using commands like **Play next (and stop)**, **Play previous (and stop)**, or **Play selected lines** (F5) would incorrectly highlight/select the next subtitle line after stopping.

Cause:
When the playback of a selected subtitle line ends, the video player pauses and is positioned exactly at `_playSelectionItem.EndSeconds` (which is often also the `StartTime` of the subsequent subtitle). Because of this, the `SelectCurrentSubtitleAtPlayhead` logic gets triggered upon pausing and auto-selects the next subtitle line instead of staying on the current one that just finished playing.

Solution:
Subtract `0.05` seconds from the final pause position `vp.Position` when stopping the playback of the selected subtitle block. This keeps the playhead strictly within the duration bounds of the current line and prevents the player from auto-selecting/shifting focus to the next subtitle.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -23304,8 +23304,8 @@ public partial class MainViewModel :
                         if (p == null)
                         {
                             PauseVideoAndFreezePlayhead(vp);
-                            vp.Position = _playSelectionItem.EndSeconds;
-                            PinPlayheadTo(_playSelectionItem.EndSeconds);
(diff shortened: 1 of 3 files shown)


niksedk:
Thx for the input - hopefully fixed via https://github.com/SubtitleEdit/subtitleedit/pull/13334

PR #13335, opened 2026-08-07
Fix #13332: TTS generation folder - honor outputFolder, configurable, cleanup after merge

### LINKED ISSUE #13332: [v5.2.0-beta6 ] TTS generation paths and CrispASR versions
Hi,
At the moment I do the generation through MOSS-TTS, but I get the files in the folder containing VoxCPM2CrispAsr. There are clearly confused paths. Also, I would ask to give the ability to assign a folder for generation manually, so as not to get hundreds of files in the folder containing the engine's files.

 Another chaos is that the folder of each engine contains a copy of CrispASR, and the versions of TextToSpeech are different from the version that is downloaded as a backend in the CrispASR folder. Is there a way to optimize this process so as not to multiply entities?
Thank you

Ironship:
Fixes points 1+2 of #13332 (confused output paths, no way to choose the generation folder, hundreds of files piling up in the engine folders). Point 3 of the issue (per-engine CrispASR copies) is not reproducible in the current code - the CrispASR runtime has been a single shared install since these engines were added; see my comment on the issue.

## Problem
Every TTS engine's `Speak(text, outputFolder, ...)` **ignored** the `outputFolder` parameter and wrote GUID-named audio files into its own data folder (`.../TextToSpeech/<Engine>/`). The folders were never cleaned, so they filled with hundreds of files, and generated audio landed in places the user didn't expect (MOSS-TTS files in the folder tree next to `VoxCPM2CrispAsr`, etc.).

## Changes
1. **Engines honor `outputFolder`** — all 21 local engines (CrispASR family, Chatterbox, Piper, EdgeTts, ElevenLabs, Azure, Google, Mistral, Murf, AllTalk, Kokoro, OmniVoice/Qwen3 cpp) now write to the folder they are given via a shared `TtsOutputFolder.Resolve(requested, engineDefault)` helper (creates the folder best-effort). Fallback to the engine folder only for direct callers that pass an empty folder. Piper additionally gets the full resolved path quoted in the `-f` argument (it used to receive only a relative name and wrote into its own folder regardless).
2. **New setting `Video.TextToSpeech.OutputFolder`** (TTS → Advanced settings, "Output folder" + Browse): empty = system temporary folder (scratch); set = per-line files are written there and kept. An unusable path falls back to the scratch folder with a log instead of blocking the TTS window.
3. **Scratch cleanup** — per-line files and audio intermediates (trim/VAD/stretch/post-process) generated this session are deleted after the merge (and on window close for sessions that never merged), so the temp folder doesn't fill with GUID wavs. Files in a user-configured folder are intentionally kept (decision snapshotted at window open).

## Behavior after this PR
- Default: per-line files go to the system temp folder (same scratch area as all other merge intermediates), the engine data folders stop accumulating, and the final audio is still saved via the existing save dialog.
- User picks a folder in advanced settings: files land there and stay.
- Imported sessions are unaffected (their files are the user's exported copies and are never touched).

## Verification
- `dotnet build src/ui/UI.csproj` — clean (0 errors, 0 warnings) on the current head.
- TTS test suite (`--filter TextToSpeech`): 111 passed, 0 failed.
- 3 parallel read-only audits (engines / ViewModel lifecycle / settings+DI+metadata): findings fixed on this head — Piper output path (HIGH), folder-creation contract, `CreateDirectory` guard, intermediate tracking, setting snapshot, cast-dialog folder consistency.
- All 4 `Speak` call sites audited; review-regenerate and cast-dialog voice test use the same session folder.
- Throwaway probe tests for `TtsOutputFolder.Resolve` + the setting default passed (3/3), removed before commit per the project's minimal-test convention.

### DIFF
--- a/src/ui/Features/Video/TextToSpeech/AdvancedTtsSettings/AdvancedTtsSettingsViewModel.cs
+++ b/src/ui/Features/Video/TextToSpeech/AdvancedTtsSettings/AdvancedTtsSettingsViewModel.cs
@@ -5,6 +5,7 @@ using CommunityToolkit.Mvvm.Input;
 using Nikse.SubtitleEdit.Features.Video.TextToSpeech.Engines;
 using Nikse.SubtitleEdit.Logic.Config;
 using Nikse.SubtitleEdit.Logic.Media;
+using System.Threading.Tasks;
 
 namespace Nikse.SubtitleEdit.Features.Video.TextToSpeech.AdvancedTtsSettings;
 
@@ -23,12 +24,16 @@ public partial class AdvancedTtsSettingsViewModel : ObservableObject
     [ObservableProperty] private string _edgeTtsVolume;
     [ObservableProperty] private bool _isEdgeTtsEngine;
     [ObservableProperty] private string _rubberbandStatus;
+    [ObservableProperty] private string _outputFolder;
+
+    private readonly IFolderHelper _folderHelper;
 
     public Window? Window { get; set; }
     public bool OkPressed { get; private set; }
 
-    public AdvancedTtsSettingsViewModel()
+    public AdvancedTtsSettingsViewModel(IFolderHelper folderHelper)
     {
+        _folderHelper = folderHelper;
         RubberbandStatus = FfmpegGenerator.IsRubberbandAvailable() ? "(installed)" : "(not found in FFmpeg)";
         var s = Se.Settings.Video.TextToSpeech;
         DoProAudioChain = s.ProAudioChainEnabled;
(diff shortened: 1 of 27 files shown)


Ironship:
Closing this as a duplicate - your #13336 covers the same fix (TtsOutputFolder.Resolve + all 21 engines honoring the caller's folder, including the Piper output-path handling) and is already merged.

One thing this branch had that #13336 does not: an optional user-configured output folder setting (TTS advanced settings, "Output folder" + Browse) that keeps generated files in a chosen location instead of the temp folder, plus cleanup of the session's scratch files after the merge. If you would like that as a follow-up, I can re-open it as a separate PR. Thanks for the quick fix!

Issue #13371, opened 2026-08-08
SE5 regressions: menu hijacks Ctrl+1, caret position persists, Shift+Enter issue


St0rmXtr00per:
Hi! Thank you for the new version! I really like it and very grateful. 

After switching to Subtitle Edit 5 from v4, I've run into a few workflow bugs regarding hotkeys and text input focus:

    1. Ctrl + 1..9 hijacked by top menu (breaks Ditto clipboard manager):
    Pressing Ctrl + 1 focuses the File tab instead of inserting clipboard items. The text field loses focus, and these shortcuts are not listed under Options -> Shortcuts to disable them.

    2. Caret position persists across subtitles:
    Moving to the next subtitle (via Enter, Down, or Alt + Down) keeps the caret at the same character index as the previous line instead of resetting to position 0. 

    3. Unwanted line break inside text via Shift + Enter:
    Pressing Shift + Enter inserts a new line inside the text box instead of doing nothing or committing the text. It often triggers by accident when typing uppercase text, and there is no way to remap or turn it off in Settings.

Please let me know if there is any way to fix it on my side or if you can fix it in the future updates. 

Thank you!

GrampaWildWilly:
Interesting.  I don't have a Ctrl+1 shortcut defined in beta 7.  The only Ctrl+digit shortcut that fires for me is Ctrl+4, which wants to import my SE 4 settings.  Maybe you migrated Ctrl+1 (and the others) from your SE 4.  In any case, I also use Ditto.  It is unfortunately true that different apps can define the same keystroke sequence for their own purposes & collisions can occur.  My take is you're out of luck with Ditto if you want to keep Ctrl+1.  I have set Ctrl+Shift+Alt+d (d for Ditto) as my keyboard shortcut.  It's a chord I can play with the fingers of my left hand.  I came up with this rather bizarre combination precisely because of a collision with a shortcut in another app.  I did have to get used to it at first but it's second nature now.  Nothing in SE collides with that . . . at the moment.  It might happen some day & then you'd have to change either the Ditto shortcut or the SE shortcut.  That's just the way it is.  I am using Ditto with SE quite successfully.

As for Shift+Enter, it does the same thing for me.  But then, Shift+Enter in Notepad++ does it as well.  I'm not sure that's a bug.

[...]

Issue #13111, opened 2026-08-03
SE 5.2.0 beta 2: Loss of keyboard navigation of the menu bar (continues #13083)


GrampaWildWilly:
Since #13083 is closed, I'm thinking it's best to just open a new issue.

Initial testing is that things are not improved.  On first open & after a couple of close/reopen cycles, Alt & F10 by themselves with no subtitle loaded just don't work.  I have to click on the menu bar with the mouse to get a menu to open.  When I do that, nothing is underlined so there's no little helpers to directly jump to menus or menu entries.  Navigation must be by arrow keys or the mouse.

By total accident I discovered that I can activate the menu bar via holding down Alt & then typing any of the 4 Arrow keys.  Then I get keyboard operation of the menus via arrows but there's still no underlines.  However, if you happen to remember what letters were underlined, are supposed to be underlined, they do still work.  For example, with the File menu open, r does cascade out the Reopen submenu.  With the Options menu open, h does open the Shortcuts dialog.  This is really weird.  After further fooling around with it, I've discovered that bare Alt is actually enough.  Follow that with an arrow key & you'll see that the menu bar is in fact activated.

With any menu open, Esc causes that menu to close & the highlight jumps back to the first entry on the menu bar (File).  But it does not deactivate the menu bar.  Alt+Tab away (to let's say Firefox) then back to SE should deactivate the menu bar, & that is supposed to be true whether a menu is open or not.  However, when you return to SE, any arrow key will confirm that the menu bar is still activated.

F10 appears to do nothing.  However, if you have the brainwave to hit F10, take your hands off the keyboard, then hit an arrow key, you will discover that F10 in fact did activate the menu bar.  You just don't get visual confirmation of this via the highlight on File that you're supposed to get.

Esc is supposed to deactivate the menu bar.  This is supposed to be true with no menu open, & with a menu open a second Esc is also supposed to deactivate the menu bar.  Not happening.

After a few closes & reopens, Alt by itself does activate the menu bar, although you don't see a highlight so you think it did nothing.  This is another one of those cases of your first execution of SE after a new release is special & odd things happen then that don't happen on later invocations.

With a menu open, opened either via Alt or F10, F10 appears to deactivate the menu bar.  Appearances are deceiving.  Even after several F10s in a row, you can still hit an arrow key & discover that the menu bar is still active.

[...]

Issue #12707, opened 2026-07-21
Incorrect cursor position in v5.1.0 rc11


lsmxpftz-cmyk:
When I use the shortcut “Go to next line and set video position” or “Go to next line” and then edit the subtitle in the text box, the cursor is not positioned at the beginning of the text. Instead, it automatically moves several characters forward.

<img width="690" height="400" alt="Image" src="https://github.com/user-attachments/assets/033efd28-1a78-4d59-99a5-c7b4719007cc" />

shanedk:
I'm seeing the same thing in English. The insertion point appears at the beginning of the text as in previous versions, but as soon as you start typing it shows up in the middle of the text somewhere.

niksedk:
Thanks for reporting :)

Fixed in #13097 - the cursor now always starts at the beginning of the text box when navigating lines, like in SE4.

Also added a new shortcut "Go to next subtitle and set cursor at end" (unbound by default, can be assigned in Options - Shortcuts).

PR #13379, opened 2026-08-08
Restore focus after activation loss and reset caret on grid navigation (#13371)

niksedk:
Fixes points 1 and 2 of #13371.

## 1. Ctrl+1..9 "hijacked by the top menu"

Not a shortcut collision — SE binds nothing on Ctrl+digit except `SetupLikeSe4Command` (Ctrl+4), menu headers use no digit mnemonics, and `MenuItem.InputGesture` is display-only in Avalonia. The reporter's Ctrl+1..9 are **Ditto's global hotkeys**, so SE never sees the keys at all.

What actually happens is a focus fallback. SE handles `Window.Deactivated` but had nothing on `Activated`. Another app taking focus and handing it straight back — a global hotkey from a clipboard manager, a launcher, a notification — leaves Avalonia without a focus target, and focus falls to the first focusable element in the window: the menu bar. "File" then looks activated, the text box loses its caret, and every following key is swallowed, because `OnKeyDownHandler` returns early while `IsMainMenuFocused()`. The same failure mode is already documented for #13111 in `InitListViewAndEditBox`.

The fix remembers the focused control on deactivation and restores it on re-activation, deferred and gated on focus having genuinely landed nowhere useful — so re-activating the window with a click keeps whatever the click focused, and the grid is the fallback.

## 2. Caret position persists across subtitles

A partial regression of the #12707 fix. The reset lives in `SelectAndScrollToRow`, which covers commanded jumps, but plain grid navigation (arrow keys, mouse) goes through `SubtitleGridSelectionChanged` instead. That path only calls `EditTextBox.ClearSelection()`, which collapses the selection to the caret without moving it — and since the text box is a TwoWay binding, Avalonia keeps `CaretIndex` when the bound text changes, so the new line inherited the previous line's clamped offset. Now reset on the same "row actually changed" condition.

## Not included

Point 3 (Shift+Enter inserting a line break) is standard `AcceptsReturn` behaviour and left as-is. The related line-limit bug — the max-lines limiter requires `KeyModifiers.None`, so Shift+Enter bypasses "limit number of lines" while plain Enter respects it — and the SE4-style "Shift+Enter goes to the next subtitle" request are better handled separately.

## Testing

`dotnet build` clean; full UI suite green (1729 passed, 1 skipped).

### DIFF
--- a/src/ui/Features/Main/MainView.cs
+++ b/src/ui/Features/Main/MainView.cs
@@ -55,6 +55,7 @@ public class MainView : ViewBase
             _vm.Window = hostWindow;
             _vm.Window.Closing += _vm.OnClosing;
             _vm.Window.Deactivated += _vm.OnWindowDeactivated;
+            _vm.Window.Activated += _vm.OnWindowActivated;
             _vm.Window.Loaded += (_, _) =>
             {
                 _vm.OnLoaded();
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -382,6 +382,7 @@ public partial class MainViewModel :
     FindViewModel? _findViewModel;
     Control? _findPreviousFocus;
     Control? _focusBeforeMainMenu;
+    Control? _focusBeforeWindowDeactivated;
     bool _altClosesMainMenuOnKeyUp;
     readonly AltMenuActivationGuard _altMenuActivationGuard = new();
     bool _findClosingProgrammatically;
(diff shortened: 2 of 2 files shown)


St0rmXtr00per:
@niksedk Thank you very much for solving these issues! Huge respect and thanks for one of my very favorite programs. SE5 has so many good things, looks and works so much better! 

As for the third "issue" (shift+enter) I agree with your verdict. If there was an option to assign shift+enter as a hotkey for jumping to the next line and thus turning off the the standard AcceptsReturn behavior it would be great in my view. I think either previous version of SE or Aegisub had this hotkey as a default that's why I adopted it. But I agree it's not that crucial, this I and others can relearn. 

Thank you once again! Will be waiting for the updates. God bless!

Issue #13325, opened 2026-08-07
SE 5.2.0 beta 6: Additional problems dealing with first caption of a subtitle file


GrampaWildWilly:
I put a problem report inside a closed Pull Request #13306.  Even though what I discovered is closely related to the issue handled by the Pull Request, it's probably a bad idea to put a problem report inside a Pull Request.  Anyway, I retested the issue in beta 6 & it's still happening.  So go read that Pull Request to find the problem report.

GrampaWildWilly:
The part of the problem in the Pull Report dealing with the obscured popup context menu also applies to any cascaded submenu subordinate to the popup context menu.

GrampaWildWilly:
This was originally framed as a problem with the first caption of a subtitle file.  But I believe that since that original problem has been fixed, the issues in question here in fact apply to any caption in a subtitle file, not just the first.

GrampaWildWilly:
In beta 7 this has not been fixed.  When you type the Delete key on a caption, the delete confirmation dialog comes up looking like it's in foreground.  But it's faking.  You can tell because the buttons in the dialog are all gray.  One of them should be whatever color you've chosen for active buttons.

The other half of the issue is also still present.  If instead you pop up the context menu on a caption in the grid, the menu is still obscured by the undocked Audio visualizer.

These things are true on all captions, not just the first.

GrampaWildWilly:
Oh no!

<img alt="#01" src="https://github.com/user-attachments/assets/d4111c4e-1b6f-42c0-b1c6-7d27452b2d34" />

The Delete key pops up the delete confirmation dialog in foreground now, but this issue with the popup context menu is still happening in beta 9.

GrampaWildWilly:
Oops.  Of course I meant beta 8.  And I did try the Delete entry on the context menu & it does also give the delete confirmation dialog in foreground.

GrampaWildWilly:
With beta 9, popping up a context menu on a line in the grid is still obscured by the undocked Audio visualizer.

PR #13399, opened 2026-08-09
Fix #13398: keep foreground on the dialog that triggered an undock rebuild

### LINKED ISSUE #13398: SE 5.2.0 beta 8: Settings focus anomaly, possibly regression
We've dealt with so many of these I can't remember if we've dealt with this particular one before.  I know the Settings dialog didn't used to come up properly in foreground on top of the undocked Audio visualizer but that's solved.  Still solved, I must add.  I don't know if this is a problem worth addressing or even a problem at all.  But when I open Settings, the undocked Audio visualizer goes out of sight.  I saw this with something else the other day as well.  I'm beginning to assume this is intentional behavior that does NOT need to be addressed.  When the current sequence is done, the Audio visualizer does float back up into view like it's supposed to.

So I traversed around the Settings dialog with Tab until I came to the end.  I had changed one setting so I typed Spacebar on Apply.  Everything flickered & the Settings dialog was still on top.  Very careful observation showed that the Audio visualizer briefly flickered into view then disappeared again.  But Settings was not in foreground.  The window title bar didn't have the right color.

It was unclear what was in foreground at that point.  The Audio visualizer was still out of sight.  The main SE window title bar color also indicated it was not in foreground.  A little experimenting (hitting Left/RightArrow) showed that the undocked video player was in foreground.  Possibly the out-of-view Audio visualizer was in foreground, but that would be rather anomalous.  I even think of that because when it is in foreground, the Arrow keys also reposition the video.  I think it's more likely that the video player was in foreground.  The Alt+Tab ribbon did show the undocked video player top left, so that's pretty strong confirmation it was in foreground.

Continuing Alt+Tab brought Settings into foreground again.  Then Tab followed by Spacebar activated OK & everything went back to normal.  The Settings dialog closed, the main window came to foreground, & the Audio visualizer hovered back into view like it's supposed to.

I thought you created some kind of central service routine that was supposed to globally fix these failed foreground issues for all windows current & future.  I guess this is one that is too special for that.

niksedk:
Fixes #13398.

## What happens

With the video controls undocked, **Settings → Apply** leaves the Settings dialog visible but *deactivated*, with an undocked window in front. @GrampaWildWilly had to Alt+Tab back to the dialog to reach OK.

## Why

`Apply` calls `MainViewModel.ApplySettings`, which in undocked mode runs `VideoUndockControls` to rebuild the layout. That closes **both** undocked tool windows and re-creates them via `WindowService.ShowIndependentWindow` — which ends in:

```csharp
window.Show();
window.Focus();
```

So the rebuild takes foreground, even though the dialog that asked for it is still open.

This also explains the flicker in the report ("the Audio visualizer briefly flickered into view then disappeared again"): the freshly shown windows come up topmost, and `KeepTopmostWhileOwnerActive` immediately drops them again because the dialog's `SuspendUndockedTopmost` is still held.

Worth saying explicitly, since the issue asks about it: the central suspension service from #13325/#13362 was working correctly here. It governs **Z-order**, not **activation** — it has no way to stop an explicit `Focus()` call on a newly shown window. That is why this one slipped past it.

## Fix

Remember which window was in front before the rebuild, and hand activation back once both undocked windows exist — posted at `Background` priority so the two `Show()`/`Focus()` calls have settled first.

The candidate must be neither the main window nor either undocked tool window. That keeps the change tightly scoped:
- undocking from the menu (no dialog in front) behaves exactly as before;
- the two windows being destroyed and re-created can never be picked as the reactivation target.

## Test plan

- [x] `dotnet test tests/UI/UITests.csproj` — 1772 passed
- [x] `dotnet build src/ui/UI.csproj -c Release`
- [ ] **Needs manual verification** (I could not reproduce this without a GUI session): undock the video controls, open Settings, change a setting, press Apply — the Settings dialog should stay in foreground, and the undocked windows should stay behind it until the dialog closes.
- [ ] Regression check: undock/redock from the menu with no dialog open still leaves the main window in front.

I have not added an automated test. The existing `SuspendUndockedTopmostTests` cover the ref-counted service, but window *activation* is not observable headlessly — `Window.IsActive` is read-only and needs a real desktop lifetime — so a test here would have to abstract away the very thing being fixed.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -6536,6 +6536,13 @@ public partial class MainViewModel :
 
         Dispatcher.UIThread.Post(() =>
         {
+            // Rebuilding the undocked windows below ends in ShowIndependentWindow, which does
+            // Show() + Focus() - so it takes foreground. Settings -> Apply reaches here through
+            // ApplySettings while its own dialog is still open, which left that dialog visible
+            // but deactivated and an undocked window in front of it (#13398). Remember a
+            // foreground dialog now and hand activation back once both windows exist.
+            var windowToReactivate = GetActiveWindowOtherThanMainAndUndocked();
+
             AreVideoControlsUndocked = true;
 
             var position = vp.Position;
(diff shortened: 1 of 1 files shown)


GrampaWildWilly:
Breathtaking speed, as usual.  You're setting a high bar for yourself.  Anyway, don't worry about test cases.  I will retest when you put up beta 9 tomorrow night.  Do you ever sleep?

niksedk:
> Breathtaking speed, as usual. You're setting a high bar for yourself. Anyway, don't worry about test cases. I will retest when you put up beta 9 tomorrow night. Do you ever sleep?

I do not sleep enough :)

Hopefully the number bugs will go down at some point... ;)

PR #13403, opened 2026-08-09
Fix ASSA tag scan in StrippableText.FixCasing

ivandrofly:
## Summary
Fix two index bugs in the ASSA tag scan of `StrippableText.FixCasing`:

- Pair each `{` with the first `}` **after it**, instead of searching both braces from the same start index. A stray `}` before the first `{` (e.g. `a} b {\i1}`) no longer aborts the scan.
- Resume the scan after the inserted id. The old `idx = end + 1` used an index into the string as it looked *before* `Remove`/`Insert` changed its length, so any tag longer than the 7 char id (`_@1000_`) made the loop skip every following tag.
- Fix the typo in the method name: `ReplacAssaTagsRemove` -> `ReplaceAssaTagsRemove`.

### Why it matters

A tag that the scan skips stays literal in `StrippedText` while casing is fixed, and `}` is one of the `breakAfterChars` - so the word after it was uppercased in the middle of a sentence:

| input | before | after |
|---|---|---|
| `one {\an8\pos(1,2)} two {\i1} three {\b1} four` | `... two {\i1} Three {\b1} Four` | `... two {\i1} three {\b1} four` |
| `a} b {\i1} c` | `a} B {\i1} C` | `a} B {\i1} c` |

This also removes an inconsistency: lines whose tags were all short (<= 7 chars) already behaved the way this PR now makes every line behave.

## Test plan
- [x] `dotnet test tests/libse/LibSETests.csproj` - 984 passed
- [x] Two new regression tests in `StrippableTextFixCasingTest`, both verified to fail on `main` and pass here
- [x] Differential run of `FixCasing` over tagged lines (`{\p1}` drawings, unclosed `{`, `{}`, multiple inline tags, leading `{\an8}`) - the only output changes are the spurious uppercases shown above

## Not addressed (pre-existing)
Tag ids start at `idName = 1000`, but the two id lookups in `FixCasing` rebuild them as `$"_@{i}_"` from the list index, so they never match a tag entry. When a tag id follows a sentence break, the loop uppercases the id's leading `_` and clears `lastWasBreak`, so the real word is never capitalized - `hello. {\i1}how are you?` keeps `how` lowercase both before and after this PR.

### DIFF
--- a/src/libse/Common/StrippableText.cs
+++ b/src/libse/Common/StrippableText.cs
@@ -200,7 +200,6 @@ namespace Nikse.SubtitleEdit.Core.Common
                             lower = StrippedText.ToLowerInvariant();
                         }
                     }
-
                     if (start + 3 > lower.Length)
                     {
                         start = lower.Length + 1;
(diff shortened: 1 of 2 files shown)


niksedk:
Rebased onto main and pushed a follow-up commit:

- kept `ReplaceAssaTagsRemove` as the name (typo fix only) — it replaces ASSA tags with ids, while `ReplaceNames1Remove` is the one that replaces names, so `ReplaceNamesWithIds` blurred the two
- `closeIndex < 0` instead of `closeIndex < openIndex`: `IndexOf` from `openIndex + 1` can never return a smaller index, so `-1` was the only case the old comparison caught
- reverted the unrelated blank line in `ReplaceNames1Remove`
- added two regression tests, both verified to fail on `main`: tags following a tag longer than the 7-char id, and a stray `}` before the first `{`

Nice catch on the resume index — the skipped tags were being casing-fixed as plain text, which is where the stray uppercase after `}` came from. Thanks!

PR #13404, opened 2026-08-09
Avoid exact floating-point equality in hide-max value converters

ivandrofly:
## Summary
- Replace `d == double.MaxValue` with `Math.Abs(d - double.MaxValue) < 0.01` in `DoubleToDisplayShortConverter`, `DoubleToNoDecimalHideMaxConverter` and `DoubleToOneDecimalHideMaxConverter`
- The `double.MaxValue` sentinel ("no value") is still detected; the tolerance form avoids floating-point equality comparison warnings

## Test plan
- [ ] `dotnet build src/ui/UI.csproj` succeeds
- [ ] Open a subtitle and check the gap column: the row after the last line (sentinel value) shows an empty cell, not a clamped number
- [ ] Regular gap/duration values still display as before

### DIFF
--- a/src/ui/Logic/ValueConverters/DoubleToDisplayShortConverter.cs
+++ b/src/ui/Logic/ValueConverters/DoubleToDisplayShortConverter.cs
@@ -20,7 +20,7 @@ public class DoubleToDisplayShortConverter : IValueConverter
         var useFrameMode = Se.Settings.General.UseFrameMode;
         if (value is double ms)
         {
-            if (ms == double.MaxValue || double.IsNaN(ms))
+            if (Math.Abs(ms - double.MaxValue) < 0.01 || double.IsNaN(ms))
             {
                 // Sentinel for "no value" (e.g. the gap after the last line) - show nothing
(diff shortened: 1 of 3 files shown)


niksedk:
Thanks for looking at this, Ivandro - but I think I will pass on this one.

The `double.MaxValue` here is a sentinel that is always assigned explicitly (see `SubtitleTextInfoHelper.cs` for the first/last row gap), never the result of a calculation, so exact equality is both correct and self-documenting. And because of the magnitude involved, `Math.Abs(d - double.MaxValue) < 0.01` matches exactly the same single value that `d == double.MaxValue` does - the next representable double below `double.MaxValue` is about 2e292 away, so the epsilon can never make a difference. It just suggests a tolerance that cannot exist.

These converters also run per cell per row, so I would rather keep the single compare.

If the motivation was a code analysis warning, a named constant like `const double NoValue = double.MaxValue;` would be a nicer way to silence it - happy to take a PR along those lines.

Thanks again for all the other work, much appreciated! :)

PR #13426, opened 2026-08-09
Fix #13407: subtitles intermittently missing in fullscreen playback

### LINKED ISSUE #13407: Subtitles intermittently disappear during fullscreen playback
In B8, subtitles can intermittently fail to appear during fullscreen playback.

I encountered this during normal watching and reproduced it several times.

When the issue occurs, subtitles that should be visible are not displayed in fullscreen.

After leaving fullscreen, the currently active subtitle is visible again in the normal video player, confirming that the subtitle and its timing are still present.

After entering fullscreen again later, subtitles can display normally, so the problem does not occur on every fullscreen session.

I also checked the same fullscreen playback scenario in B7 and could not reproduce the issue there. Subtitles are displayed normally.

So this appears to be a regression introduced in B8.

W11
SE 5.2.0 Beta8 ZIP

niksedk:
Fixes #13407.

### Root cause

Going fullscreen creates a brand-new `VideoPlayerControl` with its own mpv core, and the subtitle has to be pushed into it as an external ASS file (`sub-add`) by `MpvReloader`. That push raced the new player's startup:

- the mpv core is initialized lazily by the rendering surface, and `loadfile` runs after that - while the subtitle push fires on the next 16 ms preview tick;
- `SubAdd`/`SubReload` discarded mpv's return code, so a rejected push (core not up, or no file playing - `sub-add` is a playback-only command) was recorded as delivered: `MpvReloader` memoized the text as "what mpv is showing" and the dirty flag was cleared;
- nothing re-pushes during plain playback (only edits dirty the preview), so one mistimed push meant **no subtitles for the entire fullscreen session**, while the docked player still had them.

The intermittency comes from the 250 ms typing-settle window: entering fullscreen by shortcut delays the push long enough for the core to come up more often than entering by mouse; GPU/decoder warm-up variance does the rest. (The B7→B8 diff doesn't touch any of this code - B8 just shifted the timing. The race is latent in B7 too.)

There was also a second, narrower window: `sid=no` was set *after* `loadfile`, so a `sub-add ... select` landing in between was added and then immediately deselected - present but never drawn.

### Fix

- **`LibMpvDynamicPlayer`**: `SubAdd`/`SubReload`/`SubRemove` now return mpv's result, and answer `MPV_ERROR_UNINITIALIZED` instead of a silent success when there is no core; `sid=no` moved before `loadfile` so it can never undo a `sub-add` that got in first.
- **`MpvReloader.RefreshMpv`** returns whether mpv took the push; on a rejected push it leaves the memo fields (and retry budget) untouched so the next attempt actually retries.
- **`MainViewModel`**:
  - `RunPreviewRefresh` re-arms the dirty flag (400 ms backoff, same as the existing exception path) when the push was rejected, so it retries until the new player takes it;
  - the fullscreen path now arms the preview from the window's `onPlayerReady` callback (after `Open` + ready-wait) instead of pushing synchronously at window-open time, which could never work; same re-arm added after the layout-rebuild ready-wait;
  - `TryRefreshVideoPreview` skips pushes while no video is loaded, so the retry loop can't spin against an empty player.

### Tests

Four new tests in `MpvReloaderRetryTests` (a default-constructed player has no core - exactly the state a just-created player is in): push reports failure, rejected text is not memoized, retry budget is not spent, and the sub-commands report "not initialized" instead of success. All four fail against the old behavior; full UI suite (2255) passes.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -1094,6 +1094,11 @@ public partial class MainViewModel :
                 // (discussion #11744).
                 ReapplyPlaybackSpeed();
 
+                // The subtitle is state the fresh player starts over with too, and the refresh
+                // armed above may well have run before its mpv core was up. Arm it again now
+                // that the player is really playing (same reasoning as fullscreen, #13407).
+                RefreshSubtitlePreview();
+
                 if (savedAudioTrack != null && vp.VideoPlayer is LibMpvDynamicPlayer mpv)
                 {
                     mpv.SetAudioTrack(savedAudioTrack.Id);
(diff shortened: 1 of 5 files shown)


niksedk:
Self-review follow-up (8737e59), covering the mirror direction - **leaving** fullscreen:

- Entering fullscreen resets the reloader, which deletes the temp file behind the docked player's still-loaded subtitle. Nothing re-armed the preview on close, so edits made while fullscreen could leave the docked player stale - and with the re-add budget spent, the `sub-reload` against the deleted file would (with this PR's retry loop) have failed and retried into the void every 400 ms. The fullscreen `onClose` now re-arms `RefreshSubtitlePreview()` so the docked player gets a fresh `sub-remove` + `sub-add`.
- A failed `sub-reload` now also drops the temp-file memo so the retry takes the re-add path - a reload against a missing file can never fix itself.

Full UI suite still green (2255).

PR #13429, opened 2026-08-09
Prevent video player from showing the next subtitle or blank (no subtitle) upon stopping

ghostminhtoan:
Description: This PR addresses two minor usability issues when using the "Play next (and stop)" and "Play previous (and stop)" features:

Playhead Overshoot: Currently, when playing a subtitle line using "Play next (and stop)", the media player stops exactly at the end time boundary. Due to video rendering latency or shared boundaries, the view frequently overshoots into the next subtitle line (e.g. stopping on line 2 selects line 3).

in summary:
I am on line 1; function play next and stop, plays the next line (line 2), but the view overshoots into line 3.
I am on line 3; function play previous and stop, plays the previous line (line 2), and the view does not overshoot into line 3.

Link sample video, please test with this video.
https://drive.google.com/drive/folders/16bts6q_K6ET37uM4HNKU98s4iPKz-x8G?usp=drive_link

Fix: Added a tiny rewind offset of 0.03s (30ms) upon stopping to ensure the playhead lands strictly within the target subtitle's boundaries and stops cleanly without bleeding into the next line.
Default Shortcuts: Assigned standard default shortcuts to help keyboard-centric users navigate subtitles more easily:

Ctrl+Down for Play next (and stop)
Ctrl+Up for Play previous (and stop)
These changes improve overall playback precision and accessibility when verifying/timing subtitles.

commit 9841b62d7

### DIFF
--- a/.agents
+++ b/.agents
@@ -1,0 +1,1 @@
+R:/HDD R/ZC SYMLINK/USERS/.agents
--- a/build/build.cmd
+++ b/build/build.cmd
@@ -1,0 +1,37 @@
+@echo off
+setlocal enabledelayedexpansion
+
+echo =======================================
+echo Building Subtitle Edit...
+echo =======================================
+
+:: Define output directory in a custom folder to prevent file locking issues
+set "OUT_DIR=%~dp0..\src\ui\bin\Release\net10.0-custom"
+set "EXE_PATH=!OUT_DIR!\SubtitleEdit.exe"
+
+:: Run dotnet build on the UI project targeting the custom directory
+dotnet build "%~dp0..\src\ui\UI.csproj" -c Release -o "!OUT_DIR!"
+if %ERRORLEVEL% neq 0 (
+    echo.
+    echo [ERROR] Build failed with exit code %ERRORLEVEL%!
+    pause
+    exit /b %ERRORLEVEL%
+)
+
+echo.
+echo [SUCCESS] Build completed successfully with 0 errors!
+echo.
+
+:: Open the output folder
+echo Opening folder: !OUT_DIR!
+explorer.exe "!OUT_DIR!"
+
+:: Run the executable
+if exist "!EXE_PATH!" (
+    echo Launching: !EXE_PATH!
+    start "" "!EXE_PATH!"
+) else (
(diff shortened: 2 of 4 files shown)


niksedk:
Thanks for the report and the PR!

Two parts of this were already/now handled:

- The grid jumping to the next line after a "play and stop" was fixed in #13334 (merged Aug 7) — the play-head is pinned through mpv's post-pause wind-down so the played line stays selected.
- You're right that stopping exactly on the end time still shows the wrong thing *on the video* (blank, or the next line on a shared boundary), since mpv hides a line at its end time. I've implemented your rewind idea in #13433, but frame-rate based: a fixed 30 ms is less than one frame at 24/25 fps, so where the stop lands would depend on the video. It now parks on the line's last visible frame (end − one frame, clamped to the line start). Credit to you in the commit.

A few notes on this PR as it stands:

- The shortcut change replaces the `Ctrl+Alt+P` default for Pause rather than adding to the list — that removal looks accidental.
- `.agents` (a symlink to your local drive) and `build/build.cmd` look accidentally committed.

If you'd like `Ctrl+Down` / `Ctrl+Up` as defaults for play next/previous (and stop), a clean PR with just that change (keeping the Pause default) would be welcome. Closing this one in favor of #13433 for the playback part.

GrampaWildWilly:
Ummm.... Let there be a dissenting voice here.  I would be totally enraged if somebody suddenly shoved a Ctrl+Up/DownArrow into my SE configuration.  I already have those shortcuts bound to do 30-second skips forward or backward in the video.  I do not look forward to suddenly having to fix my configuration because somebody else sees fit to replace my shortcut definitions with something he/she thinks is better.  Better for you is almost certainly not better for others, certainly not for me.

GrampaWildWilly:
Existing shortcuts can already do what you're proposing.

Alt+DownArrow  goes to the next caption, with waveform & video repositioning to the start of the caption.
F5 then plays that caption, advancing both the waveform & the video in sync, & stops at the end of the caption.

Alt+UpArrow then F5 do the same things for previous.

Please do not supply a solution for a problem that does not exist.  Your other stuff about overshooting the caption is fine by me.  But don't be foisting shortcuts on people in a Pull Request.

PR #13465, opened 2026-08-10
Fix #13405: reclaim keyboard focus stolen from an open modal dialog

### LINKED ISSUE #13405: SE 5.2.0 beta 8: Some more window focus examples
I suspect your fix from earlier this evening might have covered these cases.  But since I don't have my hands on beta 9, I can't tell.  I'll offer these as test cases I can verify on beta 9.  Or, maybe, these are in fact not covered.  In which case, these are legitimate new bugs.

I promise you, I'm not deliberately hunting for these.  I just report them as I stumble upon them.

I accidentally stumbled on this one.  It was not the function I was trying to execute.

After Alt+Tab . . .

OK.  Now this is what I was really trying to do . . .

There is actually no visible hint, as in the other cases we've encountered, that this is not in foreground.  My first clue came when I hit Esc & nothing happened.  So then I started hitting Tab & noticed focus moving around in the main window underneath the Speech to Text window.

Just as a somewhat off-topic aside, you might notice the rather unusual Input language of Bengali.  Last time I used Speech to Text there was no way to specify no input language.  (Shows you how often I use this feature.)  I have always had `--multilingual true` in the Advanced Settings, so I always chose an input language I knew didn't occur in the input video.  But at some point in the recent past the option of selecting Auto Detect for the Input language has arrived.  That's what I will use from now on.

niksedk:
Fixes the three remaining beta 9 repros from #13405 ([comment](https://github.com/SubtitleEdit/subtitleedit/issues/13405#issuecomment-5238749288)):

- Text to Speech with nothing loaded → warning box not in foreground
- Speech to Text with no video → dialog opens not in foreground
- Post-transcribe: "1 completed" box fine, the chained "open subtitle?" box in *fake* foreground (intermittent)

### Why #13409 didn't catch these

#13409 enforces **OS activation** on the dialog. But Avalonia routes every key press to its single **app-global focused element**, regardless of which window the OS delivered the key to (`KeyboardDevice.ProcessRawEvent`). So any deferred `Focus()` into the main window — the menu bar's post-close focus restore (`DeactivateMainMenu` posts it), a posted grid focus after an edit — lands *after* the dialog opened and silently re-routes the keyboard under the dialog, while the dialog stays genuinely OS-active with an active-looking title bar. That is exactly the reported state: no visible hint, Esc dead, Tab visibly walking the main window underneath. No activation event ever fires for it, so the #13409 enforcement is blind to it. The keyboard-first workflow (Alt-menus) makes it likely: the menu's focus restore races the dialog open, which also explains the intermittency.

### The fix

`ShowModalAsync` now enforces the focus half of the invariant for the dialog's whole lifetime, alongside the activation half:

- Focus landing on the input-disabled owner while the dialog is open is handed straight back to the dialog — to the control that last held the caret, falling back to the first focusable control.
- A key press or text input that reaches the owner is swallowed instead of delivered (better a dead keystroke than one editing the subtitle under the dialog), and triggers the same repair. This also repairs key routing even when Windows refuses the `SetForegroundWindow` call, since routing follows the global focused element, not the OS-active window.
- Nested modals reclaim to the top-most dialog; focus in the independent, enabled undocked tool windows is never touched.

Like #13409, the repair is driven by the broken state itself, not by patching each producer — so all three repros (and any future producer of the same steal) are covered by one mechanism in one place.

### Tests

7 new headless tests in `ModalForegroundEnforcementTests`: steal-and-reclaim, caret-position restore, owner key swallow (and dialog keys not swallowed), independent-window focus left alone, nested-modal reclaim to top, and no interference after close. Full UI suite green: 2269/2269.

### DIFF
--- a/src/ui/Logic/WindowsService.cs
+++ b/src/ui/Logic/WindowsService.cs
@@ -1,8 +1,13 @@
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
+using Avalonia;
 using Avalonia.Controls;
 using Avalonia.Controls.Primitives;
+using Avalonia.Input;
+using Avalonia.Interactivity;
 using Avalonia.Threading;
+using Avalonia.VisualTree;
 using Microsoft.Extensions.DependencyInjection;
 using Nikse.SubtitleEdit.Features.Main.MainHelpers;
 using Nikse.SubtitleEdit.Features.Options.Settings;
@@ -371,6 +376,7 @@ namespace Nikse.SubtitleEdit.Logic
         internal static void ResetOpenModalsForTests()
         {
             _openModalCount = 0;
+            _modalFrames.Clear();
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


GrampaWildWilly:
> even when Windows refuses the SetForegroundWindow call

That sounds positively evil.  How dare they?  Anyway, as ever I'm looking forward to beta 10.  I should write a program that generates that comment.  "I'm looking forward to &currentbeta+1"  I would wear it out.

Issue #12293, opened 2026-07-09
Feature Request: histogram-based color isolation for VobSub/DVD OCR (handy in CLI/Batch mode)


tekk42:
Hi @niksedk,

I would like to propose an optimization for the VobSub (`.sub`/`.idx`) OCR pipeline, specifically to improve unattended CLI/batch processing (`/convert` or `seconv`) across both Subtitle Edit 4 and 5.

### The Problem
During unattended batch OCR of certain DVDs (such as "Lord of War"), text characters often blur together because the engine processes the gray outlines and anti-aliasing artifacts as part of the foreground text. For instance, the name "Yuri" is repeatedly misread as "Yurl" because the dot of the "i" and the spacing melt into a solid block. 

While this can be easily fixed manually in the GUI by toggling the color/transparency boxes, there is currently no index-independent way to automate this in a headless CLI batch run. This is due to the fact that DVD authoring tools frequently assign the 4 internal color indices dynamically and unpredictably per file or even per frame, making hardcoded JSON/XML custom color mappings risky for batch operations.

### The Solution: Frame-by-Frame Histogram Analysis
Since VobSub frames are strictly encoded using a 4-color index palette (2 bpp), we can utilize a mathematically stable hierarchy based on pixel frequency to isolate the text body dynamically and index-independently. 

Statistical analysis of the rendered bitmaps shows a highly reliable frequency distribution:
1. **Most frequent color:** Background
2. **Second most frequent color:** Main Text Body
3. **Third/Fourth most frequent color:** Outline / Anti-Aliasing

By implementing a quick frame-by-frame histogram check in the RAM right before passing the bitmap to the OCR engine, Subtitle Edit could automatically binarize the image based on frequency rather than rigid index positions:
* Identify the 2nd most frequent color (the text body) and map it to solid Black (or White for inversion).
* Map all other three colors (Background, Outline, Anti-Aliasing) to the opposite background color.

### The Benefit
This provides Tesseract with a perfectly skeletonized, thin, and crisp character shape. Outlines and blurred edges are physically removed before Leptonica touches the image, eliminating character-blurring bugs completely without risking index-mismatch errors across different DVDs in a batch. Besides better recognition rates this should also increase speed.

Would you be open to implementing a native toggle/feature for this dynamic histogram-based binarization in the official build?

Thanks for your amazing work on Subtitle Edit!

PR #13481, opened 2026-08-11
VobSub color isolation fixes

tekk42:
Hi Nikse,

First, I want to apologize for my poorly thought-out histogram proposal back in issue #12293. While the intention was right, the implementation had logical flaws. Thank you for following up on that and introducing the inner-pixel logic in v5.1.0-rc15!

However, the inner-search heuristic introduced a new edge case with letters like "i", "j", or umlauts ("ä", "ö", "ü"). The shadow/outline pixels completely bridge the narrow gap between the dot and the main glyph body. Since the true background color never penetrates this tiny cavity, the code flags those trapped pixels as `touchesBackground = false`, assigning them a `ratio = 0.0`. Because the code looks for the *lowest* ratio, the outline wins and the actual text body is wiped out. 

This PR brings three topological changes to resolve this:

1. **Laplace Smoothing:** We now calculate the ratio via `(b + 1) / (total + 1)`. If both the text body and a trapped cavity shadow hit $b = 0$, the larger mass in the denominator ensures that the larger text body pushes the ratio closer to 0, preventing the cavity shadow from winning.
2. **4-Directional Edge Intensity:** Instead of a binary boolean check, we sum up the actual background borders (0 to 4) per pixel to better differentiate between outlines and compact interior text shapes.
3. **`minPlane` Recalibration:** We lowered the noise floor from `16` to `8` pixels to ensure short words or isolated punctuation marks don't bypass the main loop, while safely preserving your original frequency fallback for absolute pixel-starved extreme cases.

***

**Feature Request for `seconv` / CLI:**
On a side note, since `seconv` is widely used for unattended batch conversion, would it be possible to bring the **line break separation** logic from the GUI to the CLI? Additionally, adding the **italics recognition feature** from Subtitle Edit 4 to `seconv` would greatly improve automated VobSub/PGS OCR. 

Thanks for your work on Subtitle Edit!

### DIFF
--- a/src/libse/Common/VobSubColorIsolation.cs
+++ b/src/libse/Common/VobSubColorIsolation.cs
@@ -43,7 +43,7 @@ namespace Nikse.SubtitleEdit.Core.Common
             var height = source.Height;
             var result = new SKBitmap(new SKImageInfo(width, height, SKColorType.Rgba8888, SKAlphaType.Opaque));
 
-            // Tally every opaque colour and how many of its pixels touch the transparent
+            // Tally every opaque colour and how often totally its pixels touch the transparent
             // background (4-neighbourhood; the bitmap edge counts as background). Key on RGB
             // only (alpha already gates the background) so anti-aliasing tiers that share a
             // hue but differ in coverage still merge.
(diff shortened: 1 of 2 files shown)


niksedk:
thx tekk42, looks good :)

DarkSwan86:
@tekk42 @niksedk tested on some problematic italian VOBSUB, seems to improve the recognition of the "i" and "l" characters with PaddleOCR Standalone on SE 5.2 beta 10. If i find other problematic Vobsub files, should i open a bugreport for them?

niksedk:
> @tekk42 @niksedk tested on some problematic italian VOBSUB, seems to improve the recognition of the "i" and "l" characters with PaddleOCR Standalone on SE 5.2 beta 10. If i find other problematic Vobsub files, should i open a bugreport for them?

You can try... remember to attach the vobsub file pair :)

PR #13490, opened 2026-08-11
Restore trough paging for the TableView scroll bar (#12894)

ivandrofly:
## Summary

Fixes the regression of #12894 reported in https://github.com/SubtitleEdit/subtitleedit/issues/12894#issuecomment-5238879921 (5.2.0-beta9): press-and-hold on the scroll bar trough pages right past the cursor to the end of the list again.

The DataGrid retirement (b24211a1b1) deleted `DataGridScrollBarBehavior` on the note that native ScrollViewer paging replaced it. It only half did: a trough click does page a full viewport now, but the trough of Avalonia's Fluent ScrollBar is a RepeatButton that keeps firing while `IsPressed`, and Button only re-evaluates `IsPressed` on PointerMoved — so with a stationary cursor the paging never stops. The shift+trough-click jump (#12051/#12438) had no native counterpart and was lost outright.

- New `TableViewScrollBarBehavior`: hold paging latches its direction, pages toward the cursor and pauses when the thumb reaches it (resuming only if the pointer moves further along, like the native Windows scroll bar); shift+click jumps the thumb to the click position.
- Enabled app-wide from a `TableView` style in Styles.axaml (same pattern as the old DataGrid style), which also covers the grids built with plain `new TableView()`.
- Unlike the DataGrid version there is no reflection and no LargeChange bookkeeping: the ScrollViewer template binds the scroll bar's `Value` two-way to the scroll offset and its `LargeChange` to the viewport. Handlers sit on the grid itself (tunneled press), so no template-applied wiring is needed.
- The old headless trough-paging tests return, ported to TableView, plus a shift+click jump test and an assertion that the scroll offset follows the bar.

Vertical scroll bar only, as before — the horizontal trough keeps the native behavior.

## Test plan

- [ ] Load a subtitle with a few hundred lines; press and hold the trough below the thumb with the mouse still: the view pages down and pauses when the thumb reaches the cursor instead of running to the end.
- [ ] Move the pointer further down during the hold: paging resumes; move it back up: it stays paused (no reverse).
- [ ] Release and short-click the trough: exactly one viewport per click.
- [ ] Shift+click the trough: the thumb jumps to the click position.
- [ ] Dragging the thumb still works normally.
- [ ] Same behavior in other grids, e.g. Options → Shortcuts.
- [ ] `dotnet test tests/UI/UITests.csproj --filter "FullyQualifiedName~TableViewScrollBarTroughPaging"` passes (7 tests).

### DIFF
--- a/src/ui/Logic/TableViewScrollBarBehavior.cs
+++ b/src/ui/Logic/TableViewScrollBarBehavior.cs
@@ -1,0 +1,308 @@
+using System;
+using System.Linq;
+using Avalonia;
+using Avalonia.Controls;
+using Avalonia.Controls.Primitives;
+using Avalonia.Input;
+using Avalonia.Interactivity;
+using Avalonia.Layout;
+using Avalonia.Threading;
+using Avalonia.VisualTree;
+
+namespace Nikse.SubtitleEdit.Logic;
+
+/// <summary>
+/// Gives a <see cref="TableView"/>'s vertical scroll bar Windows-standard trough behavior:
+/// press-and-hold on the trough pages toward the cursor and pauses when the thumb reaches
+/// it (#12894), and shift + trough click jumps the thumb straight to the click position
+/// (#12051/#12438).
+///
+/// Both behaviors shipped for the old DataGrid in DataGridScrollBarBehavior, which the
+/// TableView migration retired with the grid itself (b24211a1b1) on the note that native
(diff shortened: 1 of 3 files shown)


GrampaWildWilly:
I want to comment because I want to be sure to test this after it becomes available in a release I can get.  Also, why not horizontal scroll bars?  Surely it's basically all the same code with only minor adjustments.

PR #13558, opened 2026-08-12
Reduce allocations in StringExtensions.RemoveChar via string.Create

ivandrofly:
## Summary
- Convert the three `RemoveChar` overloads from allocate-`char[]`-then-copy to a counting pass + `string.Create`, writing directly into the final string's buffer
- Return the original string when nothing matches — the old code allocated a full copy even for no-op calls
- On .NET 10 the multi-char overload takes `params ReadOnlySpan<char>`, so literal char arguments no longer allocate an args array; netstandard2.1 keeps `params char[]` (its `string.Create` does not accept a `ref struct` state)

## Benchmarks

BenchmarkDotNet, .NET 10, i7-13700. "short" = 38-char subtitle line, "long" = ~200 chars, "no match" = removed chars do not occur.

| Overload | Case | Old (char[]) | New (string.Create) | Ratio | Alloc old → new |
|---|---|---:|---:|---:|---|
| 1 char | short | 58.6 ns | 65.0 ns | 1.11 | 192 B → 88 B |
| 1 char | long | 234.3 ns | 278.1 ns | 1.19 | 760 B → 344 B |
| 1 char | no match | 63.4 ns | 15.3 ns | **0.24** | 208 B → **0 B** |
| 2 chars | short | 53.0 ns | 65.2 ns | 1.26 | 184 B → 80 B |
| 2 chars | long | 217.0 ns | 318.7 ns | 1.56 | 712 B → 296 B |
| 2 chars | no match | 59.8 ns | 26.4 ns | **0.44** | 208 B → **0 B** |
| 4 chars | short | 139.0 ns | 119.7 ns | 0.86 | 208 B → 72 B |
| 4 chars | long | 423.0 ns | 587.5 ns | 1.39 | 720 B → 272 B |
| 4 chars | no match | 83.9 ns | 73.4 ns | 0.88 | 240 B → **0 B** |

Allocated bytes roughly halve when matches exist and drop to zero when nothing matches; the counting pass costs some CPU on match-heavy calls, while no-match calls get much faster.

## Test plan
- [ ] `dotnet build src/libse/LibSE.csproj` succeeds for both netstandard2.1 and net10.0
- [ ] `dotnet test tests/libse/LibSETests.csproj --filter "FullyQualifiedName~StringExtensions"` passes (RemoveChar1–4 cover all three overloads)
- [ ] Spot-check: `" Hallo  world! ".RemoveChar(' ')` returns `"Halloworld!"`; a string not containing the removed chars comes back unchanged

### DIFF
--- a/src/libse/Common/StringExtensions.cs
+++ b/src/libse/Common/StringExtensions.cs
@@ -783,52 +783,123 @@ namespace Nikse.SubtitleEdit.Core.Common
 
         public static string RemoveChar(this string value, char charToRemove)
         {
-            char[] array = new char[value.Length];
-            int arrayIndex = 0;
-            for (int i = 0; i < value.Length; i++)
+            var count = 0;
+            for (var i = 0; i < value.Length; i++)
             {
-                char ch = value[i];
-                if (ch != charToRemove)
+                if (value[i] == charToRemove)
                 {
-                    array[arrayIndex++] = ch;
+                    count++;
                 }
             }
 
-            return new string(array, 0, arrayIndex);
+            if (count == 0)
+            {
+                return value;
+            }
+
+            return string.Create(value.Length - count, (value, charToRemove), (s, state) =>
+            {
+                int index = 0;
+                foreach (var ch in state.value)
+                {
+                    if (ch != state.charToRemove)
+                    {
+                        s[index++] = ch;
+                    }
+                }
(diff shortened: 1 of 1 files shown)


ivandrofly:
Could've been 4/5 * faster using SIMD see #13561

PR #13571, opened 2026-08-12
Bring the main window to the front on startup in undocked mode

### LINKED ISSUE #13569: SE 5.2.0 beta 11: In undocked mode initial foreground window is video player instead of main window
This is totally unexpected.  I've opened & closed beta 11 several times to make sure this was not another of those one-off boundary occurrences.  Every time I open beta 11, the undocked window in foreground is the video player, as confirmed in the Alt+Tab ribbon.  Of course, this is not right.  SE has never done this before.  Whatever else was ever wrong, the main window was always in foreground.  This is a very obvious condition on startup.  It's definitely new with beta 11.  This must be a side effect of one of the moles you whacked between beta 10 & beta 11.

niksedk:
Fixes #13569

The reporter is right that it is new in beta 11, and right that it is a side effect of a mole whacked between beta 10 and beta 11.

### The regression

`f86786a5a8` ("Fix two focus/topmost violations against other applications", in beta 11, not in beta 10) removed the unconditional `Window.Activate()` from the startup sequence:

```csharp
-  Window.Activate();
-  TableViewExtras.FocusRow(SubtitleGrid);
+  if (Window.IsActive)
+  {
+      TableViewExtras.FocusRow(SubtitleGrid);
+  }
```

That was a correct fix for its own problem — the call runs a second after launch and would drag SE back over an application the user had switched to while SE was loading. But that `Activate()` was doing a second job nobody had written down: putting the main window in front of **SE's own** windows.

In undocked mode `VideoUndockControls()` runs during startup (`RememberPositionAndSize` + `UndockVideoControls`), and `WindowService.ShowIndependentWindow` ends with:

```csharp
window.Show();
window.Focus();
```

So the undocked video and waveform windows take the foreground as they are created. Before beta 11 the `Activate()` a second later quietly took it back; now nothing does, and an undocked window is left in front — exactly what the reporter sees in the Alt+Tab ribbon. The subtitle grid also never gets focused, since that is behind the same `Window.IsActive` check.

### The fix

Distinguish the two cases the single `IsActive` check was conflating:

- **Main window active** → focus the grid, as now.
- **Main window not active, but some other SE window is** → SE owns the foreground, so another of our own windows is in front. Activate the main window and focus the grid — the pre-beta-11 behaviour.
- **No SE window active** → the user is in another application. Do nothing, which is what `f86786a5a8` was protecting.

`IsAnyApplicationWindowActive()` is the new helper, sitting next to the existing `GetActiveWindowOtherThanMainAndUndocked()` which walks `desktop.Windows` the same way.

### Verification

`src/ui` builds clean. No test: this is window activation across two real top-level windows a second into startup, and the headless harness in this repo cannot give a trustworthy signal for it — the existing `UITests.Features.Main.*` window tests already fail in my environment for unrelated reasons (verified against a clean tree), so a new one there would prove nothing. Worth a quick manual check in undocked mode: launch, confirm the main window is foreground in Alt+Tab and the grid has focus; then launch and immediately click into another application, and confirm SE does *not* steal it back.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -7073,6 +7073,29 @@ public partial class MainViewModel :
     }
 
     /// <summary>
+    /// Whether SE itself holds the foreground, i.e. any of its windows is the active one. Tells
+    /// "the user switched to another application" apart from "another SE window is in front of the
+    /// main one" - cases that need opposite answers when startup claims the foreground (#13569).
+    /// </summary>
+    private static bool IsAnyApplicationWindowActive()
+    {
+        if (Application.Current?.ApplicationLifetime is not IClassicDesktopStyleApplicationLifetime desktop)
+        {
+            return false;
+        }
+
+        foreach (var window in desktop.Windows)
+        {
+            if (window.IsActive)
+            {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /// <summary>
     /// Suppress (or restore) the undocked tool windows' topmost state. They float above the
     /// main window while SE is active (KeepTopmostWhileOwnerActive, #11971), which also puts
     /// them above the main window's popups and above modal dialogs. Registered with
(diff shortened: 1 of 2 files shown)


GrampaWildWilly:
Oh you know I'll confirm.  I find it mildly amusing that somebody would launch SE & in the split second it takes for the app to open, there would be a task switch to something else.  Didn't you want to open SE?  Why would you switch to something else?  Ah me . . .  users.  Can't live with them.  Can't live without them.

PR #13589, opened 2026-08-13
Update German.json

Need74:
Just updating the character encoding and standardizing to UTF-8.


niksedk:
I really find it easier to read with utf-8 letters instead of encoding

PR #13604, opened 2026-08-13
Speed up OCR/export pixel loops with spans and Vector<T>

ivandrofly:
## Summary
- VobSubColorIsolation: replace up to 5 `SKBitmap.GetPixel` interop calls per pixel (pixel + 4 neighbours) and per-pixel `DrawPoint` rendering with row-span reads and direct pixel writes; gated to non-premultiplied Bgra8888/Rgba8888 with the original GetPixel path kept as fallback for other layouts
- NikseBitmap: fourteen whole-buffer BGRA loops now write one packed uint per pixel instead of a 4-byte `Buffer.BlockCopy`; `MakeBlackAndWhiteForOcr` and `MakeTwoColor` additionally get a `System.Numerics.Vector<T>` pass (netstandard2.1-compatible, endian-safe packed constants, scalar tail)
- SkiaExt: `GetNonTransparentHeight`/`GetNonTransparentWidth` reuse the existing pointer-based `GetNonTransparentBounds` scan instead of per-pixel `GetPixel` (the width scan was column-major on top of the interop cost)
- `ConvertTo8BitsPerPixel`: exact-colour dictionary cache in front of the linear palette scan; only provably-stable outcomes are cached, so palette contents, insertion order and the index buffer are unchanged
- BluRaySupPicture: RLE encode and both palette passes read whole rows via a small row reader instead of calling `GetPixel` per pixel of every run

Output is bit-identical everywhere: each change was verified against the original implementation with byte-comparison harnesses (all colour types incl. premul fallbacks, exhaustive 256-threshold and 767-minRgb sweeps for the Vector paths, palette + RLE byte comparison for the sup encoder).

## Benchmarks

A/B vs the branch base commit, same harness binary for both sides, median of 5 samples after warmup (ms/op; 1920×1080 bitmaps, 1920×200 for the VobSub/palette rows; Release, net10.0, AVX2):

| Method | before | after | speedup |
|---|---|---|---|
| SkiaExt.GetNonTransparentHeight+Width | 55.76 | 0.54 | 104× |
| VobSubColorIsolation.BinarizeForOcr | 34.60 | 0.71 | 49× |
| NikseBitmap.MakeBlackAndWhiteForOcr | 10.00 | 0.29 | 35× |
| NikseBitmap.Fill | 7.97 | 0.35 | 23× |
| NikseBitmap.MakeTwoColor(300) | 9.84 | 0.56 | 18× |
| VobSubColorIsolation.Isolate | 23.17 | 2.89 | 8× |
| NikseBitmap.ConvertTo8BitsPerPixel | 6.54 | 1.56 | 4.2× |
| NikseBitmap.InvertColors | 2.27 | 0.75 | 3.0× |
| NikseBitmap.GrayScale | 7.37 | 6.64 | 1.1× (float math kept verbatim; only memory access changed) |

## Test plan
- [ ] `dotnet build src/libse/LibSE.csproj` succeeds for both netstandard2.1 and net10.0
- [ ] `dotnet test tests/libse/LibSETests.csproj` passes
- [ ] `dotnet test tests/seconv/SeConvTests.csproj` passes (covers ImageToImage, VobSubColorIsolation, VobSub extraction, OCR)
- [ ] Batch-OCR a VobSub sample and export a Blu-ray .sup; output matches main byte-for-byte

### DIFF
--- a/src/libse/BluRaySup/BluRaySupPicture.cs
+++ b/src/libse/BluRaySup/BluRaySupPicture.cs
@@ -122,14 +122,17 @@ namespace Nikse.SubtitleEdit.Core.BluRaySup
             }
 
             var transparentColor = (byte)palette[palette.Count - 1];
-            var bytes = new List<byte>();
+            var bytes = new List<byte>(bm.Width * 2);
+            var reader = new BitmapRowReader(bm);
+            var row = new SKColor[bm.Width];
             for (var y = 0; y < bm.Height; y++)
             {
+                reader.ReadRow(y, row);
                 int x;
                 int len;
                 for (x = 0; x < bm.Width; x += len)
                 {
-                    var c = bm.GetPixel(x, y);
+                    var c = row[x];
 
                     byte color;
                     if (c.Alpha == 0)
@@ -147,7 +150,7 @@ namespace Nikse.SubtitleEdit.Core.BluRaySup
 
                     for (len = 1; x + len < bm.Width; len++)
                     {
-                        if (bm.GetPixel(x + len, y) != c)
+                        if (row[x + len] != c)
                         {
                             break;
                         }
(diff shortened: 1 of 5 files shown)


niksedk:
## Review

Verified with an A/B harness against the branch's merge base (`c17df8c40`) — same public APIs, two libse builds, hashed outputs:

- **NikseBitmap ops** — 12 sizes × 4 pixel-content modes over all 14 rewritten loops, plus `ChangeBrightness`/`ChangeAlpha` across 7 factors: byte-identical.
- **Threshold sweeps** — `MakeBlackAndWhiteForOcr` over 0–256 and `MakeTwoColor` over −2…768 (both overloads): byte-identical, and the `Vector<T>` and scalar paths agree (whole harness re-run with `DOTNET_EnableHWIntrinsic=0`).
- **`ConvertTo8BitsPerPixel` colour cache** — the riskiest change, since it rests on `FindBestMatch`'s early-stop invariants. 400 randomised images spanning cluster counts 1–300, i.e. both sides of the 200/255 palette limits: byte-identical.
- **`VobSubColorIsolation` + `BluRaySupPicture.CreateSupFrame`** — 8 colour/alpha formats including the Premul and non-8888 fallbacks: byte-identical.
- `dotnet test` on libse and seconv: green.

Three things came out of it, all now fixed on the branch.

### 1. `GetNonTransparentHeight`/`Width` mis-read four-byte formats without an alpha byte (76b3489)

The new pointer scan was gated on `bitmap.BytesPerPixel == 4`, but four bytes per pixel does not mean alpha lives in byte 3. `Rgb888x` has a *padding* fourth byte that `GetPixel` ignores — it always reports alpha 255 — and the `1010102` types pack alpha into the top two bits of that byte, so its low bits carry colour. An `Rgb888x` bitmap whose padding bytes are zero came back **fully transparent: height and width 0 instead of its full size**.

Gated on `Bgra8888`/`Rgba8888` instead — the two layouts that really do put alpha in byte 3, under any alpha type — plus a `GetPixels() != IntPtr.Zero` check, matching the guards `TryGetChannelOffsets` and `BitmapRowReader` already use in this same diff. Three regression tests added to `SkiaExtTrimTest.cs`; the `Rgb888x` one fails on the branch as it stood.

The one in-repo caller (`BluRaySupParser.cs:1254`) passes `Rgba8888`, so SE itself was unaffected — but these are `public` extension methods, and the fast path is unchanged for them.

### 2 & 3. Cache growth and table-build range (1f84427)

Both are on work the rewrites moved to the front of the method rather than per pixel:

- **`ConvertTo8BitsPerPixel`'s `seen` dictionary was unbounded.** Once the palette fills at 255 entries every result is frozen, so *every* distinct colour becomes cacheable — millions of entries on a photographic source. Capped at 65536; past the cap the linear scan simply runs again, so the only thing lost is the shortcut, and subtitle bitmaps stay far below it.
- **`BuildScaleTable` evaluates all 256 channel values**, not just the ones the bitmap holds, so a large enough factor could overflow the decimal multiply or the `int` cast where the old per-pixel loop only did if such a pixel existed. The factor is now clamped to 0…256 first: from 256 up every non-zero channel already saturates at 255, and a negative factor already floors at 0.

Re-verified after both: byte-identical for every sane factor and for 90k-distinct-colour bitmaps well past the cap. The only behaviour change anywhere is that factors which used to throw `OverflowException` now return the saturated result.

PR #13616, opened 2026-08-14
Refactor: replace profile-name TextBox reference in VMs with focus callback

ivandrofly:
## Summary
- Remove the `ProfileNameTextBox` control reference from `FixCommonErrorsProfileViewModel` and `ImageBasedProfileViewModel` (including the placeholder `new TextBox()` in their constructors)
- Replace it with an `Action? FocusProfileName` callback assigned by the window, matching the existing `FocusSearchBox` pattern in the Find/Replace windows
- Windows now post `nameBox.Focus()` to the UI thread dispatcher, preserving the original deferred-focus timing
- Drop the now-unused `Avalonia.Threading` usings from both ViewModels

## Test plan
- [x] Tools → Fix common errors → profiles: click "New profile" — a new empty profile is added and the name textbox receives keyboard focus
- [x] Type a name and press Enter — the dialog saves and closes (Enter-to-OK still works)
- [ ] File → Export image-based format → profiles: click "New profile" — same focus behavior in the name textbox
- [ ] Delete button enable/disable still follows profile count (>1) in both dialogs

### DIFF
--- a/src/ui/Features/Files/ExportImageBased/ImageBasedProfileViewModel.cs
+++ b/src/ui/Features/Files/ExportImageBased/ImageBasedProfileViewModel.cs
@@ -1,6 +1,5 @@
 ﻿using Avalonia.Controls;
 using Avalonia.Input;
-using Avalonia.Threading;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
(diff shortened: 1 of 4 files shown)


ivandrofly:
Eventually it would also be better to get rid of the Window instances from ViewModels as VMs are not supposed to be storing UI stuff!

PR #13674, opened 2026-08-15
Feature.romanisation [Branch Source Changed]

thando-ndlovu:
Moved here for branch change

Image below is one of the intended outcomes of a file processing.
Would be created with romanisation tool and ported into various styles by using existing tools (merge lines, merge two subtitles, edit subtitle appearance etc)

README screenshot

#### Standard romanization for subtitles

README screenshot procedure tools

romanize


niksedk:
Hi @thando-ndlovu, thanks a lot for the work you put into this - the romanizers for Cyrillic, Devanagari, Ge'ez, Greek, Hangul and Kana are clearly a lot of careful effort, and the tests are appreciated.

That said, I'm going to close this one. Romanization is a fairly niche feature, and I'd rather not grow the core app (and the translation/maintenance surface that comes with it) for it. This is exactly the kind of thing the SE 5 plugin system is made for, so my suggestion is to ship it as a plugin instead:

- SE 5 plugins are standalone executables talking to SE via JSON, so your libse-independent romanizer code can move over more or less as-is, and you can keep your own UI (language checkboxes, replace/above/below/before/after placement etc.).
- Docs: https://github.com/SubtitleEdit/subtitleedit/blob/main/docs/plugin.md
- Plugin index / listing: https://github.com/SubtitleEdit/plugins - once it's there, users can install it from **Plugins → Get plugins online...**

Also note that SE already has the "merge" side covered: **Tools → Merge two subtitles** (docs/features/merge-two-subtitles.md) combines a romanized subtitle with the original, and the "Merge lines with same time codes" / "Merge short lines" tools handle the single-line variants. So a plugin only needs to do the romanization itself, and the existing tools take care of combining the results.

Thanks again, and I'd be happy to add it to the plugin index when it's ready!

PR #13709, opened 2026-08-16
Show translation progress in batch convert, and stop progress lingering

### LINKED ISSUE #13706: Feature Request: Show translation progress percentage in Batch Convert status bar
Hi Nikse,

First of all, I want to express my sincere gratitude for creating and continuously maintaining Subtitle Edit. It is an incredibly powerful and well-designed tool that has become an essential part of my daily workflow. The fact that such a feature-rich subtitle editor is available as free, open-source software is truly remarkable. The continuous improvements — from Whisper/Parakeet integration to OCR and batch processing — speak volumes about your dedication. Thank you for your outstanding work!

### Description

I would like to suggest adding a **translation progress indicator** to the **Batch Convert** dialog when using the auto-translate feature, so users can intuitively see whether the translation is actively in progress.

### Current Behavior

When running auto-translate inside Batch Convert, there is no visible progress feedback. It is difficult to tell whether the translation is actively processing, stalled, or already completed — especially when converting a large number of files or when the translation engine (e.g., llama.cpp / local models) takes a while per file.

### Proposed Solution

Display the current file's translation completion percentage in the **status bar** of the Batch Convert window. It could be placed either before or after the existing status bar text. For example:

- Before: `45% | Translating file...`
- After: `Translating file... | 45%`

This way, users can see at a glance that the translation is progressing, without having to repeatedly check the output folder.

### Use Case

I frequently use Batch Convert to transcribe and translate multiple video files at once. Without a progress indicator, I have to keep checking the output folder to see if new translated files are appearing, which is not ideal.

### Additional Context

This is a minor UX improvement and certainly not a criticism of the current design — Subtitle Edit already works wonderfully. I just thought this small addition could make the batch translation workflow even more intuitive.

Please note that English is not my native language, and this issue was written with the help of a translation tool. I apologize for any language errors.

Thank you for considering this suggestion, and thank you again for this amazing piece of software!

niksedk:
Fixes #13706.

## Translation progress

Batch convert showed nothing while auto-translate ran: the row sat unchanged and the label stayed on `Converting 1 of 4...`, so a slow engine (local LLMs especially) was indistinguishable from a stalled one.

`DoAutoTranslate` already exposes a progress callback and seconv already uses it — batch convert just never set it. It is now wired into the item status the same way the OCR runners do, so the row reads `Translating: 45%`.

Engines forced into single-line mode (NLLB, CrispASR/Madlad) raise the callback once per subtitle line, so the status is only pushed when the whole percent changes — a 2000-line file gets at most 100 notifications instead of 2000.

## Progress no longer lingers

Translating and OCR are each one step of a run that still has the remaining convert functions and the save to go. The last progress value used to stay up for that whole stretch, so a row would show a finished-looking `Translating: 100%` / `OCR: 100%` long before it was done. Both now clear it when the step ends.

For OCR that clearing keys off the `OCR: {0}%` format string rather than a hard-coded prefix (translations move the label and the percent sign), so a terminal status a runner deliberately left behind — `Cancelled`, an error, `model likely wrong` — is never swallowed.

## Tests

- `BatchConverterAutoTranslateProgressTests` — drives a fake `IBatchContextTranslator` (the llama.cpp-advanced path the issue names) through `Convert()`: percent sequence, no repeated percent, no lingering 100%, and a negative control for translate-off. Verified non-vacuous: the two positive tests fail with the wiring removed.
- `BatchConverterOcrProgressStatusTests` — covers the progress/terminal-status discrimination directly, which is where the risk of swallowing a `Cancelled` lives.

Full UI suite: 2920 passed, 0 failed.

### DIFF
--- a/src/ui/Assets/Languages/English.json
+++ b/src/ui/Assets/Languages/English.json
@@ -612,6 +612,7 @@
     "top": "Top",
     "totalAdjustmentX": "Total adjustment: {0}",
     "translate": "Translate",
+    "translatePercentX": "Translating: {0}%",
     "translateRow": "Translate row",
     "translation": "Translation",
     "twoLetterLanguageCode": "Two-letter language code",
--- a/src/ui/Features/Tools/BatchConvert/BatchConverter.cs
+++ b/src/ui/Features/Tools/BatchConvert/BatchConverter.cs
@@ -350,6 +350,17 @@ public class BatchConverter : IBatchConverter, IFixCallbacks
             {
                 await RunOcrTesseract(imageSubtitle, item, cancellationToken);
             }
+
+            // OCR is only one step of the run - the item still goes through the convert functions
+            // and the save before it can say "Converted". Leaving the last progress value up would
+            // show a finished-looking "OCR: 100%" for that whole stretch, so put the row back to
+            // the plain working status it started this block with. A runner that deliberately left
+            // a terminal status behind (cancelled, an error, "model likely wrong") keeps it - only
+            // our own percentages are reset.
+            if (IsOcrProgressStatus(item.Status))
+            {
+                item.Status = Se.Language.General.OcrDotDotDot;
+            }
         }
 
         // Run convert functions (remove formatting, etc.)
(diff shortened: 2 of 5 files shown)


emsb8888-prog:
明白了，谢谢

Issue #13619, opened 2026-08-14
Loosing focus after splitting a row


radektuma:
Hi,
if I split text in a row, the application loses the focus. See the image - I edited a row nubmer 522, after that I stayed at the editation field of that row, but I saw a row 505 above it.

<img width="962" height="348" alt="Image" src="https://github.com/user-attachments/assets/d1b0b33b-7d0f-4f57-8421-34222ebd1b12" />

<img width="951" height="343" alt="Image" src="https://github.com/user-attachments/assets/8c39ccaa-b23e-4c82-99c3-c46a32c9ed0d" />

niksedk:
How did you split? SE can do this in a number of different ways?

radektuma:
There was a long row in the one line, so I use Enter to split it.

original state of the row 552 (I rather use English):  
I was in his mind, Rhinannon. I know him.

new state of the row 552: 
I was in his mind, Rhinannon.
I know him.

Issue #13579, opened 2026-08-13
v5.2.0-beta11 Regression in list scrolling and auto-break behavior (compared to v5.1.0)


St0rmXtr00per:
<img width="800" height="419" alt="Image" src="https://github.com/user-attachments/assets/702d0a2e-a3de-4b7b-89dc-5057214399a3" />

1. I'm scrolling through the video with the mouse wheel from bottom to top. First time at normal speed, second time sped up. You can see jittery movements of the scrollbar thumb.

Additional info: In the previous version, this wasn't an issue. However, I encountered two related issues in 5.1.0: either the subtitle list wouldn't scroll with the scrollbar thumb, or the subtitle list would get completely stuck, in which case you would need to click on the scrollbar thumb and scroll manually.

2. In the second part of the gif, you can see me quickly finishing the editing and pressing Ctrl + R for auto-break. The list view immediately jumps to a random position (you can no longer see the current line). The current line comes back into view when I continue playing the video because I have "Subtitle grid, center when selecting prev/next row" turned on.

All other applications that could influence SE like Ditto or Autohotkey were completely closed. After rolling back to v5.1.0 I turned back to the older bugs (they are more manageable)

GrampaWildWilly:
For what it's worth, all the scroll bar functions I tried work fine in beta 12.  But I don't have a wheel on my mouse.  The wheel seems to be a separate issue.

PR #13756, opened 2026-08-17
OCR window: index-mapped scrollbar for the subtitle grid

niksedk:
The OCR subtitle grid had the same jumping scrollbar thumb as the main subtitle grid (#13579): Avalonia's `VirtualizingStackPanel` estimates the total pixel extent as *realized rows + remaining × average realized height*, so with variable-height rows the extent — and the native thumb — shifts on every wheel tick. The OCR grid is the worse case: every row holds a subtitle bitmap, so heights vary far more than the main grid's one-vs-two text lines.

## Change

`MakeSubtitleView` now wraps the grid in `TableViewIndexScrollBar`, the same control the main grid already uses. It hides the native pixel-mapped vertical bar and docks a standalone one mapped to row indices:

- `Value` = index of the first visible row (+ the fraction scrolled into it)
- `Maximum` = row count − fully visible rows

Wheel, keyboard and the OCR loop's own `ScrollIntoView` still move the pixel offset; `ScrollChanged` re-derives the thumb. Trough press-and-hold (#12894) and shift+click jump come along with the control.

## Verification

New test `OcrWindow_UsesIndexMappedScrollBar` asserts, on a 500-item OCR window:

- the native vertical bar is hidden and the index bar is the visible one
- `Maximum` is in row units (400–499), while the pixel extent is many times larger
- setting `Value = 250` places row 250 at the viewport top
- scrolling the view down never moves the thumb backwards

779 Ocr/TableView UI tests pass; build clean.

## Not included

The main grid also gets `TableViewScrollAnchor.Attach` (#13619) to hold the view steady when a row changes height. The OCR grid's rows do change height as text lands during OCR, but the OCR loop already scrolls to the current line itself, so an anchor could fight it — left for a separate look.

### DIFF
--- a/src/ui/Features/Ocr/OcrWindow.cs
+++ b/src/ui/Features/Ocr/OcrWindow.cs
@@ -575,6 +575,13 @@ public class OcrWindow : Window
                 },
         });
 
+        // Index-mapped scrollbar (#13579), like the main subtitle grid: the native bar is
+        // pixel-mapped and the virtualizing panel estimates its extent from the average
+        // realized row height, so the thumb jumps around while scrolling. Rows here vary
+        // even more than in the main grid - each holds a subtitle bitmap - so hide the
+        // native vertical bar and dock a row-index one beside the grid instead.
+        var scrollBarHost = new TableViewIndexScrollBar(dataGridSubtitle);
+
         // The image thumbnails scale with Ctrl+plus/minus (Image.MaxWidth/MaxHeight are
         // bound to the VM) - keep the pixel-sized image column in step with the zoom.
         var imageColumn = dataGridSubtitle.Columns[dataGridSubtitle.Columns.Count - 2];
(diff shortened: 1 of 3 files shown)


niksedk:
Follow-up: checked how the image zoom (Ctrl+plus/minus) interacts with the index bar.

**The bar itself is fine** — a headless probe on 500 rows shows `Maximum`, `ViewportSize` and `Value` all resync when the rows are re-measured, and `Value` keeps tracking the actual first visible row.

**But the view jumped rows on zoom**, because the ScrollViewer preserves its *pixel* offset across the re-measure (pre-existing, the native bar did the same):

```
at row 250:      value=250.00 max=495 first=250  offsetY=14000 extent=28000 rowH=56
zoom in x5:      value=172.84 max=497 first=172  offsetY=14000 extent=40500 rowH=81   <- 78 rows back
zoom out x10:    value=492.00 max=492 first=491  offsetY=19165 extent=19500 rowH=39   <- pinned to the end
```

Since the bar now speaks row indices, this is cheap to fix: `TableViewIndexScrollBar.PreserveTopRow()` captures the top row before the change and re-places it once the new heights are measured (repeated calls while one is pending keep the first row, so holding the zoom key doesn't drift). `MakeSubtitleView` calls it from the existing `ImageMaxWidth` handler. Same probe after the fix:

```
zoom in x5:      value=250.00 max=496 first=250  offsetY=20250 extent=40500 rowH=81
zoom out x10:    value=400.00 max=492 first=400  offsetY=15600 extent=19500 rowH=39
```

New test `OcrWindow_ZoomKeepsTheRowAtTheTop` zooms in five notches and back out ten, asserting the row heights really change and row 250 stays at the top; it fails without the fix. 780 Ocr/TableView tests pass.

PR #13760, opened 2026-08-17
Trim embedded fonts to used characters (attachments, font collector, batch convert)

niksedk:
Implements font subsetting for embedded ASSA fonts, as hoped for in discussion #13740 - there, four embedded fonts turned a 144 KB subtitle into 28.8 MB.

## Approach

`FontTrimmer` does **glyph stripping** rather than full re-subsetting: it shapes the subtitle's visible text with HarfBuzz (per-script runs, so Arabic/Indic contextual forms and ligatures resolve to exactly the glyph IDs libass would use), adds each character's nominal cmap glyph and all composite-glyph components, then rewrites only `glyf`/`loca` with the unused outlines emptied. Every other table - `cmap`, `GSUB`/`GPOS`, `name`, `hmtx` - stays byte-identical, so glyph IDs, shaping, kerning and the family name keep working unchanged in libass and VSFilter. `DSIG` is dropped (invalid after modification). CFF-flavored OpenType and `.ttc` collections are skipped with a reason; the original bytes are kept on any failure.

**Real-world result:** Arial Unicode with two subtitle lines: 23.3 MB → 864 KB (30 of 50,377 glyphs kept).

## UI

- **ASSA attachments window**: new *Trim fonts to used characters...* button - confirms, trims all embedded fonts, reports per-font before → after sizes (or the skip reason) and updates the rows.
- **Font collector**: *Only embed used characters (trim fonts)* checkbox (persisted); the embed confirmation then already shows the trimmed sizes.
- **Batch convert**: same checkbox on the *Embed fonts* function, persisted and carried in the profile config.

The prompts warn that trimming is a final step - text added later may show missing characters.

## Verification

- New fixture font (built from scratch with fonttools, committed, 940 bytes) with a GSUB `liga` ligature and a composite glyph: tests assert the ligature output glyph and composite components survive, unused glyphs lose their ink but stay cmap-mapped, shaping (IDs + advances) is identical before/after, Skia parses the result, double-trim reports *no savings*, CFF/ttc/garbage are skipped.
- A system-font test trims a real installed font and asserts identical shaping (skips silently when no suitable font exists).
- Trimmed Arial/Arial Unicode/Times New Roman validated externally with fonttools strict checksum load + full table parse.
- Full UI test suite: 3008 passed.


mtsubs:
This feature is amazing! As a member of the CJK subtitle community, I also developed a standalone tool to implement it early on, along with some specifications, which you can see here.

Many thanks to SE for also supporting it. For a long time, this feature required either command-line access or installing separate software. Now it's integrated into subtitle software that's readily available to every subtitler, this is the right direction!!

My project and related specifications may be helpful for your continued development:
https://github.com/MontageSubs/ass-subset
https://subs.js.org/ass-subset/

Issue #13621, opened 2026-08-14
Transcribe audio > OpenRouter > Generated audio file not found


slothsh:
# Issue

When trying to transcribe a snippet of audio from the timeline via the OpenRouter API, the tool fails silently. The tool logs report that no output file could be found due to an ffmpeg error (exit code 234) and error message:

```txt
Stream map '' matches no streams.
To ignore this, add a trailing '?' to the map.
Failed to set value '0:1' for option 'map': Invalid argument
```

___

# Logs

```txt
Date: 08/14/2026 16:50:39
SE: v5.2.0-beta13 - Unix 15.7.9 - 64-bit
Generated audio file not found: /var/folders/3p/zz9849956xj5k6rf_y8z7_dw0000gn/T/se-stt-97abaf8b-6291-4678-92d6-67b4361c6314/79f9a59d-b3bf-4035-a46f-555c703f4ca3.mp3
ffmpeg: /Applications/Subtitle Edit.app/Contents/MacOS/ffmpeg
Parameters: -i "/var/folders/3p/zz9849956xj5k6rf_y8z7_dw0000gn/T/se_audioclip_1952f183-8cc5-4b6e-a5c4-3e7c65a1c79c.wav" -vn -ar 16000 -ac 1 -af volume=1.75 -c:a libmp3lame -b:a 32k -f mp3 -map 0:1 "/var/folders/3p/zz9849956xj5k6rf_y8z7_dw0000gn/T/se-stt-97abaf8b-6291-4678-92d6-67b4361c6314/79f9a59d-b3bf-4035-a46f-555c703f4ca3.mp3"
OS: Unix 15.7.9
64-bit: True
ffmpeg exit code: 234
ffmpeg log: ffmpeg version 8.0 Copyright (c) 2000-2025 the FFmpeg developers
  built with Apple clang version 13.0.0 (clang-1300.0.29.30)
[...]

PR #13784, opened 2026-08-17
Fix #13781: STT aborts on batch videos whose audio is not the picked stream

### LINKED ISSUE #13781: Speech to text fails with CrispASR v0.8.29 (cohere-transcribe.gguf) in beta 16
Hi Nikse,
I've encountered an issue in beta 16. When selecting CrispASR v0.8.29 with the cohere-transcribe.gguf model for Speech to Text, it fails to work. Every time I click convert, it waits for over 10 seconds and then aborts with an error.
However, everything works perfectly fine if I switch back to CrispASR v0.8.28.
Could you please help check if this is a bug introduced in CrispASR v0.8.29, or if it's caused by the recent updates in beta 16?
Thanks for your hard work!

niksedk:
Fixes #13781. The user's log has the whole story — the failure is in the **ffmpeg extraction step**, before CrispASR (or any engine) ever runs:

```
Parameters: -i "G:\1.mp4" ... -map 0:1? "...\bb2ae777....wav"
Stream #0:0: Audio: aac ...
Stream #0:1: Video: h264 ...
[out#0/wav] Output file does not contain any stream
ffmpeg exit code: -22
Generated audio file not found: ...
```

## Root cause

`G:\1.mp4` has its streams ordered **audio (0:0) then video (0:1)** — ffmpeg lists streams in container order, and plenty of muxers put audio first. SE remembered stream index 1 as the audio track picked in the main window (correct for *that* video, where 0:1 was the audio) and applied `-map 0:1?` to **every video in the batch**. On this file that selects the H.264 stream, `-vn` drops it, ffmpeg has nothing to write and exits -22, and the run aborts with "Generated audio file not found".

The trailing `?` (added for #13621) only helps when stream N is **missing** — here stream 1 exists, it's just the wrong kind. Reproduced exactly with a synthesized audio-first MP4: `-map 0:1?` → "Output file does not contain any stream"; without the map, ffmpeg auto-selects the audio and the extraction succeeds.

It also explains "some files work fine, but others still abort" on both CrispASR versions — it depends on each file's stream order, not the engine. And it's not new in beta 16; the pattern predates it.

## Fix

The stream index is now applied **only to the file the track was actually picked from** (tracked alongside the index; paths compared case-insensitively for Windows). Every other input — other batch videos, the demuxed `se_audioclip_*.wav` clips from "transcribe selected lines" — gets no `-map`, so ffmpeg's automatic selection picks the best audio stream. The single-video flow with a deliberately chosen track is unchanged.

The mapping decision is extracted into `BuildAudioMapParameter` so it's testable.

## About the CrispASR v0.8.29 half of the report

I A/B'd v0.8.28 vs v0.8.29 on the cohere backend with SE's exact command line — `cohere-transcribe-q4_k.gguf` *and* the full 4.1 GB `cohere-transcribe.gguf`, Metal and `--no-gpu` — all four pairs exit 0 with **byte-identical SRT**. No flags were removed in v0.8.29 and the `cohere` backend listing is unchanged. So I found no engine regression to fix; the reporter's v0.8.29-vs-v0.8.28 difference may need their crispasr log if it persists after this fix.

## Testing

- 6 new tests on `BuildAudioMapParameter`: maps for the owning video (incl. case-differing paths, track 0), empty for another batch file / a demuxed clip / no track picked.
- End-to-end ffmpeg verification of both the failing and fixed command on an audio-first MP4.
- Build clean; full UI suite 3090 passed, 0 failed.

### DIFF
--- a/src/ui/Assets/Languages/English.json
+++ b/src/ui/Assets/Languages/English.json
@@ -906,7 +906,8 @@
       "ssaStylesHint": "Sub Station Alpha styles",
       "webVttStylesHint": "WebVTT style manager",
       "ssaPropertiesHint": "Sub Station Alpha properties",
-      "ssaAttachmentsHint": "Sub Station Alpha attachments"
+      "ssaAttachmentsHint": "Sub Station Alpha attachments",
+      "getFrameRateFromVideoFileHint": "Get frame rate from video file"
     },
     "waveform": {
       "playPauseHint": "Play / Pause {0}",
--- a/src/ui/Features/Video/SpeechToText/SpeechToTextViewModel.cs
+++ b/src/ui/Features/Video/SpeechToText/SpeechToTextViewModel.cs
@@ -149,6 +149,10 @@ public partial class SpeechToTextViewModel : ObservableObject
     private string? _videoFileName;
     private string _audioFileName = string.Empty;
     private int _audioTrackNumber;
+
+    // The file _audioTrackNumber was picked from. A stream index only means anything in its own
+    // file, and batch mode reuses this view model for other videos - see GetFfmpegProcess.
+    private string? _audioTrackVideoFileName;
     private readonly List<string> _filesToDelete = new();
     private string? _sttTempFolder;
     private readonly ConcurrentQueue<string> _outputText = new();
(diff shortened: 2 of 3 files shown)


GrampaWildWilly:
Wouldn't a -map 0:a just avoid the whole business of the order of tracks?  Or is that what you did?  I suppose that would work only for a media file that does not have multiple audio tracks, but at least in the degenerate case, wouldn't 0:a avoid the question?

Issue #13738, opened 2026-08-16
Transcription is worse in Subtitle Edit than calling the same tools/models standalone - configurable wav extraction


AmineI:
**Description:**

When I use Faster-Whisper standalone (or crispASR) standalone through cli, passing the original .mp4, the transcription is accurate.
When I use the same model and parameters through Subtitle Edit, it produces more word mistakes.
I noticed that the only difference is the source files : SubtitleEdit passes preprocessed.wav files to the ASR engine. Their quality seems quite degraded compared to the original audio, causing the issues.

I tried to review Subtitle Edit ffmpeg wav extraction command, and compared it through some testing. In my case, converting the audio with the below commands before inputting them to faster-whisper produced accurate transcriptions:

```powershell
ffmpeg.exe -i ".\video.mp4" -map 0:a:0 -vn -ac 1 -af "aresample=16000:resampler=soxr:precision=28" -c:a pcm_f32le ".\extracted_mono_32.wav"
```

and:

```powershell
ffmpeg.exe -i ".\video.mp4" -map 0:a:0 -vn -ac 1 -af "volume=1.75,aresample=16000:resampler=soxr:precision=28" -c:a pcm_f32le ".\extracted_mono_32_boosted.wav"
```

The following versions produced wrong words:

- Stereo output
- `pcm_s16le` output
- Subtitle Edit’s current extraction command

Subtitle Edit currently appears to use approximately:

```powershell
ffmpeg.exe -i ".\video.mp4" `
  -vn `
  -ar 16000 `
  -ac 1 `
  -ab 32k `
  -af "volume=1.75" `
  -f wav `
  ".\extracted.wav"
```

The important differences are:

1. Subtitle Edit does not explicitly use `pcm_f32le`.
2. Subtitle Edit does not explicitly use the high-quality SoX resampler:
   ```text
   aresample=16000:resampler=soxr:precision=28
   ```
3. Explicitly selecting the audio stream may be important:
   ```text
   -map 0:a:0
   ```
4. The volume boost is optional in my testing. With background noise, could the boosted version could cause more issues ?
5. `pcm_s16le` produced worse recognition, while `pcm_f32le` was more accurate in my tests.

I'm not too familiar with all ffmpeg parameters, so I'm sure there could even be some better settings

Could Subtitle Edit’s Faster-Whisper audio extraction parameters be changed or made configurable ? 
For my scenario for example, using the parameters mentione above is more efficient.
```text
[...]

Issue #13233, opened 2026-08-05
Select audio track.


fraternl:
By default, it seems to select the last audiotrack.
I prefer it to select the first audio track of my preferred language.

<img width="1007" height="280" alt="Image" src="https://github.com/user-attachments/assets/53a276ff-a655-40c1-8768-919651a8c531" />

PR #13786, opened 2026-08-17
Only send single-audio-track files directly to XXL and CTranslate2

niksedk:
Follow-up to #13784, answering "how do Purfview XXL and CTranslate2 handle multi-track files — should we give them a WAV when there is more than one track?"

**Yes.** What each side actually does with a multi-track input:

| path | which track gets transcribed |
|---|---|
| Purfview XXL, given the source file | `--ff_track` — "1 - selects the first audio track. (default: 1)" → always the **first** audio track |
| CTranslate2, given the source file | faster-whisper's PyAV decode is hardcoded to the first audio stream — **no selector exists** |
| extracted WAV, no `-map` | ffmpeg automatic selection → the container's **default-flagged** track (verified both ways on a synthesized two-track file: flag on 0:1 → picks 0:1, flag on 0:2 → picks 0:2) |
| extracted WAV, `-map 0:N` | exactly the track the user picked |

mpv honors the default flag too, so the WAV path always matches the track SE plays and draws the waveform for — the direct path matches it only when the first track happens to be the default. On any multi-track input (video or an audio-only `.mka`/`.m4a`, which the old gate never considered), the source-file shortcut could silently transcribe a different track than the one on screen.

## Change

`CanSendSourceFileToEngine` now sends the source file only when it has **exactly one audio track**. Everything else goes through the extracted WAV. This replaces the old picked-track==first-audio-track comparison, which had two soft spots: in batch mode it compared a stale stream index from the main window's video against unrelated files, and even when it passed, "first audio track" only coincidentally matched what the user was hearing.

Single-audio-track files — the overwhelmingly common case, and the whole point of #13738's direct-source change — still skip the WAV.

## Testing

- Engine behavior established from the shipped CLI help (`--ff_track`, [PurfviewFasterWhisperXXL.txt:347](../blob/main/src/ui/Assets/SpeechToText/PurfviewFasterWhisperXXL.txt#L347)) and faster-whisper's decode path; ffmpeg default-flag selection verified empirically in both orders.
- New `ParseLog` test pinning that a multi-track audio-only container reports 2 audio tracks (the input the gate consumes).
- Full UI suite 3090 passed, libuilogic 236 passed, 0 failures.

### DIFF
--- a/src/ui/Features/Video/SpeechToText/SpeechToTextViewModel.cs
+++ b/src/ui/Features/Video/SpeechToText/SpeechToTextViewModel.cs
@@ -3565,11 +3565,14 @@ public partial class SpeechToTextViewModel : ObservableObject
     }
 
     /// <summary>
-    /// True if the source file itself can go to the engine. Neither engine can pick an audio track,
-    /// so a track other than the first one is the one thing only the extracted WAV can express -
-    /// handing over the source file there would silently transcribe the wrong track. SE 4 made the
-    /// same call, though its track number was audio-relative and 0 meant "unspecified", while
-    /// SE 5 stores the global ffmpeg stream index of an always-auto-selected track.
+    /// True if the source file itself can go to the engine, which is only safe when there is
+    /// exactly one audio track: with several, the engines read the FIRST audio stream (Purfview
+    /// XXL's --ff_track defaults to 1, CTranslate2's PyAV decode is hardcoded to the first and has
+    /// no selector) - not the container's default track that mpv plays and the waveform shows, and
+    /// not a track the user picked. The extracted WAV expresses both: the picked track via -map on
+    /// the video it was picked from, and otherwise ffmpeg's automatic selection, which honors the
+    /// default-track flag like mpv does (verified both ways on a two-track file). SE 4 drew the
+    /// same line at "anything but the default first track goes through the WAV".
     /// </summary>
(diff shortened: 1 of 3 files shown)


niksedk:
Reviewed against #13787 (merged earlier today from a concurrent session), which had taken the opposite alignment: it pinned the batch/no-pick WAV fallback to `-map 0:a:0?` so the WAV path would match the first-in-container-order track the direct-reading engines pick. That was the wrong invariant — SE's track policy is the container's **default flag**, not container order: on a fresh open the main window deliberately follows mpv's own selection because the first track can be commentary or audio description (#13233), and the waveform sits on the same track. This PR's rationale (WAV path = the track the user hears) is the correct one, independently re-verified: whisper-ctranslate2 run on a synthesized two-track mp4 transcribed track 1 even with the default flag on track 2, and the bundled ffmpeg 7.1.1 followed the default flag in both orders.

Pushed two commits to make the branch coherent with current main: a merge of main resolving the conflict in `CanSendSourceFileToEngine`, and a revert of #13787's `-map 0:a:0?` fallback (back to empty → ffmpeg automatic selection) with the remarks and `SpeechToTextAudioMapTests` restored to the empty-map expectations, now documenting why the pin must not come back. UI audio-map tests 6/6 and libuilogic track tests 6/6 pass.

Issue #185, opened 2014-06-24
Find Replace


ivandrofly:
@niksedk

There is a problem with find replace **_regex**_ let say we have a text like:
[Text]
foo bar
[Regex pattern]
(foo) (bar)
[Replace with]
$2 $1
[Output]
$2 $1
[Correct output]
bar foo

PR #13787, opened 2026-08-17
STT: pin batch/fallback audio extraction to the first audio track

niksedk:
Follow-up to #13781/#13784, aligning which audio track every speech-to-text path transcribes on multi-track files.

**The gap:** the no-map fallback from #13781 left stream selection to ffmpeg's automatic choice, which prefers the **default-disposition** track — verified on the bundled ffmpeg 7.1.1, where a default-flagged stereo track beats a non-flagged 5.1. But the two engines that receive the source file directly take the **first audio track in container order**:

- **Purfview Faster-Whisper-XXL**: "It works only on the first audio track" per its author ([whisper-standalone-win #185](https://github.com/Purfview/whisper-standalone-win/discussions/185)); `--ff_track` exists since Apr 2024 but SE doesn't pass it.
- **whisper-ctranslate2**: faster-whisper's `decode_audio` hardcodes `container.decode(audio=0)`. Verified empirically — a two-track mp4 with different speech per track transcribed track 1 even with the default flag on track 2.

So on a file whose default-flagged audio is not the first track, the same batch would transcribe different tracks depending on the engine.

**The fix:** extraction now maps `0:a:0?` for every input the picked stream index does not belong to (batch videos, demuxed `se_audioclip_*.wav` clips, no-track-picked). The audio-relative form is valid for any stream layout, so the #13781 audio-before-video case keeps working — verified against the bundled ffmpeg on three fixtures: default-flag-on-second-track (now picks the first track, matching the engines), audio-first mp4 (no crash), and a plain wav clip. The trailing `?` keeps audio-less files failing the same way as before.

The picked-track path (`-map 0:{n}?` on the file it was picked from) and the source-file gate for XXL/CT2 are unchanged; all 6 `SpeechToTextAudioMapTests` updated and passing.

### DIFF
--- a/src/ui/Features/Video/SpeechToText/SpeechToTextViewModel.cs
+++ b/src/ui/Features/Video/SpeechToText/SpeechToTextViewModel.cs
@@ -3580,7 +3580,7 @@ public partial class SpeechToTextViewModel : ObservableObject
 
         if (_audioTrackNumber < 0)
         {
-            return true; // no track picked - the engine falls back to the first one, exactly like ffmpeg would
+            return true; // no track picked - the engine takes the first audio track, the same one the WAV extraction maps
         }
 
         try
@@ -4122,8 +4122,8 @@ public partial class SpeechToTextViewModel : ObservableObject
     }
 
     /// <summary>
-    /// The "-map" argument for extracting <paramref name="inputFileName"/>'s audio, or an empty
-    /// string to leave the choice to ffmpeg's automatic stream selection.
+    /// The "-map" argument for extracting <paramref name="inputFileName"/>'s audio: the picked
+    /// stream for the file it was picked from, the first audio track for everything else.
     /// </summary>
     /// <remarks>
     /// A stream index only addresses a stream in the file it was read from, so it is applied to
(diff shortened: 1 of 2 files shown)


GrampaWildWilly:
Looking in the help embedded in the SE Speech to Text dialog for PFW-XXL, I see there is a parameter ff_track that takes values from 1 to 6.  



You have always been able to specify this value in the SE Speech to Text Advanced Settings.  So even if there are multiple audio tracks in the media file, it might not be right to just blanket decide to send to PFW-XXL an extracted WAV when there are multiple audio tracks.  PFW-XXL seems to stop at 6 audio tracks.  That seems like a lot to me but then I've never pretended to be a pro at this.  Still, always generating that WAV file doesn't feel right to me for PFW-XXL even for multi-audio files.  It's the "always" part I take issue with.  If the user requests one of the first 6 audio tracks, you should have SE code ff_track for the user & not do the extract of the WAV.  If the user requests any of audio track 7 or higher, then go ahead & do the relevant extract of a WAV.  I suppose the number 6 is subject to change.  Nobody knows if Purfview might change that in the future.  But it's 6 today.  Also, this number is not 0-based.  So there is probably some translating from what ffprobe reports & what PFW-XXL needs.  But it's only a +1 shift.  I think everybody can deal with that.  Also, I have seen MP4 files with about 8 embedded VTT tracks, & the first audio track is way past 0:1.  The embedded VTT tracks occupied the low stream identifiers.  So the first audio track might be reported in ffprobe as high as 0:16.  I have personally observed this.  So I say this needs some further thought.

Just to make things concrete, here's a real world video I have seen.

[ffprobe.txt](https://github.com/user-attachments/files/31154114/ffprobe.txt)

If you download anything other than Program 0 from this HLS master manifest, your audio file will be the 8th track, behind the 7 VTT tracks, and incidentally in front of the video track.  So despite there being lots of tracks here, there would be no reason to generate the WAV file.  Actually, this wouldn't really exercise the multiple audio track logic.  There is only one audio track in each rendition.  But if for some reason you wanted all the subtitle tracks, the audio track would be neither 0 nor 1, & it would be in front of the video.  I wonder if PFW-XXL could handle it.

Not that it's particularly relevant, but I got around all issues by running ffmpeg 3 separate times to get this: once for the video track I wanted, once for the audio track I wanted, & once for the VTT I wanted.  So I have 2 media files consisting of only one track & a text file containing the captions.  But I've never even thought about handing any of this to PFW-XXL.

Issue #13824, opened 2026-08-18
Feature Request: Add option to hide ASS/SSA override tags in the subtitle grid


MrGoraj:
When working with ASS/SSA subtitles, the inline override tags (e.g., {\an8\fnCourier New\b1\fs40...}) often clutter the subtitle list view, making it difficult to focus on the actual dialogue text that needs translation.

### Example:

Currently, a subtitle line appears as:

```
{\an8\fnCourier New\b1\bord5\fs40\t(\fs28)\shad0\c&H1CFEFC&\3c&H8B04E0&\t(2275,2276,\alpha&HFF&)\t(2850,2851,\alpha&H00&)\t(4280,4281,\alpha&HFF&)\t(5144,5145,\alpha&H00&)\t(9481,9482,\c&HE8FEDF&\3c&HF171F4&)\t(9650,9651,\c&H1CFEFC&\3c&H8B04E0&)\t(9900,9901,\c&HE8FEDF&\3c&HF171F4&)}YOUNG LADIES DON'T PLAY 
FIGHTING GAMES"
```

What the translator actually needs to see is simply:

```
YOUNG LADIES DON'T PLAY 
FIGHTING GAMES
```

The styling tags should remain untouched when saving the file — they only need to be **visually hidden** in the editor interface.

### Why this is needed

1.  **Masking/effect lines** — Some ASS subtitles use hundreds of lines containing only override tags (e.g., drawing masks, complex karaoke effects, animated shapes) with no actual dialogue text. These lines have no translatable content but clutter the grid, making it hard to locate the real text lines that need translation.
2.  **Translation workflow** — Translators working from existing subtitle files waste significant time scrolling through dozens of tag-heavy lines to find the actual dialogue. A cleaner view would dramatically improve efficiency.
3. **AI/MT translation** — When using machine translation or AI tools, feeding the raw text without override tags reduces token usage, minimizes hallucinations (since the AI doesn't have to process irrelevant formatting codes), and provides cleaner source text for better translation quality. This is a separate but related benefit.

### Proposed solution:

Add a toggle option (e.g., checkbox or menu item) to:

1. **Hide ASS/SSA override tags** from the subtitle grid/text view, showing only the plain dialogue text
2. Optionally, **completely hide lines** that contain only override tags (no text), so the translator can instantly see how many lines actually require translation
3. The tags should remain intact in the file when saving — this is purely a visual/hiding feature for the UI

### Benefits

[...]

PR #13828, opened 2026-08-18
Show the text of tag-heavy lines in the grid's "show formatting" mode

niksedk:
Follows from #13824.

The option that request asks for already exists — *Options → Settings → Appearance → "Show formatted (HTML/ASSA) text in subtitle grid" → **Show formatting*** — and it hides override tags while leaving them in the file. It just does not work on the lines the report is about, which is why it reads as missing.

### What happens

`Convert` cut the raw string at 197 characters before any tag was looked at:

```csharp
if (str.Length > 200)
{
    str = str.Substring(0, 197).TrimEnd() + "...";
}
```

The reported line is 321 characters, **283 of them a single `{\t(...)}` block** — so the closing `}` sits at index 282, inside the part that is thrown away. `MakeShowFormatting` then meets `{\`, searches for the brace, gets `tagEnd == -1`, and falls into:

```csharp
// Malformed or overly long tag - treat as regular text
```

so the markup is printed character by character. The mode whose whole job is hiding tags displays 197 characters of tag and none of the dialogue. Verified by running the new test against the current code:

```
Expected: "YOUNG LADIES DON'T PLAY"
Actual:   "{\an8\fnCourier New\b1\bord5\fs40\t(\fs28)\shad0\c"···
```

A second path reaches the same place: the `tagEnd - i < 500` cap sends any single tag block of 500+ characters down that same branch, which karaoke and effect lines exceed routinely.

### Change

- In *show formatting*, the cap applies to the **visible** text rather than the raw string, so hidden tags no longer consume a line's 200 characters. The raw string is bounded at 5000 purely against pathological input.
- The length of a tag block now decides **how much of it is parsed** into formatting state, never **whether it counts as a tag**. An over-long block is skipped like any other — printing braces as text is the one outcome a tag-hiding mode must not produce.
- The other two grid modes (*No formatting*, *Show tags*) keep the 200-character cut exactly as before.

### Testing

Three tests: the reported line renders as its dialogue; a 3000-character tag block is still hidden; a long line of plain dialogue still truncates with an ellipsis at 200 visible characters. `ValueConverterTests` 66 passed, full UI suite 3171 passed.

Point 2 of the issue — hiding lines that consist only of override tags — is not addressed here; that is a grid filter, not a rendering question.

### DIFF
--- a/src/ui/Logic/ValueConverters/TextWithSubtitleSyntaxHighlightingConverter.cs
+++ b/src/ui/Logic/ValueConverters/TextWithSubtitleSyntaxHighlightingConverter.cs
@@ -19,6 +19,19 @@ public class TextWithSubtitleSyntaxHighlightingConverter : IValueConverter
 {
     private ISpellCheckManager? _spellCheckManager;
 
+    /// <summary>Characters of a line's own text a grid cell ever shows - the rest is an ellipsis.</summary>
+    private const int MaxVisibleLength = 200;
+
+    /// <summary>
+    /// How much of a line is parsed at all in "show formatting" mode. Tags cost nothing on screen
+    /// but still have to be walked, so this only bounds the pathological case; a line of a few
+    /// hundred characters of override tags is ordinary in karaoke and effect files.
+    /// </summary>
+    private const int MaxRawLength = 5000;
+
+    /// <summary>Longest override-tag block whose contents are still parsed into formatting state.</summary>
+    private const int MaxParsedTagLength = 2000;
+
     // Pre-compiled <font> attribute patterns (reused across every grid-row render)
     private static readonly TimeSpan RegexTimeout = TimeSpan.FromMilliseconds(100);
     private static readonly Regex FontColorRegex = new(@"color\s*=\s*[""']?([^""'\s>]+)[""']?", RegexOptions.IgnoreCase | RegexOptions.Compiled, RegexTimeout);
@@ -116,17 +129,29 @@ public class TextWithSubtitleSyntaxHighlightingConverter : IValueConverter
             return new InlineCollection();
         }
 
-        // Truncate long strings for performance
-        if (str.Length > 200)
-        {
-            str = str.Substring(0, 197).TrimEnd() + "...";
-        }
-
         var formattingType = Se.Settings.Appearance.SubtitleGridFormattingType;
+
+        // "Show formatting" hides the ASSA/HTML tags and renders what they mean, so the length that
+        // matters is the dialogue, not the markup around it. Truncating the raw string first threw
+        // away the text of exactly the lines this mode is for: a karaoke or effect line carrying a
+        // few hundred characters of {\t(...)} became 197 characters of tag and no words at all
+        // (issue #13824). Only pathological lines are cut here, and the visible text is capped
+        // inside MakeShowFormatting instead.
         if (formattingType == (int)SubtitleGridFormattingTypes.ShowFormatting)
         {
+            if (str.Length > MaxRawLength)
+            {
+                str = str.Substring(0, MaxRawLength);
+            }
+
             var lines = MakeShowFormatting(str);
             return SpellCheckLines(lines);
+        }
+
+        // Truncate long strings for performance
+        if (str.Length > MaxVisibleLength)
+        {
+            str = str.Substring(0, MaxVisibleLength - 3).TrimEnd() + "...";
         }
 
         if (formattingType == (int)SubtitleGridFormattingTypes.ShowTags)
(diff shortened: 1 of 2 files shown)


Triathlon-rally:
Hi,

I have now tested the merged Teletext/EBU STL implementation in Subtitle Edit 5.2.0-beta19 on Windows in a real broadcast subtitle workflow.

First of all: the new Teletext functionality works well.

I also tested a subtitle file that was not EBU STL/Teletext compliant, corrected the subtitles and Teletext positions in Subtitle Edit, and exported the result as EBU STL. This workflow worked successfully.

While doing this practical test, I found a few additional points that are especially relevant when preparing non-compliant subtitle files for EBU STL/Teletext output.


## 1. Line length correction should optionally affect only lines that are actually too long

For our Teletext workflow the limits are normally:

- 37 visible characters without an explicit color code
- 36 visible characters when a Teletext color code is used
- maximum 2 text lines

I tested "Split/break long lines" with:

- Maximum single line length: 37
- Maximum number of lines: 2

I found a subtitle containing a single line of 40 characters.

The main editor correctly shows:

Line length: 40

However, the subtitle is not listed as something that needs to be corrected in the "Split/break long lines" window.

lange Texte trennen


If I additionally enable "Rebalance long lines", Subtitle Edit does correct this subtitle.

However, it then also rebalances many other subtitles whose individual lines are already within the 37-character limit.

For broadcast subtitles this is not always desirable.

An intentionally unbalanced subtitle can be editorially correct. For example, there may be a reason to have four words on the first line and only one word on the second line.

If neither line exceeds the configured maximum, this existing line break should be preserved.

It would therefore be very useful to have two modes/options:

- Rebalance all subtitles
- Rebalance only subtitles containing a line longer than the configured maximum

For the second option, existing line breaks should remain untouched when all lines already comply with the configured maximum length.


## 2. An overlong two-line subtitle should be split into two subtitle events

There is another important case when converting existing subtitle files to Teletext.

A subtitle may already contain two text lines, but still contain too much text to fit into the configured maximum of two lines with 37 (or 36) characters per line.

In this situation, rebalancing cannot solve the problem because no valid two-line arrangement is possible.

Ideally Subtitle Edit should then split the original subtitle into two new subtitle events.

For example:

Original subtitle:

TC IN  = 10:00:00:00
TC OUT = 10:00:08:00

Line 1: ...
Line 2: ...

If the text cannot be fitted into two valid Teletext lines, Subtitle Edit could create:

Subtitle 1:
TC IN  = original TC IN
TC OUT = calculated split point

Subtitle 2:
TC IN  = Subtitle 1 TC OUT + configured minimum gap
TC OUT = original TC OUT

The important timing rules would be:

- The first new subtitle keeps the original TC IN.
- The second new subtitle keeps the original TC OUT.
- The configured minimum gap between subtitles is respected.
- The remaining available duration is divided between the two new subtitles.
- The duration should preferably be distributed proportionally according to the amount of text in each new subtitle.
- Both resulting subtitles should respect the configured maximum number of lines and maximum line length.

For example:

available duration =
original duration - configured minimum gap

duration subtitle 1 =
available duration × text length subtitle 1 / total text length

duration subtitle 2 =
available duration - duration subtitle 1

This keeps the complete original time range intact while creating two valid subtitle events.

This would be especially useful when normalizing subtitle files from other formats before exporting them as EBU STL/Teletext.


## 3. "No color" option in the color picker

I also found a small but important issue with colors in the Teletext workflow.

For Teletext, default white text without an explicit color code is different from explicitly assigning white as a color.

When I select white in Subtitle Edit's color picker, Subtitle Edit adds:

<font color="#FFFFFF">Text</font>



code coleur blanc



For our Teletext limits this distinction matters:

- No explicit color code -> maximum 37 visible characters
- Explicit Teletext color code -> maximum 36 visible characters

At the moment I can remove the color via:

Right click
-> Remove formatting
-> Remove color

This works correctly and removes the explicit color formatting.


kontextmenue



However, this is somewhat inconvenient during normal subtitle editing.

It would be useful if the existing color picker also had a "No color" option.

Selecting "No color" should not mean black or transparent. It should remove the explicit color formatting, equivalent to:

Remove formatting -> Remove color

This would allow the user to easily return a subtitle to default Teletext white without generating an unnecessary color code.

The attached mock-up shows the idea:

model keine Farbe aktiv

model keine Farbe


## 4. "Fix common errors..." does not open the expected dialog

"Fix common errors..." is also very useful when preparing an existing subtitle file for EBU STL/Teletext.

In Subtitle Edit 4.x I could use:

Tools -> Fix common errors...

and a dialog opened where I could select the individual correction rules before applying them.

In my Windows test with Subtitle Edit 5.2.0-beta19, clicking:

Tools -> Fix common errors...

does not open this dialog.

fix common erreors



As far as I understand, the current workflow should still open the Fix Common Errors dialog where the individual rules can be enabled or disabled.

Therefore this may be a beta19 regression or another issue worth checking.


## Result of the Teletext/EBU STL test

Apart from the points above, the new Teletext implementation itself worked successfully in my tests.

I tested:

- EBU STL import
- Teletext line positions
- horizontal Teletext alignment
- editing individual and multiple subtitles
- conversion of a non-compliant subtitle file
- normalization for Teletext
- export back to EBU STL with Teletext positions

The resulting EBU STL/Teletext workflow worked correctly.

Thanks again for integrating the Teletext changes.

PR #13860, opened 2026-08-19
Update Italia GUI language and installer script

bovirus:
@niksedk 

About instalelr the change is necessary to remove from uinstall program list version after program name (version is already in Version column).

image


bovirus:
@niksedk 

I downloaded program version 5.2.0 beta 18 but when I ran it show 5.2.0 beta 17.

Issue #11476, opened 2026-06-07
SE4 style line breaking in SE5


AuroraMartell:
In SE4, we had the option to automatically break/split long lines (I would attach a screenshot, but I'm on Mac at the moment). This is not available in SE5. Do you have any plans to re-add it, @niksedk?

niksedk:
We have Tools - Split/re-balance long lines

AuroraMartell:
> We have Tools - Split/re-balance long lines

I think it's difficult to use.

xuehao:
I have the same problem with long line. I don't want to break it into multi lines, but I'd like the text widget supports line wrap so that I can see the whole line and no need to use the horizontal scroll bar.

<img width="598" height="145" alt="Image" src="https://github.com/user-attachments/assets/f06e79e8-2388-4d22-b709-eba1ac846650" />

Triathlon-rally:
I can confirm this after directly comparing Subtitle Edit 4.0.11 with Subtitle Edit 5 using the same real-world EBU STL workflow.

The old SE4 "Break/split long lines" behavior does something that the current SE5 "Split/rebalance long lines" workflow does not currently reproduce correctly.

### Example

I configured the maximum single-line length to 36 characters.

SE4 correctly identifies all subtitles containing lines longer than this limit.

More importantly, SE4 can handle an already two-line subtitle that contains too much text to fit within the configured two-line limit.

For example, the original subtitle contains:

Und, war ich in der Kantine?
Nein. Ich konnte Sie dort nicht finden.

SE4 automatically turns this into two subtitle events, for example:

Subtitle 1:
Und, war ich in der Kantine?

Subtitle 2:
Nein. Ich konnte
Sie dort nicht finden.

This is not just a line rebalance.

SE4 creates a new subtitle event when the text cannot fit into the allowed line structure. It preserves the outer timing boundaries of the original subtitle, distributes the available duration between the resulting subtitle events, and updates the subtitle numbering.

I have attached screenshots showing the behavior in SE4 and the corresponding SE5 workflow.

This is especially useful when a strict maximum line length and maximum number of lines must be respected. Rebalancing alone cannot solve a subtitle when there is simply too much text to fit within the permitted number of lines.

So this appears to be existing SE4 functionality that was lost or changed during the transition to SE5 rather than a completely new feature request.

Ideally, I think SE5 should distinguish between:

- Rebalance lines inside an existing subtitle event.
[...]

PR #13862, opened 2026-08-19
Add Teletext alignment and EBU STL editing improvements

Triathlon-rally:
## Summary

Adds a dedicated Teletext alignment workflow for broadcast subtitle editing, focused on EBU STL/Teletext use cases.

### Main changes

- Add a separate Teletext alignment dialog using the 23-line Teletext grid.
- Set Teletext line for one or multiple selected subtitles.
- Set horizontal alignment independently: left / center / right.
- Shift Teletext lines relatively.
- Replace one specific Teletext line with another (line X -> line Y).
- Add a dynamic TT column in the subtitle grid, with persistent show/hide setting.
- Add persistent Teletext alignment preview setting.
- Add Teletext line-length warning.
- Improve EBU STL Teletext handling, including preservation of box/double-height and header settings.
- Add hearing-impaired language codes used by the Teletext workflow.
- Add localization for the new UI in English, German, and French.

## Verification

Tested manually on current main / 5.2.0-beta18:

- `dotnet build src/ui/UI.csproj`
- Teletext alignment and multi-selection
- Teletext line setting
- Relative line shifting
- Line X -> Y replacement
- Left / center / right alignment
- Preview
- Persistent Preview and TT-column settings across restart
- EBU STL import/export
- macOS ARM64 build
- Windows x64 build

No new network functionality or external dependencies are introduced.

## Notes

This contribution was developed with material AI assistance and manually tested. The branch was rebuilt on the current upstream main before opening this pull request.


niksedk:
Thanks for the contribution! Nice to see broadcast/teletext workflows getting attention.

I've reviewed the PR and pushed some follow-up commits to your branch before merging - a summary of what changed and why:

**Scoping**
- The teletext editing is now gated to EBU STL (`IsFormatEbu`): the "TT" column, the alignment preview and the dialog no longer appear for other formats. The two settings still remember your choice across format switches - only the view is gated.
- The separate "Teletext alignment..." context menu item is gone; the existing **Alignment...** item opens the teletext dialog when an EBU STL file is loaded, and the normal picker otherwise.

**Correctness fixes**
- `MarginV` is only treated as a teletext row when the subtitle actually came from an EBU STL file. ASSA lines always carry a numeric `MarginV`, so exporting an .ass as teletext pinned every cue to the top row.
- Teletext rows are handled 1-based (1..23) throughout, matching the `VerticalPosition` field the rest of Ebu.cs reads/writes - the dialog and TT column were one off, and row 23 was unreachable.
- A stored row is clamped when the text has been re-wrapped since load, so extra lines can't run off the bottom of the page.
- Horizontal alignment keeps the line's vertical band (a `{\an8}` line set to right becomes `{\an9}`, not `{\an3}`).
- Teletext line edits now register as document changes (undo, modified flag, save prompt) - `MarginV` wasn't part of the change-detection hashes before.
- The line-length warning moved from a "⚠" prefix in the TT cell to the standard error tint on the Text cell, and it feeds the error list / go-to-next-error / screen-reader text like other rule breaks.
- Preview off no longer forces left alignment in the grid (broke RTL subtitles).

Plus some smaller cleanups (indentation, a few leftover German dev comments, "Linie" → "Zeile" in the German strings, invariant-culture number formatting).

Since quite a bit changed since your manual testing, it would be great if you could re-test your EBU STL workflow on the merged result (next beta) - especially import → edit lines/alignment → export round-trips against your usual broadcast tooling, and multi-selection with mixed rows. Please open an issue if anything regressed.

Triathlon-rally:
Hello NikolajThank you that you like the modifications On more thing if you don’t use a Color Code in Teletext you can use up to 37 characters instead of 36. because the color code use one positionTotal you have 40 Double High is oneColor is oneStart oneEnd oneRest 36 (characters and space)Best Regards RalphAm 19/08/2026 um 17:52 schrieb Nikolaj Olsson ***@***.***>:﻿
Merged #13862 into main.

Triathlon-rally:
Hi,

I have now tested the merged Teletext/EBU STL implementation in Subtitle Edit 5.2.0-beta19 on Windows in a real broadcast subtitle workflow.

First of all: the new Teletext functionality works well.

I also tested a subtitle file that was not EBU STL/Teletext compliant, corrected the subtitles and Teletext positions in Subtitle Edit, and exported the result as EBU STL. This workflow worked successfully.

While doing this practical test, I found a few additional points that are especially relevant when preparing non-compliant subtitle files for EBU STL/Teletext output.


## 1. Line length correction should optionally affect only lines that are actually too long

For our Teletext workflow the limits are normally:

- 37 visible characters without an explicit color code
- 36 visible characters when a Teletext color code is used
- maximum 2 text lines

I tested "Split/break long lines" with:

- Maximum single line length: 37
- Maximum number of lines: 2

I found a subtitle containing a single line of 40 characters.

The main editor correctly shows:

Line length: 40

However, the subtitle is not listed as something that needs to be corrected in the "Split/break long lines" window.

lange Texte trennen


If I additionally enable "Rebalance long lines", Subtitle Edit does correct this subtitle.

However, it then also rebalances many other subtitles whose individual lines are already within the 37-character limit.

For broadcast subtitles this is not always desirable.

An intentionally unbalanced subtitle can be editorially correct. For example, there may be a reason to have four words on the first line and only one word on the second line.

If neither line exceeds the configured maximum, this existing line break should be preserved.

It would therefore be very useful to have two modes/options:

- Rebalance all subtitles
- Rebalance only subtitles containing a line longer than the configured maximum

For the second option, existing line breaks should remain untouched when all lines already comply with the configured maximum length.


## 2. An overlong two-line subtitle should be split into two subtitle events

There is another important case when converting existing subtitle files to Teletext.

A subtitle may already contain two text lines, but still contain too much text to fit into the configured maximum of two lines with 37 (or 36) characters per line.

In this situation, rebalancing cannot solve the problem because no valid two-line arrangement is possible.

Ideally Subtitle Edit should then split the original subtitle into two new subtitle events.

For example:

Original subtitle:

TC IN  = 10:00:00:00
TC OUT = 10:00:08:00

Line 1: ...
Line 2: ...

If the text cannot be fitted into two valid Teletext lines, Subtitle Edit could create:

Subtitle 1:
TC IN  = original TC IN
TC OUT = calculated split point

Subtitle 2:
TC IN  = Subtitle 1 TC OUT + configured minimum gap
TC OUT = original TC OUT

The important timing rules would be:

- The first new subtitle keeps the original TC IN.
- The second new subtitle keeps the original TC OUT.
- The configured minimum gap between subtitles is respected.
- The remaining available duration is divided between the two new subtitles.
- The duration should preferably be distributed proportionally according to the amount of text in each new subtitle.
- Both resulting subtitles should respect the configured maximum number of lines and maximum line length.

For example:

available duration =
original duration - configured minimum gap

duration subtitle 1 =
available duration × text length subtitle 1 / total text length

duration subtitle 2 =
available duration - duration subtitle 1

This keeps the complete original time range intact while creating two valid subtitle events.

This would be especially useful when normalizing subtitle files from other formats before exporting them as EBU STL/Teletext.


## 3. "No color" option in the color picker

I also found a small but important issue with colors in the Teletext workflow.

For Teletext, default white text without an explicit color code is different from explicitly assigning white as a color.

When I select white in Subtitle Edit's color picker, Subtitle Edit adds:

<font color="#FFFFFF">Text</font>

code coleur blanc


For our Teletext limits this distinction matters:

- No explicit color code -> maximum 37 visible characters
- Explicit Teletext color code -> maximum 36 visible characters

At the moment I can remove the color via:

Right click
-> Remove formatting
-> Remove color

This works correctly and removes the explicit color formatting.

kontextmenue

sans code coleur


However, this is somewhat inconvenient during normal subtitle editing.

It would be useful if the existing color picker also had a "No color" option.

Selecting "No color" should not mean black or transparent. It should remove the explicit color formatting, equivalent to:

Remove formatting -> Remove color

This would allow the user to easily return a subtitle to default Teletext white without generating an unnecessary color code.

### Permanent standard color palette

While testing this further, I noticed another related usability issue.

The current "Recently used colors" palette changes whenever other colors are selected. This means that the original/basic colors can eventually disappear from the palette.

Instead of only adding "No color", it might be useful to have two separate color rows:

Standard colors:
No color | Black | Red | Green | Yellow | Blue | Magenta | Cyan | White

Recently used colors:
(the existing dynamic list)

The standard color row would always remain available and would never be modified by recently selected custom colors.

"No color" in this permanent palette would remove explicit color formatting, as described above.

This would be useful not only for Teletext but also for general Subtitle Edit use, because users would always have quick access to the basic colors while still keeping their recently used custom colors.

The attached mock-up shows the idea:


palette neu



## 4. "Fix common errors..." does not open the expected dialog

"Fix common errors..." is also very useful when preparing an existing subtitle file for EBU STL/Teletext.

In Subtitle Edit 4.x I could use:

Tools -> Fix common errors...

and a dialog opened where I could select the individual correction rules before applying them.

In my Windows test with Subtitle Edit 5.2.0-beta19, clicking:

Tools -> Fix common errors...

does not open this dialog.

fix common erreors


As far as I understand, the current workflow should still open the Fix Common Errors dialog where the individual rules can be enabled or disabled.

Therefore this may be a beta19 regression or another issue worth checking.


## Result of the Teletext/EBU STL test

Apart from the points above, the new Teletext implementation itself worked successfully in my tests.

I tested:

- EBU STL import
- Teletext line positions
- horizontal Teletext alignment
- editing individual and multiple subtitles
- conversion of a non-compliant subtitle file
- normalization for Teletext
- export back to EBU STL with Teletext positions

The resulting EBU STL/Teletext workflow worked correctly.

Thanks again for integrating the Teletext changes.

Triathlon-rally:
One additional Teletext workflow improvement I noticed while manually correcting subtitles in beta19:

## Automatically adjust the Teletext row when the number of text lines changes

When editing an EBU STL/Teletext subtitle manually, the Teletext row currently remains unchanged when a line break is inserted or removed.

Example:

A bottom-positioned single-line subtitle is correctly displayed as:

Teletext row: 23

If I manually insert a line break in the text editor and turn this subtitle into a two-line subtitle, the Teletext row remains at 23.

For the bottom Teletext position, it should automatically move to row 21 so that the two double-height text lines occupy the correct bottom positions.

In other words:

- bottom single-line subtitle -> TT row 23
- insert a line break -> bottom two-line subtitle -> TT row 21
- remove the line break again -> bottom single-line subtitle -> TT row 23

This would be especially useful when manually correcting imported subtitles, because changing the text from one line to two lines is a very common operation.

### Important: preserve intentional vertical positioning

I would not apply this rule to every Teletext subtitle regardless of its position.

If a subtitle has been intentionally positioned on another Teletext row, its manually selected position should not be overwritten simply because a line break is inserted or removed.

So the automatic 23 <-> 21 adjustment should apply specifically to subtitles using the normal bottom Teletext placement.

This would make manual EBU STL/Teletext editing much safer, because the text layout and the Teletext position would stay synchronized automatically.

Triathlon-rally:
One further idea came up while using the new Teletext editing workflow in practice.

This is not a bug report, but a possible concept for making manual Teletext editing much more efficient.

## Idea: optional Teletext "flow editing" mode

When manually correcting subtitles for EBU STL/Teletext, it could be useful if text editing behaved a little more like editing a document.

The difference would be that each subtitle event is effectively a small "page" with a maximum of two Teletext text lines.

For example:

Subtitle 1
Line 1
Line 2
----------------
Subtitle 2
Line 1
Line 2
----------------
Subtitle 3
...

The editor could automatically move text between these subtitle "pages" when necessary.


### 1. Automatically create a new subtitle when the text no longer fits

While editing a subtitle:

- If the text fits on one valid Teletext line -> keep one line and use the appropriate bottom TT row (normally 23).
- If the text needs two valid Teletext lines -> use two lines and the appropriate bottom TT row (normally 21).
- If the text cannot fit into the maximum two allowed Teletext lines -> automatically split the overflowing text into a new subtitle event.

The configured line-length limit should be respected, for example:

- 37 visible characters without an explicit Teletext color code
- 36 visible characters when a Teletext color code is present

The split should preferably occur at a sensible linguistic position (space, punctuation, etc.), not simply after character 74.


### 2. Keep the original overall timing when an automatic split is made

If one subtitle is automatically split into two subtitle events, the original outer timecodes should remain unchanged.

Example:

Original subtitle:

TC IN  = 10:00:00:00
TC OUT = 10:00:08:00

After the automatic split:

Subtitle 1:
TC IN  = original TC IN
TC OUT = calculated split point

Subtitle 2:
TC IN  = Subtitle 1 TC OUT + configured minimum gap
TC OUT = original TC OUT

The available duration would therefore be:

original duration - configured minimum gap

and could be distributed proportionally according to the amount of text in the two resulting subtitles.

This means that splitting the text does not change the beginning or end of the original subtitle period.


### 3. Backspace at the beginning of a subtitle could merge with the previous subtitle

The reverse operation could also be very useful.

If the cursor is at the very beginning of a subtitle and Backspace is pressed, Subtitle Edit could attempt to join the current subtitle with the previous subtitle.

This should only happen when the two subtitle events are temporally close enough to belong together.


### 4. Use the configured gap as a safety condition for Backspace merge

A subtitle should obviously not be merged with another subtitle that occurred many seconds earlier.

Therefore the Backspace merge should first check the gap between:

previous subtitle TC OUT

and

current subtitle TC IN

For example, if the allowed merge distance is configured as 5 frames:

gap <= 5 frames
-> merging is allowed

gap > 5 frames
-> do not merge the subtitle events

Optionally Subtitle Edit could display a small message such as:

"Cannot merge: gap exceeds 5 frames"

This would prevent accidental merging of subtitle events that are not temporally related.


### 5. Reflow the text after a merge

After joining two subtitle events, Subtitle Edit could treat their combined text as one continuous text flow and redistribute it according to the Teletext rules.

For example:

combined text fits on one line
-> one subtitle / TT row 23

combined text fits on two lines
-> one subtitle / TT row 21

combined text requires more than two lines
-> redistribute the text across two or more subtitle events

The timing could then be redistributed proportionally while respecting the configured minimum gap.


### 6. Automatically keep Teletext rows synchronized with the text layout

This could also solve the manual row issue mentioned in my previous comment.

For normal bottom-positioned Teletext subtitles:

1 text line
-> TT row 23

2 text lines
-> TT row 21

If an automatic split, merge, inserted line break or deleted line break changes the number of lines, the TT row could be updated automatically.

Intentionally positioned subtitles on other Teletext rows should of course keep their manually selected vertical position.


## Why this could be useful

This optional "flow editing" mode could combine several operations that currently have to be performed separately:

- correcting overlong lines
- splitting overlong two-line subtitles
- automatically creating a new subtitle when required
- merging adjacent subtitles with Backspace
- respecting the configured subtitle gap
- redistributing timing after split/merge
- automatically changing bottom Teletext row 23 <-> 21
- respecting the 36/37 character Teletext limits

It would make manual normalization of imported subtitle files feel more like editing continuous text, while Subtitle Edit automatically maintains the EBU STL/Teletext structure underneath.

I would see this as an optional Teletext editing mode rather than changing the normal Subtitle Edit editing behaviour.

niksedk:
I've added a teletext color picker - a bit like the alignment picker.

niksedk:
@Triathlon-rally Thanks again for the detailed workflow testing - it turned directly into a follow-up round, #13937, together with a code survey of the teletext/EBU paths. Status on your points:

**1. Line length correction should optionally affect only lines that are actually too long** - done in #13937. "Split/rebalance long lines" has a new persisted option, *Only subtitles with a too-long line*: rebalance then leaves existing line breaks untouched whenever every line is already within the configured maximum, so intentionally unbalanced subtitles survive. It also covers your 40-character example - a subtitle whose only problem is one overlong line is now corrected without rebalancing the rest of the file.

**Automatically adjust the Teletext row when the number of text lines changes** - done in #13937. A bottom-anchored subtitle follows your 23 ↔ 21 rule: inserting a line break in a single-liner on row 23 moves it to 21, removing the break moves it back. Subtitles on any other row are treated as intentionally positioned and left alone.

**2. Splitting an overlong two-line subtitle into two events** (with the original outer timecodes kept and the duration distributed proportionally) - not in this round; it is on the shortlist for a next one, along with more of the flow-editing ideas.

**3. "No color" / permanent standard palette** - the new teletext color picker (in beta since #13930) already has a "No color" tile that removes the explicit color code, matching your 37-vs-36 distinction. The permanent standard-color row in the general color picker is a good idea and is on the list.

**4. "Fix common errors..." not opening** - I could not reproduce this; the dialog opens fine here on the same build. If it still happens for you, could you open a separate issue with the subtitle file (or a trimmed sample) and the exact steps? It may be specific to the file or environment.

#13937 also fixes a few things the survey found: the video preview now updates immediately when a teletext row is changed, EBU STL header metadata (publisher/editor/disk fields) survives a load-save round trip, and re-saving a teletext STL no longer adds box/double-height codes that were not in the original file.

Triathlon-rally:
One useful follow-up regarding point 2 ("Splitting an overlong two-line subtitle into two events"):

I compared this workflow with Subtitle Edit 4.0.11 and found that the required behavior already existed there.

I have documented the SE4 vs SE5 comparison in #11476 ("SE4 style line breaking in SE5"), since that issue already describes the loss/change of the old SE4 line-breaking workflow.

In SE4, when a subtitle cannot fit within the configured maximum line length and maximum number of lines, the old Break/split function can create an additional subtitle event.

For example, an overlong two-line subtitle can become:

Original:

Und, war ich in der Kantine?
Nein. Ich konnte Sie dort nicht finden.

Result:

Subtitle 1:
Und, war ich in der Kantine?

Subtitle 2:
Nein. Ich konnte
Sie dort nicht finden.

SE4 also preserves the original outer timing boundaries, distributes the available duration between the resulting subtitle events, and updates the subtitle numbering.

This is particularly relevant to our Teletext workflow because we have a strict maximum of two displayed text lines. Once both lines are already occupied and one of them still exceeds the permitted length, rebalancing alone cannot make the subtitle compliant.

So point 2 may not require designing a completely new splitting algorithm. It might be possible to reuse or adapt the existing SE4 behavior for SE5.

For Teletext, the resulting events would then additionally use the existing Teletext positioning rules:

- one-line bottom subtitle -> row 23
- two-line bottom subtitle -> row 21

This would fit very naturally with the 23 <-> 21 automatic adjustment already implemented in #13937.

I think this also supports keeping "rebalance" and "split into additional subtitle events" as two distinct operations: intentional line breaks can remain untouched, while subtitles that physically cannot fit within the configured limits can still be split into new events.

subt 4


37 char 5-19

Issue #12338, opened 2026-07-10
Logitech Options+ plugin for Subtitle Edit (MX Master, MX Keys, MX Creative Console)


muaz978:
I have built a plugin that lets a Logitech device drive Subtitle Edit 5: MX Master through the on screen Actions Ring, MX Keys, or the MX Creative Console with its dials.

Repository: https://github.com/muaz978/subtitleedit-logi-plugin
Release: https://github.com/muaz978/subtitleedit-logi-plugin/releases/tag/v1.0
Licence: MIT

@niksedk sharing it here in case it is useful to others, and in case you would like it mentioned anywhere in the docs. Nothing in Subtitle Edit needs to change for it to work, so there is nothing to review unless you want there to be.

### How it works

Subtitle Edit has no automation API, so the plugin does what the Logi Actions SDK recommends for that case: it simulates the keyboard shortcuts Subtitle Edit already listens for.

The part I want to call out is where those shortcuts come from. `ShortcutsMain` registers actions by name and the key bindings live in the user's `Settings.json`, so the keys are whatever that user configured, and the primary modifier differs by platform (`Win`, meaning Command, on macOS; `Ctrl` on Windows). Hard coding a combination would break for anyone who remapped a shortcut, and break again across platforms. So the plugin reads `Settings.json` at load and resolves each action to its real keys. Remap a shortcut in Subtitle Edit and the plugin follows. An action with no keys assigned never appears, because there is nothing to send.

It handles a portable install too, where Subtitle Edit keeps `Settings.json` next to its executable rather than in the roaming folder.

### What it provides

- One action per bound shortcut, grouped as Playback, Timing, Lines, Text, File, Edit, Find and Tools.
- Four dials for the MX Creative Console: subtitle line, video position, waveform zoom, next error.
- An icon per action.
- Action names in all 26 languages Subtitle Edit ships. These are not translated again: Subtitle Edit already translates these exact command names, and `BuildCommandTranslations()` records which language string belongs to which action, so they are taken from there.

![Action icons](https://raw.githubusercontent.com/muaz978/subtitleedit-logi-plugin/main/docs/action-icons.png)

Tested on macOS and on Windows. Requires Subtitle Edit 5; the classic Subtitle Edit 4 stores its shortcuts in `Settings.xml`, in an unrelated format.

I will add screenshots of it running in Options+ shortly.

muaz978:
[...]

PR #13875, opened 2026-08-19
Bundle the Logitech Options+ plugin with official releases

muaz978:
Attaches the Logitech Options+ plugin (drives Subtitle Edit from an MX Master, MX Keys, or MX Creative Console by simulating the shortcuts already bound in Settings.json) to every official Subtitle Edit release, alongside the platform installers.

The release workflow now pulls the latest `.lplug4` from [muaz978/subtitleedit-logi-plugin](https://github.com/muaz978/subtitleedit-logi-plugin) and attaches it as a release asset. `continue-on-error: true` on that step, so a hiccup reaching the plugin's repo never blocks the main release.

No C# changes: the plugin works entirely from the Logitech side by simulating shortcuts, so nothing in Subtitle Edit itself needs to change for it to work.

The plugin is MIT licensed, is also listed on the [Logi Marketplace](https://marketplace.logi.com/plugin/SubtitleEdit/en), and was first shared in #12338.

### DIFF
--- a/.github/workflows/build-ui.yml
+++ b/.github/workflows/build-ui.yml
@@ -903,6 +903,17 @@ jobs:
           echo "Release assets:"
           ls -la ./release-assets/
 
+      - name: Download Logitech Options+ plugin
+        # Best-effort: pulls the latest release from the plugin's own repo so it
+        # ships alongside Subtitle Edit without needing a copy vendored here.
+        # continue-on-error so a hiccup fetching this optional asset never blocks
+        # the main release.
+        continue-on-error: true
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release download --repo muaz978/subtitleedit-logi-plugin --pattern '*.lplug4' --output ./release-assets/SubtitleEdit-LogiOptionsPlus-Plugin.lplug4
+
       - name: Get current date
         id: date
         run: echo "date=$(date +'%Y-%m-%d %H:%M:%S UTC')" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


muaz978:
@niksedk, can you please check this as it is an easy one and officially approved?

muaz978:
@niksedk, no rush, and this may just be low on the list. But I suspect the open question is not the workflow diff, it is whether you want official releases attaching a binary pulled from a repo of mine at release time. That is a trust call, not a code review, and no, or not yet, is a perfectly good answer.

The plugin has been live on the [Logi Marketplace](https://marketplace.logi.com/plugin/SubtitleEdit/en) since August, so it has been through Logitech's own review, and anyone can install it from there today. Nothing is lost if you would rather not bundle it.

If that is the part that gives you pause, I can pin it to a fixed tag with a SHA-256 check, so nothing I push later reaches a release you already approved. Or I link the plugin from the README and attach nothing. Or I close this. A one line reply is enough.

niksedk:
Hi @muaz978, thanks for the plugin and for the patience on this one - and you read the situation correctly in your second comment: the diff itself is fine, the hesitation is about official releases pulling a binary from a third-party repo at build time.

I've decided I'd rather not do that, so I'm closing this. Not a comment on the plugin at all - it's great that Subtitle Edit can be driven from the MX devices, and the Logi Marketplace listing is really the right home for it: users get it through Logitech's own channel, with their review and updates, without the release pipeline here depending on an external repo.

I'm happy to point people to it though - a link to the marketplace page and your repo from the docs would be very welcome if you want to open a small PR for that.

Thanks again!

muaz978:
Makes total sense, thanks for the clear reasoning. That's a supply-chain trust call, not a code one, and I get it. Opened a small docs PR linking the Logi Marketplace listing and the plugin repo instead: #14892

PR #13959, opened 2026-08-21
Allow main window interaction while Source View is open

ghostminhtoan:
## Description
Changed the **Source View** window from a modal dialog (`ShowDialogAsync`) to a modeless window (`ShowWindowInit`), allowing users to freely interact with the main window (e.g. inspecting waveforms, browsing list, navigating video) while keeping the Source View editor open side by side.

## Changes
- **Window Management**: Added `ShowWindowInit` in `IWindowService` / `WindowService` to configure the `ViewModel` before window instantiation.
- **Modeless Lifecycle**: Switched `ShowSourceView` in `MainViewModel` to modeless mode with save callback (`onSave`) and single-instance activation guard.
- **Data Synchronization**: Synchronizes updated subtitles back to the main window upon confirming changes (OK button).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@
 ## Get latest from https://github.com/github/gitignore/blob/main/VisualStudio.gitignore
 
 # User-specific files
+.agents/
+build/build.cmd
 *.rsuser
 *.suo
 *.user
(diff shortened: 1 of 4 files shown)


niksedk:
Thanks for taking the time to work on this!

I'd rather not go this way though. With the main window fully interactive while Source View is open, all the tools, the video/waveform, and the main grid can change the subtitle behind the editor - keeping the two in sync would be a a lot, and it's very easy to end up silently losing edits in one of them.

So I'm closing this as not wanted - Source View stays modal. Sorry, and thanks again for the contribution.

PR #13974, opened 2026-08-22
Update French.json

Need74:
For 5.2.0 beta 20

Generated using "Json Content Translator" 1.7, then stripped of "\u00A0" codes (non-breaking spaces). These are recognized within the JSON if replaced by the UTF-8 Alt+255 code. 
In French, it is standard practice to place a non-breaking space before the characters `:`, `!`, and `?`. 
I went through the entire file, which explains the large number of lines. 

Regards


Need74:
Thanks, "Claude"

For "Backend", I think "Côté serveur" is a good translation that is easier for a layperson to understand, though the English term is more commonly used.

Issue #10959, opened 2026-05-16
Batch Split subtitle adds break line


sl1k:
If you split a subtitle singularly, it splits it correctly in "karaoke" fashion: ie the text gets split by timecode into two subtitle bits. In batch mode it adds a line break for some reason, and splits the subtitles into two columns for no apparent reason. This makes work tedious and shite. Affected versions 3.16 up to 5 beta 27.

In beta version also adds random break lines even when no rebalancing is applied instead of splitting the lines. Also the maximum nro of lines option in the beta doesn't function with value "1" after adding higher value first and trying to revert to 1.

So frustrating

sl1k:
Hey so this fix is now commited to the source (?), but the beta 30 build doesn't show this fix in the changelog, how long would it usually take for the commits to get fixed in a next release? Or am I misunderstanding something?

niksedk:
Does beta 30 not work for you?
Not everything is included in the change log.

sl1k:
> Does beta 30 not work for you? Not everything is included in the change log.

No, sadly even the newest 5.1 beta 8 is still not functional. So how do the fixes actually get commited to the new builds?

niksedk:
> > Does beta 30 not work for you? Not everything is included in the change log.
> 
> No, sadly even the newest 5.1 beta 8 is still not functional. So how do the fixes actually get commited to the new builds?

Then explain very clearly how it should work, thx.

PR #14006, opened 2026-08-22
Improve split long lines workflow

Triathlon-rally:
This PR improves the Split/Break long lines workflow, especially for broadcast and Teletext subtitle editing.

Main changes:

- Split long subtitles into compliant subtitle events
- Rebalance split results locally so line length limits are respected
- Keep intentionally unbalanced subtitles unchanged when they already fit
- Allow rebalancing only subtitles that actually exceed the configured limits
- Add selective Rebalance with Apply checkboxes and Select all / Deselect all
- Distribute timing proportionally when a subtitle is split
- Respect the configured minimum gap between newly created subtitles
- Add an option to apply the global minimum gap to the whole subtitle file
- Add a 10 ms tolerance to minimum-gap comparisons to avoid no-op corrections
- Adjust Teletext row positions relatively when the number of text lines changes
  (for example 21↔23, 19↔21, 20↔22 for double-height Teletext)
- Automatically renumber subtitles after splitting to avoid duplicate numbers
- Add a direct EBU STL Header/Export button next to the format selector
- Add unit tests for Split/Break long lines and minimum-gap handling

The intention is to make Split long lines a more complete one-pass correction workflow:
splitting, local line balancing, timing, gaps, Teletext positioning and subtitle numbering are handled together where applicable.

The branch was rebased onto v5.2.0-beta21 and the related UI tests pass.

### DIFF
--- a/src/ui/Features/Main/Layout/InitToolbar.cs
+++ b/src/ui/Features/Main/Layout/InitToolbar.cs
@@ -548,6 +548,25 @@ public static class InitToolbar
             RoutingStrategies.Tunnel,
             handledEventsToo: true);
         stackPanelRight.Children.Add(comboBoxSubtitleFormat);
+
+        // Direct access to the existing EBU STL header/export dialog.
+        // Match the format ComboBox visually and leave a small right-side margin.
+        stackPanelRight.Children.Add(new Button
+        {
+            Content = "Header",
+            Command = vm.ExportEbuStlCommand,
+            Background = Brushes.Transparent,
+            BorderBrush = Brushes.Gray,
+            BorderThickness = new Thickness(1),
+            Padding = new Thickness(10, 4),
+            Margin = new Thickness(6, 0, 4, 0),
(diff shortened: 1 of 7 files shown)


niksedk:
Thanks for this, @Triathlon-rally - the three real bugs here (overlong lines not being split, no gap between split events, duplicate numbers after splitting) plus the per-row Apply checkboxes and the EBU STL teletext row shift are now on main via #14013, re-done on top of current main with tests.

A few parts were deliberately left out and would be welcome as separate proposals if you still want them:
- the toolbar **Header** button (needs a toolbar toggle and gating on the EBU format, and belongs in its own PR),
- the 10 ms tolerance in Apply min gap (changes existing behaviour; broadcast QC checks are usually exact, so this should be a setting if anything),
- auto-wrapping inside Split itself - split-only must not insert line breaks (#10959); instead, when Rebalance is on, subtitles that wrapping can fix are no longer cut into events.

Closed in favour of #14013.

PR #14019, opened 2026-08-23
Speech to text: transcription quality report, non-speech/repeat removal, overlap fix (#13973)

### LINKED ISSUE #13973: Speech to Text Post Processing Problems.
There are a number of problems that can occur that the current post processing does not detect or correct.

For a number of these problems it is required that SE first develop a number which represents the duration per word.  This is obtained by dividing the number of words in a subtitle into the duration.  This yields a number I call duration per word (DPW).  Here are the specific problems:

Subtitle to short:  Text to speech of produces subtitles that are two short.  The most common are short (one word subtitles) that are displayed for short time.  A subtitle with a display time of .2 sec is definitely too short to be comprehended.  I am not sure what the optimal display time should be, but the more words there are the long a subtitle needs to be.  Hence the DPW calculation.

Subtitle too long:  Any subtitle of over 10 seconds duration is definitely bad.  Especially when there are only a few words so the DPW is too high.  Again I not sure what is optimal or how to correct (i.e shorten, delete all together, request rescan of effected audio. 

Overlapping Subtitles:  Text to speech often produces subtitles that overlap with one another.  There is not warning about this.  Although overlaps are colorized differently there is not easy way to find them or quickly move from one to another.

Unwanted SDH:  It appears that Whisper at least has added an undocumented feature that if there is a long portion (over a minute) of audio that has no speaking Whisper adds a line to say what it thinks is going on.  Like Music, Waves, etc.

Subtitles that do not make sense:  Often subtiles are produced that make no sense.  Say the dectected speech is "The boy ran up the hill".  Subtitles will be produced that say:

The (one subtitle)
boy (next subtitle)
ran up the hill (next subtitle)

Some kind of cross subtitle grammar checking would yield much more meaningful subtitles.

niksedk:
Addresses the post-processing part of #13973 — the reporter's core ask was that SE should *tell you* when a transcription went badly, and fix what it safely can.

## What's new

**Quality report window** (shown after a single-file transcription when anything was found; batch runs log a one-line summary to the console instead):

- Summary cards per issue type — click to filter the table
- Table of affected lines with number, start/end, detail (duration / cps / overlap ms) and text
- "Do not show this report again" checkbox; re-enable via post-processing settings

**Detected issue classes** (thresholds come from the user's general timing settings so the report agrees with the grid colouring):
| Type | Rule |
|---|---|
| Too short | duration < min display time, or reading speed > max cps |
| Too long | duration > max display time, or ≥ 4 s with < 2 cps (a few words over a long span — typical hallucination) |
| Overlapping | end time later than next line's start |
| Non-speech | line is only `[Music]` / `(waves)` / `♪` style descriptions |
| Repeated | same text as the previous line (engine loop) |

**New post-processing options** (both off by default):
- *Remove non-speech lines*
- *Remove repeated lines* — compares against the last kept line, so a run of identical lines collapses to one

**Fix short duration** now also runs `FixOverlappingDisplayTimes` afterwards, so engine overlaps (and any created by extending short lines) don't arrive in the grid as red.

## Notes
- Report is always computed, even with post-processing off, so the user still gets told.
- `English.json` intentionally untouched (generated from the language classes).
- Not in this PR: "go to next overlap" navigation and per-engine doc links (separate follow-ups).

## Tests
- 9 new tests in `SpeechToTextPostProcessorTests.cs` (non-speech/repeat detection, analyzer per type, removal on/off, overlap fix chained, report with post-processing off)
- Layout verified with a headless `CaptureRenderedFrame()` render.

Closes #13973

### DIFF
--- a/src/ui/DependencyInjectionExtensions.cs
+++ b/src/ui/DependencyInjectionExtensions.cs
@@ -213,6 +213,7 @@ using TmpegEncXmlPropertiesViewModel = Nikse.SubtitleEdit.Features.Files.FormatP
 using VideoPlayerUndockedViewModel = Nikse.SubtitleEdit.Features.Shared.Undocked.VideoPlayerUndockedViewModel;
 using SpeechToTextAdvancedViewModel = Nikse.SubtitleEdit.Features.Video.SpeechToText.SpeechToTextAdvancedViewModel;
 using SpeechToTextPostProcessingViewModel = Nikse.SubtitleEdit.Features.Video.SpeechToText.SpeechToTextPostProcessingViewModel;
+using SpeechToTextQualityReportViewModel = Nikse.SubtitleEdit.Features.Video.SpeechToText.SpeechToTextQualityReportViewModel;
 
 namespace Nikse.SubtitleEdit;
 
@@ -571,6 +572,7 @@ public static class DependencyInjectionExtensions
         collection.AddTransient<WebVttStylePickerViewModel>();
         collection.AddTransient<SpeechToTextAdvancedViewModel>();
         collection.AddTransient<SpeechToTextPostProcessingViewModel>();
+        collection.AddTransient<SpeechToTextQualityReportViewModel>();
         collection.AddTransient<WordListsViewModel>();
     }
 
--- a/src/ui/Features/Video/SpeechToText/SpeechToTextPostProcessingViewModel.cs
+++ b/src/ui/Features/Video/SpeechToText/SpeechToTextPostProcessingViewModel.cs
@@ -1,4 +1,4 @@
-using Avalonia.Controls;
+﻿using Avalonia.Controls;
 using Avalonia.Input;
 using Avalonia.Media;
 using CommunityToolkit.Mvvm.ComponentModel;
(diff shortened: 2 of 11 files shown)


GitGianluc:
@niksedk just one question: ~is the transcription quality report (or the notification that the check has taken place) only displayed if any issues are detected?~

> **Quality report window** (shown after a single-file transcription when anything was found; batch runs log a one-line summary to the console instead)

I started a transcription using v5.2.0-beta23 with post-processing ON, but I haven’t received any notification regarding the transcription quality check; however, if you open the console log, you can see: `Quality check: no issues found in 112 lines`.

Basically, I’d be interested to know whether it’s possible to manually fire up the quality check function or read the quality check report once the speech-to-text conversion is complete. Thanks!

niksedk:
> @niksedk just one question: ~is the transcription quality report (or the notification that the check has taken place) only displayed if any issues are detected?~
> 
> > **Quality report window** (shown after a single-file transcription when anything was found; batch runs log a one-line summary to the console instead)
> 
> I started a transcription using v5.2.0-beta23 with post-processing ON, but I haven’t received any notification regarding the transcription quality check; however, if you open the console log, you can see: `Quality check: no issues found in 112 lines`.
> 
> Basically, I’d be interested to know whether it’s possible to manually fire up the quality check function or read the quality check report once the speech-to-text conversion is complete. Thanks!

You have to enable the report in the post-processing settings:
image
The report will only be shown if the transcription has any issues.

It's a lot like the "List errors" (Command+F8) from the main window.

GitGianluc:
@niksedk thanks for letting me know!

By the way, shouldn’t an important command such as ‘List errors’ be visible in the ‘Edit’ or ‘Tools’ menu? Or shouldn’t it be a customisable toolbar button?  ;-)

PR #14028, opened 2026-08-23
Fix OpenRouter STT requests for GPT transcription models

muaz978:
## Summary

- Use `json` for GPT transcription models that reject `verbose_json`.
- Omit timestamp granularity hints for those models.
- Keep verbose timestamp responses for Whisper-compatible models.
- Add request-body coverage for the supported GPT transcription model IDs.

## Testing

- `dotnet test tests/UI/UITests.csproj --no-restore --filter FullyQualifiedName~OpenRouterSttServiceTests -p:UsedAvaloniaProducts=`
- `dotnet test tests/UI/UITests.csproj --no-restore -p:UsedAvaloniaProducts=`

Both test runs pass.


muaz978:
@niksedk Quick note: I’ve added a fix for OpenRouter STT requests with GPT transcription models in this PR. These models reject `verbose_json`, so the patch switches them to `json` and keeps the existing verbose and timestamp path for Whisper models. The focused OpenRouter tests and the full UI test suite pass. Please take a look when you have a moment.

muaz978:
Working now!

image

PR #14029, opened 2026-08-23
Add configurable AI models folder

### LINKED ISSUE #14024: [Feature request] Allow changing the models folder location
I'm using Subtitle Edit on macOS. My Mac has limited internal storage, but I have plenty of space available on an external SSD.

Currently, downloaded models are stored under:

~/Library/Application Support/Subtitle Edit/

To keep these large model files off my internal drive, my current workaround is to symlink the entire Subtitle Edit folder to my external drive.

Could you add support for storing all downloaded models in a custom directory?

muaz978:
## Summary

Fixes #14024.

Adds an optional AI models folder setting so large downloaded model files can be stored on an external SSD without moving Subtitle Edit's settings, logs, dictionaries, themes, or executable folders.

## What is covered

- Adds Options > General > AI models folder with a folder picker.
- Preserves all historical locations when the setting is empty.
- Redirects managed Whisper, CTranslate2, Whisper.cpp, Purfview, CrispASR, TTS, Qwen3 ASR, llama.cpp, PaddleOCR, CrispEmbed, Tesseract, and audio.cpp model paths.
- Uses XDG_CACHE_HOME for OpenAI Whisper and HF_HOME for CTranslate2 when a custom root is selected.
- Keeps engine binaries in the normal application-data folder.
- Adds focused compatibility tests and detailed implementation documentation.

## Migration behavior

Existing models are not moved automatically. The setting changes future model resolution and downloads, while the documentation explains the manual copy layout. This avoids destructive or interrupted multi-gigabyte moves.

## Verification

AVALONIA_TELEMETRY_OPTOUT=1 dotnet test tests/UI/UITests.csproj --no-restore --filter FullyQualifiedName~DataFolderLocationTests

Result: 6 passed, 0 failed.

Also verified with git diff --check.


niksedk:
Hi @muaz978, thanks for the thorough work here - the model-path audit across all the engines alone is useful, and the "don't move multi-gigabyte files automatically" call is the right one.

I'm closing this one though. Touching every engine class (57 files) to redirect its model path is more surface than I want to carry for this, and the branch has drifted into conflicts with main since the engines have kept moving. If I add a configurable models folder, I'd like to do it in one place at the `Configuration` level, so the engines don't each need to know about it - and your list of which engines resolve their models where will be a handy map when I get to it.

I'll keep #14024 open as the feature request. Thanks again!

muaz978:
Understood, and the scope and conflict concerns are fair. 57 files is a lot of surface for this. Glad the model-path audit is useful as a reference for when you get to the Configuration-level version. I'll leave it to you, happy to help once you've settled on the shape you want.

PR #14031, opened 2026-08-23
Add WhisperX speech-to-text engine

muaz978:
## Summary

Add WhisperX as a first-class local speech-to-text engine and make it available to the existing imported-script alignment workflow.

## Implementation plan

1. Register WhisperX in the speech-to-text engine factory and expose it on Windows, Linux, and macOS.
2. Keep the installation GUI-based by creating an isolated Python virtual environment under Subtitle Edit's SpeechToText data folder.
3. Reuse Subtitle Edit's existing model picker and SRT result discovery, while allowing WhisperX to download its faster-whisper models from Hugging Face on first use.
4. Pass the selected language, model, translation task, output folder, and advanced parameters to the WhisperX CLI.
5. Expand WhisperX's language selector from the legacy ten-language subset to the full Whisper language catalog.
6. Make the engine usable from the existing Align time codes via Speech to text command, which preserves the imported text and replaces its timings.
7. Document the Apple Silicon behavior and avoid exposing an unsupported Apple GPU switch.

## What changed

- Added `WhisperEngineWhisperX` with managed virtual-environment paths, external executable detection, model and language metadata, and per-engine parameters.
- Added a GUI installer that finds Python 3, creates `.venv`, and installs or upgrades `pip` and `whisperx` without changing the system Python environment.
- Added default parameters `--device cpu --compute_type int8` for reliable macOS and CPU execution.
- Added WhisperX process handling, direct-media input support, translation support, automatic language selection, and SRT output handling.
- Added the engine to the existing engine factory and settings flow.
- Removed the legacy WhisperX language restriction so Turkish, Arabic, Russian, Persian, and the rest of the Whisper language catalog are available for transcription.
- Added embedded help, feature documentation, and focused tests.

## Apple GPU decision

WhisperX's normal faster-whisper plus alignment stack does not provide a reliable supported Metal/MPS path. This change intentionally does not force Apple GPU usage. On Apple Silicon, WhisperX uses CPU with `int8` by default. MLX Whisper remains the separate Apple-GPU option for transcription-only use, but it does not replace WhisperX alignment.

## Verification

- `dotnet build src/ui/UI.csproj --no-restore -c Release` passed with 0 errors.
- `dotnet test tests/UI/UITests.csproj --no-restore -c Release --filter FullyQualifiedName~WhisperEngineWhisperXTests` passed: 3 tests.
- Debug builds remain blocked by the pre-existing `Application.AttachDeveloperTools` error in `src/ui/Program.cs` with the current Avalonia diagnostics package. That unrelated issue is not changed here.

## Limitations

- Python 3 must already be installed for the GUI installer.
- WhisperX downloads the selected faster-whisper model on first use.
- Word-level alignment is language-dependent. Languages without a compatible WhisperX alignment model still receive normal segment timestamps.

### DIFF
--- a/docs/features/import-plain-text.md
+++ b/docs/features/import-plain-text.md
@@ -45,6 +45,8 @@ Missing models are downloaded automatically when you press OK. Progress is shown
 
 Runs normal [speech to text](speech-to-text.md) on the video, then matches your script against the transcript. Useful when the forced aligner does not support your language. Lines that cannot be matched get interpolated time codes.
 
+For the highest local transcription quality, select **WhisperX** in the speech-to-text window. It supports the full Whisper language catalog, and its language-specific word aligners improve timing where available. The imported script remains the source text; this command replaces its time codes from the WhisperX transcript.
+
 ## OK / Result
 
 Pressing **OK** replaces the currently loaded subtitle with the imported lines. Save your current work first if you need it.
--- a/docs/features/speech-to-text.md
+++ b/docs/features/speech-to-text.md
@@ -14,6 +14,7 @@ Subtitle Edit can automatically transcribe audio to text using Whisper-based and
 | Whisper CPP | Windows, Linux, macOS | Local CPU engine. On Windows the cuBLAS (NVIDIA CUDA) and Vulkan GPU backends can also be selected from the Whisper CPP backend dropdown. |
 | Purfview Faster Whisper XXL | Windows, Linux | Fast local engine, often used with NVIDIA CUDA |
 | Whisper CTranslate2 | Windows, Linux (x64), macOS (Apple Silicon) | CPU / NVIDIA CUDA depending on installation; CUDA requires [CUDA 12.x](https://developer.nvidia.com/cuda-12-0-0-download-archive) |
+| WhisperX | Windows, Linux, macOS | Installed by Subtitle Edit into an isolated Python environment; word alignment depends on language-specific WhisperX models |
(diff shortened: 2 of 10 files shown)


muaz978:
The main reason for this PR is to separate text quality from timestamp generation.

For many projects, the best workflow is to generate the transcript text with OpenRouter or another high-quality external text generator, then use WhisperX to align that existing text to the video's audio. External generators can produce better wording, punctuation, names, terminology, and translation quality, but many of them do not provide reliable word-level timestamps. WhisperX fills that missing timing step without forcing us to replace the higher-quality text with a second, lower-quality transcription.

This is especially useful for Subtitle Edit users who already have a carefully generated plain-text transcript and need accurate subtitle timings. It avoids spending hours manually shifting lines while preserving the quality of the original text, whether it came from OpenRouter, another external speech-to-text service, or a separate high-end text-generation workflow.

niksedk:
Thanks for the detailed work here, and for thinking through the Apple Silicon/CPU tradeoffs!

Unfortunately, I don't think this is a good fit for Subtitle Edit as-is. Requiring a Python installation plus a managed virtual environment (with pip installs on first run) is a significant barrier for most users — it adds a fragile dependency users have to install and troubleshoot themselves, which doesn't match how the other STT engines are packaged.

If WhisperX could be packaged as a standalone executable (e.g. via Nuitka or PyInstaller) that ships/downloads without requiring a separate Python setup, similar to how other engines are bundled, that would remove the main blocker and I'd be happy to reconsider.

Closing for now, but thanks again for the effort — feel free to revisit if a no-Python-required build becomes feasible!

PR #14032, opened 2026-08-23
Add EBU STL header toolbar button

Triathlon-rally:
**Add EBU STL header button to toolbar**

Description

Adds a small Header button to the toolbar for direct access to the existing EBU STL header/export settings.

The button is only visible when EBU STL is the active subtitle format and is placed before the format selector to keep the toolbar layout stable when switching formats.

Tested with switching between EBU STL and SRT.

No changes are made to the existing EBU STL header/export logic.


niksedk:
Implemented in other PR

PR #14037, opened 2026-08-23
Rename CharUtils.IsEnglishAlphabet to IsAsciiLetter

ivandrofly:
## Summary
- Rename `CharUtils.IsEnglishAlphabet` to `CharUtils.IsAsciiLetter`, matching the BCL name (`char.IsAsciiLetter`, .NET 7+) for the same A-Z/a-z check
- Update call sites (`FinalCutProXml14Text`, `FinalCutProXml15`) and the `CharUtilsTest` test method

## Test plan
- [ ] `dotnet build SubtitleEdit.sln` succeeds
- [ ] `dotnet test tests/libse/LibSETests.csproj` passes (verified locally: 1463 passed)
- [ ] No behavior change expected — pure rename

### DIFF
--- a/src/libse/Common/CharUtils.cs
+++ b/src/libse/Common/CharUtils.cs
@@ -17,6 +17,7 @@
         /// <summary>
         /// Checks if character is between A-Z or a-z
         /// </summary>
-        public static bool IsEnglishAlphabet(char ch) => ch >= 'A' && ch <= 'z' && (ch <= 'Z' || ch >= 'a');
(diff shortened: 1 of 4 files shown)


ivandrofly:
Note: the BCL has `char.IsAsciiLetter` with the same semantics, but it was added in .NET 7 — libse also targets `netstandard2.1`, where it is not available, hence the helper in `CharUtils` (a `<remarks>` on the method now documents this).

PR #14041, opened 2026-08-24
OCR fix replace lists: repair broken rules, remove dead entries

niksedk:
Full audit of the 19 shipped `Dictionaries/*_OCRFixReplaceList.xml` files (structural pass + .NET regex validation of every `find` pattern and `$N` group reference). ~115 defective entries fixed across 11 files.

## Actively harmful rules fixed
- **hrv**: `find="iscj?edi" replaceWith="$1scijedi"` had no capture group — .NET emits unmatched `$N` literally, so matching lines got `$1scijedi` in the subtitle text. Now `([iI])scj?edi`.
- **spa**: `¿Con cuanta → ¿Con cuántas` (spurious plural) shadowed the correct singular entry; the wrong one is removed.
- **hrb**: `casu → satu` replaced with `casu → čašu` (consistent with the list's own `cašu → čašu`).
- **swe**: ungated `\bl\b → I` (live since 2020) removed — it turned the liter symbol ("2 l mjölk") into "2 I mjölk".
- **deu**: entries that corrupted correct German words removed (gernhaben, ranzukommen, runterspült, wiederhaben, Blutgetränkte, immerwährende are all valid words).

## Silently dead rules repaired or removed
- **hrv**: `\([lL])j?ečen` and `\([pP])otc` never compiled ("too many )'s") and were silently dropped by `IsValidRegex`; now `([lL])…` / `([pP])…` and active.
- **dan**: `\b\l\b` (invalid `\l` escape, dead since 2020) removed rather than "fixed" — lone l is the liter symbol in Danish.
- 11 `<RegularExpressionsIfSpelledCorrectly>` entries were missing the required `spellCheck` attribute and silently skipped by the loader: the **eng** standalone `\bl\b → I` now has `spellCheck="I"` and is live (gated on the dictionary); the dead copies in dan/fra/nld/nob/nor/swe are removed ("I" is not a dictionary word there, lone "l" is the liter symbol). The 4 fra elision rules were double-dead (their char class was two straight apostrophes replacing with a straight apostrophe — a no-op) and redundant with the hardcoded `’ → '` normalization.

## Hygiene
- 24 no-op `from==to` entries deleted (the loader skips them — SE4's loader did too, so they were never functional). Three spa no-ops were restored to their evident intent instead (`¿a quien → ¿a quién`, `¿A quien → ¿A quién`, lowercase `¿es que no me has escuchado? → ¡¿…?!` mirroring the uppercase entry).
- ~30 exact duplicates and ~10 dead same-key conflicts (first-wins) removed, mostly hrv.
- Chains flattened so a single fix pass lands on the final form: dan `ﬁeme → fjerne`, deu `Spell/Spells → Spaß`, `weillt → weißt`, `filr'ne → für 'ne`, hrv `đta → što`, `jle → je l'`, `maiopre → malo prije`, `obiastima → područjima`.

## Verification
- All `find` patterns now compile under .NET and every `$N` reference in `replaceWith`/`spellCheck` has a matching group (checked with a scratch .NET tool).
- Structural re-audit is clean: no remaining no-ops, duplicates, dead conflicts, or attribute-incomplete gated rules.
- `dotnet test tests/libuilogic --filter FullyQualifiedName~Ocr`: 293 passed.
- `dotnet test tests/UI --filter FullyQualifiedName~Ocr.FixEngine`: 90 passed.

### DIFF
--- a/Dictionaries/dan_OCRFixReplaceList.xml
+++ b/Dictionaries/dan_OCRFixReplaceList.xml
@@ -49,7 +49,7 @@
     <Word from="Hvergang" to="Hver gang" />
     <Word from="erder" to="er der" />
     <Word from="villetilgive" to="ville tilgive" />
-    <Word from="ﬁeme" to="fjeme" />
+    <Word from="ﬁeme" to="fjerne" />
     <Word from="genopståri" to="genopstår i" />
     <Word from="svigtejer" to="svigte jer" />
     <Word from="kommernu" to="kommer nu" />
@@ -671,7 +671,6 @@
   </RegularExpressionsIfSpelledCorrectly>
   <RegularExpressions>
     <RegEx find="\b\|\b" replaceWith="I" />
-    <RegEx find="\b\l\b" replaceWith="I" />
   </RegularExpressions>
   <RegularExpressionsIfSpelledCorrectly language="da">
 
@@ -719,8 +718,6 @@
     <!-- leading l→I before 2+ lowercase (e.g. lste → Iste caught by spellcheck) -->
     <RegEx find="\bl([a-zæøå]{2,})\b" spellCheck="I$1" replaceWith="I$1" />
 
-    <!-- standalone l → I (Roman numerals, list markers, ordinals) -->
-    <RegEx find="\bl\b" replaceWith="I" />
 
     <!-- ============================================================ -->
     <!-- DANISH GENITIVE / APOSTROPHE                                 -->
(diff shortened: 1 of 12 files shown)


niksedk:
Added `ShippedOcrFixReplaceListsTests` (54b5f44d6): a CI guard that turns every silently-swallowed defect class into a test failure — non-compiling regexes, $N references without a matching capture group, gated rules missing `spellCheck`, no-op `from==to` pairs, exact duplicates, first-wins-shadowed entries, and unknown section names. Verified against the pre-cleanup lists: it fails on exactly the 11 files the audit flagged and passes the 8 clean ones. Full libuilogic suite: 524 passed.

Issue #13698, opened 2026-08-16
[Feature Request] Add automatic voice cloning and the ability to return to the dubbing phase.


invio-a11y:
The program is very comprehensive, but in my opinion, two key features are missing:

1 - the ability to automatically clone the vocal timbre of individual sentences: other free programs like Pyvideotrans and Voicestudio don't need to load sample voices to generate the clone, while in Subtitle Edit, in addition to loading the sample voice, you then have to manually edit who says what. I think this is feasible, since others are already doing it without any problems.

2 - Once the dubbed audio is generated, in Subtitle Edit you have to edit and possibly re-dub the individual sentence in the same session. From what I understand, you can't exit that phase, otherwise you have to start from scratch. And this is a problem with long videos, because the work can't necessarily be finished in one sitting.

The rest of the program is fantastic, very fast, and above all professional.

I hope these two improvements can be incorporated, as they would significantly improve production.

niksedk:
Thanks for the detailed feedback!

**2 - returning to the dubbing phase:** this already works today, it's just not named well. After generating, the review window has an **Export...** button - it writes `SubtitleEditTts.json` plus a `wav` folder with all generated audio. Next session, open "Text to speech" again and use **Import...** to pick that json: you get every line back with its audio, its engine/voice/speed, the include check boxes and the actor->voice cast mapping, and you can re-generate individual lines from there. So a long video can be done over several sittings. If the "Export/Import" wording is what made this hard to find, that's fair feedback - better names welcome.

**1 - automatic voice cloning:** a few questions so we build the right thing:

- Do you mean automatic *speaker detection* (the tool finds each speaker and clones them separately), or just a faster way to clone from the video you already have open - e.g. select a region in the waveform, right click -> "Clone to \<TTS engine\>"?
- In pyvideotrans/VoiceStudio, do you get one cloned voice per detected speaker, or is each sentence cloned from its own audio?
- Which language(s), and roughly how many speakers, do your videos have? And which TTS engine are you using in Subtitle Edit today?

Using subtitleformat "Advanced Sub Station Alpha" or "WebVTT" is probably required for this to work as these formats support actor/voice.

invio-a11y:
[...]

PR #14062, opened 2026-08-24
macOS: close the menu-bar gaps, add the Window-menu staples, fix the bundle's document types

niksedk:
A pass over the macOS-specific surface turned up three separate gaps. Each is invisible from a build or a run on Windows/Linux, which is why they had sat unnoticed.

## The three menu entries macOS was missing

`InitNativeMacMenu.cs` is a hand-maintained mirror of `InitMenu.cs` — Avalonia's `NativeMenu` is a separate API, so there is no shared builder the two could grow from. Three entries had drifted out of it:

- **SSA tools** (Styles / Properties / Attachments), the `.ssa` sibling of the ASSA menu. The ASSA menu is gated on `IsFormatAssa`, so nothing took its place: Mac users editing a plain SSA file had no menu route to those dialogs at all.
- **File ▸ Export ▸ IMSC image**
- **Video ▸ More ▸ Auto cast from video** (#13698)

They are added here together with a parity test, so the next one-sided add fails a test instead of shipping. Both files are pure builders whose entries all name a `MainViewModel` command, so the test compares the two command sets and names the offenders on failure. Deliberate differences are listed with their reason: Exit (macOS gets Quit in the app menu from Avalonia) and the parameterised audio-track command. Presence only is checked, never placement — which submenu an item lives in legitimately differs between the two, and pinning layout would fight normal edits.

Verified the test is not vacuous: dropping the new IMSC line again makes it fail naming that exact command.

## Cmd+M and Cmd+W now do something

AppKit gives an app no default Window menu and no default key equivalents — on macOS a shortcut works only where a menu item carries it. SE authored neither, so both keys were simply dead, unlike Hide and Quit which Avalonia appends for us.

Added **Minimize (⌘M)** and **Zoom** to the Window menu, and **Close window (⌘W)** to the File menu, which is where macOS keeps it. Zoom is the green button's resize, not full screen, so it toggles Avalonia's `Maximized`, which maps onto that same zoomed state.

The items are authored at build time, above the window list: once `MacWindowsMenuInterop` hands the menu to AppKit, AppKit appends its own window entries below whatever the menu already holds — the standard layout. The manual fallback in `UpdateWindowMenus` clears the whole menu on each refresh, so it re-adds them there too.

Each item acts on its own menu bar's window (macOS shows the focused window's menu bar) and does nothing while a modal dialog is open: the dialog leaves its owner input-disabled but still showing the menu bar, so an unguarded ⌘W would close the window out from under the open dialog.

These carry no view-model command — they act on the window — so the parity test cannot see them; `MacWindowMenuItemsTests` pins them instead, and runs on any OS since the builder is platform-independent (only its caller in `Program.cs` is macOS-only).

## The bundle's minimum version and document types

- **`LSMinimumSystemVersion` was 10.15.** The .NET runtime we ship is built against macOS 12 (`LC_BUILD_VERSION minos 12.0` in `libcoreclr.dylib`), so below that the app cannot load at all — the stale key only meant those users got a crash instead of Finder's "requires a newer version of macOS". Set to 12.0, with the `otool` command in a comment to re-derive it after a TFM bump.
- **Ten subtitle extensions the app reads were missing**, so Finder never offered SE for them: `ttml`, `dfxp`, `itt`, `smi`, `scc`, `mcc`, `cap`, `pac`, `lrc`, `idx`. Each is backed by a format in libse (TimedText10, ItunesTimedText, Sami, ScenaristClosedCaptions, MacCaption10, CheetahCaption/CapMakerPlus, Pac, Lrc, Idx).
- **`.xml` moved to its own entry at `Alternate` rank.** SE reads several XML subtitle formats, but at `Owner` rank it was claiming every XML file on the machine ahead of whatever really owns it. `Alternate` keeps SE in "Open With" without taking the default.

## Testing

- Full UI suite green: 3570 passed, 0 failed.
- Both new tests mutation-checked — a wrong ⌘M modifier and a removed menu entry each fail with the specific item named.
- `plutil -lint` passes, and `update-plist-version.sh` still sets both version keys on the edited plist (PlistBuddy rewrites the file, so the comments are repo-only — the built bundle is unaffected).

Not included, from the same review and worth their own PRs: the Shortcuts window's context flyout has no Ctrl+click handler, the ASSA menu sits after Help (macOS expects Help last), and macOS has no native OCR/TTS engine while Piper is hidden there despite a wired-up download path.

### DIFF
--- a/installer/macBundle/SubtitleEdit.app/Contents/Info.plist
+++ b/installer/macBundle/SubtitleEdit.app/Contents/Info.plist
@@ -20,8 +20,13 @@
     <string>SBED</string>
     <key>CFBundleIconFile</key>
     <string>SE.icns</string>
+    <!-- The .NET runtime we ship is built against macOS 12, so the app cannot load
+         below it no matter what this key says - it was 10.15 for years, which only
+         meant Catalina users got a crash instead of Finder's "requires a newer
+         version of macOS". Re-derive after a TFM bump with:
+           otool -l .../shared/Microsoft.NETCore.App/<ver>/libcoreclr.dylib | grep -A3 LC_BUILD_VERSION -->
     <key>LSMinimumSystemVersion</key>
-    <string>10.15</string>
+    <string>12.0</string>
     <key>NSHighResolutionCapable</key>
     <true/>
     <key>NSSupportsAutomaticGraphicsSwitching</key>
@@ -41,7 +46,16 @@
                 <string>vtt</string>
                 <string>sbv</string>
                 <string>stl</string>
-                <string>xml</string>
+                <string>ttml</string>
+                <string>dfxp</string>
+                <string>itt</string>
+                <string>smi</string>
+                <string>scc</string>
+                <string>mcc</string>
+                <string>cap</string>
+                <string>pac</string>
+                <string>lrc</string>
+                <string>idx</string>
             </array>
             <key>CFBundleTypeName</key>
             <string>Subtitle Files</string>
@@ -49,6 +63,22 @@
             <string>Editor</string>
             <key>LSHandlerRank</key>
             <string>Owner</string>
+        </dict>
+        <!-- .xml is its own entry, at Alternate rank: SE reads several XML subtitle
+             formats, but claiming Owner made it fight every other app on the machine
+             for every XML file. Alternate keeps SE in "Open With" without taking the
+             default away from whatever owns the file type. -->
+        <dict>
+            <key>CFBundleTypeExtensions</key>
+            <array>
+                <string>xml</string>
+            </array>
+            <key>CFBundleTypeName</key>
+            <string>XML Subtitle Files</string>
+            <key>CFBundleTypeRole</key>
+            <string>Editor</string>
+            <key>LSHandlerRank</key>
+            <string>Alternate</string>
         </dict>
         <dict>
             <key>CFBundleTypeExtensions</key>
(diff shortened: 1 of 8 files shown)


niksedk:
Follow-up pushed: **all tests pass and the solution now builds with zero warnings.**

**Tests** — every project, not just the UI one:

| Project | Result |
|---|---|
| LibSETests | 1465 passed, 0 failed |
| UITests | 3570 passed, 0 failed |
| LibUiLogicTests | 542 passed, 0 failed |
| SeConvTests | 383 passed, 0 failed |

5960 total, no failures.

**Warnings** — the solution built with five, all long-standing and all in libse. That is few enough to still be noise rather than a wall, but exactly the level at which a genuinely new warning slips by unnoticed, so the last commit takes it to zero:

- 4x CS8632 — `SubtitleFormat.FormatLimits` and Cavena890's `GreekLookup` carried `?` nullable annotations, but libse has no `<Nullable>` setting, so they are inert and only warn. Dropped, with each null contract written down instead of implied by a marker the compiler ignores.
- 1x xUnit2002 — `Assert.NotNull` on `TransportStreamSubtitle.GetScreenSize()`, which returns an `SKSize`. A struct is never null, so the assert could never fail; replaced with the no-source fallback that method actually returns, so the line now tests something.

That commit is unrelated to macOS and easy to drop or split out if you would rather keep this PR single-purpose.

Issue #13680, opened 2026-08-16
[Bug] Linux: whisper.cpp fails with "error while loading shared libraries: libwhisper.so.1: cannot open shared object file"


keremyavuz:
*I encountered a issue while I was trying to use speech to text function in Subtitle Edit. I use EndavourOS. I solved the problem by using AI. Having resolved the issue, I also wanted to create a ticket to let the developers know the issue. Here is the explanation of AI. Thank you.

```
### Environment
- **OS:** Linux (Arch Linux / EndeavourOS x86_64)
- **Subtitle Edit Version:** 5.1.0 (or latest)
- **Engine:** whisper.cpp (Speech-to-Text)

### Description
When running speech-to-text via **whisper.cpp** on Linux, the process fails immediately on launch with a missing shared library error.

### Error Log
```text
Calling speech-to-text (CPP) with : /home/user/.config/Subtitle Edit/SpeechToText/Cpp/whisper-cli --language en --model "/home/user/.config/Subtitle Edit/SpeechToText/Cpp/Models/tiny.en.bin" --output-srt --print-progress  "/tmp/se-stt-xxx/sample.wav"
/home/user/.config/Subtitle Edit/SpeechToText/Cpp/whisper-cli: error while loading shared libraries: libwhisper.so.1: cannot open shared object file: No such file or directory
Speech to text (CPP) done in 00:00:00.0961585
```
Running ldd on whisper-cli confirms missing dependencies:

```
$ ldd ~/.config/Subtitle\ Edit/SpeechToText/Cpp/whisper-cli | grep "not found"
    libwhisper.so.1 => not found
    libggml.so.0 => not found
```

Root Cause Analysis
Missing Shared Library: The automated download/extraction for the Linux whisper.cpp bundle in ~/.config/Subtitle Edit/SpeechToText/Cpp/ contains libggml*.so files, but libwhisper.so (or libwhisper.so.1) is missing entirely from the directory.

Missing ABI Symlinks: The binary specifically links against versioned SONAMEs (libwhisper.so.1, libggml.so.0). If only unversioned .so files exist, the dynamic linker (ld.so) fails to resolve them.

Missing RPATH / $ORIGIN: The provided whisper-cli binary does not have its RPATH / RUNPATH set to $ORIGIN. As a result, Linux does not search the local executable directory for bundled .so files unless LD_LIBRARY_PATH is explicitly exported.

Workaround Applied
[...]

PR #14070, opened 2026-08-24
WhisperX as a standalone (no Python) speech-to-text engine

muaz978:
Re-attempt at #14031, which you closed over a fair objection: requiring a Python install plus a managed virtual environment (with pip installs on first run) does not match how every other engine here is packaged, and is a real barrier for non-technical users.

This time WhisperX is packaged as a standalone build with PyInstaller instead, so it downloads and runs exactly like whisper-ctranslate2 or Purfview's Faster-Whisper-XXL: no Python, no pip, nothing for the user to install separately. Source and build scripts are at https://github.com/muaz978/subtitleedit-whisperx-standalone, with its own CI building and smoke-testing the package on Windows, macOS, and Linux on every change, and a v1.0.0 release with the built binaries for each platform.

## What changed here

Modeled on `WhisperEngineCTranslate2` (the closest existing engine: also a downloaded prebuilt binary, same flat-unpack layout), not on the old managed-venv approach:
- `WhisperEngineWhisperX.cs`: `CanBeDownloaded() => true`, points at a downloaded binary, no more Python discovery.
- `WhisperDownloadService.cs` / `DownloadSpeechToTextEngineViewModel.cs`: standard download-and-unpack, same pattern as CTranslate2.
- `SpeechToTextViewModel.cs`: WhisperX shells out to a real ffmpeg binary for audio loading (not bundled in the standalone build), so the process launch now puts Subtitle Edit's own configured ffmpeg on PATH for it, reusing whatever ffmpeg Subtitle Edit already has rather than requiring a separate install.

## Verified

- Full pipeline tested end to end on macOS arm64 with real audio: voice activity detection, transcription, forced alignment (real wav2vec2 model download), and diarization (real pyannote/speaker-diarization-3.1 model, correct speaker-labeled output).
- The standalone build itself builds and smoke-tests clean in CI on all three platforms.
- This repo: `dotnet build` clean, full test suite passing (3560/3560, including new tests for the engine).

## Known gap

Windows and Linux binaries are CI-smoke-tested (proving the frozen binary loads and runs) but I have not personally verified real-speech transcription on those platforms, only macOS. Flagged in the release notes too.

### DIFF
--- a/src/ui/Assets/SpeechToText/WhisperX.txt
+++ b/src/ui/Assets/SpeechToText/WhisperX.txt
@@ -1,0 +1,180 @@
+usage: whisperx-standalone [-h] [--model MODEL]
+                           [--model_cache_only MODEL_CACHE_ONLY]
+                           [--model_dir MODEL_DIR] [--device DEVICE]
+                           [--device_index DEVICE_INDEX]
+                           [--batch_size BATCH_SIZE]
+                           [--compute_type {default,float16,float32,int8}]
+                           [--output_dir OUTPUT_DIR]
+                           [--output_format {all,srt,vtt,txt,tsv,json,aud}]
+                           [--verbose VERBOSE]
+                           [--log-level {debug,info,warning,error,critical}]
+                           [--task {transcribe,translate}]
+                           [--language LANGUAGE]
+                           [--align_model ALIGN_MODEL]
+                           [--interpolate_method {nearest,linear,ignore}]
+                           [--no_align] [--return_char_alignments]
+                           [--vad_method {pyannote,silero}]
+                           [--vad_onset VAD_ONSET] [--vad_offset VAD_OFFSET]
+                           [--chunk_size CHUNK_SIZE] [--diarize]
+                           [--min_speakers MIN_SPEAKERS]
+                           [--max_speakers MAX_SPEAKERS]
+                           [--diarize_model DIARIZE_MODEL]
+                           [--speaker_embeddings] [--temperature TEMPERATURE]
+                           [--best_of BEST_OF] [--beam_size BEAM_SIZE]
+                           [--patience PATIENCE]
+                           [--length_penalty LENGTH_PENALTY]
+                           [--suppress_tokens SUPPRESS_TOKENS]
+                           [--suppress_numerals]
+                           [--initial_prompt INITIAL_PROMPT]
+                           [--hotwords HOTWORDS]
+                           [--condition_on_previous_text CONDITION_ON_PREVIOUS_TEXT]
+                           [--fp16 FP16]
+                           [--temperature_increment_on_fallback TEMPERATURE_INCREMENT_ON_FALLBACK]
+                           [--compression_ratio_threshold COMPRESSION_RATIO_THRESHOLD]
+                           [--logprob_threshold LOGPROB_THRESHOLD]
+                           [--no_speech_threshold NO_SPEECH_THRESHOLD]
+                           [--max_line_width MAX_LINE_WIDTH]
+                           [--max_line_count MAX_LINE_COUNT]
+                           [--highlight_words HIGHLIGHT_WORDS]
+                           [--segment_resolution {sentence,chunk}]
+                           [--threads THREADS] [--hf_token HF_TOKEN]
+                           [--print_progress PRINT_PROGRESS] [--version]
+                           [--python-version]
+                           audio [audio ...]
+
+positional arguments:
+  audio                 audio file(s) to transcribe
+
+options:
+  -h, --help            show this help message and exit
+  --model MODEL         name of the Whisper model to use (default: small)
+  --model_cache_only MODEL_CACHE_ONLY
+                        If True, will not attempt to download models, instead
+                        using cached models from --model_dir (default: False)
+  --model_dir MODEL_DIR
+                        the path to save model files; uses ~/.cache/whisper by
+                        default (default: None)
+  --device DEVICE       device type to use for PyTorch inference (e.g. cpu,
+                        cuda) (default: cpu)
+  --device_index DEVICE_INDEX
+                        device index to use for FasterWhisper inference
+                        (default: 0)
+  --batch_size BATCH_SIZE
+                        the preferred batch size for inference (default: 8)
+  --compute_type {default,float16,float32,int8}
+                        compute type for computation; 'default' uses float16
+                        on GPU, float32 on CPU (default: default)
+  --output_dir OUTPUT_DIR, -o OUTPUT_DIR
+                        directory to save the outputs (default: .)
+  --output_format {all,srt,vtt,txt,tsv,json,aud}, -f {all,srt,vtt,txt,tsv,json,aud}
+                        format of the output file (default: all)
+  --verbose VERBOSE     whether to print out progress and debug messages
+                        (default: True)
+  --log-level {debug,info,warning,error,critical}
+                        log level (default: info)
+  --task {transcribe,translate}
+                        whether to transcribe or translate to English
+                        (default: transcribe)
+  --language LANGUAGE   language spoken in the audio; specify None to perform
+                        language detection (default: None)
+  --align_model ALIGN_MODEL
+                        name of phoneme-level ASR model to do alignment
+                        (default: None)
+  --interpolate_method {nearest,linear,ignore}
+                        for word .srt, method to assign timestamps to non-
+                        aligned words, or merge them into neighbouring
+                        (default: nearest)
+  --no_align            do not perform alignment (default: False)
+  --return_char_alignments
+                        return character-level alignments in the output json
+                        file (default: False)
+  --vad_method {pyannote,silero}
+                        VAD method to be used (default: pyannote)
+  --vad_onset VAD_ONSET
+                        onset threshold for VAD (see pyannote.audio pipelines)
+                        (default: 0.5)
+  --vad_offset VAD_OFFSET
+                        offset threshold for VAD (see pyannote.audio
+                        pipelines) (default: 0.363)
+  --chunk_size CHUNK_SIZE
+                        Chunk size for merging VAD segments (default: 30)
+  --diarize             apply diarization to assign speaker labels to each
+                        segment/word (default: False)
+  --min_speakers MIN_SPEAKERS
+                        minimum number of speakers to in audio file (default:
+                        None)
+  --max_speakers MAX_SPEAKERS
+                        maximum number of speakers to in audio file (default:
+                        None)
+  --diarize_model DIARIZE_MODEL
+                        Name of the diarization model to use (default:
+                        pyannote/speaker-diarization-3.1)
+  --speaker_embeddings  Include speaker embeddings in JSON output (only works
+                        with --diarize too) (default: False)
+  --temperature TEMPERATURE
+                        temperature to use for sampling (default: 0)
+  --best_of BEST_OF     number of candidates when sampling with non-zero
+                        temperature (default: 5)
+  --beam_size BEAM_SIZE
+                        number of beams in beam search, only applicable when
+                        temperature is zero (default: 5)
+  --patience PATIENCE   optional patience value to use in beam decoding
+                        (default: 1.0)
+  --length_penalty LENGTH_PENALTY
+                        optional token length penalty coefficient (alpha)
+                        (default: 1.0)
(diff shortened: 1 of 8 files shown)


niksedk:
Thanks for the rework — packaging it standalone is the right direction, and the ffmpeg-on-PATH plumbing (`configureStartInfo` + `FfmpegHelper`) is clean and reuses what we already have.

I downloaded the macOS arm64 build and benchmarked it against the engines we already ship. Results below. There's one blocker that has to be fixed before this can go in.

## Blocker: the default invocation produces no output

Alignment is on by default, and the alignment model download dies:

```
urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED]
certificate verify failed: unable to get local issuer certificate (_ssl.c:1010)>
[PYI-57774:ERROR] Failed to execute script 'whisperx_entry' due to unhandled exception!
```

Exit code 1, zero SRT written. That is exactly the command this PR builds, since `CommandLineParameterWhisperX` defaults to `string.Empty` — no `--no_align`.

Root cause: the bundled OpenSSL has its CA path compiled in as

```
/Library/Frameworks/Python.framework/Versions/3.12/etc/openssl/cert.pem
```

which is the `actions/setup-python` framework path from the CI runner. It won't exist on a user's machine. HuggingFace downloads survive because `huggingface_hub` uses `requests`, which picks up bundled certifi automatically; `torch.hub` uses raw `urllib` and falls back to OpenSSL's compiled-in default, which isn't there.

Confirmed it's packaging and not my machine: pointing `SSL_CERT_FILE` at the bundle's own `_internal/certifi/cacert.pem` makes the same command exit 0 and write the SRT. Fix is a couple of lines at the top of `whisperx_entry.py`, before torch is imported:

```python
import os, certifi
os.environ.setdefault("SSL_CERT_FILE", certifi.where())
os.environ.setdefault("REQUESTS_CA_BUNDLE", certifi.where())
```

**The build CI cannot catch this**, because the smoke test passes `--no_align` and never exercises the alignment path. Worth having the smoke test run the default pipeline.

Verified on macOS arm64. I did not run the Windows/Linux binaries, but they're built the same way by the same workflow, so I'd expect the same failure there.

## Benchmark

Apple M4 (4P+6E), 24 GB, AC power. JFK inaugural address, 300.0 s, 16 kHz mono WAV. 3 timed runs each after a warm-up, using each engine's real Subtitle Edit command line. Variance under ±2%. WhisperX runs used the `SSL_CERT_FILE` workaround above, or they would not have completed at all.

| Engine | tiny.en | ×RT | large-v3-turbo | ×RT |
|---|---|---|---|---|
| whisper.cpp (Metal — our macOS default) | **2.24 s** | 134× | **21.9 s** | **13.7×** |
| whisper.cpp (CPU, `-ng`) | 4.22 s | 71× | 63.6 s | 4.7× |
| Whisper CTranslate2 (CPU, `--vad_filter True`) | 6.13 s | 49× | 86.5 s | 3.5× |
| WhisperX `--no_align` (CPU) | 23.7 s | 12.6× | 56.3 s | 5.3× |
| WhisperX full pipeline (CPU) | 30.8 s | 9.7× | 67.6 s | 4.4× |

Purfview Faster-Whisper-XXL isn't in the table — it has no macOS build, so it can't be compared here.

Two things stand out.

**The batching win is real.** WhisperX `--no_align` beats CTranslate2 by **1.54×** at large-v3-turbo (56.3 s vs 86.5 s). CTranslate2 is the same backend Purfview XXL uses, so that's a genuine advantage over the faster-whisper family — and that's the comparison that matters on Windows, where there's no Metal. I want to be fair to the PR on this point: my expectation going in was that a CPU-only build would simply be slower than everything, and that turned out to be wrong.

**But on macOS it loses to what we already ship**, by 3.1× against whisper.cpp with Metal — from a 6.8 MB binary instead of a 1.1 GB bundle.

**Fixed per-file overhead** (two-point fit, 30 s vs 300 s clip, turbo):

| | Fixed cost | Marginal throughput |
|---|---|---|
| WhisperX | **13.2 s/file** | 7.0× RT |
| whisper.cpp (Metal) | **0.6 s/file** | 14.1× RT |

That constant is why WhisperX loses badly on short clips, and it compounds in Batch convert — ~21 extra minutes of pure startup across 100 files.

## Size

The `DownloadSizeText` values are understated, and the on-disk cost isn't surfaced anywhere. I measured macOS directly and read the Linux/Windows zip central directories via HTTP range requests:

| | PR claims | Actual download | Unpacked | Removable |
|---|---|---|---|---|
| macOS | ~300 MB | 382 MB | 1.10 GB | — |
| Windows | ~750 MB | 863 MB | **3.71 GB** | **2.81 GB (76%)** |
| Linux | ~750 MB | 924 MB | **2.91 GB** | **1.22 GB (42%)** |

The removable part is dead weight that `--collect-all torch` drags in:

- **Windows**: `_internal/torch/lib/dnnl.lib` is **2.33 GB**, plus `libprotoc.lib` (157 MB), `libprotobuf.lib` (115 MB), `kineto.lib`, `torch_cpu.lib`. Static link libraries — build-time only, never loaded at runtime.
- **Linux**: `libtriton.so` ships **twice, 462 MB each** (`_internal/libtriton.so` and `_internal/triton/_C/libtriton.so`), plus `libcupti_static.a`, `libnvperf_host_static.a` and `ptxas-blackwell`. Triton is a GPU kernel compiler and the CUDA bits are unusable in a CPU-only build.

Excluding triton and pruning `torch/lib/*.lib` and bundled headers should roughly halve Windows and cut Linux ~40%.

## Other findings

- **Good news**: I checked all three archives for symlink entries (Subtitle Edit unpacks with .NET's `ZipArchive`, which writes symlinks out as plain text files — this broke our Linux whisper.cpp archives in #13680). **Zero symlinks in all three.** Not a problem here.
- **First launch took 66.8 s** on macOS while the OS scanned the unsigned 89 MB binary across ~19,000 files (warm start afterwards: 1.6 s). Users will see a minute of apparent hang on first transcribe with no progress indication.
- **Runtime model downloads add ~2.0 GB** beyond the engine (1.6 GB HuggingFace + 368 MB `torch.hub` alignment models), with no size shown in the UI. `--model_dir` is never passed, so these land in `~/.cache/huggingface` outside Subtitle Edit's control — we can't show or clean them.

## Code issues in this PR

1. **`_downloadStream` is a `MemoryStream`** ([DownloadSpeechToTextEngineViewModel.cs#L96](https://github.com/SubtitleEdit/subtitleedit/blob/main/src/ui/Features/Video/SpeechToText/DownloadSpeechToTextEngineViewModel.cs#L96)). Buffering an 863–924 MB download in memory, with `MemoryStream`'s doubling growth, means ~1.5–2 GB peak before unpacking starts. Purfview XXL streams to a file (`DownloadWhisperPurfviewFasterWhisperXxl(string destinationFileName, ...)`) precisely because of size — WhisperX needs the same treatment.
2. **Unpinned third-party URL.** `releases/latest/download/` means every user gets whatever that repo publishes at any moment, and we can't tell which version is installed. Every other engine points at an upstream release or `SubtitleEdit/support-files`, pinned by tag.
3. **No `ClearExecutableStack`** on Linux. WhisperX bundles ctranslate2 from the same wheel family that hit the glibc 2.41 `PT_GNU_STACK=RWE` problem we patch for Purfview.
4. **Model plumbing**: uses the ggml `WhisperModel` list and model folder; `WhisperEngineCTranslate2` uses `WhisperPurfviewFasterWhisperModel` for the same backend. `IsModelInstalled` always returns true, so every model shows as installed.
5. Windows gating has no arch check, so Windows ARM64 is offered the x64 build. Same as the existing CTranslate2 behaviour, so not a regression, but it's a 863 MB mistake.

## On PyInstaller vs Nuitka

Asked internally whether Nuitka would be faster. The measurement answers it: warm process start including `import torch` is 1.56 s of the 13.2 s fixed overhead — the rest is pyannote VAD load, model load and native compute, which Nuitka cannot touch. Even eliminating *all* Python startup saves ~2% of a turbo run. PyInstaller is the right choice; not worth building a second freezer.

## Where I'd like this to go

The batching result is good enough that I don't want to just close this. But as it stands it emits nothing on a clean machine, and the Windows build would cost users 3.7 GB of disk. Before merging:

1. Fix the CA bundle issue, and make the CI smoke test run the **default** pipeline (with real speech, asserting on content — the current synthesized tone only checks the exit code).
2. Prune triton and the static `.lib` files.
3. Correct `DownloadSizeText`, and surface the unpacked size and the ~2 GB of runtime model downloads.
4. Move hosting to `SubtitleEdit/support-files` behind a pinned tag, built by a workflow there like `build-whispercpp-release.yml`. I'd rather own the build and the pruning than point users at `latest` on a third-party repo. Happy to set that workflow up.
5. Fix the `MemoryStream`, the Linux execstack call, and the model-list plumbing.

Worth considering a CUDA build for Windows/Linux too — that's a far bigger lever than anything else here, and we already have the `CrispAsrWindowsVariant` pattern for letting users pick cpu/vulkan/cuda at download time.

muaz978:
Thanks for the thorough review, this was really helpful. Went through everything:

**SSL certificate blocker (the actual crash)**: fixed. The bundled OpenSSL had the CI runner's own filesystem path compiled in as its CA path, which doesn't exist on a user's machine. torch.hub (used for the alignment model) uses raw urllib and fell back to that broken default, failing with CERTIFICATE_VERIFY_FAILED. Now both SSL_CERT_FILE and REQUESTS_CA_BUNDLE are pointed at the bundle's own certifi CA bundle before any network request happens. Verified on a fresh CI run across Windows, macOS and Linux with a real-speech smoke test (transcribe -> align, default pipeline) asserting the output transcript is correct and zero certificate errors on any platform.

**Size**: removed GPU/build-time dead weight a CPU-only build never touches at runtime (Windows static .lib files, Linux triton/CUDA artifacts).
- Windows: 855 MB -> 348 MB zip
- Linux: 913 MB -> 574 MB zip
- macOS: unchanged at 365 MB (nothing to prune there)

DownloadSizeText now shows these measured numbers instead of the old guesses, and the help text notes the unpacked size (~950 MB-1.7 GB depending on platform) plus the ~2 GB of Whisper/alignment/diarization models that download from Hugging Face on first use.

**The five code issues**, all addressed:
- Model list now comes from the same faster-whisper/ctranslate2 model set WhisperEngineCTranslate2 uses, instead of whisper.cpp's ggml list.
- IsModelInstalled checks the actual Hugging Face hub cache for the resolved repo id instead of always returning true.
- Download switched to streaming straight to a file (matching Purfview's approach) instead of buffering the full archive in a MemoryStream, which was peaking at 1.5-2 GB before unpacking even started.
- URLs are pinned to a specific release tag (v1.0.1) instead of "latest".
- The Linux PT_GNU_STACK executable-stack fix now also runs for WhisperX, not just Purfview Faster-Whisper-XXL.
- Windows ARM64 is explicitly gated with PlatformNotSupportedException since no ARM64 build exists.

**Hosting migration to SubtitleEdit/support-files**: leaving that to you as you offered, happy to help however's useful once that's set up.

v1.0.1 with these fixes is published now: https://github.com/muaz978/subtitleedit-whisperx-standalone/releases/tag/v1.0.1

niksedk:
Thx, moved the build here: https://github.com/SubtitleEdit/support-files/actions/workflows/build-whisperx-standalone-release.yml

Building...

niksedk:
Thx, could you bedrift latest beta?

muaz978:
Tested beta24 (macOS arm64) end to end against the new support-files hosted build:

- Downloaded whisperx-standalone-101 from support-files and confirmed the .7z extracts correctly. One thing worth flagging: it's compressed with the newer Delta ARM64 LZMA2 filter, which the old p7zip fork (the one most Linux distros ship) can't decode - I hit "Unsupported Method" errors on ~500 files testing with it locally. Not a problem for us though: SE's own extractor only shells out to a system 7-Zip if it finds one at a hardcoded path, which almost no user will have, so it falls through to the bundled SharpCompress reader for basically everyone. I tested that path directly (extracted all 19,028 entries, zero errors, byte count matched exactly), so this isn't actually a real-world risk, just noting it in case that assumption ever changes.
- Ran the binary standalone, both --no_align and the default pipeline (transcribe + align, no flags - what a fresh install actually invokes). Default pipeline completed with zero SSL/certificate errors and produced a correctly aligned SRT.

Looks solid on macOS. Let me know if you'd like Windows/Linux checked the same way.

Issue #14053, opened 2026-08-24
subtitle edit v5.1 & 5.2 beta 12 after translation and mult-replace the original text disapears


madsoft-ha:
I am using the v5.2 beta 12 but also tested this with v5.1

After I translating the subtitles I see both the original text and the translated text version. 
When I try to fix translation errors with multi-replace it removes the original text when OK or APPLY is pressed.

Is this a setting I missed?

Reagards,

Robin

niksedk:
thx for reporting this :)
Fix will be included in tomorrows beta.

PR #14087, opened 2026-08-25
Fix five defects found in the pre-beta24 review sweep

niksedk:
A pre-release review sweep over the 27 PRs merged since v5.2.0-beta23 surfaced five defects; this fixes them before the beta24 tag. All are regressions or gaps in code merged during this beta window, so no changelog entries.

- **dots.tts consent gate** (#14083): `DotsTtsVoice` was missing from `VoiceCloningConsent.IsCloneVoice`, whose doc comment warns every cloning engine must be listed. dots.tts voices always carry a reference recording (cloning is its only mode), and the auto-seeded voices never pass the import-time gate — so a user could clone without ever seeing the consent dialog, and synthesis ran without the accepted-provenance attestations (the server then applies its own defaults: audible disclaimer or refusal, with no explanation offered in the UI).
- **Google Translate clients5 fallback** (#14050): `Trim('"')` also ate the escaped quote ending a translation — `He said \"hi\"` was left with a dangling backslash, `Regex.Unescape` threw, and the raw escapes shipped into the subtitle. Now exactly one delimiter pair is stripped; two regression tests added.
- **Fix Netflix errors vs. auto-backup** (#14054): the dialog stored the live working subtitle. An auto-backup tick while the dialog was open rebuilds every paragraph with fresh ids, so the id-based apply from #14053 matched nothing and — when RemoveEmptyLines changed the count — fell back to the full row rebuild that empties the original column. The dialog now snapshots with ids preserved, the same pattern Fix common errors uses.
- **Evenly distribute lines** (#14064): per-line whole-millisecond rounding accumulated through the running end, drifting the block end by a few ms — enough to eat the minimum gap to, or overlap, the line after the block. The last line is now pinned to the original block end; inner joins were already exact.
- **EBU STL 35/48/50/60** (#14078 made this reachable): `Ebu.FrameRate` reads these rates from the disk format code, but the export dialog's DFC list stopped at STL30 — opening properties on an STL50 file and pressing OK wrote DFC "STL25.01" with TTI frames computed at 50 fps. The DFC list (and the 35/48 frame-rate entries) now cover what the reader accepts.

Two further findings are reported in the session rather than fixed here, as they are design trade-offs: the #14064 start+duration rounding can create 1 ms overlaps at back-to-back joins in change frame rate / adjust speed (the documented equal-durations choice and exact joins are mutually exclusive under whole-ms rounding), and a theoretical ARM64 store-ordering race on the deferred video load's pending fields (#14048).

Verified: UI + libuilogic build; the 27 GoogleTranslateV1 tests (incl. 2 new) and the 180 Ebu/Netflix/Distribute/Consent-related UI tests pass.

### DIFF
--- a/src/libuilogic/AutoTranslate/GoogleTranslateV1.cs
+++ b/src/libuilogic/AutoTranslate/GoogleTranslateV1.cs
@@ -237,7 +237,14 @@ namespace Nikse.SubtitleEdit.UiLogic.AutoTranslate
                     s = inner[0];
                 }
 
-                s = s.Trim('"');
+                // The parser returns the raw JSON string with its delimiters; strip exactly that
+                // one pair. Trim('"') would also eat an escaped quote ending the translation
+                // ("He said \"hi\"" -> dangling backslash) and make Regex.Unescape throw.
+                if (s.Length >= 2 && s.StartsWith('"') && s.EndsWith('"'))
+                {
+                    s = s.Substring(1, s.Length - 2);
+                }
+
                 try
                 {
                     s = Regex.Unescape(s);
--- a/src/ui/Features/Files/ExportEbuStl/ExportEbuStlViewModel.cs
+++ b/src/ui/Features/Files/ExportEbuStl/ExportEbuStlViewModel.cs
@@ -99,6 +99,13 @@ public partial class ExportEbuStlViewModel : ObservableObject
             "STL25.01",
             "STL29.01 (non-standard)",
             "STL30.01",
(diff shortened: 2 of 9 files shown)


niksedk:
Added a second commit per the maintainer's request: the 1 ms join-overlap from the whole-ms rounding is now fixed too — change frame rate and point sync clip a previous end that rounding pushed past the next start (source overlaps are left as-is; visual sync already had its own fix-up, adjust-all-times is a pure shift). Two tests added; a 50-line back-to-back chain at 25→23.976 manufactures 7 overlaps without the clip.

PR #14086, opened 2026-08-25
Update Italian.json

bovirus:
@niksedk 

Please merge.

Didn't work anymore Tranalte option in Context Translator.

Please enable in Context Translator the view of original English text (with copy function) in the box above translated box, to enable copy/translate/paste in translated box.

Thanks.

PR #14088, opened 2026-08-25
Prepare v5.2.0-beta24

niksedk:
Release prep for v5.2.0-beta24:

- Version bump in `Se.cs` to v5.2.0-beta24.
- `English.json` regenerated — besides the version line this also picks up nine language keys from #14055, #14062 and #14082 (Window-menu items, "List errors...", and the new shortcut display strings) that were never regenerated in, so translators get them now.
- Change-log section covering the 27 PRs merged since the beta23 tag (every merge commit ancestor-checked against v5.2.0-beta23; skipped: #14084 behavior-preserving refactor, #14072 test-only, #14051 fix to a same-window translation update, and #14068 which is folded into the Apple Vision entry — net minimum stays macOS 12).

Merge #14087 (the pre-release fix sweep) before tagging; its fixes are all for in-window regressions, so they add no changelog entries.


niksedk:
Added the change-log entry for #14070 (WhisperX), which merged after the section was compiled. The Italian entry already covers both #14049 and the later #14086. Now 29 PRs in the section's window.

Issue #14092, opened 2026-08-25
Why do I get a lot of unicode during translation


fraternl:
Why do I get a lot of unicode during translation.

The quotes, dashes, the triple dot....
Can they be converted

niksedk:
Details? Screenshots?

niksedk:
Fixed in #14101, will be in the next beta.

Those characters come from DeepL itself: it writes the target language's typography, so plain ASCII input comes back with curly quotes, en dashes and the one-character ellipsis. I checked whether DeepL can be asked not to - its `preserve_formatting` option makes no difference at all, the output is identical with and without it - so the conversion has to happen on our side.

Fix common errors -> "Normalize strings" already turned the en dash into a hyphen. It now also converts:

- curly single and double quotes to straight ones, and the typographic apostrophe to '
- the one-character ellipsis to three dots
- prime and double prime to ' and ", non-breaking hyphen to -

So after translating, run Fix common errors with "Normalize strings" ticked and the text comes out plain.

One exception: German and the languages around it (cs, sk, pl, sl, hr, sr, hu, lt, et, ro, is, bg, ka, sq) quote with the low-9 marks - those are the language's own quotation marks, so they are kept there. Dutch is not in that group, so a Dutch translation gets the straight quotes.

GitGianluc:
@niksedk for some tasks relating to the education sector, it would be extremely useful if the ‘Fix common errors’ also included the reverse transformation – a sort of "Normalize strings" opposite. If you’re interested, I’m happy to share some of my “Multiple replace” RegEx rules taken from my personal template that are suitable for handling quotation marks (this exact <ins>order of execution is mandatory</ins>):
- `(^|\s)[E]\'` → `$1È` (find mispelled Latin Capital Letter E with Grave = U+00C8 at the beginning of a line or preceded immediately by a whitespace and replace with the correct glyph)
- `(^|[\s([{«\"\'])\"` → `$1“` (find a **double quote** at the beginning of a line or preceded immediately by a whitespace / an opening bracket / another quotation mark and replace with the **opening curly double quote**)
- `\"` → `”` (find the **remaining double quotes** and replace with the **closing curly double quotes**)
- `(^|[\s([{«\"\'])\'` → `$1‘` (find a **single quote** at the beginning of a line or preceded immediately by a whitespace / an opening bracket / another quotation mark and replace with the **opening curly single quote**)
[...]

PR #14111, opened 2026-08-25
Italian OCR fix list: E' with apostrophe to È, gated on the dictionary

niksedk:
Fixes the classic Italian typo / OCR misread `E'` (capital E + apostrophe standing in for È), as suggested in [this comment on #14092](https://github.com/SubtitleEdit/subtitleedit/issues/14092#issuecomment-5412577294).

**v2 (current):** two ungated line regexes in `ita_OCRFixReplaceList.xml`:

```xml
<RegEx find="(^|[\s&quot;«(\[\-–—…])E['’](?![\p{L}\p{N}])" replaceWith="$1È" />
<RegEx find="(\p{L}['’])E['’](?![\p{L}\p{N}])" replaceWith="$1È" /> <!-- COS'E' -> COS'È -->
```

- **Ungated**, so it works on installs without an it_IT Hunspell dictionary (only `en_US.dic` ships) and in seconv.
- **Explicit left context** (line start, whitespace, `"`, `«`, brackets, dashes, ellipsis) so a quoted letter — `la lettera 'E'` — is never touched; the first version's `\b` matched there and ate the closing quote.
- The second rule recovers the elision case (`COS'E'` → `COS'È`) without reopening the quoted-letter hole: an elision apostrophe is always preceded by a letter, an opening quote never is.
- The lookahead keeps glued elisions (`E'sti`), decades (`'80`) and `E'ndrangheta`-style words safe. Covers straight `'` and typographic `’`.

Verified over a 20-sentence Italian battery through the real engine, with and without a dictionary — all `E'`/`E’` forms fixed; quoted letters, elisions, and the trailing-apostrophe family (`perche'`, `po'`) untouched. See the review comment below for the full comparison that led to this shape.

Tests: 6 positive cases, a no-dictionary positive, and 5 new negative rows in the existing left-alone theory; all 24 tests in `OcrFixItalianAccentAndCapsTests` pass.

### DIFF
--- a/Dictionaries/ita_OCRFixReplaceList.xml
+++ b/Dictionaries/ita_OCRFixReplaceList.xml
@@ -114,6 +114,14 @@
     <RegEx find="\b([a-zàèéìòóù]{2,})u\b" spellCheck="$1ù" replaceWith="$1ù" />
 
     <!-- ============================================================ -->
+    <!-- E' SCRITTO CON APOSTROFO AL POSTO DI È                       -->
+    <!-- E' un problema in È un problema (issue #14092). Copre sia    -->
+    <!-- l'apostrofo dritto sia quello tipografico; non tocca la E    -->
+    <!-- seguita da un'elisione staccata (E 'sti soldi).              -->
+    <!-- ============================================================ -->
+    <RegEx find="\bE['’](?![\p{L}\p{N}])" spellCheck="È" replaceWith="È" />
+
+    <!-- ============================================================ -->
     <!-- SPAZIO PERSO TRA PREPOSIZIONE/ARTICOLO E NOME PROPRIO        -->
     <!-- daRoma in da Roma : solo se la parola con maiuscola è nel    -->
     <!-- dizionario.                                                  -->
--- a/Dictionaries/ita_OCRFixReplaceList.xml
+++ b/Dictionaries/ita_OCRFixReplaceList.xml
@@ -64,7 +64,12 @@
     <RegEx find="\b\|\b" replaceWith="I" />
     <RegEx find="([\p{Ll},] )(I)([oai][, \.])" replaceWith="$1l$3" />
     <RegEx find="\b(I)([àì]|'[ \r\n])\b" replaceWith="l$2" />
-    <RegEx find="([\p{Ll},] )(II)\b" replaceWith="$1Il" />  
+    <RegEx find="([\p{Ll},] )(II)\b" replaceWith="$1Il" />
+    <!-- E' scritto con apostrofo al posto di È (issue #14092). Il contesto sinistro esplicito
+         evita di mangiare la virgoletta di chiusura di una lettera citata ('E' resta 'E');
+         il lookahead protegge le elisioni attaccate (E'sti). -->
+    <RegEx find="(^|[\s&quot;«(\[\-–—…])E['’](?![\p{L}\p{N}])" replaceWith="$1È" />
+    <RegEx find="(\p{L}['’])E['’](?![\p{L}\p{N}])" replaceWith="$1È" /> <!-- COS'E' in COS'È -->
   </RegularExpressions>
   <RegularExpressionsIfSpelledCorrectly>
     <!-- Italic l misread as i (#13660): trailing i to l, double ii to ll.  -->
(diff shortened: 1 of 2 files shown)


niksedk:
@GitGianluc: Or should we use `<Word from="E'" to="È"/>` instead?

niksedk:
*I am Claude (an AI assistant), posting this review summary at the maintainer's request.*

A code review of the first version of this PR found a confirmed bug and two design problems, so the rule was reworked. The reasons, with the empirical comparison that drove the final shape:

**1. The original gated regex (`\bE['’](?![\p{L}\p{N}])`) ate closing quotes.** `\b` matches after an apostrophe, so the closing mark of a quoted letter matched too: `Ha detto la lettera 'E'.` → `Ha detto la lettera 'È.` The dictionary gate could not prevent this — it compares the literal token `E'`, which is never a dictionary word.

**2. The gated section made the fix invisible to most users.** `RegularExpressionsIfSpelledCorrectly` only runs with an installed Italian Hunspell dictionary, and SE ships only `en_US.dic`. A fresh install would have seen no fix at all — the same failure class as #12126. The gate was also degenerate here: `spellCheck="È"` is a constant, so it never did per-occurrence validation like the section's other rules.

**3. A `WholeWords` entry (`<Word from="E'" to="È"/>`, the French `e'`→`é` precedent) was tested as the replacement — and rejected on evidence.** The word tokenizer treats a trailing apostrophe as part of the word, so `'E'.` tokenizes as `'` + `E'` + `.` and the entry eats the closing quote exactly like the regex did. It also cannot see the typographic `E’` (the `’` is not a word character, so the token is just `E`).

Results of the same sentences run through the real OCR fix engine under all three variants (✅ correct, ❌ corrupted/missed):

| Input | v1 gated regex | WholeWords | **v2 (merged)** |
|---|---|---|---|
| `E' un problema.` | ✅ `È un problema.` | ✅ | ✅ |
| `E’ tardi.` (typographic) | ✅ `È tardi.` | ❌ unchanged | ✅ |
| `Ha detto la lettera 'E'.` | ❌ `…lettera 'È.` | ❌ `…lettera 'È.` | ✅ unchanged |
| `La lettera ‘E’ canta.` | ❌ `…‘È canta.` | ✅ unchanged | ✅ unchanged |
| `COS'E' successo?` | ✅ `COS'È…` | ❌ unchanged | ✅ `COS'È…` |
| `E'sti soldi` (glued elision) | ✅ unchanged | ✅ unchanged | ✅ unchanged |
| `Negli anni '80` | ✅ unchanged | ✅ unchanged | ✅ unchanged |
| **Without an it_IT dictionary** | ❌ does nothing | ✅ works | ✅ works |

**v2** keeps the line-stage regex (the only stage with left context and `’` visibility) but makes it ungated and replaces `\b` with explicit left context, plus a second rule for the elision case — the only variant with zero corruptions in the battery, full apostrophe coverage, and no dictionary requirement.

Deliberately out of scope, noted for follow-up: lowercase `e'` → `è`, and the trailing-apostrophe accent family (`perche'` → `perché`, `citta'` → `città`) — the same convention on longer words, which would pair with the existing gated accent rules.

GitGianluc:
@niksedk I’m not sure I know what `<Word…/>` means in this context. If it means that any word containing `E'` must be replaced with ‘È’, that’s not right: there are words in Italian – such as ‘perché’ (= why / because) – which have an acute accent on the last ‘e’. So PERCH**E'** -> PERCH**È** won’t be correct

niksedk:
> @niksedk I’m not sure I know what `<Word…/>` means in this context. If it means that any word containing `E'` must be replaced with ‘È’, that’s not right: there are words in Italian – such as ‘perché’ (= why / because) – which have an acute accent on the last ‘e’. So PERCH**E'** -> PERCH**È** won’t be correct

Claude review'ed itself, and it sounds correct to me now... with 3 regular expressions instead.

GitGianluc:
OK, v2 works much better than “WholeWords” version… Claude is so incredibly clever and well-read! :)

niksedk:
> OK, v2 works much better than “WholeWords” version… Claude is so incredibly clever and well-read! :)

He, he. Especially when it reviews itself.
But surely amazing how fast AI can implement advanced stuff. Still needs a bit direction and testing - for now...

GitGianluc:
@niksedk whether Claude is any good or not, thank you for taking charge of this improvement to the OCR’s functionality.

If you think it might be worth it, feel free to use the [RegEx expressions (#14092)](https://github.com/SubtitleEdit/subtitleedit/issues/14092#issuecomment-5412577294) relevant to the reverse process of “Normalize strings” with regard to single or double quotation marks.

PR #14129, opened 2026-08-25
seconv: add `seconv mcp`, a Model Context Protocol server over stdio

ivandrofly:
## Summary

- New subcommand **`seconv mcp`** runs seconv as an [MCP](https://modelcontextprotocol.io) server over stdio, so AI clients (Claude Desktop/Code, Cursor, ...) can inspect and convert subtitles without a shell: `{"mcpServers":{"seconv":{"command":"seconv","args":["mcp"]}}}`.
- Seven tools, each a thin adapter over the Core helpers the CLI subcommands already use (`SubtitleInfoGatherer`, `SubtitleLinter`, `SubtitleConverter`, `FixCommonErrorsRunner`, ...), so behaviour cannot drift between the two entry points:
  - `list_formats` (optional substring filter), `subtitle_info`, `read_subtitle` (paged paragraphs of any supported format), `lint_subtitle`
  - `convert_subtitle` — format, output folder/name, encoding, overwrite, offset, fps/targetFps, renumber, duration/gap adjustments, delete first/last/contains, `operations`, FixCommonErrors / RemoveFormatting rule specs, container track numbers, OCR engine/language, `timeCodesOnly`, resolution
  - `list_fix_common_errors_rules`, `list_remove_formatting_rules`
- Stdout is reserved for JSON-RPC: `Console.Out` is redirected to stderr and the converter runs quiet; logs go to stderr (`--verbose` for debug level).
- Tool failures return `isError` with the real message (unknown operation, missing file, ...) rather than the SDK's generic placeholder; unknown operations are hard errors, not silent no-ops, matching the CLI's strict parsing.
- Result JSON uses the same field names as the corresponding subcommand's `--json` output.
- `mcp` is listed in `--help`, `--help-json`, the seconv README and `docs/reference/command-line.md`.
- Dependencies: `ModelContextProtocol` 2.2.0, `Microsoft.Extensions.Hosting` 10.0.11 (no trimming/AOT, so reflection-based tool discovery is unaffected by the single-file publish).

## Test plan

- [x] `dotnet build src/seconv/SeConv.csproj` — clean, no warnings
- [x] `dotnet test tests/seconv/SeConvTests.csproj` — 393/393 pass, including 10 new `McpToolsTest` cases (one spawns `dotnet seconv.dll mcp`, lists tools via the MCP client SDK, calls `subtitle_info` on a fixture and on a missing file)
- [ ] Register in Claude Code with `claude mcp add seconv -- <path-to>/seconv mcp`, then ask it to describe an `.srt` and convert it to WebVTT with a 1 s offset — the output file appears and its time codes are shifted
- [ ] `seconv --help` and `seconv --help-json` list `mcp`
- [ ] Existing conversions and subcommands behave exactly as before (no change to any non-`mcp` code path)

https://claude.ai/code/session_01V1UEPy85BqcK2D9dHgfymB

### DIFF
--- a/docs/reference/command-line.md
+++ b/docs/reference/command-line.md
@@ -84,6 +84,7 @@ seconv list-rf-rules        # list remove-formatting rule IDs
 seconv dump-settings        # print a full --settings JSON with libse defaults
 seconv info <file>          # print format/encoding/duration/language for a file
 seconv lint <pattern>       # validate subtitle(s); exit 1 if issues found
+seconv mcp                  # run as a Model Context Protocol server over stdio
 seconv --help               # show help (same text as -h, /? and /help)
 seconv --help-json          # print the whole command-line schema as JSON
 seconv --version            # print version and exit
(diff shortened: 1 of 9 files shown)


niksedk:
Thanks for the careful work here — it merges cleanly onto today's main, builds with zero warnings, the stdio test passes, and the stdout/stderr handling and error reporting are done right.

I'm going to close it anyway, because the intended AI surface for seconv is the CLI itself plus `seconv --help-json` and the `--json` subcommand output. Shell-capable agents (Claude Code, Codex, Cursor agent mode) read the schema once and then get every option; a hand-mirrored MCP tool set can only ever be a subset.

That subset has already drifted. `convert_subtitle` maps 28 of the 60 conversion options, and the three options added to the CLI since this PR opened (`--ocr-prompt`, `--translate-prompt`, `--output-filename-append`) aren't in it. Everything translate-related, the OCR URL/model/prompt options, `--settings`, PAC/EBU/ASSA/custom-format/multiple-replace files, teletext and plain-text options are all missing — and those are the parts where an agent most needs the full engine. Each future option would need a second parameter, description and test to stay in sync.

The dependency cost also matters for a single-file tool: ModelContextProtocol + Microsoft.Extensions.Hosting add 31 DLLs (~3 MB) and the whole config/DI/logging stack to run one stdio loop.

If MCP support for shell-less clients (Claude Desktop etc.) turns out to be wanted, I'd rather see it as a small separate project that generates its tool schema from `CliSchema` and exposes a single "run seconv with these args" tool, so there's one source of truth — but not as part of the seconv binary.

Issue #14131, opened 2026-08-26
Improve Video OCR accuracy by reusing VideOCR's pipeline


caiweihan:
# Improve Video OCR accuracy by reusing VideOCR's pipeline

Thanks for implementing burned-in subtitle OCR in #12270, based on the request in #10944.

I tested Subtitle Edit's current Video OCR against [[VideOCR](https://github.com/timminator/VideOCR)](https://github.com/timminator/VideOCR) on the same videos. Although both can use PaddleOCR, VideOCR is noticeably more accurate in practice.

**Note:** The possible reasons below were identified by GPT after comparing the source code of Subtitle Edit and VideOCR, so they are implementation-based hypotheses and may need verification.

The main differences seem to be:

- **Frame selection:** Subtitle Edit groups similar frames and OCRs the middle frame, while VideOCR tries to select a better frame using text-detection confidence.
- **Text-aware filtering:** VideOCR first runs PaddleOCR text detection and compares the actual detected subtitle regions, instead of relying mainly on the whole subtitle ROI.
- **More precise similarity checking:** VideOCR performs additional SSIM comparison on tight text bounding boxes, reducing interference from changing backgrounds.
- **Temporal redundancy:** Subtitle Edit discards most repeated frames before OCR, while VideOCR can retain multiple useful OCR observations from the same subtitle.
- **Voting/confidence:** VideOCR can choose the final text using repeated OCR results and recognition confidence, which helps correct occasional character errors.
- **Bounding-box processing:** VideOCR reconstructs text lines and reading order using word positions.
- **Post-processing:** VideOCR also contains additional language-specific text cleanup.

Because #12270 was already inspired by VideOCR, I wonder if it would be better to **directly reuse or port the relevant VideOCR source code**, where licensing permits, instead of independently reimplementing only part of its pipeline.

This could make Subtitle Edit's Video OCR behavior much closer to VideOCR and likely improve accuracy substantially.

niksedk:
I will look into it.  
SE have multiple OCR engines so I would not go for a switch to that engine, but I'm pretty sure we can improve the SE video OCR with your suggestions :)
SE does have this post-processing already from normal OCR.

niksedk:
Should be much improved in next beta :)

See:
- https://github.com/SubtitleEdit/subtitleedit/pull/14138
- https://github.com/SubtitleEdit/subtitleedit/pull/14143
- https://github.com/SubtitleEdit/subtitleedit/pull/14146

[...]

PR #14138, opened 2026-08-26
Video OCR: fix brightness-mask fidelity, adopt VideOCR's confidence/masking techniques (#14131)

niksedk:
Addresses #14131 — measured against real footage with real SRTs burned in for exact ground truth, and head-to-head against [VideOCR](https://github.com/timminator/VideOCR) itself.

### The dominant defect was ours, not a missing VideOCR stage

The frame grouper thresholded brightness on a 96px-wide thumbnail. At that scale, anti-aliasing blends white glyphs with their dark outlines to below any sensible brightness minimum — whole subtitles read as "blank" and were never OCR'ed, and the surviving masks were noise-dust that shattered every subtitle into single-frame groups the minimum-duration filter then deleted. On real footage **51–63% of all subtitles vanished**. The mask is now thresholded at ~360px (where glyph cores survive) and max-pooled down to the comparison grid.

### Adopted from comparing with VideOCR

- **Recognition confidence is used** (it was parsed and thrown away): regions under 75% are dropped as clutter in Video OCR (never when nothing clears the bar — that erased short subtitles the engine hesitated on), and the line builder's majority vote weights each observation by duration × confidence.
- **Paddle's OCR input is brightness-masked** (VideOCR's `brightness_threshold`): its detector reads darker scene text — a shirt print polluted 10+ of 57 subtitles at 0.99 confidence, untouchable by any confidence cut, while sitting well below the brightness minimum. Paddle path only; Apple Vision measured better on natural frames.
- **Line reconstruction by vertical-midpoint overlap** instead of the average-height walk, plus right-to-left word order for Arabic-script languages.
- **Junk-bridging in the line builder**: one scene-text-polluted observation ("14 Wait." between clean "Wait." reads) no longer severs a subtitle into sub-minimum fragments that silently vanish.

### Measured (240s clips of real footage + a full 21-min episode)

| clip / engine | before | after | VideOCR (same engine family) |
|---|---|---|---|
| ep. clip A, Apple Vision | 16/57 exact, 36 missed, CER 0.69 | **47/57, 0 missed, CER 0.045** | — |
| ep. clip A, PaddleOCR | — | **42/57, 0 missed, CER 0.035** | 34/57, 0 missed, CER 0.049 |
| ep. clip B (yellow subs), Vision | 32/77 exact, 39 missed | **67/77, 0 missed, CER 0.017** | — |
| stress clip, PaddleOCR | — | **21/24, 0 missed, CER 0.006** | 21/24, 0 missed, CER 0.009 |
| full 21-min episode, Vision | — | **238/270 exact, 2 missed, CER 0.038, ±72ms** | — |

Start-time error dropped from ~1.8s to ~70–90ms. (VideOCR comparison ran through its own pipeline with pip PaddleOCR behind a CLI-compatible shim, since it has no macOS build.)

Remaining known limitations are shared with VideOCR: back-to-back near-identical cues merge, and genuine scene text inside the scan area (credits, ad boards) can pollute lines. A possible follow-up is wiring the existing OcrFixEngine (replace lists + spell check) into Video OCR results.


niksedk:
Added timing refinement (7c600b55a): the scan grid quantized every boundary to 200ms bins at the default 5 fps. After line building, each subtitle's two boundaries are re-extracted at 50 fps (crop only, no OCR — two tiny ffmpeg calls per subtitle) and moved to the frame where the brightness mask starts/stops matching the subtitle's reference frame.

Measured on real footage: average start/end error 88/74ms → **58/36ms** and 64/60ms → **50/40ms** on the 4-min clips; **72/70ms → 51/40ms** on the full 21-minute episode for +20s wall time (562 boundaries). This closes the gap to VideOCR's timing (33/38ms on the same clip — and that residual is a consistent +33ms late bias, while SE's is symmetric jitter).

niksedk:
Two more improvements (1205a2209):

- **Apple Vision scene-text filter**: Vision's per-observation boxes are checked against the brightness minimum — sub-threshold scene text (shirt prints, dark credits) is dropped from the results while the subtitle is untouched. This is the Vision counterpart of the Paddle input masking (masking Vision's input measurably hurt it — it wants natural frames). Full 21-min episode: exact lines 238/270 → **247/270 (91%)**, CER 0.038 → **0.016**.
- **Timing on real frame timestamps**: the refiner now reads each boundary-window frame's actual PTS from ffmpeg's `showinfo` instead of assuming a uniform 50 fps grid (which was off by up to one source frame, and wrong for VFR). A subtitle now ends exactly where the first frame without it begins. Episode boundary error: **49/39ms** with bias down to +19/−7ms — at the noise floor of the burned-in ground truth itself (libass renders a cue on the first frame at-or-after its start time, ~±17ms).

Follow-up candidates left for separate PRs: wiring OcrFixEngine (replace lists + spell check) into Video OCR results, and a dictionary/language picker for it.

PR #14148, opened 2026-08-26
Video OCR: retry other frames of a group when OCR reads nothing

niksedk:
Detection follow-up to #14131, measured against the same burned-in ground truth.

An empty OCR result on a group whose brightness mask says text is on screen is often just an unlucky representative frame — the one subtitle a 21-minute episode lost was white text over a white wall, unreadable on the group's middle frame but read perfectly from the frame at ¾ of the group (verified frame by frame). When the result is empty and the group spans ≥3 sampled frames, the per-frame engines now try the ¾ and ¼ frames before giving up.

Episode (Apple Vision): missed 2 → **1** (the survivor is cut off by end-of-file), exact 247 → **248**/270, CER 0.016 → **0.013**, +3 s wall. Short-clip fixtures unchanged.

Also measured and *not* changed: raising the scan rate to 10 fps only bought +1 exact line and ~20 ms timing for **+60% OCR calls** — wrong trade for a default, left as a tip for the fast engines.

### DIFF
--- a/src/ui/Features/Video/VideoOcr/VideoOcrFrameGroup.cs
+++ b/src/ui/Features/Video/VideoOcr/VideoOcrFrameGroup.cs
@@ -28,6 +28,16 @@ public class VideoOcrFrameGroup
     /// </summary>
     public double Confidence { get; set; } = 1.0;
 
+    /// <summary>
+    /// File name of another sampled frame belonging to this group, by its 0-based frame
+    /// index (the extraction writes "img000000.jpg" ... into one folder).
+    /// </summary>
+    public string GetSiblingFrameFileName(int frameIndex)
+    {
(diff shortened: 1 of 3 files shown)


niksedk:
Second commit adds two-frame verification: a non-empty read on a ≥3-frame group is checked against a second frame. Near-identical reads are arbitrated by the spell check (fewer unknown words wins — fixes the "I'think"-class jitter when a dictionary is loaded); dissimilar short reads are dropped as hallucination ghosts (15 → 13 phantom lines on a GLM-OCR stress clip; the survivors consistently read a real ad board, which no frame comparison can reject); long reads always survive disagreement so credit-polluted lines and unreadable-verify-frame cases lose nothing. Accuracy-neutral for Apple Vision on all fixtures, ~+10% OCR calls.

Issue #14167, opened 2026-08-26
Subtitle Edit 5.2.0 Beta 24 Missing Features and Playback Issues


kadrimarzouki:
### A. Shortcut for “Go to next line from current position, highlight the text, play, and pause”

Where can I find the shortcut for **“Go to the next line from the current position, highlight the text in the text box, play it, and return the cursor to the beginning of the subtitle”**?

This exists in Subtitle Edit 4.0.16 and previous versions. Whenever I press **Alt + Down Arrow**, the next line is played, the text is selected in both the text box and the list, and the cursor returns to the beginning of the subtitle in the waveform.

This is useful because I can immediately start typing, and the text changes on the video as I type.

I am not able to do this in **Subtitle Edit 5.2.0 Beta 24**.

### B. Double-click playback behavior

Another matter: when **double-clicking a line in the ListGrid**, it starts playing and continues playing after the selected line, even though **“Pause after playing”** is enabled.

This might need to be looked at.

### C. Auto Repeat option

In Subtitle Edit 4.0, the **Translate/Create/Adjust** section contains an **Auto Repeat** option, which could be set to **0** to prevent the video from autoplaying after pressing **Alt + Up Arrow** or **Alt + Down Arrow**.

Where can I find this option in Subtitle Edit 5.0?

### D. Toggle parentheses and quotation marks

I also cannot find a shortcut for toggling **parentheses `( )`** around selected text, or **quotation marks `" "`** around selected text.

Was this shortcut removed, or is it located somewhere else in Subtitle Edit 5.0?

### E. Left/Right Justification of Subtitles in Video

Please integrate the subtitle left/right justification option added by @OmrSi in 2021.

In Subtitle Edit 4.0, this can only be enabled manually through the settings file using:

<MpvExtraOptions>--sub-ass-justify=yes --sub-justify=right</MpvExtraOptions>

This option is not yet available in Subtitle Edit 5.2.0 Beta 24. It would be useful to have it integrated into the settings/interface rather than requiring manual editing of the settings file.

@niksedk, @OmrSi

niksedk:
[...]

PR #14196, opened 2026-08-27
Add mpv preview line justification setting (#14167)

niksedk:
Item **E** of #14167: left/right justification of the subtitle in the video preview, which in SE4 was only reachable by hand-editing `<MpvExtraOptions>--sub-ass-justify=yes --sub-justify=right</MpvExtraOptions>`.

Alignment was already there (all 9 ASS positions), so this adds the missing part next to it: **Options → Settings → Video → mpv preview → "Justify lines"** (Auto / Left / Center / Right).

Alignment moves the whole text block and rides along in the generated ASS style header. Justification only decides where the shorter lines sit *inside* that block, and it has no ASS style field — it goes to mpv as `sub-justify`, together with `sub-ass-justify`, which mpv defaults to `no` and which is what otherwise keeps the option from reaching the preview (the preview is an ASS subtitle).

- Default is `auto`, i.e. mpv's own behaviour — nothing changes unless the user picks something.
- Both options are written on every call, so switching back to *Auto* restores mpv's defaults instead of leaving the last value stuck.
- Applied on every preview refresh as well as on load, so changing it takes effect on the video that is already open — same as "margin is part of the subtitle area".

mpv preview only; VLC and burn-in/export are untouched.

### DIFF
--- a/src/ui/Features/Options/Settings/MpvJustifyDisplay.cs
+++ b/src/ui/Features/Options/Settings/MpvJustifyDisplay.cs
@@ -1,0 +1,32 @@
+﻿using Nikse.SubtitleEdit.Logic.Config;
+
+namespace Nikse.SubtitleEdit.Features.Options.Settings;
+
+/// <summary>
+/// One value of mpv's "sub-justify" option - how the lines of a multi-line preview subtitle are
+/// justified inside the text block. "auto" leaves it to the alignment, which is mpv's own default.
+/// </summary>
+public class MpvJustifyDisplay
+{
+    public string Code { get; }
+    public string DisplayName { get; }
+
+    public MpvJustifyDisplay(string code, string displayName)
+    {
+        Code = code;
+        DisplayName = displayName;
+    }
+
+    public override string ToString() => DisplayName;
+
+    public static MpvJustifyDisplay[] GetAll()
+    {
+        return
+        [
+            new MpvJustifyDisplay("auto", Se.Language.General.Auto),
+            new MpvJustifyDisplay("left", Se.Language.General.Left),
+            new MpvJustifyDisplay("center", Se.Language.General.Center),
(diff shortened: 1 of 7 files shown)


GrampaWildWilly:
You mean this was being ignored?

GrampaWildWilly:
Environment:  SE 5.2.0 beta 26, controls undocked, Windows 11 Pro 64-bit version 25H2 build 26200.9278.

In an excess of energy, I went through all 9 entries in the drop-down list I show in the image in my post immediately above this.  They do all appear to work.  I'll spare you the screenshots.  You'll have to trust me that I was able to position my subtitles in all 9 locations on the video player.  I was using a plain VTT that goes along with an opera video (MP4) of a bit over 3 hours duration.  Every time I selected a new position, then clicked Apply, then OK, the video player immediately started showing the subtitles in the specified area of the screen.  What exactly did you add?  From where I'm sitting, the functionality is already there.  Or is there something else I'm missing?

A couple of odd things I did notice.  The first few times I repositioned the subtitles, Apply/OK made the video player & the Audio visualizer rewind to time index 0.  However, the selected subtitle in the grid was where it had been before I opened the Settings dialog to choose another position for the subtitles.  So a quick double click in the grid resynced all 3 windows to the previous time index.  Well, actually, close to that time index.  I paused playback a bit after the most recently displayed subtitle so the double click actually did a small, harmless rewind.  But after I had done 3 or 4 of these repositions, that stopped happening.  Apply/OK left all 3 of the grid, the video player, & the Audio visualizer at the same time index in the middle of the opera, where it was when I paused playback in order to make the Settings change.

But there was another anomaly.  Apply/OK sent the video player into foreground.  I had the main window in foreground when I paused playback & opened Settings.  But Apply/OK shifted the video player into foreground.  A simple Alt+Tab sent the main window back to foreground, which is what I wanted so I could open Settings again & try the next subtitle position.

PR #14236, opened 2026-08-29
Center the grid on every prev/next row again (#14231)

### LINKED ISSUE #14231: Subtitle grid scrolling and centering issues during consecutive navigation
1. 

https://github.com/user-attachments/assets/bceb706f-864a-4749-9045-98564550ccdd

The setting "Subtitle grid, center when selecting prev/next row" is checked, but the grid fails to center when navigating downwards consecutively.

2. I haven't noticed what triggers it, but there's still an issue where the list view suddenly jumps straight to the bottom

niksedk:
Fixes #14231 (part 1).

## The bug

With **Subtitle grid, center when selecting prev/next row** on, the grid only centered when the row being selected happened to be *off screen*. Stepping down through a file therefore did nothing at all until the selection had walked from the middle of the view all the way to the bottom edge — at which point the view jumped a half screen and the walk started over. That is the stutter in the reporter's video.

## Causes

**1. The already-visible guard swallowed the navigation case.**
`SelectAndScrollToRow` skips all scrolling for a row that is already fully visible. That guard was added for the delete/insert commands in 447dc7aa8 — the line after a deleted one is visible right below it, and re-centering for it moves the rows out from under the user. It should never have covered prev/next-line navigation, which is precisely what the setting is about.

The guard stays exactly where it was. `SelectAndScrollToRowCentered` is the navigation entry point, used by go-to-previous/next-line (plain, "and set video position", and "from video position") and by plain PageUp/PageDown.

**2. Plain arrow keys never reached the centering code at all.**
Up/Down in the grid falls through to TableView's own list navigation, which scrolls only when the next row is off screen — so the setting did nothing for the most common way of moving between lines. The grid key handler tunnels, so it now posts the centering to run after the built-in navigation has moved the selection.

## Tests

`tests/UI/Features/Main/SubtitleGridCenterSelectedRowTests.cs` — six headless tests. Four fail on `main` (the selected row drifts one row height per step: 21, 43, 65, 87, … px away from the middle) and pass here. The other two are regression guards: with the setting off the arrow key must not move the offset, and delete with the setting **on** must still leave the offset alone.

Full UI suite passes.

## Not covered

The second half of #14231 — "the list view suddenly jumps straight to the bottom" — is not addressed. The reporter has not identified a trigger and I could not reproduce it (arrow-key stress runs of 200 steps over 2000 lines, and the Ctrl+R case from #13579, all stay put). It needs its own repro.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -15540,7 +15540,7 @@ public partial class MainViewModel :
             return;
         }
 
-        SelectAndScrollToRow(idx);
+        SelectAndScrollToRowCentered(idx);
 
         if (WaveformCenter)
         {
@@ -15600,7 +15600,7 @@ public partial class MainViewModel :
             return;
         }
 
-        SelectAndScrollToRow(idx);
+        SelectAndScrollToRowCentered(idx);
 
         if (WaveformCenter)
         {
@@ -15632,7 +15632,7 @@ public partial class MainViewModel :
             return;
         }
 
-        SelectAndScrollToRow(idx);
+        SelectAndScrollToRowCentered(idx);
         vp.Position = Subtitles[idx].StartTime.TotalSeconds;
 
         if (AudioVisualizer != null && AudioVisualizer.WavePeaks != null)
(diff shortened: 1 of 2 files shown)


St0rmXtr00per:
I've downloaded the newest beta (28) with this fix listed but the issue persists

PR #14258, opened 2026-08-29
Center the grid for the last prev/next-line paths too (#14231)

niksedk:
Follow-up to #14236. That PR taught the six go-to-previous/next-line commands and the grid's arrow/page keys to re-center on every step with **"Subtitle grid, center when selecting prev/next row"** on. The reporter of #14231 says it still misbehaves on beta 28 — and the fix *is* in beta 28 (the tag's commit includes it), so three more ways of stepping a line at a time were simply left on the old behaviour: they scrolled only once the target row had walked off the bottom of the view, and then jumped half a screen.

### What was still missing

| Path | Before | After |
| --- | --- | --- |
| `PlayNextParagraph` / `PlayPreviousParagraph` — "play next/previous line (and stop/loop)" | sets the grid's `SelectedItem` + `ScrollIntoView` itself, so it never reached the centering in `SelectAndScrollToRow` | new `CenterOrEnsureRowVisibleInSubtitleGrid` follow-up |
| `GoToNearestTimeCode` — "go to next/previous time code" | `SelectAndScrollToRow` | `SelectAndScrollToRowCentered` |
| "set end time and go to next" (position timer key-up handling) | `SelectAndScrollToRow` | `SelectAndScrollToRowCentered` |

The play commands are the interesting ones: with a video open, SE 4's Alt+Down default — "go to next subtitle (play translate)" (#14167) — runs `PlayNextParagraph`, i.e. the line-by-line translate/sync loop where this setting matters most. They move the selection synchronously on purpose (a posted selection would be nulled by the `SelectionChanged` handler's `ResetPlaySelection` and break stop/loop), so they get the same *posted* scroll follow-up `SelectAndScrollToRow` does: center when the setting is on, otherwise only nudge the row fully into view.

### Tests

Three cases added to `SubtitleGridCenterSelectedRowTests`, each measuring the selected row's middle against the viewport's middle over 8–16 steps. All three fail on `main` with the pre-fix signature (a clean drift of one row height per step) and pass with the change.

Note: `LibMpvEventLoopTests.Pause_RightAfterSeek_KeepsReportingTheSeekTarget` fails on a clean checkout of `main` as well — pre-existing and unrelated to this change.

The second symptom in #14231 ("the list view suddenly jumps straight to the bottom") is not addressed here; it still has no known trigger.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -1470,6 +1470,7 @@ public partial class MainViewModel :
         // Change selection first, then assign _playSelectionItem.
         SubtitleGrid.SelectedItem = next;
         SubtitleGrid.ScrollIntoView(next);
+        CenterOrEnsureRowVisibleInSubtitleGrid(next);
         vp.Position = next.StartTime.TotalSeconds;
         PinPlayheadTo(next.StartTime.TotalSeconds);
         _playSelectionItem = new PlaySelectionItem(new List<SubtitleLineViewModel> { next }, next.EndTime, loop);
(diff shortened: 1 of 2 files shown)


St0rmXtr00per:
To scroll down I use the following shortcut:

Video - Everywhere - Play next (and stop) **Alt + Down** 

As for the second issue (scrolling to the very bottom of the list), it's a completely separate issue. I will report when I notice what triggers it

PR #14308, opened 2026-08-30
Updated french.json for beta29.

Need74:
Hello

Added missing lines and tweaked a few imperfect phrases.

I also noticed that the new OCR function for decoding hard coded subtitles isn't working very well (tested with several OCR engines). Additionally, when downloading the engine for the first time, untranslatable messages appear in English.

Best regards,


niksedk:
Ocr of burned-in subtitles works well here... can you upload a small video sample?

PR #14329, opened 2026-08-31
feat: handle colors for imsc rosetta format

Scryper:
Hello,

I did a small improvement in `TimedTextImscRosetta` subtitle format to be able to keep font colors.
In my case it was to convert from EBU STL to TTML-Rosetta.

Best regards

### DIFF
--- a/src/libse/SubtitleFormats/TimedTextImscRosetta.cs
+++ b/src/libse/SubtitleFormats/TimedTextImscRosetta.cs
@@ -352,6 +352,17 @@ namespace Nikse.SubtitleEdit.Core.SubtitleFormats
                     ConvertParagraphNodeToTtmlNode(child, ttmlXml, span);
                     ttmlNode.AppendChild(span);
                 }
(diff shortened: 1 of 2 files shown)


niksedk:
Thanks a lot for the contribution and for flagging the missing color support! 🙏

After checking against the [IMSC Rosetta spec](https://github.com/imsc-rosetta/imsc-rosetta-specification/blob/main/documents/styles.md), inline `tts:color` on spans unfortunately isn't valid Rosetta — all styling in `<body>` must be referential via the fixed `s_fg_xxx` style ids (that's also what lets Rosetta→STL/PAC converters map colors back to teletext, which matters for your EBU STL use case). The verbatim color value was also a problem: EBU import emits capitalized names like `Yellow`, which isn't a valid TTML color.

This is superseded by #14337, which snaps font colors to the nearest of the eight fixed `s_fg_xxx` styles, handles colors spanning line breaks, and fixes reading colors from spec-compliant files too. Closing this one — thanks again for getting the ball rolling!

PR #14345, opened 2026-08-31
Prevent duplicate macOS launch and Matroska opens

marcpbailey:
## Summary

On macOS, opening a file as Subtitle Edit launches can deliver the same path through two independent routes: the command-line startup argument and Avalonia's `IActivatableLifetime` file-activation event. The activation route is dispatched asynchronously and can be delivered again when the application regains focus.

The existing duplicate-open check depends on a subtitle already being loaded, so it does not protect an open that is still in progress. It fails especially badly for a multi-track Matroska file because the track picker opens before `Subtitles` is populated. This change records the file currently being opened and the active Matroska path before asynchronous picker work begins, compares normalised and symlink-resolved paths, and suppresses matching repeated requests. A request for a different file continues through the normal open flow.

### User impact

This is more disruptive than a harmless duplicate event. Launching Subtitle Edit with a multi-track Matroska file can queue multiple track pickers for the same container. Switching to another application and returning to Subtitle Edit can queue more pickers while an earlier picker is still open. The resulting dialog storm causes phantom or stacked windows, focus flicker, and unstable window ordering, and forces the user through repeated track-selection dialogs for a file they opened once.

The repeated requests make launch and document-open behavior unreliable and leave it unclear which pending dialog controls the document being loaded. Standalone subtitle files are also exposed to repeated activation delivery, although the Matroska picker makes the failure most visible. The expected behavior is one open operation, and at most one track picker, for each canonical path regardless of how many macOS delivery events report it.

### Before and after

Before, concurrent paths could race. After, in-flight, picker, and already-loaded matches return early.

## Implementation

Track canonical opening paths, compare resolved paths, and set and clear state around open/reset.

### Dependencies and order

The change is self-contained within the existing file-open and reset paths. It does not add settings, dependencies, or public APIs.

### Out of scope

Dialog modality, ownership, and window-parenting changes are not included. This change only prevents repeated requests to open the same file.

### Files changed

`src/ui/Features/Main/MainViewModel.cs`

### Automated

The change was reapplied cleanly to current upstream `main` and checked for whitespace errors. No automated test currently covers operating-system file-activation delivery on macOS.

### Manual

Verified manually on macOS 26.6.1 and macOS 26.6.2: launching with a multi-track Matroska file opened one track picker, and repeatedly deactivating and reactivating Subtitle Edit did not open additional pickers.

### Platform coverage

The affected code is not macOS-version-specific, but manual verification was performed only on macOS 26.6.1 and macOS 26.6.2.

### UI evidence

The app appears to open many windows and be dysfunctional as a result when you deactivate and reactivate the app. Before the fix, deactivating and reactivating the application could open additional Matroska track pickers, leaving multiple windows and making the application difficult to use. After the fix, repeated activation does not create additional pickers.

### Compatibility and migration

Distinct paths still open. No setting or migration is required.

### Risks and rollback

The risk is limited to path identity and open-state guards. Revert the commit to restore prior open handling.

### Assumptions, limitations, and open questions

Manual verification completed: command-line launch with a multi-track Matroska file and repeated application activation while its picker was open. Verification remains pending for Finder-initiated opens, standalone subtitle files, symlinked paths, and paths differing only by case.

### AI assistance

Claude Opus 4.8 materially assisted the original implementation. OpenAI Codex assisted with source audit, rebasing onto current `main`, and PR preparation.

## Checklist

- [x] Add or confirm macOS test evidence.
- [ ] Review path normalisation on supported macOS filesystems.
- [ ] Run project checks.
- [ ] Attach the recording.
- [x] State exactly which verification was performed and which remains pending.
- [x] Disclose material AI assistance in the submitted PR.
- [ ] If an AI agent owns implementation through merge, complete the required Copilot review cycle.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -642,6 +642,13 @@ public partial class MainViewModel :
     private long _lastKeyPressedMs;
     private bool _loading;
     private bool _opening;
+    // Dedupe state for the macOS launch double-open (a launched-with file arrives via BOTH
+    // the CLI argument AND the "open file" activation event, which re-fires on every focus).
+    // _openingFileName marks the file whose open is in flight; _openMatroskaFileName marks
+    // the currently-open Matroska container — set synchronously before the async track
+    // picker is posted, so it survives past the open and dedupes the picker storm.
+    private string? _openingFileName;
+    private string? _openMatroskaFileName;
     private PointerEventArgs? _lastTextEditorPointerArgs;
     private string? _textBoxLookUpWord;
     private PointerEventArgs? _lastSubtitleGridPointerArgs;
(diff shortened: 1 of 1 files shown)


niksedk:
Thanks for the report and the detailed analysis - the underlying dedupe gap is real, and it was nastier to spot than expected: the duplicate track picker opens at exactly the same position/size as the first, so it looks like a single dialog.

I ran an instrumented build (macOS 26.5 arm64, logging every activation event, `SubtitleOpen` call, and the open-window list) to check the delivery routes:

- Launching via Finder/`open` delivered the file **once**, via the activation event only - the CLI/argv route never fired, so no launch double-delivery.
- Focus changes delivered only `Background`/`Reopen` activations, which the existing handler already filters - the file event did not re-fire on focus.
- The stacking reproduced only when the same file was explicitly opened again (e.g. double-clicked in Finder) while the picker was still open: second `Kind=File` activation → second picker at the identical position.

Since only in-flight opens need protecting, the permanent guards here would also block legitimate re-opens (re-opening the same Matroska to pick a different track, re-opening after cancelling the picker, reloading the current file from disk - the latter on all platforms). I've merged #14353 instead, which keeps your in-flight guard idea (cleared in the existing `finally`) and adds a picker-scoped guard that clears when the dialog closes, so deliberate same-file re-opens keep working.

Closing this one in favor of #14353 - thanks again for surfacing it, the stacked-picker failure mode was well worth fixing.

PR #14352, opened 2026-08-31
Fix subtitle grid jumping to the end of the list after cut + auto-break

niksedk:
Fixes the remaining part of #14231 (the sudden jump to the bottom of the list), using the repro St0rmXtr00per provided in https://github.com/SubtitleEdit/subtitleedit/issues/14231#issuecomment-5481529046: cut the selected line, then auto-break, near the end of the file.

## Root cause

`TableViewScrollAnchor` remembered a single anchor: the row at the top of the viewport. Cutting the selected line removes exactly that row *and* changes the item count, so both the item lookup and the same-length index fallback in `ResolveAnchorIndex` failed. The next extent re-estimate — triggered by auto-break changing the row's height (the VirtualizingStackPanel estimator, upstream AvaloniaUI/Avalonia#17831) — then went uncorrected; when the estimate shrinks, the unchanged pixel offset maps to later rows, and near the end of the file the clamp lands at `extent − viewport`: the very bottom.

## Fix

`Remember()` now records every visible row with its viewport top, ordered top first. `Restore()` migrates to the first remembered row that still exists and puts it back at its own old top, so removing the top row hands the anchor to its nearest surviving neighbor instead of abandoning the restore. The re-estimate trigger, ordinary scrolling, ScrollIntoView and the index scrollbar suspend path are unchanged.

## Tests

Two new headless tests in `TableViewScrollAnchorTests`:

- `CutOfTheTopRowThenAutoBreak_DoesNotJumpToTheEndOfTheList` — the #14231 repro; fails against the old code with the view at the end of the list, passes with the fix.
- `RemovingRowsAboveTheAnchor_KeepsTheSameRowInView` — guard for deletes above the viewport followed by a height change.

All 12 scroll-anchor-related tests pass.

### DIFF
--- a/src/ui/Logic/TableViewScrollAnchor.cs
+++ b/src/ui/Logic/TableViewScrollAnchor.cs
@@ -2,6 +2,7 @@ using Avalonia;
 using Avalonia.Controls;
 using Avalonia.VisualTree;
 using System;
+using System.Collections.Generic;
 using System.Linq;
 
 namespace Nikse.SubtitleEdit.Logic;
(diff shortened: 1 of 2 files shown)


St0rmXtr00per:
@niksedk 
The issue still persists closer to the end of the list.

niksedk:
Thanks for the retest - you were right, and this fix was aimed at the wrong thing. Looking at your recording frame by frame there is no cut involved at all: the row just grows on auto-break, and the short lines at the end of the file are what sends the grid to the last page. Fixed properly in #14626, will be in the next rc - please give the same steps another go then.

PR #14353, opened 2026-08-31
Dedupe duplicate same-file opens (Matroska picker stacking) - counter-proposal to #14345

niksedk:
## Summary

Counter-proposal to #14345 (thanks @marcpbailey for the report and analysis - the underlying dedupe gap is real).

On macOS, opening the same file from Finder again while it is already being opened re-delivers a file-activation event into `SubtitleOpen`. For a multi-track Matroska the async track picker then opens a **second, pixel-identical picker exactly on top of the first** (verified: both windows at the same position/size), so it looks like one dialog, one Cancel appears to do nothing, and each stacked picker holds its own `MatroskaFile` handle.

## What the instrumented repro showed

Tested on macOS 26.5 (arm64) with a build of current `main` that logged every activation event, `SubtitleOpen` call, and the open-window list:

- **Launch via Finder/`open` delivers the file once** - only the activation route fires; the file never arrives via argv, so there is no CLI+activation double delivery at launch.
- **Focus changes deliver only `Background`/`Reopen` activations**, which the existing handler already filters. The file event does not re-fire on focus.
- **The only reproducible stacking**: explicitly opening the same file again while the picker is open - second `Kind=File` activation → second `SubtitleOpen` → second picker at the identical position.

So only the in-flight open and the picker window itself need guarding - permanent state is not required.

## Implementation

Two transient markers checked at the top of `SubtitleOpen`, compared with normalized/symlink-resolved case-insensitive paths:

- `_openingFileName` - the file whose `SubtitleOpen` is in flight; set in the existing `try`, cleared in its `finally` (covers the Matroska early-return and exceptions).
- `_pickMatroskaTrackFileName` - the file whose track picker is showing; set before the picker is posted, cleared in a `finally` when the dialog closes.

Because both markers clear when their operation finishes, the legitimate re-open flows that #14345's permanent guards blocked keep working:

- re-opening the same Matroska to pick a different track,
- re-opening after cancelling the picker,
- re-opening the currently loaded file to reload it from disk,
- opening the same file after File > New.

## Test plan

- `dotnet build src/ui/UI.csproj` - clean (0 warnings/errors).
- Manual, macOS 26.5 arm64 bundle: launch with a 2-subtitle-track mkv → one picker; re-send the same file via `open` while the picker is open → no second picker; cancel picker and open the file again → picker shows again; pick track A, open the file again → picker shows again (different-track flow intact).

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -642,6 +642,15 @@ public partial class MainViewModel :
     private long _lastKeyPressedMs;
     private bool _loading;
     private bool _opening;
+    // Transient dedupe for repeated open requests of the same file. macOS delivers a second
+    // file-activation event when the user opens an already-opening file from Finder again;
+    // for a multi-track Matroska the (async) track picker then opens a second, pixel-identical
+    // picker exactly on top of the first. _openingFileName covers the in-flight SubtitleOpen
+    // call and _pickMatroskaTrackFileName covers the posted track picker, which outlives
+    // SubtitleOpen. Both are cleared when their operation finishes - deliberately opening the
+    // same file again afterwards (e.g. to pick another track) works as always.
+    private string? _openingFileName;
+    private string? _pickMatroskaTrackFileName;
     private PointerEventArgs? _lastTextEditorPointerArgs;
     private string? _textBoxLookUpWord;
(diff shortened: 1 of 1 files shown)


marcpbailey:
I'll check it out @niksedk and report back here. Thank you.

Issue #14168, opened 2026-08-26
SE 5.2.0 beta 25: Importing from spreadsheets?


GrampaWildWilly:
I was skimming over #14164 & I noticed a reference to ODS files.  I thought that sounds interesting.  So I opened SE.  I happened to have nothing loaded due to a Ctrl+n before closing it the last time.  I hit F1 & that opened the expected documentation page in Firefox.  I started hunting around for the discussion of importing.  I found a discussion of importing plain text.  But I didn't find anything about importing open document spreadsheets.  Along the way, I didn't find anything about importing CSV files, either.  There must be conventions governing the formats of these files.  I didn't find any explanations.  I imagine the explanations require barely 3 lines of text, but still, I didn't find the explanations.

And then I discovered a bug that I have to say should go on the list at the very lowest priority.  When I closed Firefox to return to SE, the undocked Audio visualizer was in foreground instead of the main SE window.  Low low low priority.  Barely even a problem.

niksedk:
Both parts are fixed in main.

**Docs** — there is now a page for it: [Import Spreadsheets](https://subtitleedit.github.io/subtitleedit/features/import-csv-xlsx.html), covering both ways in (just opening a `.csv`/`.xlsx`/`.ods`, and *File → Import → CSV/XLSX/ODS with custom columns*), the header/column-name rules the automatic import needs, and how time codes, frames and millisecond columns are read. F1 in the import window opens it. The File menu page also got the four Import entries that were missing.

ODS was supported all along — every label just said "CSV/XLSX", which is why you found nothing. The menu item, window title and file filter now say **CSV/XLSX/ODS**. And `.xlsx`/`.ods` were in no file-picker filter, so File → Open never listed a spreadsheet; there is now a "Spreadsheet files" type there.

**The visualizer stealing the foreground** — fixed too, not as low priority as you thought: the undocked tool windows sit above the main window in the z-order, so when the other application closes, Windows hands them the foreground. Activation is now handed back to the main window, unless you actually aimed at the tool window.

Also found and fixed five real bugs in the importers while writing the docs (pipe-delimited files importing as one column, a crash on a CSV header with an unterminated quote, a blank cell turning frames into milliseconds, the wrong worksheet being read from a workbook, and merged ODS cells shifting the columns).

PRs #14173 and #14175. Thanks for the report.

GrampaWildWilly:
[...]

PR #14368, opened 2026-09-01
Decide undocked tool-window activations from the foreground history

### LINKED ISSUE #14354: SE 5.2.0 beta 30: Suddenly Alt+Tab refuses to bring undocked windows to foreground
I have something in which the audio is terrible plus the participants are being really bad about enunciating properly.  Even though it's in English, I was having a terrible time puzzling out what they were saying.  So I used SE to transcribe it, then I got into a long editing session fixing the subtitles because the dialog was so bad, even the machine transcription had trouble in places.  And I'm not so sure I've got it right even now.  The point is I spent a couple of hours correcting a subtitle file.  The long session might be a factor.

At some point, I noticed that Alt+Tab was refusing to bring the Audio visualizer to foreground.  It would flicker into foreground, then the main window would grab foreground.  I have a shortcut for toggling between the waveform & the grid, but that was doing the same thing.  The Audio visualizer would flicker into foreground then relinquish foreground to the main window.  I decided to check the video player & it was doing the same thing.

Then I clicked the mouse on the Audio visualizer.  That brought it durably to foreground.  I could then Alt+Tab between it & the video player, & even other apps, & the two subordinate SE windows would accept foreground assignment.  But as soon as I sent Alt+Tab to the main SE window, the other 2 windows went back to refusing to come to foreground via Alt+Tab.

Then I discovered that I had fat-fingered something & I had 2 instances of SE running.  I've asked about this before.  Can we please get a setting that says only 1 SE at a time, yes or no?  I don't know if that messed something up with the foreground state.  I closed the second SE & it didn't help.  I closed the remaining SE & that didn't help.  I redocked the controls, closed SE, reopened SE, undocked the controls again, & the problem persists.  On the good side, undocking after docking remembered my old window sizes & positions, including the video player on the second monitor in fullscreen.

I'm pretty sure I could Alt+Tab to the 2 subordinate windows when I first started using beta 30.  Not 100% certain but I think it was working properly.  But now it's not & I can't figure out a way to make it go back to working properly.  I have a faint recollection we fixed this in some past version of SE 5.  Might have even been back in the Preview days.  But I could easily be wrong.

niksedk:
Alt+Tab to an undocked waveform or video player window still flickers into the foreground and loses it a beat later in beta 31 — [#14354](https://github.com/SubtitleEdit/subtitleedit/issues/14354#issuecomment-5489349992).

### Why the beta 30 fix wasn't enough

The watcher that hands the foreground back when the OS activates a tool window by itself (#14168) classifies activations from the physical pointer buttons, and Alt+Tab has none down. PR #14363 added a task-switch signal on top of `EVENT_SYSTEM_SWITCHSTART` / `SWITCHEND` — and those never reached SE on the reporter's machine. His screenshot shows the Windows 11 XAML switcher, not the legacy switch window those two events were built for.

Everything else in the chain is confirmed by his report: the flicker is the 200 ms `RunOnce` → `mainWindow.Activate()`, a click fixes it because `IsAnyPointerButtonDown()` is true, and Alt+Tab back to the main window re-arms the claim in `OnWindowActivated`. On Windows `IsAnyPointerButtonDown()` never returns null, so the hover fallback is dead there — the task-switch signal was the only thing that could have rescued Alt+Tab, and it returned false.

### What this does instead

Judge **what became of the previous foreground window**. The OS only has to pick a new foreground window when the old one went away, so:

- previous window still alive, visible, not minimized → the user aimed here, whatever gesture they used;
- previous window closed or minimized → the handover this watcher exists to correct.

That covers Alt+Tab, the taskbar button, Alt+Esc and Win+Tab without any of them announcing itself — the three gaps PR #14363 knowingly left open. `ForegroundWindowTracker` (renamed from `TaskSwitchDetector`) keeps the history from an `EVENT_SYSTEM_FOREGROUND` hook, seeded with `GetForegroundWindow()` at install so the first switch has something to compare against. The switcher's own window is recognised by class (`XamlExplorerHostIslandWindow`, `MultitaskingViewFrame`, `TaskSwitcherWnd`, `#32771`) and kept out of the history — it is gone by the time the chosen window is activated, which would otherwise read as "the previous window went away". Its taking the foreground is itself a task-switch signal, so the reporter's case is now fixed twice over, whichever way his switcher behaves. The `SWITCHSTART`/`SWITCHEND` hook stays: it costs nothing and is a positive signal where it does fire.

Three supporting changes:

- **The verdict moved into one place, a beat after the activation.** Only the buttons can be sampled while the activation is delivered; the press event (touch) and both WinEvent callbacks arrive after it — the callback is delivered through the same message queue as the activation it describes, so at `Activated` time the history can still be a window behind. Reading it is gated on `GetForegroundWindow() == ` the recorded current window, so a stale history answers "unknown" rather than describing the wrong window.
- **Nothing is decided unless the tool window is still the one in front.** This also fixes a live hazard the deferred read would otherwise have introduced *and* one already present in beta 31: topmost churn while another application takes over used to disarm the claim for the whole round (the #14168 beta 26 failure). Now a bounce leaves it exactly as it found it.
- **Alive but not visible answers "unknown", not "handover".** Closing destroys a window and minimizing leaves it visible-and-iconic, so neither user action looks like that — it is a helper or island window, and the class list only knows the switchers by name. Unknown falls back to the old pointer rules, i.e. today's behaviour, so an unrecognised switcher degrades to the status quo instead of asserting a handover.

`SetWinEventHook` failures are now logged. A silent failure was indistinguishable from an event that never fires, which is a large part of why this needed a second round-trip with the reporter.

### Testing

`tests/UI/Logic/UndockedForegroundStealTests.cs` covers the new rule and its precedence — history evidence both ways, a click outranking a vanished previous window, and the fallback to the pointer rules when the history is unknown. Full UI suite: 4488 passed, 1 skipped.

The WinEvent half cannot be exercised headlessly. Manual check on Windows: undock the video and waveform controls, then Alt+Tab between the main window, the visualizer and the video player — each should take and keep the foreground — and confirm the original #14168 case still works by switching to another application and closing it, which must return to the main window rather than a tool window.

Fixes #14354


GrampaWildWilly:
I do hope this works.  One question that floated across my mind.  You're keeping a sort of window history list now in SE, if I'm reading this right.  If I'm away in other tasks, Alt+Tabbing back & forth to who knows how many windows, your list in SE is unaffected by all that, right?  So when I do try to come back to one of the 3 SE windows, it's supposed to know how things were & bring to foreground whichever of the 3 I choose now, regardless of which one was last in foreground before, right?

Issue #14330, opened 2026-08-31
SE 5.2.0 beta 30: Audio dropouts


GrampaWildWilly:
This is not new with this beta.  It's been going on for rather a long time but I'm only just now getting around to mentioning it.

With the MPV video player I find that every time I pause & every time I resume playback, there's a couple of seconds of delay while audio is not heard.  This also happens in MPV.NET but not to the same extent it does in SE.  In MPV.NET, when I go from not fullscreen into fullscreen, there's a momentary audio dropout.  Then again, when the video finishes playing, going out of fullsceen triggers another audio dropout.  It's not a big deal in MPV.NET because I have it set to not autoplay.  So the audio dropout happens & then I start playback.  But I don't get audio dropouts every time I pause & resume playback in MPV.NET.  So there is something the other video player is telling MPV that SE is not telling it.

It may sound strange to say there is an audio dropout when I pause.  Nothing is playing.  How can there be an audio dropout when nothing is playing?  I notice it from the display on my surround receiver.  It briefly displays "No signal" on every pause.  I can't hear anything then, but the receiver is giving evidence that the audio dropout is happening.  When I resume playback in SE, there's another brief "No signal" and this time, there is a gap in the audio.  If I want to hear what I missed, I have to rewind a few seconds without pausing playback.  Then the video player replays the same section with the audio.

I believe this has something to do with my audio drivers.  I have an NVidia GeForce RTX 4070.  It does double duty as both my video driver AND my audio driver.  I have the HDMI port of the NVidia connected to my surround receiver.  When I play anything back, the NVidia adapter sends both audio & video over the HDMI cable to the receiver.  The receiver then splits off the video & sends it to my TV, at the same time splitting off the audio & sending it to my speakers (or headphones).  I make it a point to be on the latest NVidia drivers at all times.  Current is 615.56 which came out just this past Wednesday.

Since MPV.NET is able to tame these dropouts, SE should be able to do the same.  When you recently included a new MPV with SE, I copied it into my MPV.NET directory.  So the 2 apps are using the same MPV.  MPV.NET has minimal dropouts, only before the start & after the end of playback.  That is completely harmless.  SE has dropouts at every pause & every resume.  That can get extremely annoying.  There has to be something you can do about this.

GrampaWildWilly:
[...]

PR #14370, opened 2026-09-01
mpv: replace the dead audio-keep-open option with audio-stream-silence

niksedk:
Follow-up to #14335 for #14330.

## The option we shipped does not exist

#14335 set `audio-keep-open`. mpv has never had that option — it is in neither `DOCS/man/options.rst` nor `options/options.c` upstream. `mpv_set_option_string` answers an unknown name with `MPV_ERROR_OPTION_NOT_FOUND`, which SE only logs, so the checkbox was a no-op in both positions. The reporter said exactly that: *"It actually looks like this toggle has utterly no effect whatsoever."*

The test added with it pinned that the call was *made*, never that mpv *accepted* it, so it passed the whole time.

## What actually causes the dropouts

mpv doesn't merely close the device at EOF — it stops it on pause and resets it on every seek. On Windows that's `ao_wasapi`'s `set_pause` → `IAudioClient::Stop()`. Over HDMI the receiver's link goes idle (hence its "No signal", a few seconds later as the driver times out), and `IAudioClient::Start()` on resume triggers a re-handshake the receiver eats — the 3–4 seconds of missing audio in the report.

`--audio-stream-silence=yes` is mpv's answer, and its manual describes this exact hardware. In `audio/out/buffer.c` the flag gates four paths: the pause branch is skipped, `ao_reset` no longer stops the device on seeks, the fill loop keeps writing silence while paused or idle, and the stream stays up at EOF. The seek case matters as much as pause here — SE seeks constantly while timing subtitles, and each one currently resets the AO.

## What this PR does

- Removes `MpvAudioKeepOpen`, its checkbox, and its language strings (including three translations that had already picked the string up). Settings-safe: both the settings and language JSON contexts use `UnmappedMemberHandling.Skip`, so old files carrying the property still load.
- Adds `Se.Settings.Video.MpvAudioStreamSilence` (default `false`), set before `mpv_initialize` and only when asked for.
- No checkbox, following `MpvAudioBufferSeconds`. mpv's manual calls `audio-stream-silence` **"strongly discouraged"** — it changes A/V-sync and underrun handling, and mpv warns about it in its own log at every AO init — and it only helps the HDMI-receiver case. If it works for the reporter, promoting it to a checkbox is a follow-up decision.

## Test

The failure mode worth guarding is a silent no-op, not a wrong value. `EveryMpvOptionNameSet_IsARealMpvOption` extracts every option name SE passes to mpv and requires it to be on a list verified against mpv's manual, so a new name has to be looked up before it can pass. Confirmed it fails on the exact bug: reinstating `audio-keep-open` produces *"mpv option name(s) not on the verified list: audio-keep-open"*.

The other 22 names SE sets were all checked against the manual — `audio-keep-open` was the only invented one.

Full UI suite: 4489 passed, 1 skipped, 0 failed.

## For the reporter

`audio-stream-silence` is off by default, so this needs `"MpvAudioStreamSilence": true` under `Video` in `Se.json`, plus a restart — it is a pre-init option, so toggling it can never affect a running player.

Still open from the beta 31 report and not addressed here: the Settings/Apply freeze that left the video player straddling both monitors.

### DIFF
--- a/src/ui/Assets/Languages/English.json
+++ b/src/ui/Assets/Languages/English.json
@@ -2937,7 +2937,6 @@
       "showFullscreenButton": "Show full-screen button",
       "fullscreenHideControls": "Hide video controls in full-screen",
       "autoOpenVideoFile": "Auto-open video file when opening subtitle",
-      "mpvAudioKeepOpen": "Keep the audio device open while paused (mpv)",
       "downloadMpv": "Download mpv",
       "downloadVlc": "Download VLC",
       "allowSingleLetterShortcutsInTextbox": "Allow single-letter shortcuts in text box",
--- a/src/ui/Assets/Languages/Italian.json
+++ b/src/ui/Assets/Languages/Italian.json
@@ -2930,7 +2930,6 @@
       "showFullscreenButton": "Visualizza pulsante 'Schermo intero'",
       "fullscreenHideControls": "Nascondi controlli video a schermo intero",
       "autoOpenVideoFile": "All'apertura dei sottotitoli apri automaticamente file video",
-      "mpvAudioKeepOpen": "Mantieni aperto il dispositivo audio durante la pausa (mpv)",
       "downloadMpv": "Download mpv",
       "downloadVlc": "Download VLC",
       "allowSingleLetterShortcutsInTextbox": "Consenti nella casella di testo scorciatoie con una sola lettera",
--- a/src/ui/Assets/Languages/Korean.json
+++ b/src/ui/Assets/Languages/Korean.json
@@ -2937,7 +2937,6 @@
       "showFullscreenButton": "전체 화면 버튼 표시",
       "fullscreenHideControls": "전체 화면에서 비디오 컨트롤 숨기기",
       "autoOpenVideoFile": "자막 파일을 열 때, 비디오 파일도 자동으로 열기",
-      "mpvAudioKeepOpen": "일시 정지 중에도 오디오 장치를 열어 두기 (mpv)",
       "downloadMpv": "mpv 다운로드",
       "downloadVlc": "VLC 다운로드",
       "allowSingleLetterShortcutsInTextbox": "입력칸에 단일 문자 단축키 허용",
--- a/src/ui/Assets/Languages/Turkish.json
+++ b/src/ui/Assets/Languages/Turkish.json
@@ -2937,7 +2937,6 @@
       "showFullscreenButton": "Tam ekran butonunu göster",
       "fullscreenHideControls": "Tam ekranda video kontrollerini gizle",
       "autoOpenVideoFile": "Alt yazı açıldığında video dosyasını otomatik aç",
-	  "mpvAudioKeepOpen": "Duraklatıldığında ses aygıtını açık tut (mpv)",
       "downloadMpv": "mpv indir",
       "downloadVlc": "VLC indir",
       "allowSingleLetterShortcutsInTextbox": "Metin kutusunda tek harfli kısayollara izin ver",
--- a/src/ui/Features/Options/Settings/SettingsPage.cs
+++ b/src/ui/Features/Options/Settings/SettingsPage.cs
@@ -447,7 +447,6 @@ public class SettingsPage : UserControl
             MakeCheckboxSetting(Se.Language.Options.Settings.ShowFullscreenButton, nameof(_vm.ShowFullscreenButton)),
             MakeCheckboxSetting(Se.Language.Options.Settings.FullscreenHideControls, nameof(_vm.FullscreenHideControls)),
             MakeCheckboxSetting(Se.Language.Options.Settings.AutoOpenVideoFile, nameof(_vm.AutoOpenVideoFile)),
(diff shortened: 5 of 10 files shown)


GrampaWildWilly:
I spent a lot of time with Google AI getting MPV.NET to work the way I want when I first got it.  As it happens, advice I got was to place 2 lines in the configuration file mpv.conf:

audio-stream-silence=yes
audio-wait-open=0.5

So that's how I've been running MPV.NET for a few months now.  (I discovered MPV.NET only a few months ago.  I got it on a recommendation from Google AI as well.  My thought process was SE uses MPV.  I'd like a way to use MPV directly myself instead of VLC.)  The MPV manual may strongly discourage audio-stream-silence=yes, but I'm here to tell you it's not caused me any problems.  Like I said, I still see the "No Signal" message on my receiver when I load a video into MPV.NET, & again when I clear the MPV.NET playlist.  In between, I get 0 audio dropouts, not when launching playback from time index 0, not when doing pause/resume in the middle of the file, not when the file runs to the end.  If you can make SE act the same, that will be what I will call a win on all counts.

Now, you mention something about changing a line in Settings.json that will control one of these parameters until you decide whether to make something a new option in the SE Settings.  What should I be coding in there in advance of beta 32?  Here's my best guess.  I've found a line in there that reads:

`  "Video": {`

About 160 lines below that I see:

`    "VideoPlayer": "mpv-wid",`

I'm thinking right below that I will need to add:

`"MpvAudioStreamSilence": true`

In fact I've gone ahead & done that now with the beta 31 I have.

OK.  You may now laugh yourself silly.  That was a distinctly bad idea.  I've just gone through the exercise of putting all my settings & all my sh9rtcuts back the way I had them.  Of course, I was too cocky to make a safe copy of Settings.json before I did that. Yeah,  Right.  I'm not a rookie . . .

Although I did notice 2 settings in there that appear to have picked up 2 of my settings from my mpv.conf file over in my MPV.NET directory tree.  That's interesting . . .  Or maybe your defaults just happen to accidentally match my settings.  The lines are:

```
    "MpvAudioBufferSeconds": 0.05,
    "MpvAudioKeepOpen": true
```

Wait.  That second line is the one you're going to deimplement, isn't it?

Anyway, my question still stands.  Am I supposed to add that line to Settings.json manually . . . AFTER I get beta 32 (since sad experience shows it's a terrible idea to try it with beta 31)?

GrampaWildWilly:
Hold the phone.  I just discovered this in my error log:

```
-----------------------------------------------------------------------------
Date: 09/01/2026 05:53:20
SE: v5.2.0-beta31 - Microsoft Windows NT 10.0.26200.0 - 64-bit
'"' is invalid after a value. Expected either ',', '}', or ']'. Path: $.Video | LineNumber: 3185 | BytePositionInLine: 4.
   at System.Text.Json.ThrowHelper.ReThrowWithPath(ReadStack& state, JsonReaderException ex)
   at System.Text.Json.Serialization.JsonConverter`1.ReadCore(Utf8JsonReader& reader, T& value, JsonSerializerOptions options, ReadStack& state)
   at System.Text.Json.Serialization.Metadata.JsonTypeInfo`1.ContinueDeserialize[TReadBufferState,TStream](TReadBufferState& bufferState, JsonReaderState& jsonReaderState, ReadStack& readStack, T& value)
   at System.Text.Json.Serialization.Metadata.JsonTypeInfo`1.Deserialize(Stream utf8Json)
   at Nikse.SubtitleEdit.Logic.Config.Se.LoadSettings(String settingsFileName)
```

I copy/pasted exactly the line you suggested out of this discussion into my Settings.json.  I apparently got this error next time I tried to run SE.  This is telling me there's some kind of syntax error in the json line you've given me.  SE reacted by trashing my Settings.json & giving me what must be the factory default Settings.json.  That's why I had to reconstruct all my settings & all my shortcuts.  And some other minor things, like undocking.  This is not nice.  I don't have a lot of customizations to settings & shortcuts so it took me only about 15 minutes to reconstruct my environment.  Along the way, I think I improved my waveform color scheme.  So it wasn't a total loss.   But I can see others being entirely enraged by such an event.

You can be sure that tonight when beta 32 lands, I'm going to make a safe copy of my Settings.json.  But still, you should not be trashing my file just because I've entered something that doesn't parse.

GrampaWildWilly:
I think I see what I did wrong.  I didn't terminate the line with a comma.  I'm not an expert on json.  I just noticed now that every line in a block ends with a comma.  Only the last line in a block ends with no punctuation.  Still, SE completely obliterated my Settings.json file because of my typo.  That can't be the desired behavior.

Also, I'm not finding this Pull Request in searches.  I don't know why.  I think I'll have to add a comment somewhere to draw your attention to this.  Or maybe I'll just say @niksedk & hope that works.

GrampaWildWilly:
I finally made this appear in a search by using these minimalist search criteria:

sort:updated-desc state:merged

With these, it's not found, but I can't explain why not:

is:pr sort:updated-desc commenter:@me

GrampaWildWilly:
It looks like there is an MpvAudioStreamSilence option now in Settings.json.  But it's defaulting to FALSE.  Why?  Shouldn't that be TRUE?  Are you telling me I should manually change that one line to read TRUE?

GrampaWildWilly:
I believe #14799 is your response to the problem I had here of losing my Settings.json file.  I've noted that in #14861.

Issue #13373, opened 2026-08-08
SE 5.2.0 beta 7: Shortcuts suggestions


GrampaWildWilly:
I was about to say there was a regression, that Ctrl in the search box in the Shortcuts dialog was not accepted as a synonym for Control.  But then I pressed+ & realized it doesn't recognize the synonym until you close it with a +.  Interesting to know.  This is not a problem, just an aha moment I needed to experience.

There were some things cut off in the dialog.  So I adjusted some column widths to make them visible.  The column I was dealing with was Shortcut.  It's on the right end so you can't grab its right border unless you first drag its left border a bit to the left.  Then the right border comes into view & then you can drag it.  Again, not a problem, just a realization.

But what I really wanted to do was autofit the column.  Make it as wide as the widest entry automatically.  This is available as a standard elsewhere in Windows, for example in the file manager.  If you double click one of the vertical lines that separates column headings, it autofits the column to the left of the separator you double clicked.  Apparently this isn't available in SE.  I'm suggesting it should be.

I know most of the things you do in the Shortcuts dialog (aside from actually binding a shortcut) are forgotten when you close the dialog.  This is as it should be.  But I'm putting forward the thought that column sizing might be one that ought to be remembered across invocations of the dialog.

And while I'm here, I'd like to repeat a couple of other ideas I've brought up before.  It should be possible to drag some of the columns around so they can be reordered.  And it would be helpful if the `Active in` column allowed the user to select the contexts in which any given shortcut fired.

GrampaWildWilly:
@niksedk ?

GrampaWildWilly:
<img alt="#01" src="https://github.com/user-attachments/assets/a813f21c-b107-4342-b002-c284ff0dc2f2" />
<img alt="#02" src="https://github.com/user-attachments/assets/06889e0c-e74d-4897-b0f7-c1c26dc5a4fb" />
<img alt="#03" src="https://github.com/user-attachments/assets/bf7d4d21-3267-4314-b36d-9a7d3f113651" />

This brings into focus a few issues.  First, things should not be hidden by default like this.  Second, double clicking the divider between column headings to autofit the column would be more convenient.  Third, when I close the Shortcuts dialog & reopen it, the column width has reverted to the original size which hides important information.

[...]

PR #14371, opened 2026-09-01
Add a "Columns..." dialog to reorder the subtitle grid columns

### LINKED ISSUE #14369: can’t change the order of the columns in Grid
Image

In SE5, I can’t change the order of the columns (Hide, Forced, Gaps, CPS, etc.).

niksedk:
Fixes #14369.

In SE5 the subtitle grid's column order was fixed - the TableView has no column dragging, and columns were built in a hard-coded sequence. Right-clicking a column header now offers **Columns...**, a dialog that handles show/hide and ordering in one place:

- Every column listed with a visibility checkbox plus **Move up** / **Move down** / **Reset** (same shape as the waveform toolbar items dialog).
- Number and Text cannot be hidden; the content-driven columns (Original text) show a disabled checkbox. The teletext preference is toggleable here even though the column stays format-gated.
- The Style/WebVTT-style and Actor/WebVTT-voice pairs are one entry each - they are format-gated variants of the same logical column sharing a toggle, so they move as one.

The order is stored as column keys in `SeGeneral.SubtitleGridColumnOrder` and re-applied on every grid rebuild. `TableViewColumnManager.ApplyOrder` does the permutation: `Sync()` can only insert missing columns (a naive permutation through it duplicates them), so the live column list is cleared and rebuilt; keys a saved order doesn't know (columns added in a later version) keep their default position instead of being displaced or hidden.

Tests: 4 headless unit tests for `ApplyOrder` (permutation, hidden columns, unknown/missing keys, empty order) and 2 integration tests that build the real main window and assert the saved order - including an "old version" order - lands in the grid. Full UI suite: 4493 passed (one unrelated known-flaky undo/redo timing test failed in-suite, passes in isolation).

Verified visually via headless screenshots: the dialog renders correctly and a saved `#, Duration, Show, Hide, Text` order shows up in the grid.

### DIFF
--- a/src/ui/Assets/Languages/English.json
+++ b/src/ui/Assets/Languages/English.json
@@ -127,6 +127,8 @@
     "colorWhite": "White",
     "colorYellow": "Yellow",
     "column": "Column",
+    "columns": "Columns",
+    "columnsDotDotDot": "Columns...",
     "consoleLog": "Console log",
     "contentAlignment": "Content alignment",
     "continueFindTitle": "Continue Find?",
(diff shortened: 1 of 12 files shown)


GrampaWildWilly:
This encourages me to point out my request for the same thing in the Shortcuts dialog: #13373.

PR #14383, opened 2026-09-01
macOS: default the video picker to the subtitle's own folder, and open files in a new window

muaz978:
## Summary

Two small macOS-only fixes, both reported from real day-to-day use on a Mac:

1. **Video picker defaults to the wrong folder.** After opening a subtitle and clicking the empty waveform (or Video > Open video) to load its matching video, the file picker opens wherever a video was last picked from, not the subtitle's own folder. If you work across several project folders (e.g. Documents one day, Downloads the next) this means re-navigating every time, even though the video almost always lives right next to the subtitle.
2. **Opening a subtitle while Subtitle Edit is already running replaces the window you're working in.** On Windows, double-clicking an .srt in Explorer while SE is running does not touch any open window - it is a straightforward consequence of Explorer launching a whole new process per double-click. On macOS, the whole app is one process, so the existing Finder-activation handler was loading the newly-opened file into whichever window happened to be `lifetime.MainWindow`, discarding whatever was open there. Same underlying issue for the Dock icon: reactivating the app via the Dock did nothing useful instead of giving you a fresh window.

## What changed

**Video picker folder (`IFileHelper.PickOpenVideoFile`, `FileHelper.cs`, `MainViewModel.CommandVideoOpen`)**
Added an optional `lastOpenedFilePath` parameter, mirroring the pattern `PickOpenSubtitleFile` already uses (`SuggestedStartLocation` resolved from the file's directory via `StorageProvider.TryGetFolderFromPathAsync`). `CommandVideoOpen` now passes the currently open subtitle's path, so the picker suggests that folder. Backward compatible - the parameter defaults to `null`, so every other `PickOpenVideoFile` call site (BurnIn, ReEncodeVideo, VisualSync, etc.) is untouched.

**Open-in-new-window on macOS (`Program.cs`, `MainWindowFactory.cs`)**
- Added `MainWindowFactory.OpenNewWindowWithFile(string filePath)`, which creates a new editor window (same as the existing `OpenNewWindow()` used by File > New window) and loads the given file into it.
- The macOS `IActivatableLifetime.Activated` handler for `ActivationKind.File` now branches on whether the app has already started (`lifetime.MainWindow != null`): during cold startup it still hands the file to the primary window that is about to be created (unchanged), but once the app is already running it opens the file in a brand new window instead of touching the existing one.
- Added handling for `ActivationKind.Reopen` (Dock icon activation while already running), which now opens a new window the same way `CommandFileNewWindowCommand` does.

## Testing

All of this was verified end to end on my own Mac, not just via the unit suite - both because the bug only reproduces through real OS-level events (Finder file-open activation, Dock reopen) that unit tests can't exercise, and because I wanted to be sure before asking for review.

**Build and existing suite**
- `dotnet build src/ui/UI.csproj -c Debug` - clean, 0 warnings introduced (only the 4 pre-existing unrelated nullable-annotation warnings in libse), 0 errors.
- `dotnet test SubtitleEdit.sln` - full suite green after the merge with latest main: 4510/4511 passed (the 1 skip is the existing `AppleVisionOcrTests.IsAvailable_OffMacOs_IsFalse`, which only runs off-macOS). Had to update the `PickOpenVideoFile` signature in the four `IFileHelper` test doubles (`StubFileHelper`, `FontCollectorEmbeddedMatchTests`, `SettingsImportExportShortcutSlotsTests`, `SettingsImportSe4XmlTests`) to match the new optional parameter.

**Manual verification, feature 1 (video picker folder)**
1. Created two throwaway .srt files in two different real folders (`~/Downloads/se-test-folder-1/test1.srt` and `~/Documents/se-test-folder-2/test2.srt`).
2. Opened `test1.srt` via File > Open (Cmd+O) into a fresh window, confirmed via the app's own status bar that it loaded from `~/Downloads/se-test-folder-1/test1.srt`.
3. Clicked the empty waveform to trigger the video/audio picker.
4. Result: the picker opened directly on `se-test-folder-1`, listing `test1.srt` right there - not the unrelated folder ("Fathi Dub Check") that a previous, unrelated video pick had left as the sticky default. Confirms the picker is now keyed off the open subtitle's folder rather than global picker-open history.

**Manual verification, feature 2 (new window instead of hijacking the current one)**
1. With one window open showing `test1.srt`, opened Finder to `~/Documents/se-test-folder-2` and double-clicked `test2.srt` - i.e. the exact real-world gesture (Finder file-open activation) the bug report was about, not a simulated event.
2. Result: a second window opened titled `test2.srt - Subtitle Edit`, loaded correctly (status bar confirmed the full path). The first window, still showing `test1.srt`, was untouched and visible behind it.
3. Confirmed via Accessibility (`System Events`) that both windows existed simultaneously: `test2.srt - Subtitle Edit v5.2.0-beta31, test1.srt - Subtitle Edit v5.2.0-beta31`.
4. Then activated the app via its Dock icon twice (while both windows were still open). Each activation opened its own new, empty editor window - two additional windows appeared, for four total, and neither of the two original file windows was closed or altered. Confirmed again via `System Events`: `Untitled, Untitled, test2.srt - Subtitle Edit v5.2.0-beta31, test1.srt - Subtitle Edit v5.2.0-beta31`.

This was done on the actual local build installed on my machine (a full `dotnet publish -r osx-arm64 --self-contained` release build copied into `/Applications/Subtitle Edit.app`, re-signed), not just `dotnet run` in a dev shell, so the behavior matches what a real user would see after installing a release build.

## Scope

Both fixes are additive and platform-scoped where it matters: the video-picker change is a plain optional parameter that only affects behavior when a `lastOpenedFilePath` is actually passed (currently just `CommandVideoOpen`), and the window-opening change only touches the macOS-only `IActivatableLifetime.Activated` handler in `Program.cs` - Windows and Linux startup/file-open paths are untouched.


muaz978:
Follow-up: a plain single click on the Dock icon was also opening a new window, which isn't right - a single click should just bring SE back to the front like any other app. Root cause: AppKit sends the same Reopen activation on every click of a running app's Dock icon, single or double; there's no separate "this was a double-click" event to key off. Fixed by tracking the time between consecutive Reopen activations and only opening a window when a second one lands within the standard double-click interval (500ms) of the last one.

Worth calling out for context: a single click just activating the app without opening anything new is already how the Windows taskbar behaves - clicking a running app's taskbar icon brings it to the front, it doesn't launch a new window. That was already normal on Windows; macOS's Dock icon had no equivalent of it before this change, since every click (not just a double-click) was treated as "open a new window." This fix brings the Mac side in line with the behavior Windows already had, rather than introducing something new.

Verified on my machine, same installed build as before:
- One click on the Dock icon: window count stayed at 1 (`Untitled - Subtitle Edit`).
- Waited past the interval, clicked once more in isolation: still 1 window.
- Two clicks back to back (an actual double-click): a second window opened (`S2 E1 Short.srt - Subtitle Edit`, `Untitled - Subtitle Edit`) - confirmed via `System Events`.

Full suite still green after this change: 4510/4511 (the 1 skip is the pre-existing macOS-only `AppleVisionOcrTests` gate).

GrampaWildWilly:
This is actually pretty close to something I've mentioned recently.  And even though I'm running Windows, I think the issue is platform-independent.

There needs to be a setting in SE that says `Permit multiple instances of SE` with a checkbox.  It is not always desirable to permit multiple instances of SE.  I see the process working like this.

If you have the setting box checked, then any attempt to open SE, whether that be by double clicking a file or launching the app from the Start Menu or maybe invoking it from a .bat file or . . . I'm sure I don't have the imagination to enumerate all the possibilities.  But with this proposed new setting enabled, yes, go ahead & open a new instance of SE.

If you do not have the setting box checked, then any attempt to open SE must be directed to the existing, running instance.  There should be a check that pops up a dialog box asking, `You have a running SE.  Do you want to replace your current task with this new file?`  There should be a guard that allows you to save unsaved changes if you have any.  The dialog should also give you the option of rejecting the request.  Maybe it was an oops & you didn't mean to do that.  I hit precisely this situation just yesterday.  I fat-fingered something & I had a second instance of SE running & I didn't know it.

So there needs to be the acknowledgement that opening a second instance of SE must be at the discretion of the user.  The user must always be aware of the ramifications of this choice.  A new setting in SE seems to be in order here.

muaz978:
@GrampaWildWilly thanks for taking the time to write this up, it's a fair point and worth being explicit about here.

To scope what this PR actually touches: it doesn't change the underlying "should opening a file while SE is already running spawn a new window/instance" decision at all. That already happens unconditionally on Windows today - double-clicking an .srt in Explorer while SE is running opens a new process, no confirmation, no setting. This PR only brings macOS in line with that existing default (macOS was instead silently replacing whatever was open in the first window, which is arguably worse since nothing indicates a switch happened at all). So the surprise-without-confirmation concern you're describing predates this PR and applies equally with or without it.

Your proposed setting - a checkbox to permit/deny multiple instances, with a confirmation dialog and a chance to save unsaved changes when denied - sounds like a reasonable feature on its own merits, and it's genuinely platform-independent like you said. But it's a bigger product decision (what the default should be, what the dialog says, whether it also needs to cover Finder/Explorer drag-and-drop, etc.) than a Mac-parity fix, so I'd rather not fold it into this PR unprompted. @niksedk, worth its own issue/discussion if you think it's a direction worth pursuing - happy to pick it up if so.

GrampaWildWilly:
Yes.  Absolutely.  A product-level setting would definitely be outside the scope of this PR.  I personally don't like the current behavior in Windows, that it's possible to have multiple instances of SE running.  There might be situations in which I would want that on purpose.  But most of the time, I would not.  It looks to me like if a product-level setting is implemented, the code in this PR would be gated by that product-level setting.  They wouldn't be mutually exclusive, just cooperating.

PR #14384, opened 2026-09-01
Keep displayed playhead monotonic during playback

marcpbailey:
## Summary

This is a follow-up augmentation to #14346, which fixed selected-cue progression at cue boundaries and when timer callbacks arrive late. That PR corrected which cue repeat playback chooses. It did not change the separate `MainViewModel` estimator that decides where the displayed playback cursor appears.

Once the selection fix made repeated playback progress correctly, sustained manual playback revealed a second problem: the displayed playhead could still move backwards when a late or stale mpv position report pulled the estimator below its prior displayed value. This is particularly apparent during repeat playback and at higher playback speeds, where position callbacks can lag the timer-driven estimate.

This change keeps the displayed estimate monotonic while playback continues, while retaining deliberate seeks and normal reinitialization behavior.

### User impact

When timing, reviewing, or repeatedly playing subtitles, the cursor can jump back into material it has already passed even though #14346 has selected the correct next cue. That makes playback look unreliable and can cause the subtitle selection view to move unexpectedly. The issue is most noticeable during repeat playback and high-speed review.

### Why this appeared after #14346

The original reproduction was a cue-selection failure: at a boundary, the completed cue could be selected again or late callbacks could schedule already-completed material. That behavior masked the estimator problem because the playback sequence itself was already wrong. #14346 fixed those transitions with focused `PlaySelectionItem` tests. Longer normal and repeat-playback runs, especially at 3x, then isolated the remaining backward movement to the displayed playhead rather than cue selection.

## Implementation

- Preserve the prior displayed estimate as a floor during an uninterrupted playback run.
- Reset the estimator when a new media file is opened, so a legitimate new position is not held to the prior file's position.
- Pin the estimate when repeat-selection playback deliberately seeks to a subtitle start.
- Scale the frozen-raw-position extrapolation hold by playback speed, so its limit remains in media time rather than wall-clock time.

The floor is not applied to explicit seeks or a playback restart, so intentional backward movement remains immediate.

### Edge cases covered by this follow-up

- A stale raw mpv position during uninterrupted normal playback must not pull the visible cursor backwards.
- The same stale-position case during repeat playback must not undo the correctly selected cue progression from #14346.
- At high playback speed, the raw-position freeze hold must be limited in media time; otherwise the estimate can extrapolate for too long before resynchronizing.
- A repeat-selection seek is intentional backward movement and must immediately pin the display to the selected cue start.
- Restarting playback or replacing the media must clear the prior floor so the new legitimate position is shown.

### Automated

Focused verification against current upstream `main`:

```text
dotnet vstest tests/UI/bin/Debug/net10.0/UITests.dll --TestCaseFilter:FullyQualifiedName~PlayheadFrameStepTests --Logger:console;verbosity=minimal

Passed: 16, Failed: 0, Skipped: 0
```

The focused class covers deliberate backward seeking, normal and repeat playback with a stale raw position, playback restart, high-speed frozen-position handling, and media replacement.

`RepeatPlayback_DoesNotSnapTheEstimateBackToAStaleRawPosition` is the direct regression test for the reported failure. It starts a 3x repeat-playback tick with a displayed estimate of 1.600 seconds and a stale raw player position of 1.100 seconds. Before this change the tick returns 1.100 seconds and fails the assertion that the displayed estimate must remain at or above 1.600 seconds. With this change, the assertion passes.

### Manual

On macOS arm64, normal and repeat playback were exercised at 1x and 3x after building the application. The displayed playhead did not move backwards during the tested runs.

## Out of scope

This change only affects presentation of the playback position and the related estimator lifecycle.

## Risks and rollback

The change is confined to the display estimator. Explicit seeks and new media playback have focused regression coverage. Revert this commit if cursor behavior regresses on an untested playback backend.

## AI assistance

OpenAI Codex materially assisted with implementation audit, focused verification, and PR preparation.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -24390,6 +24390,7 @@ public partial class MainViewModel :
         _videoOpenTokenSource?.Cancel();
         _audioTrack = null;
         await vp.Open(videoFileName, startPositionSeconds);
+        ResetPlayheadForMediaChange();
         _videoFileName = videoFileName;
         RefreshSubtitlePreview();
 
@@ -28736,6 +28737,11 @@ public partial class MainViewModel :
 
         if (isPlaying && !_pauseRequested && _playheadValid && _playheadLastRealSeconds >= 0)
         {
+            // Keep a continuously advancing playhead from visibly snapping back to a stale mpv
+            // position. Explicit seeks take the pin path above, and a playback restart re-seeds
+            // before this floor applies, so intentional position changes still take effect.
+            var playbackFloorSeconds = _playheadEstimateSeconds;
+            var preservePlaybackFloor = _playheadWasPlaying && !_playheadResyncOnPlay;
             var elapsedSeconds = (nowTimestamp - _playheadLastTimestamp) / (double)Stopwatch.Frequency;
             if (elapsedSeconds < 0)
             {
@@ -28802,7 +28808,7 @@ public partial class MainViewModel :
             // the same spot. Extrapolating through that freeze races the cursor ahead of the still-frozen
             // frame and then forces a visible sub-1x crawl to resync. Holding while frozen and gliding
             // once mpv resumes is seamless, since it resumes from where the cursor is held.
-            if (rawFrozenSeconds < PlayheadFreezeHoldSeconds)
+            if (rawFrozenSeconds < PlayheadFreezeHoldSeconds / speed)
             {
                 _playheadEstimateSeconds += elapsedSeconds * speed;
             }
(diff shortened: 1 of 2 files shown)


marcpbailey:
Update: upstream PR #14385 identifies a narrower confirmed cause for the repeat-playback cursor reversal addressed here. It pins the playhead at the repeat-selection seek site and uses SeekTo, preventing the estimate from remaining stale while mpv reports the pre-seek position.
That overlaps the repeat-seek portion of this draft and may make the broader estimator changes here unnecessary or require them to be revised. I am leaving this PR open while #14385 is reviewed, then will reconcile this proposal with the upstream result.

PR #14416, opened 2026-09-02
Video & Shortcut UX fixes: Open recent video menu, zero-time preview fix, and immediate key-up end timing

ghostminhtoan:
### Overview of Changes

This PR incorporates three clean, targeted improvements and bug fixes kept in separate, atomic commits (Please check video if needed: https://drive.google.com/drive/folders/1TsY3iSZ0Ah1PDKB7odK5zRvBklSpN-_w?usp=drive_link)

---

### 1. Fix "Set end time on key up" shortcut timing & navigation (`dec386a83`)
* **Problem**: Setting end time and auto-advancing to the next row on key up relied on the 50ms periodic UI timer tick loop. This introduced a slight timing delay/lag and unnecessary work inside the polling loop.
* **Fix**: Directly set `EndTime` from current player position and advance selection in `OnKeyUpHandler` upon key release. Cleaned up obsolete timer-tick code.

---

### 2. Fix subtitle preview disappearing when trailing zero-time lines exist (`1385b6305`)
* **Problem**: When a subtitle file contains entries at the end with start and end times equal to zero (`00:00:00,000` -> `00:00:00,000`, e.g., newly inserted draft rows or empty template lines), `libass` event timeline sequencing caused line 1 not to render in video preview during playback.
* **Fix**: Filter out zero-time paragraphs (`StartTime <= 0 && EndTime <= 0`) when preparing the ASS preview string in `MpvReloader.cs` and `VlcReloader.cs`. Left `AudioVisualizer.cs` untouched since waveform sorting already handles display order safely.

---

### 3. Add "Open recent video" menu option (`3359c5087`)
* **Problem**: While subtitle files had a `File > Reopen` menu, there was no quick way to reopen recently used standalone video files from the **Video** menu.
* **Fix**:
  - Added an "Open recent video" dynamic submenu under the **Video** menu (supported on both standard Avalonia and macOS native menu bars).
  - Maintained dedicated `Se.Settings.Video.RecentFiles` (`List<string>`) so that clearing recent videos (`Clear recent videos`) never clears or breaks `Se.Settings.File.RecentFiles` subtitle-media associations.
  - Added dedicated localization keys: `OpenRecentVideo` and `ClearRecentVideos`.

---

### Commits in this PR (in chronological order):
1. `dec386a83` - `fix(shortcut): set end time and advance to next subtitle on key up instead of timer tick`
2. `1385b6305` - `fix(preview): filter out zero-time paragraphs for video preview`
3. `3359c5087` - `feat(video): add open recent video menu option`

---

### Verification & Testing
- [x] Tested "Set end time on key up" shortcut: End time is immediately registered on key release and focus advances instantly.
- [x] Tested zero-time subtitle preview: Line 1 correctly previews over video even when rows 7+ have zeroed timestamps.
- [x] Tested "Open recent video" submenu & "Clear recent videos": Videos reopen properly without interfering with `File > Reopen` subtitle links.
- [x] Built in Debug & Release configurations with **0 Warnings, 0 Errors**.

### DIFF
--- a/src/ui/Assets/Languages/English.json
+++ b/src/ui/Assets/Languages/English.json
@@ -1,4 +1,4 @@
-﻿{
+{
   "title": "Subtitle Edit",
   "version": "v5.2.0-beta32",
   "translatedBy": "",
@@ -2545,6 +2545,8 @@
     "openSecondarySubtitleOnVideoPlayerDotDotDot": "Open second subtitle file...",
     "openSecondarySubtitleOnVideoPlayer": "Second subtitle file (on video player)",
     "removeSecondarySubtitleOnVideoPlayer": "Remove second subtitle file",
+    "openRecentVideo": "Open recent video",
+    "clearRecentVideos": "Clear recent videos",
     "cutVideoTitle": "Cut video",
     "cutVideoDotDotDot": "Cut video...",
     "embedSubtitlesDotDotDot": "Add/remove embedded subtitles...",
--- a/src/ui/Assets/Languages/Vietnamese.json
+++ b/src/ui/Assets/Languages/Vietnamese.json
@@ -2235,6 +2235,8 @@
     "openSecondarySubtitleOnVideoPlayerDotDotDot": "Phụ đề phụ trên trình phát video, mở...",
     "openSecondarySubtitleOnVideoPlayer": "Phụ đề phụ (trên trình phát video)",
     "removeSecondarySubtitleOnVideoPlayer": "Phụ đề phụ trên trình phát video, xóa",
+    "openRecentVideo": "Mở video gần đây",
+    "clearRecentVideos": "Xóa danh sách video gần đây",
     "cutVideoTitle": "Cắt video",
     "cutVideoDotDotDot": "Cắt video...",
     "embedSubtitlesDotDotDot": "Thêm/xóa phụ đề nhúng...",
--- a/src/ui/Features/Main/Layout/InitMenu.cs
+++ b/src/ui/Features/Main/Layout/InitMenu.cs
@@ -1,4 +1,4 @@
-﻿using Avalonia;
+using Avalonia;
 using Avalonia.Controls;
 using Avalonia.Data;
 using Avalonia.Data.Converters;
@@ -33,6 +33,13 @@ public static class InitMenu
         };
 
         UpdateRecentFiles(vm);
+
+        vm.MenuRecentVideos = new MenuItem
+        {
+            Header = Se.Language.Video.OpenRecentVideo,
+        };
+
+        UpdateRecentVideos(vm);
 
         var menu = vm.Menu;
         menu.DataContext = vm;
@@ -751,6 +758,7 @@ public static class InitMenu
                         },
                     },
                 },
+                vm.MenuRecentVideos!,
                 menuItemAudioTracks,
                 new Separator(),
                 new MenuItem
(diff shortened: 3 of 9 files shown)


niksedk (changes requested):
Thanks for the work on this. Before going into details: **could you please split this into separate PRs, one per change?** Three unrelated changes (plus the BOM stripping) in one PR are hard to review, revert, and bisect, and a problem in one part blocks the others. The recent-videos menu in particular is close to mergeable on its own once the points below are addressed.

Below are the findings per change (line numbers refer to the PR head, `3359c5087`).

### 1. Set end time on key up (`dec386a83`)

- **No lower bound on the end time** (`MainViewModel.cs:28146`). `OnKeyUpHandler` writes `EndTime = position` unconditionally. A quick tap of "set start / set end on key up / go to next" now yields a ~40 ms (or, after a pause + step back, negative-duration) line and discards the line's previous end time. SE4 guarded this with `if (p.StartTime.TotalSeconds + 0.1 < videoPosition) SetEndTime()`. The same applies to the insert variant, where the tap overwrites the `NewEmptyDefaultMs` default.
- **Live tracking while the key is held is gone**, and the PR doesn't mention it. The removed tick block updated `EndTime` every 50 ms for the insert variant, so the new line's waveform bar grew with the playhead while held. Now it sits at the default length and snaps at release. The go-to-next variant never really tracked live (the tick cleared it immediately), so moving *that* to key-up is a genuine fix. Suggest keeping the per-tick `EndTime = position` update for the armed line and doing only the final set + advance (with the SE4-style guard) on key-up.
- **Clock mismatch** (`MainViewModel.cs:28146`). Key-down sets `StartTime` from `vp.Position` (SMPTE-adjusted, rounded to whole ms), but key-up reads `vp.VideoPlayer.Position` (raw mpv seconds). With SMPTE timing enabled the end lands ~0.1% late (about 3.6 s at the one-hour mark), and it always carries fractional milliseconds. The expression is inherited from the old tick code, but since it's being moved it should use `vp.Position` with the same rounding as the key-down side.

### 2. Zero-time preview filter (`1385b6305`)

- **The secondary subtitle isn't filtered** (`MpvReloader.cs:183`, `VlcReloader.cs:39`). `SecondarySubtitleMerger.AddSecondarySubtitle` appends the second subtitle's paragraphs unfiltered after the `RemoveAll`, so the same broken event stream is produced when the zero-time rows are in the second subtitle file.
- **The predicate is narrower than the problem.** `Start <= 0 && End <= 0` only drops lines sitting at exactly 0→0. A zero/negative-duration line at any other position still becomes a `Dialogue` event, and a line with `Start=-500, End=200` (common after a shift) is kept and serialised as `0:00:00.-50`. A single check like `DurationTotalMilliseconds <= 0 || StartTime < 0` (or clamping negative starts) would cover all of these. Also, please put it in one shared place instead of pasting the same line into both reloaders. The two already duplicate the SMPTE stretch loop; a small `PreparePreviewCopy` helper next to `SecondarySubtitleMerger` would be the natural home.
- It would help to state the actual mechanism in the commit message. "libass event timeline sequencing" isn't specific enough for someone revisiting this later.

### 3. Open recent video (`3359c5087`)

- **Entries are recorded even when the open fails** (`MainViewModel.cs:24613`). Nothing between `vp.Open` and the recording block reports failure (`LoadFile` only logs, `WaitForPlayersReadyAsync` just times out). The drop handler forwards every non-subtitle extension to `VideoOpenFile`, so dropping a `.zip` or `.png` puts it at the top of the recent-videos menu. Record only after a successful load, or gate on `VideoPlayer.Duration > 0`.
- **Clear never saves** (`MainViewModel.cs:8384`). `CommandVideoClearRecentFiles` clears the list in memory but doesn't call `Se.SaveSettings()`, while the add path saves immediately. A crash or force-quit after clearing brings the list back.
- **The list leaks into settings export/import** (`SettingsImportExportViewModel.cs:373`). `SeVideo` is exported and imported as a whole object, and `SeFile.RecentFiles` is deliberately excluded from export, so the new list becomes the only recent-file history that travels in an exported settings file, and an import overwrites the local list with foreign paths.
- **Not covered by the settings reset** (`SettingsViewModel.cs:2651`). Options > Reset > "Reset recent files" only clears `File.RecentFiles`; the new list should be included there too.
- Smaller points: the `|| IsValidUrl(f)` branches in both menu builders and in `CommandVideoReopen` are unreachable, since `VideoOpenFile` returns early for URLs before the recording block (and the recorder excludes them anyway). Either record URLs too, so "Open from URL" streams show up, or drop the branches and keep `IsValidUrl` private. `UpdateRecentVideos` is a near-verbatim copy of `UpdateRecentFiles` in both menu hosts; a shared fill helper per host would avoid four copies of the same item/separator/clear tail. `VideoOpenFile` now calls `Se.SaveSettings()` twice per open when a subtitle is loaded (once here, once in `AddToRecentFiles`). And the `File.Exists` sweep over up to 25 paths on the UI thread now runs on every video open; with dead NAS/UNC entries that stalls the window for the SMB timeout each.

### BOM stripping

All seven `.cs` files and `English.json` lost their UTF-8 BOM. `English.json` in particular is rewritten with a BOM by both `EnglishLanguageFileUpToDateTests` and the Save-language-file command, so the next string change will flip line 1 back and add noise to an unrelated PR. Please restore the BOMs (or keep encoding changes out of feature PRs).

Issue #13341, opened 2026-08-07
"Point sync...." only works when a video is loaded.


fraternl:
I was quite baffled that "Point sync..." wasn't working when it was added just recently.

But it only works when a video is loaded.
Maybe the function should be grayed out when no video is loaded.

I wanted to enter a value from another instance of SE, but when I saw a value of 00 00 00.000 I stopped.

<img width="707" height="633" alt="Image" src="https://github.com/user-attachments/assets/c13b1d81-b433-4f5b-bced-9895ea0d9c26" />

<img width="740" height="676" alt="Image" src="https://github.com/user-attachments/assets/07d7d0ca-fb81-45e7-a973-3007bafdd13b" />

PR #14419, opened 2026-09-02
Sync dialogs: make the waveform under the video drag-resizable

### LINKED ISSUE #14414: Point sync waveform cannot be resized vertically
In the new beta, the waveform displayed in the Point sync window still has a fixed vertical size.

It would be very useful to make the waveform vertically resizable, similarly to the main waveform in Subtitle Edit. When the window contains several UI elements, the waveform can become quite small, which makes it harder to identify precise sync points.

The main Subtitle Edit waveform already supports vertical resizing, so having the same option in the Point sync window would make this feature much more practical.

I have checked the latest beta and I cannot find any way to resize this waveform vertically.

Thanks!

niksedk:
Fixes #14414

## Problem

The **Set sync point** dialog (Point sync) and the **Visual sync** dialog pinned their waveforms at a fixed 80 px under the video player. With a video open the player took all the free height and the waveform stayed a thin strip, which makes precise sync points hard to pick. The main window's waveform is drag-resizable; these dialogs had no way to resize theirs.

## Fix

Both dialogs now host the player and the waveform in a shared `VideoWaveformSplitGrid` (`src/ui/Features/Sync/`), laid out the way the main window lays out its video and waveform:

- A horizontal `GridSplitter` between the player and the waveform. The waveform row is pixel-sized and the video row is Star, so a drag only moves the waveform, as in the main window. The video keeps a minimum height so the waveform cannot eat it entirely.
- The rows and the handle collapse when either side is hidden: no waveform lent by the main window, or the videoless Set sync point dialog. Without a player, a lent waveform fills whatever height the window has, and the videoless height cap from #13341 is lifted for it.
- Visual sync's two panes follow each other's handle, so the start and end waveforms stay the same height.
- The dragged height is remembered per dialog (`SetSyncPointWaveformHeight` / `VisualSyncWaveformHeight` under the synchronization settings), since Point sync opens its dialog twice per sync.

## Tests

- `VideoWaveformSplitGridTests`: handle/row state for every visibility combination, drag bookkeeping, pane linking without echo, bad saved heights falling back to the default.
- `SyncWaveformResizeTests`: both windows construct with the split grid, the handle appears once peaks arrive, the height is persisted on close, Visual sync panes move together and collapse together.
- Full UI test project: 4575 passed, 1 skipped. Layout verified with headless Skia captures of both dialogs before and after a drag.

### DIFF
--- a/src/ui/Features/Sync/PointSync/SetSyncPoint/SetSyncPointWindow.cs
+++ b/src/ui/Features/Sync/PointSync/SetSyncPoint/SetSyncPointWindow.cs
@@ -64,7 +64,6 @@ public class SetSyncPointWindow : Window
 
         vm.AudioVisualizer = new AudioVisualizer
         {
-            Height = 80,
             Width = double.NaN,
             IsReadOnly = true,
             DrawGridLines = Se.Settings.Waveform.DrawGridLines,
@@ -131,11 +130,19 @@ public class SetSyncPointWindow : Window
         var buttonCancel = UiUtil.MakeButtonCancel(vm.CancelCommand);
         var buttonPanel = UiUtil.MakeButtonBar(buttonOk, buttonCancel);
 
+        // The waveform used to be pinned at 80 px under the player, too little to pick a precise
+        // sync point from; the drag handle between them mirrors the main window (issue #14414).
+        var split = new VideoWaveformSplitGrid(vm.VideoPlayerControl, vm.AudioVisualizer, Se.Settings.Synchronization.SetSyncPointWaveformHeight)
+        {
+            IsVideoVisible = vm.IsVideoVisible,
+            IsWaveformVisible = vm.IsAudioVisualizerVisible,
+        };
+
         // A Star row keeps its share of the height even when its child is hidden, so the row itself
         // has to collapse - otherwise the videoless dialog is mostly empty space.
         var rowVideoPlayer = new RowDefinition
         {
-            Height = vm.IsVideoVisible ? new GridLength(1, GridUnitType.Star) : new GridLength(0),
+            Height = vm.IsVideoVisible || vm.IsAudioVisualizerVisible ? new GridLength(1, GridUnitType.Star) : new GridLength(0),
         };
 
         vm.VideoPlayerControl.WithBindIsVisible(nameof(vm.IsVideoVisible));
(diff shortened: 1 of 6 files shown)


perkesfurmah-collab:
Thanks for the implementation, but I think there is a misunderstanding about the requested behavior

The new splitter allows the waveform panel itself to be resized vertically, but this is not quite what I was asking for.

What I meant was the vertical zoom of the waveform itself, like in Subtitle Edit's main window.

In the main window, the waveform can be vertically zoomed (for example from 136% to 500%) while the height of the video area remains unchanged. The waveform becomes much more readable without taking space away from the video.

In the Point Sync dialog, moving the new splitter upwards makes the waveform panel larger, but this reduces the height of the video. This can actually make the situation worse, because you have to sacrifice video space in order to get a reasonably readable waveform.

So what would be ideal for Point Sync is the same behavior as the main window: keep the video/waveform layout height unchanged, but allow the waveform itself to be vertically zoomed/scaled within its existing area.

The attached screenshots show the difference:

Main window at 136%: waveform is small.
Main window at 500%: waveform is much more readable, while the video area is unchanged.
Point Sync with the new splitter: increasing the waveform size reduces the video area.

Thanks again for working on it! I just wanted to clarify the intended behavior.
photo_5821245716788088883_y
photo_5821245716788088890_y
photo_5821245716788088876_y
photo_5821245716788088880_y

niksedk:
Thanks for the clarification! You're right, the resizable splitter in this PR trades video height for waveform height, which isn't what you wanted.

I've opened #14487 to add the vertical-zoom behavior you described: Shift+/Shift- (mirroring the main window's shortcut) now scales the waveform's amplitude in place, without touching the video/waveform layout height at all. It turns out Shift+scroll-wheel already did this here (AudioVisualizer handles it internally), it just had no keyboard shortcut - #14487 adds that, and extends it to Cut video's waveform too, which has the same video+waveform pairing.

The splitter from this PR stays as an independent, complementary control - useful when you do want more room for the waveform panel itself (e.g. with no video, or when zoom alone isn't enough).

perkesfurmah-collab:
Thank you for listening to and addressing our requests and for the great work you're doing with Subtitle Edit

perkesfurmah-collab:
Thanks! I tested RC2.

Vertical waveform zoom works correctly in the main Subtitle Edit window using Shift + mouse wheel, but it does not work at all in the Point Sync dialog.

I also tried Shift + + and Shift + -, but neither shortcut has any effect in Point Sync.

So the new vertical zoom behavior doesn't seem to be working in the Point Sync waveform on my side, although it works normally in the main window.

I’m testing this on a Spanish keyboard layout.

PR #14442, opened 2026-09-02
Add FixType property to IFixCommonError

ivandrofly:
## Summary
- Add a `FixType` enum (`Time`, `Formatting`, `Dialog`, `Punctuation`, `Casing`, `Spacing`, `Characters`, `Ocr`) in `src/libse/Enums`.
- Add `FixType FixType { get; }` to `IFixCommonError`, so the compiler enforces it on every implementation.
- Every fix rule (39 in libse plus `FixCommonOcrErrors` in the UI) declares the kind of fix it performs, so the UI and CLI can group, filter or sort rules by category.

No behaviour change on its own; the first consumer is the "Type" filter in the Fix common errors window, stacked on this PR.

## Test plan
- [ ] `dotnet build SubtitleEdit.sln` compiles (every `IFixCommonError` implementation has the new member)
- [ ] Spot-check a few classes: `FixCommas` is `Punctuation`, `FixShortGaps` is `Time`, `FixUnneededSpaces` is `Spacing`, `FixCommonOcrErrors` is `Ocr`
- [ ] `dotnet test tests/libse/LibSETests.csproj` passes

https://claude.ai/code/session_01SXEWrgv5uAnSzDgxFupSHZ

### DIFF
--- a/src/libse/Enums/FixType.cs
+++ b/src/libse/Enums/FixType.cs
@@ -1,0 +1,14 @@
+﻿namespace Nikse.SubtitleEdit.Core.Enums
+{
+    public enum FixType
+    {
+        Time,
+        Formatting,
+        Dialog,
+        Punctuation,
+        Casing,
+        Spacing,
+        Characters,
+        Ocr,
+    }
+}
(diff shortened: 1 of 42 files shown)


niksedk:
Hi @ivandrofly, thanks for this! The categorisation itself is sensible, but I'm going to close it together with #14443: I'd like to keep the Fix common errors window as it is for now, so I don't want to add the `FixType` member to the interface and all 40 rule classes without a consumer for it. If a category filter ends up on the roadmap, the mapping you did here (rule -> type) is easy to pick up again.

Thanks again!

PR #14443, opened 2026-09-02
Add a FixType filter to the Fix common errors rule list

ivandrofly:
Stacked on #14442 (the `FixType` property on `IFixCommonError`); this PR's own change is the top commit.

## Summary
- Add a "Type" combo to the step 1 toolbar of Fix common errors that filters the rule list by the rule's `FixType` (All, Time, Formatting, Dialog, Punctuation, Casing, Spacing, Characters, OCR).
- The type filter is combined with the "Search rules..." text box, which was created but never added to the window; both now sit in the step 1 toolbar and hide in step 2.
- `FixRuleDisplayItem` resolves its `FixType` once from the fix classes (one lazily built name-to-type map), so libse stays the single source of truth; the copy constructor carries it into profile clones.
- Filtering always reads the profile's full rule list, so hidden rules keep their selection and apply/save still see them. The active filter follows a profile switch.
- New language strings: `FixTypeFormatting`, `FixTypeDialog`, `FixTypePunctuation`, `FixTypeOcr` (the other four reuse existing General strings); `English.json` regenerated.
- Tests: `FixCommonErrorsRuleFilterTests` (type filter, AND with search, profile switch, hidden selection kept, type resolution for every default and language-specific rule, window wiring and step visibility).

## Test plan
- [ ] Open a subtitle, Tools > Fix common errors: the search box and a "Type" combo appear in the top row of step 1 and are hidden in step 2
- [ ] Pick "Time": only the four timing rules remain; type "short" in the search box: two remain; pick "All": the full list is back
- [ ] With "Time" selected, click Select all / Invert selection: only the visible rows change
- [ ] Select a profile, choose a type, switch profile: the filter stays applied
- [ ] Filter to "Time", untick a rule, pick "All", then "Go to apply fixes": rules hidden by the filter still run according to their tick state
- [ ] `dotnet test tests/UI/UITests.csproj`: `FixCommonErrorsRuleFilterTests` pass

https://claude.ai/code/session_01SXEWrgv5uAnSzDgxFupSHZ

### DIFF
--- a/src/libse/Enums/FixType.cs
+++ b/src/libse/Enums/FixType.cs
@@ -1,0 +1,14 @@
+﻿namespace Nikse.SubtitleEdit.Core.Enums
+{
+    public enum FixType
+    {
+        Time,
+        Formatting,
+        Dialog,
+        Punctuation,
+        Casing,
+        Spacing,
+        Characters,
+        Ocr,
+    }
+}
--- a/src/libse/Forms/FixCommonErrors/AddMissingQuotes.cs
+++ b/src/libse/Forms/FixCommonErrors/AddMissingQuotes.cs
@@ -1,4 +1,5 @@
 ﻿using Nikse.SubtitleEdit.Core.Common;
+using Nikse.SubtitleEdit.Core.Enums;
 using Nikse.SubtitleEdit.Core.Interfaces;
 using System;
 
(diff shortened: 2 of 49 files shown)


niksedk:
Hi @ivandrofly, closing this along with #14442 - see my note there. I'd rather keep the Fix common errors rule list simple for now, so I'm not going to add a type filter at this point. The observation that the "Search rules..." box was never added to the window is a good catch though, so I'll look at that separately.

Thanks for the work on it!

PR #14454, opened 2026-09-02
Waveform click/wheel landed a GOP away and hopped a frame forward (#14441)

### LINKED ISSUE #14441: 5.0.2 Beta32 The cursor is not positioned in the area I want, it moves as it pleases
The cursor is not positioned in the area I want, it moves as it pleases, I don't know if it is a new option that I was looking for to configure and I didn't find anything, although there are parts of options that are not translated into Spanish so I don't know if it could be one of those, but when I want to put the cursor at a certain point it doesn't do so and moves wherever it wants.

https://github.com/user-attachments/assets/3c21d768-3e3e-490f-ad6b-b54920b15ba2

niksedk:
Fixes #14441 (beta32 regression: "the cursor is not positioned where I want", wheel steps that keep landing on the same spot).

## What was wrong

**1. Mid-burst keyframe seeks never got their exact landing.** Beta32's two-tier scrub seeking serves a seek issued while another is in flight with `absolute+keyframes`. On a long-GOP file that lands on the keyframe *before* the target, seconds back. The deferred exact seek that was supposed to fix that up was skipped every time: mpv reports the seek target as `time-pos` while a seek is in progress, and the observed-property cache only refreshes after the event queue drains, so at the restart event the "already landed close enough" check compared the target with itself. Wheel notches then kept landing on the same keyframe, and scrolling back jumped a whole GOP (the GIF in the issue).

**2. A one-frame hop forward on every click in frame mode.** The Position getter skipped the paused seek-target cache in frame mode (forced on for EBU STL), so after a click the reported position moved from the click to the first decoded frame at or after it, and the waveform cursor followed.

## Fix

- The exact follow-up is unconditional once a burst settles (its restart arrived, nothing newer replaced it). Never gate a seek decision on mpv's cached time-pos.
- A burst needs three seeks in a row: a waveform click issues two seeks in one input event (pointer release, then tap), and that pair must stay exact. Drags and wheel spins pay two exact seeks at the start instead of one.
- The paused seek-target cache is served in frame mode too.

## Verification

Live libmpv core with a 30 s h264 file (keyframes every 10 s): a raw `seek 12.5 absolute+keyframes` lands at 10.000, and before the fix a click pair left mpv at 0.000 with the follow-up debt standing. After: click pairs, 100 ms wheel notches and 31-step drags (playing and paused) all land on the final target with nothing owed.

Tests: `ScrubSeekPolicyTests` updated for the new rules; three new live-core tests in `LibMpvEventLoopTests` (burst pays exactly one follow-up, click pair stays exact, frame-mode paused seek reports the target). 33 mpv/scrub/playhead tests pass.

### DIFF
--- a/src/ui/Logic/VideoPlayers/LibMpvDynamic/LibMpvDynamicPlayer.cs
+++ b/src/ui/Logic/VideoPlayers/LibMpvDynamic/LibMpvDynamicPlayer.cs
@@ -74,6 +74,7 @@ public sealed class LibMpvDynamicPlayer : IDisposable, IVideoPlayer
     // has already been and gone.
     private long _scrubFollowUpSeekId;     // keyframe seek owing an exact landing, 0 if none
     private long _scrubFollowUpTargetBits; // that seek's target, as double bits
+    private bool _lastSeekIssuedInFlight;  // the newest seek was issued while a seek was in flight (under _seekStateLock)
 
     // Guards every publish and claim of the seek generation + follow-up debt (IssueSeek, the
     // follow-up/settle/cancel paths). The id, target and debt slot are three separate fields, so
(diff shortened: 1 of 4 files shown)


Surlime:
Confirmed. Fixed in v5.2.0-rc1 (at least on my end). TYSM 🙏

PR #14505, opened 2026-09-04
Sync/Cut video: zoom the waveform on Shift + main-row plus/minus too

niksedk:
Follow-up to #14487, from the [rc2 test report](https://github.com/SubtitleEdit/subtitleedit/pull/14419#issuecomment-5537409823) on #14419.

## Problem

Shift + plus/minus vertical waveform zoom in **Set sync point**, **Visual sync** and **Cut video** only matched `Key.Add` / `Key.Subtract`, which only the numeric keypad produces. The plus and minus keys on the main key row report as `Key.OemPlus` / `Key.OemMinus`, so on a laptop or a Spanish keyboard the shortcut could never fire. Confirmed with a headless probe against rc2: Shift+NumPadAdd changed the zoom, Shift+OemPlus did not.

## Fix

- Set sync point and Visual sync: match `OemPlus` / `OemMinus` alongside `Add` / `Subtract`.
- Cut video: it resolves the user's configured main-window zoom shortcut first (unchanged). When no shortcut claimed the key, Shift+`OemPlus` / `OemMinus` now zoom as a fallback, so the three dialogs behave the same.

Not touched: the main window's default binding is still Shift+Add / Shift+Subtract, rebindable in Settings.

## Not fixed here

The same report says Shift + mouse wheel does nothing in Point sync. The Avalonia path works in a headless test (Shift+wheel over the dialog waveform changes the zoom), so that one is environmental. Most likely the mpv native child window holding keyboard focus on Windows after a click on the video, so wheel and key messages never reach Avalonia. Needs a repro on Windows.

## Tests

- `SyncWaveformVerticalZoomKeyTests`: Set sync point zooms from both key pairs and ignores plus without Shift; Visual sync zooms only the focused pane from the main-row keys.
- `CutVideoVerticalZoomKeyTests`: main-row keys zoom via the fallback; plus without Shift does nothing.
- Sync test folder: 29 passed.

### DIFF
--- a/src/ui/Features/Sync/PointSync/SetSyncPoint/SetSyncPointViewModel.cs
+++ b/src/ui/Features/Sync/PointSync/SetSyncPoint/SetSyncPointViewModel.cs
@@ -591,12 +591,12 @@ public partial class SetSyncPointViewModel : ObservableObject
             e.Handled = true;
             SetVideoPosition(CurrentPositionSeconds + 0.5);
         }
-        else if (e.Key == Key.Add && e.KeyModifiers.HasFlag(KeyModifiers.Shift))
+        else if ((e.Key == Key.Add || e.Key == Key.OemPlus) && e.KeyModifiers.HasFlag(KeyModifiers.Shift))
         {
             e.Handled = true;
             WaveformVerticalZoomIn();
         }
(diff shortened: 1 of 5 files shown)


perkesfurmah-collab:
It still doesn't work on my Linux distributions; it's not a feature I use often, but I think it made the waveform implementation in the point-time synchronization window more complete.

PR #14507, opened 2026-09-04
Batch convert: hide the main window while the tool window is open

### LINKED ISSUE #14502: SE5 not minimized when I open Batch Convert
In SE4, when I open Batch Convert, the main SE4 window is minimized and only the Batch Convert window remains visible. This is convenient because I can easily drag and drop files into it.

However, in SE5, the main SE5 window stays open. Is this intentional behavior?

niksedk:
Fixes #14502

SE4 hid the editor while Batch convert was open, so only the tool window was on screen (the reporter's drag-and-drop workflow). SE5 kept the main window visible behind the modal dialog.

## Why not just hide/minimize around the existing modal

- Avalonia's `Window.Hide()` cascades to every owned window, so hiding the owner takes the modal dialog down with it.
- `Window.ShowDialog(owner)` throws on a hidden owner, so hide-then-open fails too.
- Minimizing is no better: on Windows an owner's owned windows go with it.

## Change

- New `IWindowService.ShowWithOwnerHiddenAsync` + static `WindowService.ShowWithWindowsHiddenAsync`: shows the window as an unowned top-level (`CenterScreen`, remembered position still wins), hides the owner and any *visible* companion windows first, awaits `Closed`, then re-shows them and activates the owner. Restoration runs in a `finally`, so the hidden windows can't be lost if showing throws.
- `ShowToolsBatchConvert` uses it, passing the undocked video/waveform windows as companions so they hide and return with the editor (matching SE4, where they were owned forms). Video is paused first as with the modal path; shortcut keys are cleared and focus restored afterwards.
- A full-screen main window keeps the plain modal path (hiding a full-screen window leaves an empty space on macOS).
- Test fakes of `IWindowService` gained the new member.

## Tests

- New `tests/UI/Logic/ShowWithWindowsHiddenTests.cs`: owner hidden while open and back + active on close; visible companions restored, never-shown ones left hidden; minimized owner comes back as Normal.
- Full `UITests` project: 4749 passed, 1 skipped.
- Not verified interactively on Windows/macOS. Worth a manual check on Windows that a maximized main window returns maximized (Avalonia's Win32 impl re-shows with the stored `WindowState`, so it should).

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -7177,7 +7177,29 @@ public partial class MainViewModel :
     [RelayCommand]
     private async Task ShowToolsBatchConvert()
     {
-        await ShowDialogAsync<BatchConvertWindow, BatchConvertViewModel>();
+        // As in SE4, the editor gets out of the way while Batch convert is open: the main window
+        // (and the undocked video/waveform windows) are hidden and only the tool window is on
+        // screen, then everything comes back when it closes (#14502). A full-screen main window
+        // is left where it is - hiding one leaves an empty space on macOS - so that case keeps
+        // the plain modal dialog.
+        if (Window == null || Window.WindowState == WindowState.FullScreen)
+        {
+            await ShowDialogAsync<BatchConvertWindow, BatchConvertViewModel>();
+            return;
+        }
+
+        var videoPlayerControl = GetVideoPlayerControl();
+        if (videoPlayerControl != null)
+        {
+            PauseVideoAndFreezePlayhead(videoPlayerControl);
+        }
+
+        await _windowService.ShowWithOwnerHiddenAsync<BatchConvertWindow, BatchConvertViewModel>(
+            Window,
+            [_videoPlayerUndockedViewModel?.Window, _audioVisualizerUndockedViewModel?.Window]);
+
+        _shortcutManager.ClearKeys();
+        RestoreFocusIfLost();
     }
 
     [RelayCommand]
--- a/src/ui/Logic/WindowsService.cs
+++ b/src/ui/Logic/WindowsService.cs
@@ -73,6 +73,27 @@ namespace Nikse.SubtitleEdit.Logic
             Action<TViewModel>? configureViewModel = null, Action<TWindow>? configureWindow = null)
             where TWindow : Window
             where TViewModel : class;
+
+        /// <summary>
+        /// Shows a window of type TWindow as a stand-alone top-level window while
+        /// <paramref name="owner"/> - and any of <paramref name="companions"/> currently on
+        /// screen - are hidden, and brings them back once it closes. This is how SE4 ran
+        /// Batch convert: the editor disappears and only the tool window remains (#14502).
+        ///
+        /// It can't be a modal dialog: Avalonia's <see cref="Window.Hide"/> cascades to every
+        /// owned window (the dialog would vanish with its owner) and
+        /// <see cref="Window.ShowDialog(Window)"/> refuses a hidden owner. Minimizing the owner
+        /// is no better - on Windows an owner's owned windows go with it.
+        /// </summary>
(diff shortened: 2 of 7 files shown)


GrampaWildWilly:
> Worth a manual check on Windows that a maximized main window returns maximized

You know I can do that for you.

My first test was to open my usual undocked SE (RC 2 at the moment).  Then I clicked the Minimize button with the mouse (I know, sometimes I do break down & use the mouse -- reluctantly) for the main window.  The main window closed.  The Audio visualizer remained open on the primary monitor.  The video player remained in fullscreen on the secondary monitor.

Second test.  Open undocked RC 2.  Ctrl+b to open the Batch convert dialog.  This causes the Audio visualizer window to go out of sight.  Alt+Tab to the Audio visualizer does bring it back into view on top of the Batch convert window.  But Alt+Tab back to the Batch convert window makes the Audio visualizer disappear again.  Ctrl+Alt+Tab shows the Audio visualizer is second on the Task Switch ribbon.  I don't think this means it's minimized but I'm not sure.  Clicked the Minimize button of the main window.  The Batch convert dialog window title bar flashed a few times.  That is the signal that Windows has refused to minimize the main window when the Batch convert window is open.

Third test.  Maximize the main window.  Then minimize it.  Same as before: the other 2 windows stay open.  Alt+Tab back to the main window does bring it back still maximized.

Fourth test.  Maximize the main window.  Ctrl+b.  Click the minimize button for the main window.  Same as before: the minimze request is rejected & the Batch convert window flashes.

Between tests, restore main window to its normal size.  For me that's not quite to the sides of the display but Win+Shift+UpArrow to snap the top & bottom of the main window to the edges of the viewable aperture: top of display, top edge of the Task Bar.  You need the sides to be just slightly within the edges of the screen for the vertical snap to work.  So I normally have the main window almost-but-not-quite maximized.

Fifth test.  Open RC 2 in its near-maximized state.  Ctrl+b.  Click the Minimize control on the Batch convert window.  That window minimizes AND the main window also minimizes.  Since the Audio visualizer was already out of sight, I don't know what it did.  Checking the Ctrl+Alt+Tab ribbon at this point shows the main SE window in position 2, the Audio visualizer in position 3, & the video player in position 4.  None of these tests minimized the video player.  It remained solidly in fullscreen on the secondary monitor at all times throughout all these tests.  Curiously, with the Batch converter window minimized, it is nowhere to be found on the Ctrl+Alt+Tab ribbon.  Alt+Tab to the main SE window magically resurrects the Batch convert window on top, like it was when I minimized it.

Not sure but I think some of this contradicts some of what you said about minimizing also taking down children.  Some of it.  Not all of it.  It's all rather strange behavior.  I have to say I never maximize or minimize any of these windows, nor anything else in my system for that matter.  I run most apps in that near-maximized state, snapped to the top & bottom edges.  If I want to work in another app, I just Alt+Tab to it.  It then sits on top & whatever is in background is just not visible.  I have to say I wouldn't care much if you did anything particular in SE to handle maximizing & minimizing.  I wouldn't use it.

Strangely, I now have a small tooltip that says "Minimize" hovering near the upper right corner of my screen.  It has survived several close/open/close cycles of SE.  Swiping the mouse over it does not make it go away.  It appears over every single open window in my system . . .  except, of course, anything on the secondary monitor.  NVidia just notified me there's a new driver update so I'll be rebooting here shortly.  I'm sure that will clear it.  This sounds like a Windows problem, not an SE problem.  Still, it was triggered by minimizing the main SE window.  I don't quite know what to make of this.

You know what?  I clicked on the little tooltip.  Then I hit Alt+F4.  And it went away.  I have no idea what window it was attached to.  It wasn't anything in my Alt+Tab ribbon.  I just went through every open task on my system verifying that the tooltip was visible over every window.  When I hit Alt+F4, it didn't close any of those tasks.  I have no idea what that tooltip was attached to.  Maybe you have a clue.  I sure don't.  So weird . . .  I guess that's Windows for you.

PR #14567, opened 2026-09-05
Google Cloud STT: send lossless audio, allow dynamic batching, recover truncation

muaz978:
Four follow-ups to #14561, from reviewing it against the runs the accuracy figures in that
PR come from. Details and evidence are in my comment there; this is the code.

### 1. Send lossless audio instead of 32 kbit/s mp3

`isOpenAiEngine` is `GetEffectiveSelectedEngine() is IOnlineSttEngine`, which is true for
`GoogleCloudSttEngine`, so extraction fell through to the `"mp3"` branch.
`GetFfmpegTranscodeFormatString` targets ~32 kbit/s there, and says why: it
*"keeps a 2-hour video well under OpenAI's 25 MB upload limit"*.

That limit does not apply to this engine. The audio goes to a Cloud Storage bucket, and it
is already split into 18 minute chunks. More importantly, the numbers quoted for this
engine, the 13,175 timed words and 54.3% speech density, were all measured on 16 kHz mono
**flac**. I have never measured `chirp_3` on 32 kbit/s mp3, so I cannot say what it costs,
only that it is not the input the evidence describes.

Measured on a real file, an 18 minute chunk is **20.7 MB as flac against 4.1 MB as mp3**.
For a bucket upload that difference is not worth any accuracy risk.

The new `"flac"` case passes `-sample_fmt s16` explicitly. That is not defensive padding:
without it ffmpeg encodes 24-bit flac from an AAC source, which is 78% larger than the
16-bit file and larger than the raw PCM it replaces. Verified on ffmpeg 8.1 and 9.0.1.

### 2. Allow `DYNAMIC_BATCHING`

`BuildRequestBody` sent no `processingStrategy`, so every transcription bills at $0.016 per
minute instead of $0.003. On my 139.74 minute episode that is $2.23 against the $0.42 I
actually paid.

It carries no latency guarantee, so I left it **off by default** and added a checkbox
rather than changing behaviour under anyone. For what it is worth it measured 13.6x
realtime on that episode, but Google promises nothing, and the default timeout is an hour.

### 3. Recover silent truncation instead of only logging it

The detection you added is correct. My concern was what follows: the short transcript is
returned as though it were complete, with one line in the tools log.

This is the defect that cost me 11.4 minutes of dialogue out of a single 18 minute chunk,
with the operation reporting success. The user gets a subtitle that looks finished and has
a hole in the middle, and finds out while watching.

The tail is now re-cut from a second before the last good word, resubmitted once, and
merged back with the offsets shifted. One attempt only; if the retry truncates too there is
nothing further to do automatically and the log line still records it.

### 4. Bound the word range check when no billed duration is reported

```csharp
if (start < 0 || end < start || (billed > 0 && end > billed + 1))
```

If `totalBilledDuration` is absent, `billed` is 0 and this degrades to "not negative and
not inverted". The offset that motivated the guard, a word claiming 6,324 s inside a
1,080 s chunk, would pass. It now falls back to the largest `resultEndOffset`, which is
reported per result and does not depend on the billing field being there.

## Testing

282 SpeechToText tests pass, including four new ones covering the strategy field being
present only when asked for, and the range check with and without a billed duration.

Still outstanding on my side is running this against the live API, which I owe you from
#14561. I am setting up a project key for that and will report back either way, including
on the `languageCodes: ["auto"]` default, which none of my runs exercised.

### DIFF
--- a/src/ui/Features/Video/SpeechToText/Engines/GoogleCloudSttEngine.cs
+++ b/src/ui/Features/Video/SpeechToText/Engines/GoogleCloudSttEngine.cs
@@ -67,6 +67,9 @@ public class GoogleCloudSttEngine : IOnlineSttEngine
         "Long audio must be read from a Cloud Storage bucket, so one named <project>-subtitle-edit-stt is created " +
         "on first use (another name can be set as GoogleCloudSttBucketName in Settings.json). Uploaded audio is " +
         "deleted after each run.\n\n" +
+        "Cost: about $0.003 per minute of audio, because dynamic batching is used by default. Google gives no " +
+        "latency guarantee for it, so set GoogleCloudSttDynamicBatching to false in Settings.json to transcribe " +
+        "at the normal rate of about $0.016 per minute instead.\n\n" +
         "Region: 'us' or 'eu' for chirp_3, or a specific region for other models. " +
         "Model: chirp_3, chirp_2, long, latest_long... " +
         "Language hint: leave empty for automatic detection (Chirp models), or a BCP-47 code such as en-US.");
--- a/src/ui/Features/Video/SpeechToText/GoogleCloud/GoogleCloudSttService.cs
+++ b/src/ui/Features/Video/SpeechToText/GoogleCloud/GoogleCloudSttService.cs
@@ -25,7 +25,7 @@ public class GoogleCloudSttSettings
     public string Language { get; set; } = string.Empty;
     public string BucketName { get; set; } = string.Empty;
     public int TimeoutSeconds { get; set; } = 3600;
-    public bool DynamicBatching { get; set; }
+    public bool DynamicBatching { get; set; } = true;
     public Action<string>? Logger { get; set; }
 
     public static GoogleCloudSttSettings FromConfiguration() => new()
@@ -396,8 +396,8 @@ public class GoogleCloudSttService : ISttTranscriber
         var files = new[] { new { uri = gcsUri } };
         var output = new { inlineResponseConfig = new { } };
 
-        // DYNAMIC_BATCHING is billed at about a fifth of the normal rate. Only sent when
-        // asked for: it carries no latency guarantee, and the default timeout is an hour.
+        // DYNAMIC_BATCHING is billed at about a fifth of the normal rate, so it is the
+        // default. It carries no latency guarantee, which is why it stays switchable.
         return settings.DynamicBatching
             ? JsonSerializer.Serialize(new { config, files, recognitionOutputConfig = output, processingStrategy = "DYNAMIC_BATCHING" })
             : JsonSerializer.Serialize(new { config, files, recognitionOutputConfig = output });
--- a/src/ui/Features/Video/SpeechToText/GoogleCloud/GoogleCloudSttService.cs
+++ b/src/ui/Features/Video/SpeechToText/GoogleCloud/GoogleCloudSttService.cs
@@ -25,6 +25,7 @@ public class GoogleCloudSttSettings
     public string Language { get; set; } = string.Empty;
     public string BucketName { get; set; } = string.Empty;
     public int TimeoutSeconds { get; set; } = 3600;
+    public bool DynamicBatching { get; set; }
     public Action<string>? Logger { get; set; }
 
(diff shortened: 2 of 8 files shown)


muaz978:
Following up with the comparison I owed you: Subtitle Edit's output against the pipeline
that produced the results I quoted, run over the **same** API response so nothing but the
processing differs.

Method: captured a real `batchRecognize` response (3 minutes, 322 words, `chirp_3`, live
API), then ran Subtitle Edit's own `ParseResponse` over it with #14567 applied, and ran the
reference pipeline's cue builder over the identical words.

### Transcript

| | Reference pipeline | Subtitle Edit with #14567 |
|:--|:--|:--|
| Words in | 322 | 322 |
| Text agreement | \multicolumn 99.7% | |
| First cue starts | 0.000 s | 0.000 s |
| Last cue ends | 179.960 s | 179.960 s |

The 0.3% is a single word, and Subtitle Edit is the one that gets it right. Google returned:

```json
{ "startOffset": "162.760s", "endOffset": "160.320s", "word": "Sağ" }
```

It ends 2.4 seconds before it starts. Your `end < start` check discards it; the reference
pipeline had no range check and kept it, which is what put a corrupt cue in its output.
That is 1 word in 322, consistent with the 0.8% rate I measured across a full episode, so
the guard earns its place on this sample alone.

### Cue segmentation

This is where the two genuinely differ:

| | Reference | Subtitle Edit |
|:--|--:|--:|
| Cues | 66 | 43 |
| Mean cue length | 1.61 s | 2.61 s |
| Longest cue | 6.1 s | 6.1 s |
| Shared cue start times | \multicolumn 36 | |

`BuildSegmentsFromWords` breaks on a 0.5 s gap or 80 characters. The reference also broke on
sentence-ending punctuation and a 6 second cap, which is what produces the shorter cues.

Both are perfectly usable, and I deliberately did **not** change the shared helper, since it
is used by the OpenAI and OpenRouter paths too and consistency across engines seems like the
better default. But if you would rather this engine produced denser cues now that it has
real punctuation and real word timings to work with, breaking on sentence endings is a
two line change and I am happy to add it. Your call.

Net: same API, same model, same words, same timeline anchors. The only content difference
went in Subtitle Edit's favour.

niksedk:
Thx for the PR - No new UI features/language tags until 5.3 - do you want to change some defaults until then?

muaz978:
Understood on the freeze. I have dropped the checkbox and the language tag, so the PR now
touches no UI and adds no strings. `GoogleCloudSttDynamicBatching` stays as a
`Settings.json` key with the same shape `GoogleCloudSttBucketName` already has, and it is
documented in the engine's own `GetHelpText()` rather than in a new language string. The
checkbox can be a two line follow up once 5.3 opens, if you want it then.

To your question, yes, there are two default changes here and they are the substance of the
PR rather than the UI:

**Audio format, mp3 to flac.** This is the one I would push for. The engine currently
inherits the ~32 kbit/s mp3 path whose stated reason is OpenAI's 25 MB upload limit, which
does not apply when the audio goes to a Cloud Storage bucket. I measured it against the live
API on identical source audio: **85.6% word agreement between the two encodings**, with real
substitutions like `gözünün` against `ağzının` and `ceketini` against `kravatını`. An 18
minute chunk is 20.7 MB as flac against 4.1 MB as mp3, which costs nothing on a bucket
upload. No UI, no strings, just a better default.

**Dynamic batching, now on by default.** $0.003 per minute instead of $0.016, so a 2.5 hour
episode costs about $0.44 instead of $2.32. My reasoning for making it the default rather
than opt in: it is a 5x difference in money the user spends without ever being asked about
it, and almost nobody will discover a `Settings.json` key. It measured 13.6x realtime on a
140 minute episode.

The honest counter argument, which is yours to weigh rather than mine: Google publishes no
latency guarantee for batch mode, so a slow queue would surface as the existing one hour
timeout rather than as a clear message. If you would rather it defaulted off and stayed
opt in, say so and I will flip it, it is one line.

The remaining changes in the PR have no default or UI dimension at all, they are corrections:
cues cut from word timings instead of one cue per result, the first word of each result no
longer dropped, truncation recovered rather than only logged, and the range check bounded
when no billed duration is reported.

284 SpeechToText tests pass.

PR #14582, opened 2026-09-05
Google Cloud STT: allow gcloud credentials, and object level access to the bucket

muaz978:
Two setup barriers that stop this engine being usable inside an organisation that has a real
cloud administrator. Both surfaced while testing #14561 and #14567 against a live project
that is managed by someone other than me, which turned out to be a more revealing
environment than a project I own.

No UI, no language tags, so this should be fine before 5.3.

## 1. A service account key was the only accepted credential

`GoogleCredential.FromFileAsync(_settings.KeyFile)` was the whole auth story. That works
when the person transcribing owns the cloud project. It does not work in a managed org:

- Creating a key is commonly blocked by the `constraints/iam.disableServiceAccountKeyCreation`
  org policy. The project I tested on has it.
- Issuing one needs `iam.serviceAccounts.create`, a role most people who want to transcribe
  a video will not hold. I do not hold it on that project, and neither will most users.

So the engine fails at setup rather than at transcription, which is a harder problem to
diagnose and a much harder one for a user to fix on their own.

An empty key file now falls back to **Application Default Credentials**, which
`gcloud auth application-default login` writes and which every Google client library already
understands. Nothing else changes: with a key file configured the behaviour is exactly as
before, including your `CredentialFactory.FromFileAsync<ServiceAccountCredential>` change
from 61197617f, which I kept.

**Verified end to end against the live API with no key file at all:**

```
resolved project: <redacted>
progress: Uploading audio to Cloud Storage...
progress: Submitting transcription...
progress: Waiting for transcription to complete...
Google Cloud: dropped word 'Sağ' with impossible timing 162.76-160.32 s
segments: 43
first cue: 0.000-2.600 Ay, ay, ay, ay, ay.
```

ADC rarely names a project, so `ResolveProjectId` now tries, in order: the new
`GoogleCloudSttProjectId` setting, the key file, `GOOGLE_CLOUD_PROJECT`, `GCLOUD_PROJECT`,
and the ADC file's `quota_project_id`. When none of those exist it says so, instead of
failing inside a JSON parse.

## 2. The bucket probe rejected object level access

`EnsureBucketAsync` threw on any non 404 failure of its metadata `GET`. Reading bucket
metadata needs `storage.buckets.get`, which object level roles such as **Storage Object
Admin do not include**.

So a user whose administrator pre-creates the bucket and grants object access to it, which
is the normal arrangement in a locked down org, is stopped by a check they were never meant
to pass, despite being able to upload and delete objects in that bucket perfectly well.

A 403 on a bucket **the user named explicitly** is now treated as "it exists and is not mine
to inspect". A 403 on a bucket this code derived still fails, because that is a genuine
misconfiguration rather than a permissions arrangement.

## Testing

287 SpeechToText tests pass, 3 new ones covering project resolution from each source.

Two notes, neither caused by this PR:

- **`SpeechToTextWindowLayoutTests.EveryChild_SitsInsideTheDefinedRows` is currently failing
  on main.** I confirmed it fails on a clean `upstream/main` checkout with none of my
  branches applied. It is a teardown race rather than a layout assertion:
  `CheckWhisperCppForUpdateAsync` shows a MessageBox after the test's window has closed, and
  Avalonia throws `Cannot show a window with a closed owner`. Flagging it since it makes a
  green run hard to recognise.
- Thanks for cleaning up the nullable warnings in my truncation recovery in 61197617f. Those
  were mine and I should have caught them; I had been filtering build output for errors only.

## What is waiting for 5.3

Three settings are reachable today only by editing `Settings.json`, and all three are things
a user configuring this engine will actually need:

| Setting | Why it matters |
|:--|:--|
| `GoogleCloudSttProjectId` | Required when signing in with ADC, since those credentials usually name no project |
| `GoogleCloudSttBucketName` | Required by anyone whose administrator pre-creates the bucket, which is the case this PR unblocks |
| `GoogleCloudSttDynamicBatching` | The difference between about $0.44 and $2.32 for a 2.5 hour episode |

I have the settings rows and language tags written and building on a branch already, held
back because of the freeze. Say the word once 5.3 opens and I will send it as a small PR; it
is three rows and three strings, no behaviour change.

### DIFF
--- a/src/ui/Features/Video/SpeechToText/Engines/GoogleCloudSttEngine.cs
+++ b/src/ui/Features/Video/SpeechToText/Engines/GoogleCloudSttEngine.cs
@@ -37,8 +37,10 @@ public class GoogleCloudSttEngine : IOnlineSttEngine
         // No key file is fine when the machine already has Application Default Credentials,
         // which is often the only route available: creating a service account key is
         // commonly blocked by org policy.
-        var hasKeyFile = !string.IsNullOrWhiteSpace(settings.KeyFile) && File.Exists(settings.KeyFile);
-        if (!hasKeyFile && !GoogleCloudSttService.HasApplicationDefaultCredentials())
+        // A configured key file must exist though: the service uses it whenever one is set,
+        // so a stale path would otherwise fail at transcription time instead of here.
+        var hasKeyFile = !string.IsNullOrWhiteSpace(settings.KeyFile);
+        if (hasKeyFile ? !File.Exists(settings.KeyFile) : !GoogleCloudSttService.HasApplicationDefaultCredentials())
         {
             configErrorMessage = Se.Language.General.OnlineSttApiKeyMissing;
             return null;
--- a/src/ui/Features/Video/SpeechToText/Engines/GoogleCloudSttEngine.cs
+++ b/src/ui/Features/Video/SpeechToText/Engines/GoogleCloudSttEngine.cs
@@ -33,7 +33,12 @@ public class GoogleCloudSttEngine : IOnlineSttEngine
     public ISttTranscriber? CreateTranscriber(out string? configErrorMessage)
     {
         var settings = GoogleCloudSttSettings.FromConfiguration();
-        if (string.IsNullOrWhiteSpace(settings.KeyFile) || !File.Exists(settings.KeyFile))
+
+        // No key file is fine when the machine already has Application Default Credentials,
+        // which is often the only route available: creating a service account key is
+        // commonly blocked by org policy.
+        var hasKeyFile = !string.IsNullOrWhiteSpace(settings.KeyFile) && File.Exists(settings.KeyFile);
+        if (!hasKeyFile && !GoogleCloudSttService.HasApplicationDefaultCredentials())
         {
             configErrorMessage = Se.Language.General.OnlineSttApiKeyMissing;
             return null;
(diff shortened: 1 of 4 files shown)


muaz978:
The `test` check on this PR failed, and it is not this PR. Recording the evidence so it does
not eat your review time.

Two tests failed:

- `FixCommonErrorsUnfixableErrorsTests.Scan_ReportsErrorsThatCouldNotBeFixed`
- `SpeechToTextQualityReportTests.Fix_FixShortDuration_RunsAfterSplitLines`

**The identical pair fails on main, on work unrelated to this.** The run for #14583, the
blank video PR, failed on exactly those two tests. Across the last eight runs on main, three
failed, including #14577 for binary edit, which touches nothing near speech to text.

Locally both tests pass on a clean `upstream/main` checkout in isolation, and they pass with
this branch applied. They only fail inside the full suite, and the two attempts in this PR's
own run failed on *different* tests, which is the signature of shared state between tests
rather than a real regression. The workflow's built in retry hit it twice.

This branch touches `GoogleCloudSttService`, `GoogleCloudSttEngine`, `SeTools` and the
Google Cloud tests. It has no path to `FixCommonErrors`.

Happy to dig into the flakiness as a separate PR if it is getting in your way. It looks like
static state on `Se.Settings` leaking between tests, given both are order dependent and
neither reproduces alone, but I did not want to widen this PR to chase it.

PR #14620, opened 2026-09-06
Add option to show original subtitle on waveform

pdjdev:
## Summary

Adds an option to show the original subtitle on the waveform.

## Background

Subtitle Edit can already show the working subtitle and the original subtitle side by side in the subtitle grid, but the waveform only shows the subtitle currently being edited.

That makes it hard to compare timing with the original subtitle.

VisualSubSync, which I used for a long time <sub>before fully switching to SE5 ;)</sub>, has a `Show/Hide Reference VO` option that displays the reference subtitle directly on the waveform. This PR adds a similar workflow to SE5.

The option is available from the waveform context menu and Settings, and is disabled by default.

## Implementation

When enabled:

* The working subtitle keeps its existing waveform rendering and editing behavior
* The original subtitle is drawn as a semi-transparent overlay in the lower half of the waveform
* Original cues are display-only and are not included in hit testing or waveform editing
* Existing waveform text wrapping behavior is reused
* Original rendering uses the same visibility culling logic as normal waveform paragraphs

The original overlay uses the normal waveform subtitle colors at 50% opacity.

## Screenshots

||
|-|
|<video src="https://github.com/user-attachments/assets/3698415e-bb08-4fa6-91a7-f5304ae54252">|

## Notes

The original subtitle on the waveform is currently display-only.

Editing the original directly from the waveform is not included in this PR, but could be considered later if useful.

## Testing

Manually verified with:

* Editable original subtitles
* Read-only original subtitles
* `Make new empty translation from current subtitle`

Also verified normal waveform selection, dragging, resizing, and multiline text rendering.

Tested on Fedora 44 x64 (local build).

### DIFF
--- a/src/ui/Assets/Languages/English.json
+++ b/src/ui/Assets/Languages/English.json
@@ -2668,6 +2668,7 @@
     "cloneVoiceExtractFailed": "Could not extract audio for the selected line.",
     "toggleShotChange": "Toggle shot change",
     "resetWaveformZoomAndSpeed": "Reset waveform zoom & speed",
+    "showOriginalSubtitle": "Show original subtitle",
     "showOnlyWaveform": "Show only waveform",
     "showOnlySpectrogram": "Show only spectrogram",
     "showWaveformAndSpectrogram": "Show waveform and spectrogram",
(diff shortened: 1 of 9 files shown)


niksedk:
Thx for the PR :)

 I will take a look after 5.2 final (no new language strings/features atm)

PR #14623, opened 2026-09-06
Toggle dialog dashes: honor focused text box, per-column direction, never stack dashes

### LINKED ISSUE #14621: Unwanted Behavior Regarding Toggle Dialogue Dashes
Hello!
When toggling dialogue dashes in **SE4**, the dashes are applied only to the two lines in the **Text** box. Nothing happens to the text in the **Original text** box unless the cursor is placed inside that box. In that case, pressing the **Toggle dialogue dashes** shortcut affects the original text.

**SE5 behaves differently:** dialogue dashes are added to the text in the **Original text** box as well.

Additionally, **Toggle dialogue dashes** should actually toggle the dashes: it should only add them when they are missing and remove them when they are already present. It should not keep adding more dashes to lines that already have them.

For example, we should never end up with consecutive dashes like this:

```text
- - - - - Elle a des ecchymoses puis
- - - - - une blessure à la tête.
```

At the moment, when the cursor is inside the **Original text** box, it seems that **Toggle dialogue dashes** is not actually toggling anything, it simply keeps adding dashes every time the shortcut is pressed.

In my case, **Toggle dialogue dashes** is assigned to **Alt+NumPadDecimal (Decimal)**.

A video demonstrating the problem is attached.

https://github.com/user-attachments/assets/2f2ce99d-9c86-43c1-be84-6201bce95561

Image

niksedk:
Fixes #14621

**Problem.** The toggle decided add/remove from the Text column only and applied that direction to the Original column as well. When Text was empty, a single line, or more than three lines, it never gained a dash, so every press was "add" and the Original column grew `- - - - -`. The add helper also prepended `- ` without stripping an existing dash.

**Changes**
- Like SE 4, when one of the edit text boxes has focus only that column is toggled. With focus elsewhere (the grid) both columns follow.
- Each column decides its own direction from the first selected line that has text.
- Adding dashes strips existing leading dashes first, so repeated presses collapse stacked dashes instead of growing them.
- Unit tests for add/remove idempotence, stacked-dash collapse, start tags, single line.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -14290,31 +14290,49 @@ public partial class MainViewModel :
             return;
         }
 
-        // Determine direction from the first selected line: if any of its lines
-        // already start with a dash, toggle OFF for every selected line; else
-        // toggle ON. Matches SE 4 behaviour so a single hotkey flips the whole
-        // selection in one direction.
-        var firstLines = selectedItems[0].Text?.SplitToLines() ?? new List<string>();
-        var hasStartDash = firstLines.Any(l =>
-            HtmlUtil.RemoveHtmlTags(l, true).TrimStart().StartsWith('-'));
+        // Like SE 4: when one of the edit text boxes has focus only that column
+        // is toggled; with focus elsewhere (e.g. the grid) both columns follow.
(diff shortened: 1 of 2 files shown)


GrampaWildWilly:
So it looks like I'm to expect some new things logged in . . . the error log?  I'll report whatever I find.  Once RC 5 lands.

PR #14658, opened 2026-09-08
obsolete

tekk42:
obsolete

PR #14660, opened 2026-09-08
seconv: use the GUI's PaddleOCR install and add a run timeout

niksedk:
Follow-up to the review of #14658, kept to two fixes on a fresh branch from main.

- **Find the GUI's PaddleOCR install.** seconv only looked for `paddleocr` on PATH. It now checks the standalone install the GUI downloads first (portable `OCR/PaddleOCR3-7` next to the exe, then `%APPDATA%/Subtitle Edit/OCR/PaddleOCR3-7`), and passes its bundled det/rec/cls models explicitly with the same per-script model mapping the GUI uses (PP-OCRv6 medium for Chinese/English/Japanese/Latin, PP-OCRv5 per-script models for Arabic, Cyrillic, Korean, Devanagari, etc., v3 for Georgian). A PATH install keeps the plain invocation.
- **Never hang.** Each run is bounded to 10 minutes and the process tree is killed on timeout with a clear error. `PADDLE_PDX_DISABLE_MODEL_SOURCE_CHECK` is set, like in the GUI, to avoid the offline hang at "Initializing...".

Not included: batch-per-process OCR and JSON result reading from #14658.

### DIFF
--- a/src/libse/Common/PaddleOcrModels.cs
+++ b/src/libse/Common/PaddleOcrModels.cs
@@ -1,0 +1,168 @@
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Nikse.SubtitleEdit.Core.Common
+{
+    /// <summary>
+    /// Maps a PaddleOCR language code to the detection/recognition model names shipped in the
+    /// "PaddleOCR.PP-OCRv6.support.files" bundle (standalone PaddleOCR 3.7). Shared by the GUI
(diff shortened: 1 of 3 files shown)


tekk42:
i'm sorry, i messed up. pull request is now in https://github.com/SubtitleEdit/subtitleedit/pull/14664 but appears to be obsolete regarding your changes :o

PR #14664, opened 2026-09-08
align seconv PaddleOCR with GUI logic

tekk42:
Use the same models and OCR options as the GUI
Apply the GUI's double-border preprocessing
Read recognition results from the generated JSON instead of stdout
Report OCR progress during batch processing
Handle UTF-8 output correctly on Windows


tekk42:
appears to be obsolete with https://github.com/SubtitleEdit/subtitleedit/pull/14660

PR #14668, opened 2026-09-08
perf: single reference-keyed regex cache in LanguageAutoDetect

ivandrofly:
## Summary

`LanguageAutoDetect` kept two caches for the same regex: a `ConditionalWeakTable` keyed on the word-list array and a `ConcurrentDictionary` keyed on the joined pattern string. The weak table could never release anything, because the dictionary pinned every `Regex` for the app lifetime anyway. It only existed to absorb the throwaway `params` arrays built at eleven inline call sites.

- Hoist the inline word lists into static `AutoDetectWords*` fields. Several were duplicated between `AutoDetectGoogleLanguage` and the code-page probing path.
- With every key long-lived, the weak table is the only cache `GetCount` needs. `BuildWordCountRegex` constructs the `Regex` directly, and the string-keyed dictionary remains only for `GetCountContains`, so no lookup does a `string.Join` any more.
- `GetCount` takes `string[]` instead of `params`, so an inline list can no longer bypass the reference-keyed cache and recompile a compiled regex on every call.
- Cache the `CreateValueCallback` delegate. On the netstandard2.1 (C# 8) build a method-group argument allocated one per lookup.
- Drop character classes and duplicate entries made redundant by `IgnoreCase` (`[Cc]'est`, `[Vv]reau`, `honor`/`honour` listed twice, `Какво`/`какво`). Match counts are unchanged.

## Benchmarks

BenchmarkDotNet, .NET 10.0.11, i7-13700 (hybrid P/E cores, so every run reports a multimodal distribution and a StdDev of 0.5 to 1 ms). `main` was built in a separate worktree and each pair was run back to back. 800-line subtitles, `AutoDetectGoogleLanguage(subtitle, 500)`.

| Corpus | main | this PR | Allocated (main → PR) |
|---|---:|---:|---:|
| English (existing `HotPathRound8Benchmarks.AutoDetectLanguage`), two runs | 3.13 ms, 3.10 ms | 3.50 ms, 3.24 ms | 101.56 KB → 101.56 KB |
| Danish-heavy (reaches the Norwegian-only probe) | 6.72 ms | 6.23 ms | 121.87 KB → 121.27 KB |
| Spanish-heavy (reaches the French-only and Portuguese-only probes) | 6.14 ms | 5.61 ms | 120.17 KB → 119.06 KB |

Every time difference is inside the noise band, so the honest reading is "no regression". The English corpus never reaches the hoisted probes, which is why its allocation is byte-identical. The Danish and Spanish corpora do, and their allocation drops by about 1 KB per detection: that is the `string.Join` cache key that no longer gets built, and it is the only deterministic signal in the numbers.

Cold start, measured with a small harness against the branch build: the first detection in a fresh process costs about 79 ms against roughly 35 ms warm. `RegexOptions.Compiled` accounts for about 1.7 ms of JIT per word list (about 100 ms if all 58 lists are probed), and in return matches about 36% faster than the interpreter (308 ms vs 485 ms for all lists over a 198 K-character text). It pays for itself within the first detection on a large file, so `Compiled` stays.

## Test plan

- [x] `dotnet build src/libse/LibSE.csproj -c Release` succeeds for both `netstandard2.1` and `net10.0` with no warnings
- [x] `tests/libse` passes (1977/1977), including `LanguageAutoDetectTest`, `LanguageAutoDetectLanguagesTest`, and `LanguageAutoDetectShortFileTest`
- [x] `tests/seconv` language detector tests pass (the single `ProgressLineTest` failure is pre-existing on `main`)
- [x] Benchmarked against `main` (table above): no regression, fewer allocations on the corpora that reach the hoisted probes
- [ ] Open subtitles in Danish/Norwegian, Spanish/Portuguese/Italian/French, and Russian/Bulgarian and confirm auto-detected language is unchanged
- [ ] Open a Windows-1251 or Windows-1250 encoded file without a BOM and confirm the encoding guess is unchanged

https://claude.ai/code/session_01FYrDKC6UJJFmDhxQXh7ywg

### DIFF
--- a/src/libse/Common/LanguageAutoDetect.cs
+++ b/src/libse/Common/LanguageAutoDetect.cs
@@ -20,22 +20,26 @@ namespace Nikse.SubtitleEdit.Core.Common
         // built-in regex cache (Regex.CacheSize is 15), so with the static Regex.Matches API
         // every AutoDetectGoogleLanguage call re-parsed every large alternation pattern from
         // scratch. The word lists are fixed for the app lifetime, so cache compiled instances.
-        private static readonly ConcurrentDictionary<string, Regex> WordCountRegexCache = new ConcurrentDictionary<string, Regex>();
-
-        /// <summary>
-        /// The same regex again, keyed on the word-list array itself. Nearly every call site
-        /// passes one of the static AutoDetectWords* fields, so the array reference alone
-        /// identifies the pattern - and looking it up that way skips the string.Join plus
-        /// concat that built a fresh multi-kilobyte cache key on every probe (one detection
-        /// run does ~150 of them). Weak keys, so the handful of call sites that pass inline
-        /// words - which allocate a throwaway params array - cannot grow this without bound;
-        /// those still land on the pattern-keyed cache below and never recompile a regex.
-        /// </summary>
+        //
+        // Keyed on the word-list array itself: every call site passes one of the static
+        // AutoDetectWords* fields, so the array reference alone identifies the pattern and the
+        // lookup never has to string.Join a multi-kilobyte key (one detection run does ~150
+        // lookups). The keys are static fields, so entries live for the app lifetime.
         private static readonly ConditionalWeakTable<string[], Regex> WordListRegexCache = new ConditionalWeakTable<string[], Regex>();
 
-        private static int GetCount(string text, params string[] words)
+        // Cached once: on the netstandard2.1 build (C# 8) a method-group argument would
+        // otherwise allocate a fresh delegate on every lookup.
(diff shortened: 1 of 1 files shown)


niksedk:
Thanks for digging into this. The benchmarks here show the regex cache lookups were not where the time goes: the detector still ran one compiled regex per word list, so every detection scanned the text 40+ times and paid regex JIT per list on first use. #14756 replaces that with a single tokenizing pass over the text (one hash lookup per word for all 58 lists), which takes warm detection from ~5 ms to ~0.15 ms and the cold first call from ~230 ms to ~35 ms, with identical counts. It also keeps the hoisted static word lists from this PR. Closing in favour of #14756.

PR #14720, opened 2026-09-09
Unify PaddleOCR Standalone/Python JSON readout and use async polling to avoid result-order races

tekk42:
### Unify PaddleOCR Standalone/Python JSON readout and use async polling to avoid result-order races

**Background**
PaddleOCR 3.x supports writing one **<index>_res.json** result file per input image via **--save_path**.
PaddleOCR Standalone previously used its stdout output for batch results. This worked, but using the JSON output provides a more modern result interface and allows the Standalone and Python implementations to use the same result handling path.
This change therefore switches PaddleOCR Standalone to JSON readout and unifies it with the existing Python JSON result handling.

**Observed result-order issue**
During testing of the JSON-based Standalone path, an intermittent ordering problem was observed:
- Input images had the correct sequential indices.
- The generated <index>_res.json files appeared in the expected order.
- The polling/logging showed the expected sequential indices and result processing.
- Nevertheless, two adjacent subtitles were occasionally swapped in the resulting subtitle output.
- The problem was non-deterministic and could disappear or reappear between otherwise identical runs.
- Different polling intervals changed the frequency with which the problem was observed.
No sorting of the resulting subtitles was introduced as a workaround. OCR processing must preserve the original input order; sorting by timestamp, end time, or paragraph number remains an independent, user-selectable Batch Converter operation.

**Changes**
- Switch PaddleOCR Standalone from stdout result parsing to PaddleOCR's <index>_res.json output.
- Use the same JSON-based result handling for PaddleOCR Standalone and Python, simplifying and unifying the two paths.
- Continue draining PaddleOCR's stdout even though OCR results are no longer parsed from it. This prevents the stdout pipe from filling up and potentially blocking the PaddleOCR process.
- Replace the previous repeated full-batch JSON polling with per-result polling.
- Wait asynchronously for the JSON belonging to the current batch index.
- Use a 5 ms delay only while the expected JSON is not yet available or complete.
- Process already-existing subsequent JSON files immediately, without an additional polling delay.
- Add a 20-second timeout for an individual result. On timeout, report [OCR TIMEOUT] and continue with the next index.
- Keep the existing IProgress<PaddleOcrBatchProgress> mechanism unchanged.

**Async result polling**
The new implementation waits for results in the existing batch-index order without serializing PaddleOCR's actual OCR processing.
PaddleOCR can continue processing and writing multiple images independently while Subtitle Edit waits for the currently expected result. Once that result is available, it is reported immediately. If subsequent result files are already available, they are consumed immediately afterwards.
This avoids the intermittent result-order races observed with the previous polling approach.

**Timeout handling**
A single missing or incomplete result should not block the batch indefinitely.
Each expected JSON therefore has a 20-second timeout. If the file does not become available and readable within that time, the affected subtitle receives:
**[OCR TIMEOUT]**
and processing continues with the next batch index.
The explicit marker is intentional: an empty subtitle would make an OCR failure indistinguishable from a legitimately empty result and make failures difficult to locate in a large batch.

**Performance**
The async polling does not introduce a performance penalty in testing. In fact, the complete OCR run was slightly faster than before, although the measured improvement was below 5%.
With approximately 1500 difficult VobSub subtitles, a complete batch took about 72 seconds on a RTX5060.
The 5 ms delay is only used while the currently expected JSON is not yet available or complete. If the file already exists, it is processed immediately.

**Result**
The JSON output path is now shared by PaddleOCR Standalone and Python, simplifying and unifying the two implementations.
The intermittent subtitle swaps observed during testing of the JSON-based Standalone implementation were no longer reproduced with the asynchronous per-result polling approach, while preserving the original input order and the user's independent control over subtitle sorting.

### DIFF
--- a/src/ui/Features/Ocr/Engines/PaddleOcr.cs
+++ b/src/ui/Features/Ocr/Engines/PaddleOcr.cs
@@ -77,6 +77,9 @@ public partial class PaddleOcr
         };
     }
 
+    // Maximum time to wait for a single PaddleOCR result file.
+    private const int PaddleOcrResultTimeoutSeconds = 20;
+
     // The engine archives all wrap their content in a folder named after the archive itself,
     // so the root folder is derived rather than repeated; the models archive is the one that
     // does not follow that rule (".VideOCR" is in the file name only) and passes it in.
@@ -302,7 +305,7 @@ public partial class PaddleOcr
         // that OutputHandlerBatch parses. So for the Python engine we let it write one
         // "<index>_res.json" per image with --save_path and read those instead.
         string? saveFolder = null;
-        if (engineType == OcrEngineType.PaddleOcrPython)
+        if (engineType == OcrEngineType.PaddleOcrStandalone || engineType == OcrEngineType.PaddleOcrPython)
         {
             saveFolder = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
             Directory.CreateDirectory(saveFolder);
@@ -312,7 +315,10 @@ public partial class PaddleOcr
             // PP-OCRv5 models (NotImplementedError: ConvertPirAttribute2RuntimeAttribute ...).
             // The bundled standalone build is known-good and faster with MKL-DNN, so only
             // disable it for the Python engine.
-            parameters += " --enable_mkldnn False";
+            if (engineType == OcrEngineType.PaddleOcrPython)
+            {
+                parameters += " --enable_mkldnn False";
+            }
         }
 
         var process = new Process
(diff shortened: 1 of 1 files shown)


tekk42:
wrong pr, will do again

niksedk:
OK.

tekk42:
clean change-only (w/o merge) now in https://github.com/SubtitleEdit/subtitleedit/pull/14721

tekk42:
fixed correctly now in https://github.com/SubtitleEdit/subtitleedit/pull/14723

PR #14721, opened 2026-09-09
Unify PaddleOCR Standalone/Python JSON readout and use async polling to avoid result-order races

tekk42:
### Unify PaddleOCR Standalone/Python JSON readout and use async polling to avoid result-order races

**Background**
PaddleOCR 3.x supports writing one **<index>_res.json** result file per input image via **--save_path**.
PaddleOCR Standalone previously used its stdout output for batch results. This worked, but using the JSON output provides a more modern result interface and allows the Standalone and Python implementations to use the same result handling path.
This change therefore switches PaddleOCR Standalone to JSON readout and unifies it with the existing Python JSON result handling.

**Observed result-order issue**
During testing of the JSON-based Standalone path, an intermittent ordering problem was observed:
- Input images had the correct sequential indices.
- The generated <index>_res.json files appeared in the expected order.
- The polling/logging showed the expected sequential indices and result processing.
- Nevertheless, two adjacent subtitles were occasionally swapped in the resulting subtitle output.
- The problem was non-deterministic and could disappear or reappear between otherwise identical runs.
- Different polling intervals changed the frequency with which the problem was observed.
No sorting of the resulting subtitles was introduced as a workaround. OCR processing must preserve the original input order; sorting by timestamp, end time, or paragraph number remains an independent, user-selectable Batch Converter operation.

**Changes**
- Switch PaddleOCR Standalone from stdout result parsing to PaddleOCR's <index>_res.json output.
- Use the same JSON-based result handling for PaddleOCR Standalone and Python, simplifying and unifying the two paths.
- Continue draining PaddleOCR's stdout even though OCR results are no longer parsed from it. This prevents the stdout pipe from filling up and potentially blocking the PaddleOCR process.
- Replace the previous repeated full-batch JSON polling with per-result polling.
- Wait asynchronously for the JSON belonging to the current batch index.
- Use a 5 ms delay only while the expected JSON is not yet available or complete.
- Process already-existing subsequent JSON files immediately, without an additional polling delay.
- Add a 20-second timeout for an individual result. On timeout, report [OCR TIMEOUT] and continue with the next index.
- Keep the existing IProgress<PaddleOcrBatchProgress> mechanism unchanged.

**Async result polling**
The new implementation waits for results in the existing batch-index order without serializing PaddleOCR's actual OCR processing.
PaddleOCR can continue processing and writing multiple images independently while Subtitle Edit waits for the currently expected result. Once that result is available, it is reported immediately. If subsequent result files are already available, they are consumed immediately afterwards.
This avoids the intermittent result-order races observed with the previous polling approach.

**Timeout handling**
A single missing or incomplete result should not block the batch indefinitely.
Each expected JSON therefore has a 20-second timeout. If the file does not become available and readable within that time, the affected subtitle receives:
**[OCR TIMEOUT]**
and processing continues with the next batch index.
The explicit marker is intentional: an empty subtitle would make an OCR failure indistinguishable from a legitimately empty result and make failures difficult to locate in a large batch.

**Performance**
The async polling does not introduce a performance penalty in testing. In fact, the complete OCR run was slightly faster than before, although the measured improvement was below 5%.
With approximately 1500 difficult VobSub subtitles, a complete batch took about 72 seconds on a RTX5060.
The 5 ms delay is only used while the currently expected JSON is not yet available or complete. If the file already exists, it is processed immediately.

**Result**
The JSON output path is now shared by PaddleOCR Standalone and Python, simplifying and unifying the two implementations.
The intermittent subtitle swaps observed during testing of the JSON-based Standalone implementation were no longer reproduced with the asynchronous per-result polling approach, while preserving the original input order and the user's independent control over subtitle sorting.

### DIFF
--- a/src/ui/Features/Ocr/Engines/PaddleOcr.cs
+++ b/src/ui/Features/Ocr/Engines/PaddleOcr.cs
@@ -77,6 +77,9 @@ public partial class PaddleOcr
         };
     }
 
+    // Maximum time to wait for a single PaddleOCR result file.
+    private const int PaddleOcrResultTimeoutSeconds = 20;
+
     // The engine archives all wrap their content in a folder named after the archive itself,
     // so the root folder is derived rather than repeated; the models archive is the one that
     // does not follow that rule (".VideOCR" is in the file name only) and passes it in.
@@ -302,7 +305,7 @@ public partial class PaddleOcr
         // that OutputHandlerBatch parses. So for the Python engine we let it write one
         // "<index>_res.json" per image with --save_path and read those instead.
         string? saveFolder = null;
-        if (engineType == OcrEngineType.PaddleOcrPython)
+        if (engineType == OcrEngineType.PaddleOcrStandalone || engineType == OcrEngineType.PaddleOcrPython)
         {
             saveFolder = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
             Directory.CreateDirectory(saveFolder);
@@ -312,7 +315,10 @@ public partial class PaddleOcr
             // PP-OCRv5 models (NotImplementedError: ConvertPirAttribute2RuntimeAttribute ...).
             // The bundled standalone build is known-good and faster with MKL-DNN, so only
             // disable it for the Python engine.
-            parameters += " --enable_mkldnn False";
+            if (engineType == OcrEngineType.PaddleOcrPython)
+            {
+                parameters += " --enable_mkldnn False";
+            }
         }
 
         var process = new Process
@@ -341,7 +347,10 @@ public partial class PaddleOcr
         // We always pass explicit local model dirs, so skip PaddleX's online model-source
         // connectivity check - otherwise it can hang the OCR run at "Initializing...".
         process.StartInfo.EnvironmentVariables["PADDLE_PDX_DISABLE_MODEL_SOURCE_CHECK"] = "True";
-        process.OutputDataReceived += OutputHandlerBatch;
+        if (saveFolder == null)
+        {
+            process.OutputDataReceived += OutputHandlerBatch;
+        }
         process.ErrorDataReceived += ErrorHandler;
         _textDetectionResults.Clear();
         lock (_errorLock)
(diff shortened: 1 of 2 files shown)


src/ui/Features/Ocr/Engines/PaddleOcr.cs
```diff
@@ -373,35 +382,87 @@ await Parallel.ForEachAsync(bitmaps, parallelOptions, async (input, ct) =>
...
+                        {
+                            // file is probably still written to
+                        }
+                    }
+                    if (waitStopwatch.Elapsed >= TimeSpan.FromSeconds(PaddleOcrResultTimeoutSeconds))
+                    {
+                        json = string.Empty;
+                        break;
+                    }
+                    await Task.Delay(5, cancellationToken);
```
niksedk, line 425: Regression: the old loop caught `TaskCanceledException` here and fell through to `WaitForExitAsync`, whose catch calls `KillProcessTree`. Now the exception escapes `OcrBatch` before that point, so on cancel the launcher and its worker process keep running.

src/ui/Features/Ocr/Engines/PaddleOcr.cs
```diff
@@ -373,35 +382,87 @@ await Parallel.ForEachAsync(bitmaps, parallelOptions, async (input, ct) =>
...
+                            {
+                                break;
+                            }
+                        }
+                        catch
+                        {
+                            // file is probably still written to
+                        }
+                    }
+                    if (waitStopwatch.Elapsed >= TimeSpan.FromSeconds(PaddleOcrResultTimeoutSeconds))
```
niksedk, line 420: A fixed 20 s per file is wrong in both directions:

- Cold GPU start (first run, CUDA init) can easily exceed 20 s before the first JSON appears, so the leading lines get stamped as timeouts even though they will be produced.
- If the worker dies mid-batch, every remaining index waits its own 20 s. With 1000 lines left that is over five hours before the run "finishes".

The old code sized the wait on `process.HasExited` (short grace after exit, long safety net otherwise). The loop should bail out once the process has exited and the file has not appeared after a short grace, and the pre-first-result wait should be generous.

src/ui/Features/Ocr/Engines/PaddleOcr.cs
```diff
@@ -373,35 +382,87 @@ await Parallel.ForEachAsync(bitmaps, parallelOptions, async (input, ct) =>
...
+                if (string.IsNullOrEmpty(json))
                 {
-                    roundsSinceProgress = 0;
+                    reportedStems.Add(stem);
+
+                    _batchProgress?.Report(new PaddleOcrBatchProgress
+                    {
+                        Index = input.Index,
+                        Item = input.Item,
+                        Text = "[OCR TIMEOUT]",
```
niksedk, line 441: Hard-coded English text written into subtitle output. At minimum route it through `Se.Language`; I'd rather report an empty text with confidence 0 and write the failure to the tools log, since consumers already treat empty results as blank lines.

niksedk (changes requested):
Thanks - unifying the standalone and Python engines on the `_res.json` readout is a good direction and I'd like to land that part. A few things need fixing first, and I think the ordering problem is actually in the batch-convert consumer rather than in the poller:

**CI is red.** `ReportNewPaddleOcrPythonResults` was removed but `tests/UI/Features/Ocr/Engines/PaddleOcrResultParserTests.cs` still calls it, so the test project does not compile. `InitializeForTest` is kept but now unused. Please factor the new per-index wait into a testable method and rewrite that test against it rather than dropping it.

**The swap race lives in the consumer.** The old loop already reported in index order (`OrderBy(p => p.Index)`), so the producer was not reordering anything. Batch convert runs inside `Task.Run` with no SynchronizationContext, which means each `Progress<T>.Report` is queued to the thread pool independently and the callbacks can execute out of order. `BatchConverter.RunPaddleOcr` then does `item.Subtitle.Paragraphs.Add(paragraph)` in arrival order. That explains every symptom you describe, including the sensitivity to the polling interval: the 5 ms cadence spaces reports out and makes reordering rarer, but it cannot eliminate it. The fix belongs in `BatchConverter` (place paragraphs by `p.Index`, or invoke the sink synchronously instead of via `Progress<T>`). The OCR window consumers key by index already and are unaffected.

Inline comments cover the cancel regression and the timeout shape.

tekk42:
Hi nikse, thanks for your reply. I saw the producer delivering the results in order, but a switch to synchronous invoke did not fix the mixed subtitles for me. But maybe I did something wrong there. I also did not consider your correct thoughts regarding the async pull. I'll close this PR and start back from the original code - switching Standalone to JSON is minor changes in PaddleOcr.cs only & then I'll go for BatchConverter.cs. Thank you.

tekk42:
fixed correctly now in https://github.com/SubtitleEdit/subtitleedit/pull/14723

PR #14722, opened 2026-09-09
feat(waveform): sync video/audio scrubbing on Ctrl+Shift waveform drag

ghostminhtoan:
### Summary of Changes:
- **Video & Audio Preview Sync (`Ctrl + Shift` Drag)**:
  - Synchronized video frame seeking when holding `Ctrl + Shift` and dragging subtitle `StartTime`, `EndTime`, or the subtitle `Body` (moving whole subtitle block).
- **Auto Play / Pause Support**:
  - Automatically starts video/audio playback when `Ctrl + Shift` dragging starts from a paused video state.
  - Automatically pauses video playback upon mouse release/capture loss (`OnDragEnded`).
- **Audio Scrubbing Optimization**:
  - Reduced `ScrubSeekMinIntervalMs` to `30ms` for smooth real-time sound feedback while dragging.
- **Smooth Dragging Fix**:
  - Bypassed neighbour boundary overlap restrictions during `Ctrl + Shift` dragging to eliminate stuttering between closely positioned subtitles.

### DIFF
--- a/src/ui/Controls/AudioVisualizerControl/AudioVisualizer.cs
+++ b/src/ui/Controls/AudioVisualizerControl/AudioVisualizer.cs
@@ -1,4 +1,4 @@
-﻿using Avalonia;
+using Avalonia;
 using Avalonia.Controls;
 using Avalonia.Input;
 using Avalonia.Media;
@@ -544,6 +544,7 @@ public class AudioVisualizer : Control
     public class PositionEventArgs : EventArgs
     {
         public double PositionInSeconds { get; set; }
+        public bool IsCtrlShift { get; set; }
     }
 
     public class ContextEventArgs : EventArgs
@@ -586,6 +587,7 @@ public class AudioVisualizer : Control
     /// move/resize drag. Lets hosts select the paragraph the user grabbed before the drag
     /// mutates it (#14000) - a click is delivered via <see cref="OnPrimarySingleClicked"/> instead.</summary>
     public event ParagraphEventHandler? OnDragStarted;
+    public event EventHandler? OnDragEnded;
 
     /// <summary>Raised when the user clicks the empty waveform to generate it on demand
     /// (shown only when auto-generate is off and there are no cached peaks).</summary>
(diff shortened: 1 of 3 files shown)


src/ui/Features/Main/MainViewModel.cs
```diff
@@ -31541,6 +31541,13 @@ internal void AudioVisualizerOnVideoPositionChanged(object sender, AudioVisualiz
         // Throttled: the pin below keeps the cursor/centered view exact per input event; the
         // actual mpv seek is issued at most every ScrubSeekMinIntervalMs, and a deferred target
         // is landed by the trailing check in the cursor timer. See the fields for the why.
+        var isCtrlShift = AudioVisualizer != null && AudioVisualizer.IsEditingWithPointer; // checks pointer drag
```
niksedk: This variable is named `isCtrlShift` but it checks `IsEditingWithPointer`, which is true for any pointer drag. With `SetVideoPositionOnMoveStartEnd` enabled, a plain edge resize on a paused video now starts playback and pauses again on release. The control should tell the host whether Ctrl+Shift is held (e.g. a flag on `PositionEventArgs` or a separate event) so the auto-play only fires for the new mode.

src/ui/Features/Main/MainViewModel.cs
```diff
@@ -711,7 +711,7 @@ private List<SubtitleLineViewModel> GetSelectedSubtitlesInOrder(bool includeRefe
     // to the input, and it is local and cheap, so it still runs per event; the real seek only
     // needs to keep up at ~10/s, newest target wins, with a trailing seek (issued from the
     // cursor timer) landing the final position. Single clicks pass through on the leading edge.
-    private const double ScrubSeekMinIntervalMs = 100;
+    private const double ScrubSeekMinIntervalMs = 30;
```
niksedk: This lowers the throttle for every scrub path (timeline drag, edge resize), not just Ctrl+Shift drags, and the comment above still says ~10/s. If 30 ms is needed for the new mode, consider a separate interval for it, or at least update the comment and check the seek backlog on keyframe-only sources.

src/ui/Controls/AudioVisualizerControl/AudioVisualizer.cs
```diff
@@ -1495,7 +1503,7 @@ private void OnPointerMoved(object? sender, PointerEventArgs e)
         var previous = currentIndex > 0 ? _displayableParagraphs[currentIndex - 1] : null;
         var next = currentIndex < _displayableParagraphs.Count - 1 ? _displayableParagraphs[currentIndex + 1] : null;
 
-        if (_isShiftDown || Se.Settings.Waveform.AllowOverlap)
+        if (_isShiftDown || Se.Settings.Waveform.AllowOverlap || (_isCtrlDown && _isShiftDown))
```
niksedk: The added `|| (_isCtrlDown && _isShiftDown)` is unreachable: the condition already begins with `_isShiftDown ||`. Can be dropped.

niksedk (commented):
Thanks for this. Three notes from a read of the diff; the first one is the blocker.

PR #14734, opened 2026-09-09
SeConv PaddleOCR: ~60× faster with batch processing and JSON polling

tekk42:
**Summary**
Change seconv to process PaddleOCR input in batch mode, matching the approach used by the Subtitle Edit GUI.

**Changes**
- Process all subtitle images in a single PaddleOCR invocation instead of starting PaddleOCR separately for every image.
- Write prepared images to a temporary in directory and read PaddleOCR's generated JSON results from an out directory via polling.
- Add the same double-border preprocessing used by the GUI: 10 px black border surrounded by 10 px transparent border.
- Keep the existing VobSub/PGS colour-isolation preprocessing unchanged.
- Preserve the existing standalone PaddleOCR model discovery and configuration.
- Retain process timeouts and redirected stdout/stderr handling.
- Report OCR progress as batch results become available.

**Performance**
A typical movie with around 1,500 subtitle images previously took more than 90 minutes to process. Now it takes around 90 seconds.

### DIFF
--- a/src/ui/Features/Main/MainViewModel.cs
+++ b/src/ui/Features/Main/MainViewModel.cs
@@ -603,6 +603,9 @@ public partial class MainViewModel :
     Control? _focusBeforeMainMenu;
     Control? _focusBeforeWindowDeactivated;
     bool _altClosesMainMenuOnKeyUp;
+    bool _altChordCancelsMainMenuActivation;
+    bool _mainMenuActiveBeforeAltRelease;
+    Control? _focusBeforeAltChord;
     readonly AltMenuActivationGuard _altMenuActivationGuard = new();
     bool _findClosingProgrammatically;
     ReplaceViewModel? _replaceViewModel;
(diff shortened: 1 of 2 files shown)


tekk42:
Hi nikse, thanks (again) for making proper improvements out of my inexperienced mess. Indeed I forgot about the BluRay & DVB part. Regarding the speed: I assume you use the CPU version of Paddle, probably on an older CPU. On my 16C/32T CPU, 1 VOBSUB takes around 400 ms and 1 PGS around 500 ms for the Paddle CPU version, so that with the 2.6-2.7s load time the new code already brings a 6-7x speedup. For the Paddle CUDA 12 version, the difference is even higher: the load times are around 4 s, while the subtitle images (either VOBSUB or PGS) just take around 50 ms. So here the new batch code delivers a speedup of around 60-80x.

niksedk:
Thanks for your PRs :)
You are correct, I do have a older Windows laptop with only CPU - nice to hear to good performance with CUDA and the new batch-mode :)

tekk42:
paddle
Hi nikse, here are the measurements. I'm posting this as I think it's interesting how 2 parallel seconv runs only slow down each by 20-40%. If your 18s paddle runs do not fully load your CPU, you might gain a total speedup by splitting the paddle folder in 2 parallel batches.

PR #14747, opened 2026-09-10
Paddle OCR: batch every seconv image path, share the engine plumbing with the OCR window

niksedk:
**Stacked on #14732** (base branch is `claude/paddle-standalone-json-cleanup`; it retargets to `main` once that merges). Supersedes #14734, whose direction this takes.

## seconv was only half-batched

#14734 batched one of the three OCR loops in `ImageOcrLoader`. The two it left alone cover most of what people actually convert:

| loop | formats | before |
|---|---|---|
| `BitmapItemsToSubtitle` | VobSub, MKV VobSub, MKV DVB-sub, MP4 VobSub, XSub | batched by #14734 |
| `PcsListToSubtitle` | **Blu-ray SUP, MKV PGS** | one process per image |
| `LoadTransportStreamDvbSub` | DVB-sub from .ts | one process per image |

All three now go through one run.

## What the measurements actually say

Real engine (PaddleOCR-Standalone v3.7.0 CPU), 40-image Blu-ray SUP: **14m03s → 12m18s**.

That is a much smaller win than #14734's "90 min → 90 s", and the reason matters: process start + model load is only ~2.6 s per image here, while *recognising* one full-HD frame costs ~18 s. Batching removes the former, and the latter now dominates.

The reason recognition is so expensive is that **seconv hardcodes the `server` model tier** while the OCR window uses `Se.Settings.Ocr.PaddleOcrMode` (mobile by default). Same two frames, one process:

| models | time |
|---|---|
| `PP-OCRv6_medium` (what seconv uses) | 40.8 s |
| `PP-OCRv6_small` (what the GUI defaults to) | 11.2 s |

**3.6× on the same images.** That is now the dominant cost in seconv and is worth a follow-up — deliberately not changed here, since silently switching model tiers changes output quality and that is your call.

## Shared plumbing instead of two drifting copies

The GUI and seconv kept two copies of the same logic; #14732 had already changed one of them. The shared parts move to `libuilogic/Ocr/Paddle`, which both projects already reference:

- **`PaddleOcrResultJson`** — reading a `<stem>_res.json`, plus the "finished being written" test the poller needs.
- **`PaddleOcrResultPoller`** — the whole poll loop, one directory listing per round.
- **`PaddleOcrTextLayout`** — grouping boxes into visual lines. seconv concatenated `rec_texts` in *detection* order, so a two-line cue could come out upside down; it now orders by geometry like the OCR window.
- **`PaddleOcrImagePrep`** — the double border and the input/result naming.

With the json path shared, seconv's stdout parsing is gone — and with it a real bug: the old parser never unescaped Python's `repr`, so apostrophes came through as `it\'s`. The sample SUP now converts identically through seconv and the OCR window. (`PaddleOcrResultParser` goes too; #14732 left it with no production caller.)

## Review findings from #14734, fixed

- **`--quiet` printed OCR progress.** `ProgressLine`'s own docs say "callers suppress it in quiet/JSON mode" and the batch callback didn't. Verified: a `--quiet` run now emits zero OCR lines. The closing 100% is also no longer printed twice.
- **A short result count no longer throws the run away.** Missing lines come back blank with a warning; only a run that produced *nothing* is an error.
- **Memory no longer scales with subtitle length.** Bitmaps are fetched one at a time and released immediately, via a streaming `RecognizeBatch(count, writePreparedImage, progress)` overload, instead of holding every decoded frame.

## Testing

24 new tests in `tests/libuilogic` for the shared json/poller/layout: out-of-order arrival, files caught mid-write, the annotated `_ocr_res_img.png` files saved next to the results, and an unreadable result keeping its slot so later lines aren't misattributed. Plus UI tests for the batch-position → line-index mapping and seconv tests for the arguments.

The real-engine integration tests from #14732 still pass (3/3), including the sample Blu-ray SUP end to end.

Pre-existing failures on `main`, unchanged here and verified from a clean `origin/main` worktree: 25 in `tests/UI`, `ProgressLineTest.Report_Redirected` (a Windows line-ending assertion) and `NOcrTrainerTests.TrainedDb_RoundTrips`.

## One behaviour change to be aware of

seconv's single-image path never applied the double border; the batch path does (as #14734 intended, matching the GUI). On the sample SUP that changes two characters — a closing `"` read as `'`, and `days...` as `days..` — in both cases now matching what the OCR window produces for the same file. Consistency was the goal; flagging it since it is visible.


tekk42:
Hi nikse, thanks (again) for making proper improvements out of my inexperienced mess. Indeed I forgot about the BluRay & DVB part. Regarding the speed: I assume you use the CPU version of Paddle, probably on an older CPU. On my 16C/32T CPU, 1 VOBSUB takes around 400 ms and 1 PGS around 500 ms for the Paddle CPU version, so that with the 2.6-2.7s load time the new code already brings a 6-7x speedup. For the Paddle CUDA 12 version, the difference is even higher: the load times are around 4 s, while the subtitle images (either VOBSUB or PGS) just take around 50 ms. So here the new batch code delivers a speedup of around 60-80x.

tekk42:
paddle
Hi nikse, here are the measurements. I'm posting this as I think it's interesting how 2 parallel seconv runs only slow down each by 20-40%. If your 18s paddle runs do not fully load your CPU, you might gain a total speedup by splitting the paddle folder in 2 parallel batches.

PR #14750, opened 2026-09-10
Settings: Redesign toolbar settings as an icon toggle grid

pdjdev:
## Summary

Replace the toolbar settings checkbox list with an icon toggle grid, making toolbar items easier to recognize and customize.

## Background

The previous list of checkbox labels was difficult to scan at a glance. Inspired by the icons at the top of the [Shortcuts dialog](https://github.com/SubtitleEdit/subtitleedit/blob/main/docs/screenshots/shortcuts.png), the new layout lets users identify toolbar items visually and toggle them directly.
## Implementation

- Display toolbar icons with labels, tooltips, and a checkmark for enabled items.
- Keep `Encoding` and `Frame rate` as checkboxes, since they have no icons.
- Centralize settings mappings in `ToolbarSettingItem.cs`, removing repetitive properties and load/save assignments from the settings view model.
- Reuse existing localized labels and remove redundant toolbar language keys.
- Preserve settings search support.

### How to add a new toolbar item

1. Add the toolbar icon asset.*
2. Add a visibility property to `SeAppearance` if needed, then register its icon, label, and settings mapping in `ToolbarSettingItem.cs`.
3. Add the actual toolbar button in `InitToolbar.cs`.

*For items without an icon, set `ImageName` to `null` to render a checkbox.

## Screenshot

|Before|After|After (Light)|
|-|-|-|
|image|image|image|

## Testing

- Verified correct display in both light and dark themes.
- Confirmed toolbar settings load, toggle, and apply correctly.


niksedk:
Looks good :)

Are the new language strings needed?

pdjdev:
No new language strings are needed.
I reused the existing menu/general labels, so the redundant `showToolbar*` strings can be removed instead.

niksedk:
You have many changes to English.json + LanguageSettings.cs... why?

pdjdev:
Previously, each toolbar item had a separate string, like `ShowToolbarBeautifyTimeCodes = "Show beautify time codes icon"`.

The new UI reuses the existing label, such as `Se.Language.Main.Menu.BeautifyTimeCodes`, since it represents the same action. The same applies to most of the other toolbar items, so the old `ShowToolbar~` strings are no longer used.

That's why `LanguageSettings.cs` and `English.json` show quite a few changes — they're just removing those now-unused strings.

niksedk (commented):
Nice work — I built the branch, rendered the new section headlessly in both themes and probed the behaviour with temporary `AvaloniaFact` harnesses. Build is clean (0 warnings) and the full UI suite passes (5046 tests).

### What works well

- **Complete and correctly ordered**: all 27 `SeAppearance.ToolbarShow*` flags are represented, in the same order `InitToolbar` emits the buttons.
- **State moved to the model**: because `IsVisible` lives on `ToolbarSettingItem` instead of on the control, a pending toggle survives the page rebuild that every search keystroke triggers (verified: toggle "Open video" on, type in the search box, clear it — still on).
- Load/save round-trip verified: toggling two tiles and running `SaveSettings` flips exactly the right `SeAppearance` flags.
- Localization is a net win — 26 hand-written "Show … icon" strings replaced by already-translated menu labels. `English.json` is in sync (`EnglishLanguageFileUpToDateTests` and `LanguageJsonFilesTests` pass), and the keys left behind in the other translation files are harmless since the language context uses `JsonUnmappedMemberHandling.Skip`.
- Tiles get `AutomationProperties.Name`, and putting the foreground override in `Resources.ThemeDictionaries` means it also behaves correctly under the "System" theme.

### Findings

**1. Icons go stale — and nearly invisible — when the theme changes while the dialog is open.** `InitToolbar.MakeImage` bakes the bitmap from the process-wide `_imagePath` cache, and `MakeOneColor` recolors for the theme active *at build time*. Pressing Apply after changing **Theme** (or **Icon theme**) runs `MainViewModel.ApplySettings` → `RebuildToolbar`, but nothing rebuilds the settings page — so the tiles keep the previous theme's icons: white glyphs on light cards (or dark glyphs on dark cards). The card and text colours *do* update live via the theme dictionaries, which makes the mismatch stand out. I reproduced this headlessly: build the section under Dark, switch to Light, and every glyph washes out.

   Cheapest fix: have `SettingsPage` rebuild its sections on `ActualThemeVariantChanged` (or after Apply). Ideally also drop the `_imagePath` cache in `MakeImage` so a rebuild cannot pick up the old theme folder.

**2. The tooltip ignores the "show button hints" setting.** `[ToolTip.TipProperty] = item.Name` is unconditional. The convention in this repo — used ten lines away in this same file, in the proxy bypass-list item — is `if (Se.Settings.Appearance.ShowHints) { ToolTip.SetTip(...); }`.

**3. Two labels are ellipsized**: "Generate video with burned-i…" and "Point sync via other subtitle" don't fit `MaxLines = 2` in a 120x85 tile. Right now the tooltip is the only way to read them, so once (2) is fixed a user with hints off has no disclosure at all. A third line, or wider tiles, would cover it.

**4. Small search regression for generic terms.** Verified against `main`: searching "icon" or "show" used to surface the whole Toolbar section (the labels were "Show … icon"); with this PR both return nothing there. Specific terms ("burn", "spell", "point sync") still work, and section titles aren't searched on either branch, so "toolbar" misses both ways. If it matters, an optional search-alias string on `SettingsItem` would restore it.

**5. Hardcoded accent colour.** `#3285D8` / `#263285D8` are baked in, whereas the shortcuts dialog that inspired this uses `UiUtil.GetAccentBrush()` (see `ShortcutGroupTile` and `ShortcutsWindow.cs:55`). Using the helper keeps the tiles on the system accent and consistent with that window.

**6. Docs need a touch-up**: `docs/features/settings.md:117` still opens with "One checkbox per toolbar button".

### Nits

- Every search keystroke re-reads 25 PNGs from disk and re-runs the recolor loop: the Toolbar section costs ~4.0 ms per rebuild vs ~1.1 ms for the comparable Network section. Not painful, but a static bitmap cache — or an `ItemsControl` + `FuncDataTemplate` over `ToolbarItems`, the way `ShortcutsWindow` builds its group tiles — makes rebuilds free. (If you go the template route: bind every per-item property and pass `supportsRecycling: true`, otherwise container recycling re-runs the template with a null item.)
- `Panel itemsPanel = WrapItems ? new WrapPanel { … } : Panel;` in `SettingsSection.Build` shadows the type name with the property. It compiles, but it reads badly.
- The new inline checkbox factory omits the `VerticalAlignment = Center` that `MakeCheckboxSetting` sets, so Encoding/Frame rate align slightly differently from every other checkbox row.
- Those two checkboxes also end up visually orphaned below the grid — a generic-glyph tile, or a small sub-header, would tie them back in.

Nothing here is a correctness bug in load/save, and none of it blocks the idea — finding 1 is the one I'd want fixed before this ships, and 2 and 6 are one-liners.

pdjdev:
Thanks for the detailed review!

I've addressed findings 1, 2, 5, and 6 in the latest commits.

PR #14758, opened 2026-09-10
Source view: Fix selection with extra original rows

pdjdev:
## Problem

Extra original rows in the subtitle grid caused Source View to jump to a later cue.

## Fix

Exclude extra original rows when calculating the source view index. Selecting an extra original row uses the preceding working cue, or the first cue if none precedes it.

## Screenshots

| Before | After |
|-|-|
| image | image |
| Selecting cue 193 jumps to cue 222 (29 cues ahead). | Selecting cue 193 correctly jumps to cue 193. |

## Testing

- Manually verified correct Source View navigation in SRT, VTT, and SAMI.


niksedk:
thx :)

PR #14841, opened 2026-09-13
FFmpeg player: D3D11VA hardware decoding on Windows

niksedk:
Follow-up to #14838. The hardware-decoding path is now platform-generic: a static picks the device type (VideoToolbox on macOS, D3D11VA on Windows, none elsewhere) and the matching surface pixel format (`AV_PIX_FMT_VIDEOTOOLBOX` / `AV_PIX_FMT_D3D11`). The get_format callback, the `av_hwframe_transfer_data` copy to system memory (NV12 / P010) and the software fallback are shared.

Needs a Windows test: play an H.264 or HEVC file with the FFmpeg player and check that CPU use drops and that no "D3D11VA decoding failed" line appears in the error log. Requires the bundled FFmpeg build to have d3d11va enabled (the standard Windows builds do).


niksedk:
Superseded by #14840 (D3D11VA/DXVA2 hardware decoding on Windows), which is merged.

Issue #14845, opened 2026-09-13
Feature Request: Letterboxing Ribbon


kadrimarzouki:
Hello!

Could it be possible to add a **Letterboxing** option to the **Video** menu?

The idea would be to add black cinematic bars to the top and bottom of the video, similar to the letterboxing used in movies.

Ideally, selecting **Video → Letterboxing** would open a dialog box where we can:

* Adjust the **height of the top and bottom black bars**.
* See a **live preview** of the result while changing the dimensions.
* Use a slider/drag control to adjust the bar height more easily.
* Click **Apply** to preview/apply the effect while keeping the dialog open.
* Click **OK** to validate the settings and close the dialog.

The layering should work like this:

* **Layer 0:** Black letterboxing bars
* **Layer 1:** Primary and secondary subtitles

This is important because the subtitles should appear **on top of the black bars**, rather than being pushed upward/downward by them. This would make it possible to cover **hardcoded subtitles** that are already present in the video without having to change the subtitle's positioning level.

The feature could also be useful for videos that **don't originally have letterboxing**, when the subtitler simply wants to work with a cinematic presentation.

The settings should also be **saved**, so we don't have to recreate the same letterboxing dimensions every time we open Subtitle Edit.

A few important points:

* This should work **only in Normal mode**.
* It is **only a visual overlay inside Subtitle Edit** 
* The primary and secondary subtitle tracks should both remain above the black bars.
* The attached pictures illustrate the desired effect.

No Letterboxing 

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/5ef84c97-66e9-4bbf-b1f3-6d9e6d3055e3" />

---

Letterboxing Exemple 1
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/27acae6b-aa6d-447b-b4bf-07376b5cd014" />

---

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/dc9c95a0-0a6c-48e2-8c23-b508ffcff630" />

@muaz978, @mtsubs, @niksedk, @OmrSi, @pdjdev

muaz978:
Hi @kadrimarzouki, thanks for writing this up so clearly, the screenshots make the intent obvious right away. I'm going to pick this up.

[...]

PR #14872, opened 2026-09-14
Video menu: add a Letterboxing ribbon (#14845)

muaz978:
## Summary

Adds **Video > Letterboxing...** - virtual black bars drawn over the video preview, adjustable top and bottom height, with an Apply (keep tweaking) / OK (commit and close) split, settings persisted between sessions. Addresses #14845.

## How it works

The bars are fed into mpv as a `vf` filter (`drawbox`), not an Avalonia overlay. Two things rule the overlay approach out:

- On the native-window video backends (`mpv-wid` on Windows, VLC everywhere), an Avalonia control can't reliably paint over the embedded video surface at all - this repo already has a name for it ("the airspace problem") and a workaround used elsewhere (`BurnInLogoWindow`).
- On every backend, the preview subtitle is already burned into the decoded frame by mpv/VLC's own renderer (`sub-add` + libass) before Avalonia ever sees pixels. An overlay drawn "above" the video would also sit above whatever subtitle text is already flattened into it.

Routing the bars through mpv's own `vf` chain sidesteps both: `vf` filters run before mpv's subtitle compositing by default, so bars-then-subtitles falls out for free on every backend, with zero change to subtitle margins or position.

I verified this against a real mpv/ffmpeg pixel test before writing any player code: a synthetic clip with `drawbox` placed before a `subtitles` filter in the chain shows subtitle text surviving on top of the box; with the order reversed, the box fully erases the subtitle. Same underlying libass renderer mpv uses for `sub-add`.

Heights are stored as a percentage of the video's own height and written into the filter as ffmpeg expressions (`ih*0.15`), so nothing here needs to look up a video's actual pixel dimensions.

The dialog doesn't open its own video player - it drives the already-open docked player (which `ShowDialogAsync` already pauses for the dialog's duration), so dragging a slider is a live preview against the real editing view, including whatever subtitle happens to already be showing - the actual scenario the issue describes (covering a hardcoded subtitle already burned into the source video).

## Tests

- `LetterboxFilterBuilderTests` - the filter-string logic (clamping, empty-when-off, color formatting).
- `LetterboxViewModelTests` - settings persistence through Apply/OK, and that Cancel reverts.
- `LetterboxWindowTests` - window construction (headless Avalonia).
- Full suite: `dotnet test tests/UI/UITests.csproj` passes (5201 tests).

## Known gaps / follow-ups

- VLC's filter equivalent is unverified - mpv is the default on every platform, so this isn't blocking, but VLC users may not see the bars yet.
- No drag-directly-on-the-video handle in this pass, only sliders (0-40% each).
- This build doesn't have a persistent "mode" concept to gate the issue's "only in Normal mode" on, so the menu item is available whenever a video is open - happy to adjust based on feedback about what that should mean here.
- No ratio-preset dropdown (2.35:1, 2.39:1, ...) yet, just raw top/bottom percentages.

Strings added to `English.json` only, matching this repo's translation workflow.

## Test plan

- [x] `dotnet build` (full solution) - 0 errors, no new warnings
- [x] `dotnet test tests/UI/UITests.csproj` - 5201 passed
- [ ] Manual click-through in a running build - I don't have a GUI environment to verify visually in this session, so a sanity check here would be appreciated before merge


niksedk (changes requested):
Thanks - the mpv `vf`/`drawbox` approach is the right call, and I confirmed it locally: the filter string parses in a real mpv (both boxes show up in the `[vf]` user filter list), and a rendered frame with `--sub-pos=100` shows the subtitle text drawn on top of the bottom bar, so the layering the issue asks for holds. Build is clean and the new tests plus the English.json up-to-date test pass here.

A few things before this can go in:

**Please add screenshots.** The PR says no GUI run was possible. Before merge I'd like to see the dialog itself, and the docked player with bars on (ideally with a subtitle line showing over the bottom bar) - on at least one of Windows/macOS/Linux.

**Menu placement.** This is fairly niche, so please move it out of the top level of the Video menu and into **Video > More** (the `videoMore` list in `InitMenu.cs`, which sorts itself alphabetically). Same for the native mac menu.

**Other video players.** Right now the menu item is enabled for every backend, but only mpv does anything - with the ffmpeg player or VLC the dialog opens, the sliders do nothing, and the setting is still saved. Please handle this:
- **ffmpeg player**: this one is easy. `FfmpegSoftwareControl.Render` already draws the frame with `DrawImage` and then the preview subtitle with `DrawText`, so two `FillRectangle` calls in between (using the same percent settings) give the same result as mpv. Please add that.
- **VLC**: I don't expect a VLC implementation here (there is no `drawbox` equivalent, and `--video-filter` has to be set at `libvlc_new` time). But the dialog shouldn't pretend to work - either disable the menu item when the current player isn't mpv/ffmpeg, or show a short "not supported with the VLC player" note in the dialog.

**Closing the window with the title-bar X leaves the live values in place.** `ApplyLive` writes every slider tick straight into `Se.Settings` and the player, but only `Cancel`/Escape reverts. Close via the window's X button and the player keeps the last dragged values, and `Se.Settings` holds them in memory - the next `Se.SaveSettings()` from anywhere persists them. Hook `Closing` (or `Closed`) and revert to the snapshot when `OkPressed` is false.

**Apply then Cancel is inconsistent.** Apply saves to disk, Cancel then reverts memory (and the player) to the pre-dialog snapshot, so the settings file and the running app disagree until the next save. Apply should refresh the snapshot (or Cancel should re-save).

**Bars appear in every mpv instance in the app, not just the main preview.** `ApplyLetterboxRibbon()` runs inside `LibMpvDynamicPlayer.LoadFile`, and burn-in, transparent subtitles, cut video, binary edit and the text-to-speech dialogs all create their own `LibMpvDynamicPlayer`. The burn-in preview in particular will show bars that won't be in the encoded output, which is misleading. Please keep this to the main window's player (e.g. only call it from `InitVideoPlayer`/`MpvReloader` for the main player, or make it opt-in per instance) rather than inside `LoadFile`.

**Filter reset on every subtitle refresh - please verify no stutter.** `MpvReloader.RefreshMpv` now sets `vf` on every refresh, and a subtitle refresh happens on every text edit, including during playback. mpv should treat an unchanged option value as a no-op, but a `vf` change does reinit the filter chain, so please confirm in a running build that editing text while the video plays doesn't hiccup with bars enabled.

Minor:
- `LetterboxWindowTests.BuildWindow` disposes its `SettingsScope` before the test body runs (`using var` inside the helper), so it doesn't actually scope anything.
- Code/XML comments say "ribbon" everywhere while the UI says "Letterboxing"; fine to keep, but the settings class doc comment "Reapply the last bar heights automatically whenever a video is (re)loaded" reads like a different setting than "Show letterbox bars" - please align the wording.
- Branch is 13 commits behind main; it merges cleanly, so no action needed unless CI asks for it.

muaz978:
Thanks for the thorough review, and especially for verifying the vf/drawbox approach independently. Good to have that confirmed outside my own testing.

Working through the list: menu moves to Video > More (in-window and native Mac), ffmpeg player support via the two FillRectangle calls you sketched, VLC gets disabled or marked unsupported instead of silently no-oping, Closing reverts unsaved live values, Apply refreshes the snapshot so Cancel behaves consistently, ApplyLetterboxRibbon gets scoped to just the main player instead of every LoadFile call, and I'll add a guard so vf is only rewritten when the filter string actually changes, which should also help the stutter concern, plus verify it live. Fixing the SettingsScope bug in LetterboxWindowTests and aligning the doc comment wording too.

Screenshots and the live-playback check need an actual GUI run. Working on that now, will follow up here once I have them.

muaz978:
I pushed a commit addressing the review items.

- Moved the menu entry from the top level Video menu into Video > More, next to the other dialog launchers
- Added ffmpeg player support, so the bars and live preview also work outside mpv
- VLC has no equivalent of mpv's filter chain, so the dialog now shows an explanatory note and disables the controls when the current player does not support it, instead of silently doing nothing
- Closing the dialog without pressing OK now reverts any dragged values, and Apply/Cancel agree on what counts as committed
- The mpv filter only applies to the main preview player, so a burn-in or TTS dialog's own throwaway player never shows bars it would not actually produce
- Skips redundant vf writes when nothing changed

I also tested this locally in a running build (isolated from my own working data so it would not touch anything else on my machine). Two screenshots from that session:

Letterboxing dialog with a player that supports it:

Letterboxing dialog

The main preview with the dialog open, bars enabled top and bottom, and a subtitle rendering on top of the bottom bar:

Bars with subtitle on top

Let me know if anything else needs a look before merge.

niksedk:
we already have a setting for this... margin is part subtitle area

PR #14906, opened 2026-09-15
feat(tts): multi-voice per actor review with inline voice/language selector and actor filtering

ghostminhtoan:
### Video:

https://github.com/user-attachments/assets/e4669b82-3d86-4ab8-84ca-3bb3f41a5322

### Summary of Changes

- **Multi-voice per actor review workflow**:
  - Automatically detect and map different voices per actor during TTS batch generation.
  - Allow user to review and regenerate audio segments per individual voice or actor.

- **Review Window UI & UX Enhancements**:
  - Removed redundant `Engine` column to streamline the review grid.
  - Replaced static text with interactive `ComboBox` for **Voice** and **Language** on each subtitle row, allowing users to freely customize the voice and language per line.
  - Implemented seamless two-way synchronization between row-level ComboBoxes and the left editing panel with zero lag.
  - Added right-click context menu on `Enabled` column and grid: `Select all`, `Select none`, `Invert selected`, and dynamic `Select actor: <Name>` options.

- **Voice Design Synchronization**:
  - Automatically synchronize Voice Design instructions to all lines assigned to the `Default` voice when parameters (Gender, Age, Pitch, Accent, Whisper) change.
  - Respect per-line custom voice and language selections during single or batch regeneration.

### Verification
- Preserved exact original UTF-8 BOM formatting across all 14 modified files.
- Release build passes with **0 Warnings, 0 Errors**.


niksedk:
Hi @ghostminhtoan, thanks a lot for the PR and the video, it's clear a lot of work went into this! 🙏

I'm going to close it, though, because it changes a core idea of the review window: the list view is a record of how each subtitle line **was generated** (voice, language, etc.). Because of that, those columns shouldn't be combo boxes or editable.

When you want to change something for a line, use the settings on the left side, and regenerate from there. That way the grid always shows exactly what produced the audio you're listening to.

Thanks again for contributing, and feel free to open smaller PRs or issues for the other ideas here!

Issue #14842, opened 2026-09-13
SE5 More Options for the Second Subtitle File


kadrimarzouki:
**Request:**
When opening a second subtitle file, allow its **font size and Line justification (left/center/right)** to be adjusted independently, without interfering with the primary subtitle.

As shown in the attached video, the second subtitle can be **previewed before pressing OK**, but there is currently **no option to change its text alignment/justification**.
(_**The line justification option for two-line subtitles exists in the main subtitle settings of the video player**_)

The main issue is that, once the dialog is closed by pressing **OK**, there is no way to go back and adjust the second subtitle's settings. If you want to change its size or other options later, you have to go through the whole process again.

It would be useful to have a **separate dialog box that remains accessible as long as the second subtitle file is loaded**, with:

* Font size
* Text alignment (Auto/left/center/right)
* **Apply** button to preview the changes immediately
* **OK** button to validate and close the dialog

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/25de904b-5e2f-4711-8b0d-1c0177350249" />

This would allow us to **go back at any time and adjust whatever settings we need**, while keeping the second subtitle independent from the primary one.

https://github.com/user-attachments/assets/4180afa0-7fa4-4d80-bf2c-365128dbc9da

kadrimarzouki:
Hello guys!
@muaz978, @mtsubs, @niksedk, @OmrSi, @pdjdev
Is there any hope for this one?

niksedk:
@kadrimarzouki "Justify lines" for the second subtitle file is now in, via #14997. Big thanks to @muaz978 for the work! 🙏

The "Second subtitle file" dialog has a new **Justify lines** option: auto, left, center or right. It only affects the second subtitle, not the main one, and it's saved with the other remembered settings.

Two follow-ups went in with it: justified lines now sit exactly where the unjustified block would, and italic or colour that spans a line break is kept on every line (#15244).

It will be in the next beta. Let us know how it works for you!

PR #14921, opened 2026-09-15
Secondary subtitle: keep style on reopen

muaz978:
Partial fix for #14842.

Following niksedk's review, I've split this down to just the part he confirmed is solid: the "reopen resets to defaults" bug. The "Justify lines" feature that was here has real problems (wrong PlayRes for non-ASS primaries, an untested non-ASS live-preview branch, and it isn't actually independent of the global Text Justify setting once that's not Auto) and needs more work before it's ready, so I've pulled it out entirely rather than patch it in place. It'll come back as its own PR once reworked, likely using per-line `\pos` on split single-line events so it's genuinely independent of `sub-ass-justify` regardless of the global setting.

## What's in this PR now

The "Second subtitle file (on video player)" dialog already has Font size, Bold, Border style, Color and Alignment. None of them survived reopening the dialog to adjust an already-loaded secondary subtitle, they reset to defaults every time. Now they persist:

* `SecondarySubtitleStyle` (new): the dialog's style choices as a small value, so `MainViewModel` can remember them between reopens of the same secondary subtitle file.
* `OpenSecondarySubtitleViewModel.Initialize(...)` now accepts the previous style and restores it instead of always resetting to defaults.
* `MainViewModel` keeps the last-applied secondary-subtitle style and passes it back in when the same file is reopened; clears it when the secondary subtitle is removed.

## Testing

* `dotnet test tests/UI/UITests.csproj`: 5270 passed, 1 pre-existing unrelated failure in `SpeechToTextWindowLayoutTests` (confirmed it also fails on a clean `upstream/main` checkout with no changes).
* Screenshots from an earlier local test run of this exact mechanism are in the comment below; niksedk separately confirmed the approach itself is correct ("the file is re-parsed on each reopen, so no tag stacking").


muaz978:
(Screenshot: reopening the "Second subtitle file" dialog for the same already-loaded secondary subtitle now shows "Justify lines" set to "Left" (and Font size/Bold/Alignment as last chosen), instead of resetting to defaults. I have the screenshot on hand if a maintainer wants it attached here, just let me know.)

muaz978:
I tested this locally in a running build (an ad-hoc self-contained bundle assembled from `dotnet publish`, not the signed installer, with a throwaway primary/secondary subtitle and `tests/libse/Files/sample_MP4.mp4` as the video). Two screenshots from that session:

**First open** - a secondary subtitle just loaded for the first time, "Justify lines" defaults to Auto like the rest of the dialog's fields:

Second subtitle file dialog on first open, Justify lines set to Auto

**Reopened to adjust** - after picking Left and pressing OK, reopening the dialog for the same already-loaded secondary subtitle now starts from what was last chosen (Justify lines: Left, plus Font size/Bold/Alignment) instead of resetting to defaults:

Second subtitle file dialog reopened, Justify lines persisted as Left

niksedk (commented):
Thanks for the thorough write-up! The "keep style on reopen" part looks good (the file is re-parsed on each reopen, so no tag stacking). The "Justify lines" part has a few problems though:

**1. `\pos` is off-screen for SubRip (and most non-ASS) primaries**
`SecondarySubtitleJustifier` computes `\pos` in the secondary's PlayRes, i.e. the video dimensions (e.g. 1920x1080). But for a SubRip primary, `MpvReloader.BuildPreviewText` uses `MpvPreviewStyleHeader` (`AdvancedSubStationAlpha.HeaderNoStyles`), which has no PlayRes, so libass uses 384x288. `SecondarySubtitleMerger.AddSecondarySubtitle` resamples the secondary *style* (font size, margins) to the target PlayRes, but not the `\pos` inside the paragraph text. So something like `\pos(960,1070)` lands far outside a 384x288 script, and justified multi-line secondary lines are pushed off the video. The same applies to an ASSA primary whose PlayRes differs from the video size. (WebVTT primaries are fine since `WebVttToAssa` uses the video dimensions.)

**2. The dialog's live preview doesn't apply it for non-ASS primaries**
In `BuildAssaSubtitle(true)`, `SecondarySubtitleJustifier.Apply` is only called in the ASSA-primary branch, so with e.g. a SubRip primary the preview in the dialog never shows the justify change. This is probably why #1 wasn't visible while testing.

**3. Not independent of the global "Text Justify" setting**
When the global Text Justify is anything but Auto, `LibMpvDynamicPlayer` sets `sub-ass-justify=yes`, which overrides line justification for all ASS events, including the secondary's `\an7`/`\an9` ones.

**Minor**
- Width is measured with Skia at `style.FontSize`, but libass scales fonts differently (by ascender+descender), and outline/box width isn't included, so the block shifts a bit from where the alignment alone would put it.
- `\pos` disables libass collision handling, so overlapping secondary events will overlap on screen instead of stacking.
- The prepended `{\anN\pos(...)}` wins over any `\an`/`\pos` already in the secondary file's text.

**Suggestion:** it would be easiest to split this. The reopen-keeps-style fix could go in on its own, and the justify feature needs `\pos` rescaled to the target PlayRes (or applied after merging), plus the live preview covering the non-ASS branch.

muaz978:
@niksedk Thanks for the detailed review, this all checks out, I verified each point directly in the code before acting on it:

- Confirmed `AdvancedSubStationAlpha.HeaderNoStyles` (what a SubRip primary's preview header uses) carries no PlayResX/PlayResY, so libass falls back to 384x288 while my `\pos` was computed against the real video dimensions. That would indeed push justified secondary lines off-screen on anything but a WebVTT primary.
- Confirmed `SecondarySubtitleJustifier.Apply` was only wired into the ASSA-primary merge branch and `Ok()`, never the non-ASS branch, so my own local testing (SubRip primary) never exercised the broken path.
- Agreed on independence from the global Text Justify setting: keeping the multi-line text as one event with a repositioned anchor means `sub-ass-justify=yes` still overrides it. Doing this properly needs per-line `\pos` on split single-line events, which I'd scoped out early on for simplicity and shouldn't have.

I've pulled "Justify lines" out of this PR entirely rather than patch it here. What's left is exactly the part you confirmed: Font size, Bold, Border style, Color and Alignment on the "Second subtitle file" dialog now persist across reopens instead of resetting to defaults. Build and full test suite (`dotnet test tests/UI/UITests.csproj`) still pass after the removal.

I'll bring Justify lines back as its own PR once it's reworked with the PlayRes rescaling, the missing non-ASS preview coverage, and the split-event approach for real independence from the global setting.

muaz978:
Follow-up is up: #14928 — reworks Justify lines addressing all three points from the review (PlayRes rescaling, the missing non-ASS preview branch, and independence from the global Text Justify setting via splitting multi-line paragraphs into single-line events).

niksedk:
Thanks @muaz978 for digging into this! Superseded by #14986, which keeps the second subtitle style between opens by saving it to settings (plus an optional "Do not show this dialog again"), so it survives restarts and works for any second subtitle file. That also avoids the unrelated English.json changes here: that file gets rewritten when the UI tests run, so it's best left out of feature PRs.

muaz978:
Thanks for the kind words, and for taking this the rest of the way. Persisting to settings (surviving restarts, plus the "Do not show this dialog again" option) is a real improvement over what I had, and `SecondarySubtitleStyler` is a cleaner shape for the dialog and no-dialog paths to share. Glad #14842 is sorted. Noted on English.json, I will leave that file alone in future PRs here.

PR #14928, opened 2026-09-16
Secondary subtitle: independent line justification (reworked)

### LINKED ISSUE #14842: SE5 More Options for the Second Subtitle File

muaz978:
Reworked version of the "Justify lines" feature split out of #14921 (which now carries only the "keep style on reopen" fix), addressing niksedk's review there: https://github.com/SubtitleEdit/subtitleedit/pull/14921#pullrequestreview-5212696357

Fixes #14842.

## Recap of the mockup

The "Second subtitle file (on video player)" dialog already has Font size, Bold, Border style, Color and Alignment. This adds the missing "Justify lines" (Auto/Left/Center/Right) from the mockup in the issue, independent of both the primary subtitle and the video player's own global "Text Justify" setting.

## Addressing the three problems from review

**1. `\pos` was off-screen for non-ASS primaries.** `SecondarySubtitleJustifier` bakes `\pos` into the secondary's paragraph text, stated in the secondary's own PlayRes (the real video dimensions). `SecondarySubtitleMerger` already resamples the secondary's *style* (font size, margins) to the merge target's PlayRes when it differs - a SubRip primary's preview header has no PlayRes at all, so libass falls back to 384x288 - but it never touched `\pos` in paragraph text, so a justified secondary would render far off-screen against anything but a WebVTT primary. `SecondarySubtitleMerger` now rescales `\pos` the same way, reusing `AssaResampler`'s existing override-tag resampler, and is careful not to mutate the shared secondary paragraphs while doing it (covered by `SecondarySubtitleMergerTests`).

**2. The dialog's own live preview never exercised the non-ASS branch.** `BuildAssaSubtitle` only applied the justify transform in the ASSA-primary merge branch, so testing against a SubRip primary (which is what my own local testing used) never hit the broken `\pos` path - that's exactly why I didn't catch problem 1 myself. Justify is now computed once per call and reused by every branch, ASS or not, merged or not.

**3. Not independent of the global Text Justify setting.** mpv's `sub-ass-justify` only ever arranges multiple lines *within a single ASS event*. The previous version kept the secondary's multi-line text as one such event with a repositioned anchor, so mpv's global setting still overrode it whenever that setting wasn't Auto. `SecondarySubtitleJustifier` now splits a multi-line paragraph into one dialogue event per physical line, each with its own `\an7\pos(x,y)` - there is nothing left for `sub-ass-justify` to arrange within a one-line event, regardless of what the global setting is.

## What's new

* `SecondarySubtitleJustifier` (rewritten): for a multi-line paragraph whose justify isn't "auto", measures each line (Skia, tags stripped, outline width included) and splits it into one single-line event per physical line, each anchored via `\an7\pos`. The block's overall position still comes entirely from the paragraph's own alignment (top/middle/bottom, left/center/right) - only which corner of the block individual lines line up against changes. 19 unit tests.
* `SecondarySubtitleMerger`: rescales `\pos` on merge when the secondary's and the target's PlayRes differ, mirroring the existing style resample. 6 new unit tests, including a regression test that the shared secondary paragraph is never mutated in place (it's read on every preview refresh).
* `OpenSecondarySubtitleViewModel` / `OpenSecondarySubtitleWindow`: the "Justify lines" row (Auto/Left/Center/Right, reusing `MpvJustifyDisplay`).
* `SecondarySubtitleStyle`: `JustifyCode` alongside the other style fields already kept and restored on reopen by #14921.

## Known remaining approximation

Skia's font metrics differ from libass's own, so the measured block is close to, but not pixel-identical to, an unjustified render of the same lines. `\pos` also replaces libass's own collision handling for these events, so two secondary paragraphs that overlap in time will overlap on screen instead of stacking - a pre-existing trade-off of manual positioning, not something specific to this change.

## Testing

* `dotnet build` (UI project): 0 errors, only pre-existing unrelated warnings.
* `dotnet test tests/UI/UITests.csproj`: 5293 passed (0 failed this run; the one flaky, pre-existing, unrelated `SpeechToTextWindowLayoutTests` failure I flagged on #14921 shows up intermittently regardless of this change).
* 25 new/updated unit tests (19 for `SecondarySubtitleJustifier`, 6 for `SecondarySubtitleMerger`) covering the exact three problems above: PlayRes rescaling, shared-paragraph safety, and the split-event geometry (alignment/justify combinations, vertical anchor preservation, single-line and auto no-ops).


niksedk:
Thank you so much @muaz978 for the work you've put into this, and for addressing the review feedback so thoroughly - the PlayRes rescaling and split-event approach were well thought through.

I'm closing this for now, as #14986 has since reworked the second subtitle dialog: the style is now remembered in settings (with an optional "Do not show this dialog again"), and the style building moved into a shared `SecondarySubtitleStyler`. That means this PR no longer applies cleanly on main.

If you'd like to bring "Justify lines" back, a fresh PR on top of current main would be very welcome - ideally with justify saved alongside the other `SecondarySubtitle*` settings, and without the `English.json` changes (that file gets rewritten when the UI tests run, so it's best left out of feature PRs).

Thanks again for helping improve Subtitle Edit!

muaz978:
Really appreciate the thorough review on this one, it caught real bugs before they shipped. Glad the PlayRes rescaling and split-event reasoning held up.

Makes sense to close this given #14986 changed the ground it was standing on. I will bring "Justify lines" back as a fresh PR on top of current main, built against `SecondarySubtitleStyler` and saved alongside the other `SecondarySubtitle*` settings, and I will leave English.json alone this time.

Thanks for the guidance, and for your patience cooperating with me on improving Subtitle Edit.

PR #14961, opened 2026-09-16
Allow frame-based values in Bridge gaps (#14959)

### LINKED ISSUE #14959: Allow frame-based values in «Bridge gaps»
I would like to ask whether it would be possible to bring back frame-based input for the «Bridge gaps» function.

In 5.2.0, «Bridge gaps smaller than» uses milliseconds only. As I often work with style guides that define these gaps in whole frames, I really miss being able to simply enter the number of frames to be bridged, as in the 4.x versions.

If there is a reason to keep milliseconds, perhaps an optional choice between milliseconds and frames could be a possible solution.

Thanks again for your work on Subtitle Edit.

niksedk:
Closes #14959

In 5.2.0, **Tools → Bridge gaps** only takes milliseconds, while style guides give these gaps in whole frames (SE 4 took frames in HH:MM:SS:FF mode).

## Changes
- **Frame mode:** "Bridge gaps smaller than" and "Minimum gap" are entered in frames (labels say so) and counted at the current frame rate - in the dialog and in Batch convert's Bridge gaps step (which uses the batch's output frame rate: the change-frame-rate target, else the current one).
- **Separate storage:** new nullable `BridgeGapsSmallerThanFrames` / `MinGapFrames` in `SeBridgeGaps`, so switching the time format never reads a frame value as milliseconds (the bug that made the old conversion turn 2000 ms into 80 000 ms). Null derives from the ms value on first use. Saving frames also updates the ms keys, which Merge short lines reads as its threshold; saving ms leaves the frame keys alone.
- **Calculator:** in millisecond mode both values get the "..." frame rate calculator (as in Settings and Apply min gap). `MinGapCalculateViewModel.Initialize` gains an overload for title / frames label / question so it can name the bridge gap limit; the existing callers are unchanged.
- Docs: `docs/features/bridge-gaps.md` updated (screenshot not refreshed).

## Tests
`BridgeGapsViewModelTests` (9): labels per mode, frames converted at 25 fps, gap wider than the frame limit left alone, first frame-mode use derived from ms, per-unit save/load, calculator text overrides and defaults. Related suites (BridgeGaps, MinGapCalculate, SettingsPersistence, BatchConvert, ApplyMinGap, MergeShortLines) pass locally: 138/138.

Changelog entry left to the maintainer.


GitGianluc:
Hi @niksedk! I just wanted to let you know that some strings in the interface for the new `Bridge gaps` feature are displayed in English without having been localised.

#### Steps to reproduce
1. Go to `Tools > Bridge gaps`.
2. Click on the three dots button next to «Bridge gaps smaller than (ms)».

Bridge-gaps_01

3. As you can see, a couple of new strings have not been localised.

Bridge-gaps_02

Perhaps they need to be added to this section of the `English.json` file?
https://github.com/SubtitleEdit/subtitleedit/blob/a94e9fbe4ae7f60c3a7aa1eb5ce809f220755baf/src/ui/Assets/Languages/English.json#L1895-L1901

Issue #14966, opened 2026-09-17
delete selection for everywhere shortcut


ruonghieu:
Hi I would like to request a "delete slection" for everywhere shortcut.

There is already the "ripple delete selection" for everywhere, but I do not use ripple delete.

When I edit subs I usually work mostly with the waveform, so it would be nice to have either the "delete slection" 
shortcut "for the waveform", 
or "for everywhere" in case I am scrolling through the list view.

Thank you for your precious work.

PR #14989, opened 2026-09-17
Add Actor picker window for easier actor assignment (SSA/ASS)

zbik1:
### Actor picker

I added a new non-modal tool window (Tools menu) for easier actor assignment in ASS/SSA files.

Setting actors is not the most convenient now. It requires either using a context menu in the grid, or using the shortcuts, which is tricky, as you can't always see which shortcut is assigned to which actor. You need to open the context menu to check it anyway, and the order changes, as the actors are always sorted alphabetically, This tool makes actor assigning faster. The idea is to keep it open next to the main window while it is needed.

### Features:
- fast one-click assignment of existing actors to selected subtitle lines
- buttons for adding a new actor or clearing the actor in the currently selected subtitle
- shortcut badges showing the current SetActor 1-10 bindings.
- the actor assigned to the currently selected subtitle is dynamically highlighted in the tool window
- the window is non-modal, designed to keep it open while working with actors. It restores main window focus on click, which makes it easy to assign actors with mouse while still using shortcuts to control video playback
- context menu for batch renaming existing actors across all lines (uses an already existing feature)
- the tools menu entry is only visible when ASS/SSA file is open

### Demonstration:
I'm attaching a screenshot and a video to show you how it works.

https://github.com/user-attachments/assets/f4110015-a5b5-4211-b901-f1eb69f62e89

actor_picker

### Note on shortcuts:
For better keyboard-only workflow, it would be best if actor assigning shortcuts were set to general category (so that they could be used while the focus is on the waveform). Alternatively, another set of actors 1-10 keys could be added, one that would work everywhere, like you did today with a new Delete everywhere shortcut (#14966). I did not want to implement this change without consulting with you first, though.

This was developed and tested with assistance from Claude Code.


niksedk:
Thanks for the work and the detailed write-up, @zbik1 – the demo looks nice.

However, I'm going to pass on this one: actor assignment is already covered by the grid context menu and the "Set actor 1-10" shortcuts, and a separate tool window ends up too close to that existing feature to justify the extra UI surface and maintenance.

If the current shortcut workflow feels awkward (e.g. not seeing which actor a shortcut maps to, or the shortcuts not working with focus on the waveform), feel free to open an issue about that – improving the existing feature is something I'd be happy to look at.

zbik1:
Hi @niksedk, I understand. I opened an issue (#14989) where I tried to explain the awkward points in the actor workflow and the possible solutions I can think of.

PR #14995, opened 2026-09-17
Background music generation (Video > More and TTS window)

niksedk:
Generates instrumental background music locally with **ACE-Step 1.5 Turbo** (MIT) through the shared audio.cpp runtime, loops it seamlessly to the video's length, and puts it on the video.

### Video > More > Generate background music...
- Presets, sorted alphabetically: Cooking / baking (default), Corporate / explainer, Custom, Forest / nature, Lo-fi vlog, Relaxing piano, Tech / gadget review. Prompt, tempo (BPM), generate length, seed (random or fixed), music volume.
- Generate → Play/Stop preview (looped to the video's length) → **Save audio...** / **Add to video...**
- **Remove existing audio tracks** is on by default. Turn it off to mix the music under the first audio track, with its own volume setting.
- Progress shows each phase (loading model → composing → generating → decoding) with "Time remaining" from `ProgressHelper`.

### TTS window
- New **Add background music** checkbox with a settings button. The button opens the same dialog in TTS mode: preset/prompt/tempo/seed, and the music volume under speech (default 30%).
- The runtime and model are checked *before* speech synthesis starts. After the merge, the music is looped to the video's length (or the speech's), then mixed under the speech with sidechain ducking, so it dips while someone talks. Music previewed in the dialog is reused. If the music step fails, the speech-only file is kept.
- Whether the original audio is replaced or ducked is unchanged: it still follows the existing Advanced > Audio ducking setting.

### Looping (`MusicLooper`, libuilogic)
ACE-Step clips end in a 6–9 s outro fade. The tempo lands a few percent off the request (asked for 110 bpm, got 113), and sections repeat sometimes sample-for-sample, sometimes only musically. The looper therefore:
- measures bar length (onset autocorrelation) and beat phase and downbeat (short-window onsets, chroma novelty);
- drops the outro;
- scores seams on rhythm and harmony, preferring whole 4-bar phrases;
- crossfades one bar: equal-gain for identical material, equal-power otherwise;
- normalizes to about -20 dBFS RMS with a -1 dBFS peak cap.

Seam scores on six real 60 s preset clips: corporate 1.00, lo-fi 0.98, tech 0.89, cooking 0.87, forest 0.71, piano 0.62. Finding the loop takes about 0.3–0.8 s per clip.

### Runtime
- Pins support-files `audiocpp-indextts25-2026-09-17b`: the same upstream v0.8.0 @ 4af14322 as today's pin, with `ace_step` compiled in. Existing installs get the usual "audio.cpp update required" prompt through the BUILD-INFO family gate. The TTS engines run the same code.
- Model: `ace-step-1.5-turbo-q8_0.gguf` (5.8 GB, SHA-256 verified), downloaded on first use after a prompt.
- On an M4 (Metal), a 60 s clip takes about 100 s: load 20–30 s, planner 25 s, DiT 8 s, VAE decode 37 s. Peak memory is about 7–8 GB.
- audiocpp_cli writes 16-bit with hard clipping, and ACE-Step decodes above full scale. When the runtime's `--help` lists `--out-format`, the CLI is called with `--out-format float32` (upstream 0xShug0/audio.cpp#591, not merged yet). Until then output can clip slightly.

### Tests
- libuilogic `MusicLooperTests`: synthetic songs cover bar alignment, tempo drift (110 requested → 113 measured), outro trimming, render length and level, normalization, and WAV round-trip including float32 above full scale.
- UI `BackgroundMusicTests`: CLI arguments, progress parser, add-to-video and mix-under-speech ffmpeg arguments (per-branch aformat for ffmpeg 4.x, padded sidechain), presets, settings defaults.
- `AudioCppRuntimeBuildInfoTests` now asserts that `ace_step` is in the pinned build.
- Manually tested on macOS: the dialog, generation, playback, and a TTS run with background music. The ffmpeg mix was checked on 4.4.6 and 9.0.1.

The changelog entry is left to the maintainer. English.json is not regenerated.


GitGianluc:
@niksedk these new strings are also missing from the `English.json` file:

https://github.com/SubtitleEdit/subtitleedit/blob/7a6bff3118a4a6ed5c2034dc1142350816c31945/src/ui/Logic/Config/Language/Video/LanguageBackgroundMusic.cs#L49-L94

See also [#14961](https://github.com/SubtitleEdit/subtitleedit/pull/14961#issuecomment-5711468299) (comment).

niksedk:
@GitGianluc thanks, you were right - `English.json` is generated from the language classes and had not been regenerated since beta5, so the new strings were in the code but not in the file.

It is regenerated now in #15016 (merged): the `backgroundMusic` section, the Bridge gaps frame/calculator strings from #14961, plus everything else added since beta5 (3D export, move-lines hints, online-only video, second subtitle dialog options, etc. - about 90 keys). One key was renamed to match the code: `setActorXY` -> `setActorX`.

Issue #14990, opened 2026-09-17
SE v5.3.0-beta 4: Shift+space can't be used as Toggle play/pause when in text box


zbik1:
Hi,

when the Toggle play/pause shortcut is assigned to shift+space, it can't be used when the focus is in the text box. Instead, a standard space is typed into the text box.  This does not happen e.g. with ctrl+space.

By the way, there are two shortcut entries in the settings for play/pause that appear to be exactly the same. I'm not sure if this is intended.

Thank you for your great work!

niksedk:
Shift+Space is a typing key (Shift+Space types a space, e.g. while typing uppercase words with Shift held), so it is treated like Shift+letter in the text box. Two options:

- Turn on "Allow single-letter shortcuts in text box" in Settings, or
- Bind play/pause to Ctrl+Space instead, which never types.

niksedk:
Oh, the two lines in the shortcuts window allow for binding to two different shortcuts.

zbik1:
Hi, understood, the single-letter shortcut setting solves this for me. Thanks.

PR #14996, opened 2026-09-17
Shift+space shortcut fix (fixes #14990)

zbik1:
Fixes issue #14990 

This adds Space to the keys where Shift means a shortcut, not typing, same as Delete and Backspace already work. Without this, a shortcut bound to Shift+Space (e.g. play/pause) never fired in a text box, it just typed a space.


niksedk:
Closing: Space is a typing key, so Shift+Space should stay typeable in the text box. See the reply in #14990 for the two working options.

PR #15011, opened 2026-09-18
Feat/tts review audio segments enhancements

ghostminhtoan:
### Video:
https://drive.google.com/file/d/1OnwZt5PVZXExfck5PZz4uhR8xO_tc7SJ/view?usp=drive_link

### Summary of Changes

This PR significantly improves the **Review Speech (TTS audio segments)** window with multi-actor workflow support, review table columns, search filters, undo/redo history, and critical bug fixes for audio regeneration and voice precedence:

---

#### 1. 🎭 TTS Review audio segments: 
- Full integration for multi-speaker subtitles in the Review Speech window.
- Automatically identifies actor tags from subtitle lines, enabling clear actor-based grouping, voice assignments, and line management.

#### 2. 📊 Review Columns & Searchable Filter Menus
- **New Columns**: Added dedicated `Actor`, `Voice`, `Engine`, and `Language` columns directly into the review table view, providing full visibility into per-line configuration.
- **Searchable Menus**: Context submenus for `Voice`, `Language`, and `Engine` now feature a real-time search/filter box with scrolling support.
- **Click Fix**: Fixed an issue where clicking directly inside the search/filter `TextBox` caused the context menu to unexpectedly close.

#### 3. ↩️ Undo / Redo (`Ctrl+Z`, `Ctrl+Y`) & Selection Shortcuts
- **Undo / Redo**: Added full history support (`Ctrl+Z` and `Ctrl+Y`) for checkbox state toggles and batch assignment actions in the review grid.
- **Batch Selection**: Added context menu shortcuts: **Select all**, **Select none**, **Invert selection**, **Select by actor**, and **Select by voice**.
- **Highlight Synchronization**: Synchronized row selection highlight styling with checkbox cell background states.

#### 4. 🔄 Regenerate Audio & Audio Segment Preservation
- **Unified Action**: Streamlined the context menu to a single, clear `"Regenerate audio"` command (removing redundant and confusing duplicate actions).
- **Segment Preservation (Critical Fix)**: Regenerating only selected lines now strictly preserves all previously generated audio segments for unselected lines. The final audio export cleanly merges untouched and newly generated segments without any audio loss.

#### 5. ⚙️ Voice & Engine Precedence (No-Actor vs. Multi-Actor)
- **Subtitles Without Actors (`!hasActor`)**:
  - The left panel is treated as the **Single Source of Truth**.
  - Clicking rows no longer overwrites the left panel with stale per-line metadata, and regenerating lines strictly applies the active voice, engine, and language selected on the left panel.
- **Subtitles With Actors (`hasActor`)**:
  - The right grid columns take precedence to maintain per-actor voice mapping and custom line overrides.

---

### Verification
- Builds cleanly via `dotnet build "src/ui/UI.csproj" -c Release` (.NET 10.0, 0 Errors).
- Clean merge against latest `upstream/main` with zero merge conflicts.
- Preserved exact UTF-8 BOM encoding across all modified files matching repository standards.


niksedk:
Thanks for the work on this (and the video) - there are good ideas here, but I can't merge it as is. The core problem is that it changes what the **Include** checkbox means.

**1. Include is "include this line's audio in the final merge", not "selected for regenerate"**

Today `Ok()` publishes only the rows with Include ticked, so unticking a row is how you leave that line's audio out of the merged track (and the flag round-trips through Export/Import for that reason). The PR changes `Ok()` to return all rows and only copies the flag to `StepResult.Include` - but `MergeAudioParagraphs` never looks at `Include`, so after this PR an unticked row is merged anyway. The "segment preservation" fix is really removing the exclude feature - the unticked rows were dropped on purpose.

**2. Regenerate now regenerates everything by default**

Every row starts with `Include = true`, and `RegenerateAudio(null)` / `RegenerateSelectedLine()` now go to `RegenerateSelectedLines()` whenever `Lines.Any(l => l.Include)`. So opening the window and pressing the regenerate shortcut (or the context menu item) re-synthesizes every line of the subtitle instead of the current line. With ElevenLabs/Murf/Azure etc. that costs real money, and with local engines it can take a very long time.

Tying the grid selection to the Include flag (`SyncGridSelectionFromIncluded`) has the same problem from the other side: selecting rows and being included in the output are two different things.

If "regenerate several lines" is the goal, it should work from the grid's multi-selection (`lineGrid.SelectedItems`) and leave Include alone.

**3. `RegenerateSelectedLines` is a second copy of `RegenerateAudio`**

It duplicates ~150 lines but without several of the guards the single-line path has (empty-text message, the "only swap StepResult once the whole pipeline succeeded" logic, error reporting - failures are just `continue`d silently). The batch path should loop over the existing single-line implementation rather than fork it.

**4. Smaller things**

- `Header = "Include"` is a hard-coded string (was `Se.Language.General.Enabled`).
- Ctrl+Z / Ctrl+Y are hard-coded with `KeyModifiers.Control`, so no Cmd on macOS.
- Undo only covers checkbox/assignment state, not regenerated audio, which will surprise people since regenerate is the main action in this window.

**5. Scope**

This is five features in one ~1800 line PR, which makes it very hard to review and to revert a part of. Could you split it up? I'd be happy to look at these separately:

1. Actor / Engine / Language columns (read-only, shown when there are multiple actors) - this part looks fine.
2. Multi-select + "regenerate selected lines", based on grid selection and reusing `RegenerateAudio`.
3. Per-row voice/engine/language assignment with the searchable menus.

Undo/redo and the select-by-actor/voice helpers can follow once 2+3 are in.

niksedk:
What are you trying to do exactly?
I am closing this for now.

ghostminhtoan:
Hi @niksedk ,

I apologize for the delay in following up—I stepped away for a bit while refactoring the branch based on your detailed feedback.

To answer your question directly about what I am trying to achieve:

### The original problem & goal:
When working with multi-speaker / multi-actor subtitles (e.g., using "Set up cast" where different characters have distinct voices, engines, or languages), the **Review Audio Segments** window currently provides no visibility into **which actor belongs to which line**, or **which engine/language** generated that segment. Reviewing dialogue in multi-voice dubbing was essentially blind—you couldn't tell at a glance if a line had the correct character voice without clicking through each row.

### Where I went wrong initially:
In my enthusiasm to also add batch-regeneration and per-row assignment (as shown in the video), I over-scoped the PR into 5 features (~1,800 lines) and mistakenly coupled row selection with the `Include` flag, breaking its fundamental merge-exclusion purpose. You were 100% right to reject that approach.

### What is in this PR now (strictly Part 1):
I have stripped out all the auxiliary features and returned strictly to the **core problem of visual clarity during review**:
1. **Actor, Engine, and Language columns**:
   - Added as clean, read-only columns in the review grid.
   - Automatically shown only when `HasMultipleActors` is true (via `TableViewColumnManager`), keeping the view identical to before for single-voice sessions.
2. **Restored original `Include` semantics**:
   - `Ok()` retains `StepResults = Lines.Where(p => p.Include).Select(p => p.StepResult).ToArray();`. Unticked lines are excluded from the merged output as originally intended.
   - Completely disconnected grid selection from `Include`.
3. **Restored localization**:
   - Checkbox header uses `Se.Language.General.Enabled`.
4. **Minimal, surgical diff**:
   - Dropped from ~1,800 lines down to just **+84 / -9 lines** across 4 files (no duplicate regeneration code, no undo/redo, no menu bloat).

The updated commit (`94509207b`) is already pushed. Could you please consider reopening this PR to review this clean foundation? (Or let me know if you would prefer a fresh PR).

It's gonna be 0:00 am now (UTC+7) so I'm gonna go to bed. I'll be online in 9 hours later.


Thank you for your patience and for keeping the project's architecture clean!

nikse:
> Hi @nikse,

wrong guy 😅

if i had a nickel for every time you all tagged me instead of @niksedk...

